PR #39, opened 2024-12-22
[WIP] Feat/s3 destination

shubham19may:
# Description

Added S3 as a writer in destinations

# How Has This Been Tested?

- [X] Tried to upload twitter_data with it, it created 1 file. Also deleted local temp file.

# Screenshots or Recordings

Screenshot 2024-12-22 at 10 48 45 PM


hash-data:
seems to be a duplicate code from the local writer, can we do something by providing an option in the local writer itself, whenever the file gets closed it can be moved to s3. Local Writer configuration can detect if needs to push to s3 or not.

writers/s3/s3.go
```diff
@@ -0,0 +1,269 @@
...
+		logger.Infof("Successfully uploaded file to s3://%s/%s with %d records",
+			s.config.Bucket, s.destinationFilePath, s.records.Load())
+	}
+	logger.Infof("Deleting temp file %s", s.tempFilePath)
+
+	return nil
+}
+
+func (s *S3) ReInitiationOnTypeChange() bool {
+	return false
```
shubham19may, line 241: mark true

writers/s3/s3.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+func (s *S3) ReInitiationOnNewColumns() bool {
+	return false
```
shubham19may, line 245: mark true

hash-data:
Done via other PR closing it

PR #40, opened 2024-12-24
fix: Writer Concurrency Issues 🎉 

hash-data:
# Description

Fixed Concurrency Issues
-> not closing of rawRecord and flattenRecord Channel on Empty batch
-> unbuffered channel errorChan issue
-> pool.wait() not waiting 
-> CDC should start after full load
-> Parquet Schema 
-> redundant context cancellation and channel closing

 

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/protocol/writers.go
+++ b/protocol/writers.go
@@ -10,7 +10,6 @@ import (
 	"github.com/datazip-inc/olake/types"
 	"github.com/datazip-inc/olake/typeutils"
 	"github.com/datazip-inc/olake/utils"
-	"github.com/piyushsingariya/relec/safego"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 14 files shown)


protocol/writers.go
```diff
@@ -78,62 +85,80 @@ func NewWriter(ctx context.Context, config *types.WriterConfig) (*WriterPool, er
...
 	for _, one := range options {
 		one(opts)
 	}
 
 	var thread Writer
-	threadInitialized := make(chan struct{}) // to handle the first initialization
-
+	threadInitialized := make(chan struct{})   // to handle the first initialization
+	processedRecord := make(chan types.Record) // To be given to Writer
+	rawRecord := make(chan types.Record)       // To be given to Reader
```
shubham19may: this name makes it confusing.

protocol/writers.go
```diff
@@ -170,72 +195,69 @@ func (w *WriterPool) NewThread(parent context.Context, stream Stream, options ..
...
-			return false, err
-		default:
-			if !safego.Insert(frontend, record) {
-				return true, nil
-			}
+	mainWriter()
 
-			return false, nil
-		}
+	return &ThreadEvent{
```
shubham19may, line 191: This structure is so complex. Why there has to be a frontend and writer?

What is the benefit that we are getting in terms of passing through two async processes/goroutines?

hash-data: I have simplified it more. 
Previously it was taking time max(flattening, writing in file), but now it will take flattening+writing for each record. 
given the fact that time(writing file) >>> time(flattening)

PR #61, opened 2025-01-15
feat: Resumable Full Load In Mongodb

vikash390:
# Description

Added logging for failed chunks in the terminal and implemented handling to retry processing them efficiently.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


drivers/mongodb/internal/backfill.go
```diff
@@ -42,6 +42,36 @@ func (m *Mongo) backfill(stream protocol.Stream, pool *protocol.WriterPool) erro
...
+		failedChunks = failedChunks[1:]
+		//stream.SetChunksFromStreamState(failedChunks)
+		return false, &chunk, nil
+	}), m.config.MaxThreads, func(ctx context.Context, chunk *types.Chunk, number int64) error {
+
+		minId, err := primitive.ObjectIDFromHex(chunk.Min)
+		if err != nil {
+			return err
+		}
+		maxId, err := primitive.ObjectIDFromHex(chunk.Max)
```
hash-data: once think can we only use single utils.concurrent function to get the results, basically chunks can be fetched if state not provided else state chunks will be used (We can also modify the concurrentC funciton to make it happen)

types/state.go
```diff
@@ -81,36 +81,57 @@ func (s *State) MarshalJSON() ([]byte, error) {
 	return json.Marshal(p)
 }
 
+// DualSyncMap struct to hold Cursor and Chunks
+type DualSyncMap struct {
+	Cursor sync.Map `json:"_data"`  // Represents a map of arbitrary data (not directly serialized as-is)
+	Chunks []Chunk  `json:"chunks"` // Represents a flat slice of chunks
```
hash-data: chunks also need to inside sync.map (reason we can modify them in parallel syncs)

hash-data (commented):
More improvements required

types/state.go
hash-data: if you want to understand it more we can connect

types/state.go
```diff
@@ -81,36 +82,69 @@ func (s *State) MarshalJSON() ([]byte, error) {
 	return json.Marshal(p)
 }
 
+// DualSyncMap struct to hold Cursor and Chunks
+type DualSyncMap struct {
```
shubham19may: What is DualSyncMap name meaning? lets give names which are understandable by humans.

shubham19may: Can be StateElements

types/state.go
```diff
@@ -81,36 +82,69 @@ func (s *State) MarshalJSON() ([]byte, error) {
...
+	Cursor sync.Map `json:"_data"`  // Represents a map of arbitrary data (not directly serialized as-is)
+	Chunks sync.Map `json:"chunks"` // Represents a flat slice of chunks
+}
+
+// Chunk struct that holds status, min, and max values
+type Chunk struct {
+	Status string              `json:"status"`
+	Min    primitive.ObjectID  `json:"min"`
```
shubham19may: Is this chunk only going to be used for MongoDB, as the datatype of Min & Max is MongoObject id? We should plan to use it generally for every other DB.

Also we should consider the scenario for cases where mongodb id is not an object id.

types/state.go
```diff
@@ -129,11 +166,15 @@ func (s *StreamState) UnmarshalJSON(data []byte) error {
...
-	for key, value := range aux.State {
-		s.State.Store(key, value)
+	// Populate Cursor sync.Map with the data from the "cursor" field
+	for key, value := range aux.State.Cursor {
+		s.State.Cursor.Store(key, value)
 	}
 
+	// Populate the Chunks slice
+	for _, chunk := range aux.State.Chunks {
+		s.State.Chunks.Store(chunk.Min, chunk)
```
shubham19may: ideally we should go newest to oldest. Which means that newest chunks will come first. 

This will be particularly helpful as 
1. people mainly care about newer data,
2. when we have to skip on older data (ex, many people want to filter for only last 2 years worth of data)
3. Can generate a better schema as newer data has high probability to have all fields compared to older one.

protocol/interface.go
```diff
@@ -85,6 +86,9 @@ type Stream interface {
 	SetStateCursor(value any)
 	SetStateKey(key string, value any)
 	Validate(source *types.Stream) error
+	AppendChunksToStreamState(newChunk types.Chunk)
+	GetChunksFromStreamState() []types.Chunk
```
shubham19may: if its returning failed chunks, then why the name is not GetFailedChunksFromStreamState?

drivers/mongodb/internal/backfill.go
```diff
@@ -20,102 +20,132 @@ import (
...
-	if err != nil {
-		return err
+	var chunks []types.Chunk
+	failedChunks := stream.GetChunksFromStreamState()
+	if len(failedChunks) > 0 {
+		//process failed chunks
+		logger.Infof("Processing %d failed chunks for stream [%s]", len(failedChunks), stream.ID())
+		chunks = append(chunks, failedChunks...)
 	}
+	if len(chunks) == 0 {
```
shubham19may: else {}

protocol/interface.go
vikaxsh: it will return failed and scheduled both chunks

types/state.go
shubham19may: We can keep it as an improvement further. Resolving this for now.

drivers/mongodb/internal/backfill.go
```diff
@@ -20,102 +20,144 @@ import (
...
+					Min: generateMinObjectID(start),
+					Max: "",
+				}
+				logger.Info("scheduling last full load chunk query!")
+			} else {
+				boundry = types.Chunk{
+					Min: generateMinObjectID(start),
+					Max: generateMinObjectID(end),
+				}
+			}
```
shubham19may: can be simplified. Created boundry struct at the end.

drivers/mongodb/internal/backfill.go
```diff
@@ -20,102 +20,144 @@ import (
...
+			chunks = append(chunks, boundry)
+			chunk := types.Chunk{
```
shubham19may: why are you again creating an chunk var?

drivers/mongodb/internal/backfill.go
```diff
@@ -20,102 +20,144 @@ import (
...
+				Status: "scheduled",
+				Min:    boundry.Min,
+				Max:    boundry.Max,
+			}
+			stream.AppendChunksToStreamState(chunk)
+			start = end
+		}
+
 	}
```
shubham19may: what if node crashes here, where are we storing the chunks into state file?

types/state.go
```diff
@@ -81,36 +81,69 @@ func (s *State) MarshalJSON() ([]byte, error) {
 	return json.Marshal(p)
 }
 
+// DualSyncMap struct to hold Cursor and Chunks
```
shubham19may: change comment. Please self-review PR line by line.

types/stream_configured.go
```diff
@@ -90,32 +92,57 @@ func (s *ConfiguredStream) InitialState() any {
...
+	var chunks []Chunk
+	s.streamState.State.Chunks.Range(func(_, value any) bool {
+		if chunk, ok := value.(Chunk); ok {
+			chunks = append(chunks, chunk)
+		}
+		return true
+	})
+	return chunks
+}
+func (s *ConfiguredStream) UpdateChunkStatusInStreamState(chunkID string, newStatus string) {
```
shubham19may: we need to persist to disk here. as OOM can occur at any place.

shubham19may (changes requested):

hash-data:
Will add defaultFlag for partitioning path in this PR

drivers/mongodb/internal/backfill.go
```diff
@@ -53,111 +50,91 @@ func (m *Mongo) backfill(stream protocol.Stream, pool *protocol.WriterPool) erro
...
 
-	stream.UpdateChunkStatusInStreamState(minStr, "succeed")
-	return nil
+	return utils.Concurrent(context.TODO(), chunks.Array(), chunks.Len(), func(ctx context.Context, one types.Chunk, number int) error {
+		err := processChunk(ctx, pool, stream, collection, one.Min, &one.Max)
+		if err != nil {
+			return err
+		}
+		// remove success chunk from state
+		stream.RemoveStateChunk(one)
```
shubham19may, line 135: what if full load finished and incremental fails, 
this will make the stream remove all chunks in state so it will resync again.

logger/logger.go
```diff
@@ -180,7 +124,8 @@ func FileLogger(content any, filePath string, fileName, fileExtension string) er
 
 func Init() {
 	// Configure lumberjack for log rotation
-	timestamp := fmt.Sprintf("%d-%d-%d_%d-%d-%d", time.Now().Year(), time.Now().Month(), time.Now().Day(), time.Now().Hour(), time.Now().Minute(), time.Now().Second())
+	timeStamp := time.Now().UTC()
```
shubham19may: name it something else. one can not have same name but camel case

drivers/mongodb/internal/backfill.go
hash-data: in this case chunks key itself is not present and GetStateChunks returns nil which results in full load only.

logger/logger.go
hash-data: will do it

drivers/mongodb/internal/backfill.go
shubham19may: No, but full load is done properly, so why it should do it again?

shubham19may (approved):

PR #69, opened 2025-01-23
feat(writer): Add option to choose normalization or denormalization

hash-data:
# Description
Normalization is an option to provide level 1 flattening.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


writers/parquet/config.go
```diff
@@ -5,12 +5,13 @@ import (
...
 
 type Config struct {
-	Path      string `json:"local_path,omitempty"` // Local file path (for local file system usage)
-	Bucket    string `json:"s3_bucket,omitempty"`
-	Region    string `json:"s3_region,omitempty"`
-	AccessKey string `json:"s3_access_key,omitempty"`
-	SecretKey string `json:"s3_secret_key,omitempty"`
-	Prefix    string `json:"s3_path,omitempty"`
+	Path          string `json:"local_path,omitempty"` // Local file path (for local file system usage)
+	Normalization bool   `json:"normalization,omitempty"`
```
shubham19may, line 9:
```suggestion
	Normalize bool   `json:"Normalize,omitempty"`
```

writers/parquet/parquet.go
```diff
@@ -34,7 +35,8 @@ type Parquet struct {
 	closed              bool
 	config              *Config
 	file                source.ParquetFile
-	writer              *goparquet.FileWriter
+	normalizeWriter     *goparquet.FileWriter // TODO: implement parquet writer as interface (dependency injection)
```
shubham19may:
```suggestion
	normalizedWriter     *goparquet.FileWriter // TODO: implement parquet writer as interface (dependency injection)
```

protocol/writers.go
```diff
@@ -96,99 +98,124 @@ func (w *WriterPool) NewThread(parent context.Context, stream Stream, options ..
...
-					return err
+		// add constants key fields
+		flattenedData[constants.OlakeID] = rawRecord.OlakeID
+		flattenedData[constants.OlakeTimestamp] = rawRecord.OlakeTimestamp
+		if rawRecord.DeleteTime != 0 {
+			flattenedData[constants.CDCDeletedAt] = rawRecord.DeleteTime
+		}
+
+		// schema evolution
+		fields := make(typeutils.Fields)
```
shubham19may: would it make any difference if this is now in the normalizeFunc?
because it was outside in the previous code.

types/data_types.go
```diff
@@ -23,6 +23,21 @@ const (
 
 type Record map[string]any
 
+type RawRecord struct {
+	OlakeID        string         `parquet:"olake_id"`
+	Data           map[string]any `parquet:"data,json"`
+	DeleteTime     int64          `parquet:"cdc_deleted_at"`
+	OlakeTimestamp int64          `parquet:"olake_timestamp"`
```
shubham19may: rename `olake_insert_time`

utils/utils.go
```diff
@@ -214,3 +218,26 @@ func IsJSON(str string) bool {
...
+// GetKeysHash returns md5 hashsum of concatenated map values (sort keys before)
+func GetKeysHash(m map[string]interface{}, keys ...string) string {
+	sort.Strings(keys)
+
+	var str strings.Builder
+	for _, k := range keys {
+		str.WriteString(fmt.Sprint(m[k]))
+		str.WriteRune('|')
+	}
+	//nolint:gosec,G115
```
shubham19may, line 231: can we not use if its flagging for it?

protocol/writers.go
hash-data: let us test on schema changes, as it might increase the time and can save memory to have fieldSchema every time

utils/utils.go
hash-data: it is for md5 hash, we are using it as it generates 32 char string, If we use sha256 that generates longer string than it.

writers/parquet/config.go
hash-data: in json we are taking normalization: true, normalization false which seems to be good what do you say ? @vikash390

protocol/writers.go
hash-data: i am making it outside for now as per discussion. 
If it is inside it will create new files even schema is being changed by any other thread.

shubham19may (approved):

PR #74, opened 2025-01-27
chore: Exponential Backoff For Backfill In Mongodb

hash-data:
# Description
Exponential backoff 

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


drivers/mongodb/README.md
```diff
@@ -42,7 +42,8 @@ Add MongoDB credentials in following format in config.json file
       "server-ram": 16,
       "database": "database",
       "max_threads": 50,
-      "default_mode" :"cdc"
+      "default_mode" :"cdc",
+      "backoff_retry_count": 2
```
shubham19may, line 46: Need to explain in the doc how it works. @zriyanshdz

drivers/mongodb/internal/backfill.go
```diff
@@ -74,46 +75,49 @@ func (m *Mongo) backfill(stream protocol.Stream, pool *protocol.WriterPool) erro
...
+			if err != nil {
+				return fmt.Errorf("collection.Find: %s", err)
 			}
-
-			handleObjectID(doc)
-			exit, err := insert.Insert(types.CreateRawRecord(utils.GetKeysHash(doc, constants.MongoPrimaryID), doc, 0))
+			defer cursor.Close(ctx)
+
+			waitChannel := make(chan struct{})
+			defer func() {
```
shubham19may: This will be called even if it returns the error right?

drivers/mongodb/internal/mon.go
```diff
@@ -54,6 +55,11 @@ func (m *Mongo) Setup() error {
 	m.client = conn
 	// no need to check from discover if it have cdc support or not
 	m.CDCSupport = true
+	// check for default backoff count
+	if m.config.RetryCount == 0 {
```
shubham19may: what if person has actually set this as 0 intentionally as he doesn't want to retry it.

drivers/mongodb/internal/backfill.go
```diff
@@ -74,46 +75,49 @@ func (m *Mongo) backfill(stream protocol.Stream, pool *protocol.WriterPool) erro
...
+			insert, err := pool.NewThread(threadContext, stream, protocol.WithNumber(number), protocol.WithWaitChannel(waitChannel))
 			if err != nil {
-				return fmt.Errorf("failed to finish backfill chunk %d: %s", number, err)
+				return err
 			}
-			if exit {
-				return nil
+			defer insert.Close()
+
+			for cursor.Next(ctx) {
```
shubham19may, line 108: why can't we simply retry the cursor.next ? 
Because thats the only thing worth a retry right?

If other things fail, its fine? because we are not writing to an warehouse, we are writing to s3 so if it fails, we should let it fail.

drivers/mongodb/README.md
zriyansh: will write the doc as soon as PR gets merged

shubham19may (approved):

PR #79, opened 2025-01-30
Feat: Source Postgres

hash-data:
# Description
Added Source Postgres with Full Load & CDC Features

Features 
- Resumable Full Load: Supports restarting full loads without reprocessing all data.
- Fast Full Load via Chunking:
  • CTID-based chunking
  • Single split-column chunking
  • Evenly distributed chunks (for int/float PKs)
  • Uneven chunking support
- WAL2JSON-based CDC Sync for incremental updates.
- New streams are automatically synced when added.
- LSN mismatch handling triggers a full resync.

## Type of change

# How Has This Been Tested?
- Full Load with Two Tables: Verified successful completion of the full load process.
- Sync Interruption & Resume: Ensured that sync resumes from the last completed chunk when interrupted.
- CDC for Two Tables (Update, Delete, Insert): Verified that all changes (updates, deletions, inserts) are correctly captured and synced.
- Adding New Streams in CDC: Ensured that only new streams are synced without reprocessing existing ones.
- LSN Mismatch Handling: Validated that when an LSN mismatch occurs, a full resync is triggered.
- Using a Committed LSN Again: Confirmed that reusing a committed LSN results in an LSN mismatch and triggers a full resync.

# Screenshots or Recordings


drivers/postgres/go.mod
```diff
@@ -0,0 +1,80 @@
+module github.com/datazip-inc/olake/drivers/postgres
+
+go 1.22
+
+toolchain go1.22.5
+
+require (
+	github.com/datazip-inc/olake v0.0.0-20230630130252-054496f39abb
+	github.com/lib/pq v1.10.9
```
naveen246, line 10: https://github.com/lib/pq#status

pq is in maintenance mode, [pgx](https://github.com/jackc/pgx) might be a better choice

hash-data: will check this

hash-data: it is only being used for quoting literals
```
params += fmt.Sprintf("%s=%s ", pq.QuoteIdentifier(k), pq.QuoteLiteral(v))
```

will remove the package later on

drivers/postgres/internal/backfill.go
```diff
@@ -0,0 +1,260 @@
...
+	"github.com/datazip-inc/olake/protocol"
+	"github.com/datazip-inc/olake/types"
+	"github.com/datazip-inc/olake/utils"
+)
+
+// Simple Full Refresh Sync; Loads table fully
+func (p *Postgres) backfill(pool *protocol.WriterPool, stream protocol.Stream) error {
+	backfillCtx := context.TODO()
+	var approxRowCount int64
+	approxRowCountQuery := jdbc.PostgresRowCountQuery(stream)
```
shubham19may, line 22: what if this is not allowed for the user (Basically the creds provided might not have access to this?)

drivers/postgres/internal/backfill.go
```diff
@@ -0,0 +1,260 @@
...
+		return nil
+	}
+	return utils.Concurrent(backfillCtx, splitChunks, p.config.MaxThreads, processChunk)
+}
+
+func (p *Postgres) splitTableIntoChunks(stream protocol.Stream, approxRowCount int64) ([]types.Chunk, error) {
+	generateCTIDRanges := func(stream protocol.Stream) ([]types.Chunk, error) {
+		var relPages uint32
+		relPagesQuery := jdbc.PostgresRelPageCount(stream)
+		err := p.client.QueryRow(relPagesQuery).Scan(&relPages)
```
shubham19may, line 106: what if relPages == 0?

drivers/postgres/internal/backfill.go
```diff
@@ -0,0 +1,260 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to get relPages: %v", err)
+		}
+		var chunks []types.Chunk
+		batchSize := uint32(1000)
```
shubham19may: Too many small pages might happen.
Suggested by AI : 
```
		if relPages < 1000 {
			batchSize = uint32(relPages/10 + 1) // Ensure at least 1
		}
```

drivers/postgres/internal/backfill.go
```diff
@@ -0,0 +1,260 @@
...
+		for start := uint32(0); start < relPages; start += batchSize {
+			end := start + batchSize
+			if end >= relPages {
+				end = ^uint32(0) // Use max uint32 value for the last range
+			}
+			chunks = append(chunks, types.Chunk{Min: fmt.Sprintf("'(%d,0)'", start), Max: fmt.Sprintf("'(%d,0)'", end)})
```
shubham19may, line 118: could it be possible that end might be processed twice? as ranges will be 
0,0 - 1000,0
1000,0 - 2000,0

drivers/postgres/internal/backfill.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+		return chunks, nil
+	}
+
+	splitEvenlySizedChunks := func(min, max interface{}, approximateRowCnt int64, chunkSize, dynamicChunkSize int) ([]types.Chunk, error) {
```
shubham19may: add comment on basics of what is min, max & other params

utils/utils.go
```diff
@@ -240,3 +246,43 @@ func GetHash(m map[string]interface{}) string {
...
+	default:
+		return nil, fmt.Errorf("failed to add contant values to interface, unsupported type %T", val)
+	}
+}
+
+func CompareInterfaceValue(a, b interface{}) int {
+	switch av := a.(type) {
+	case int:
+		if bv, ok := b.(int); ok {
+			return av - bv
```
shubham19may: lets make it similar to conditions below and use a simple function to not iterate same code again.

drivers/postgres/internal/datatype_conversion.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+var pgTypeToDataTypes = map[string]types.DataType{
+	// integers
+	"bigint":      types.Int64,
+	"tinyint":     types.Int64,
+	"integer":     types.Int64,
+	"smallint":    types.Int64,
+	"smallserial": types.Int64,
+	"int":         types.Int64,
```
shubham19may, line 14: Int64 might be much higher size that required.
Lets add todo to change it to proper types.

drivers/postgres/internal/datatype_conversion.go
```diff
@@ -0,0 +1,86 @@
...
+	"tsquery":           types.String,
+	"tsvector":          types.String,
+	"xml":               types.String,
+	"enum":              types.String,
+	"tsrange":           types.String,
+
+	// date/time
+	"time":                        types.Timestamp,
+	"timez":                       types.Timestamp,
+	"date":                        types.Timestamp,
```
shubham19may, line 76: we also need a proper datetime support

drivers/postgres/internal/config.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (c *Config) Validate() error {
+	if c.Host == "" {
+		return fmt.Errorf("empty host name")
+	} else if strings.Contains(c.Host, "https") || strings.Contains(c.Host, "http") {
+		return fmt.Errorf("host should not contain http or https")
+	}
+
+	if c.SSLConfiguration == nil {
+		return fmt.Errorf("ssl config not set")
```
shubham19may: lets have a default value for this.

drivers/postgres/internal/config.go
```diff
@@ -0,0 +1,102 @@
...
+	}
```
shubham19may, line 54:
```suggestion
	}
		// Validate port
	if c.Port <= 0 || c.Port > 65535 {
		return fmt.Errorf("invalid port number: must be between 1 and 65535")
	}
		// Set default values if not provided
	if c.BatchSize <= 0 {
		c.BatchSize = 10000 // default batch size
	}
	if c.MaxThreads <= 0 {
		c.MaxThreads = 2 // default number of threads
	}
```

drivers/postgres/internal/config.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// construct the connection string
+	connStr := fmt.Sprintf("postgres://%s:%s@%s:%d/%s", c.Username, c.Password, c.Host, c.Port, c.Database)
```
shubham19may:
```suggestion
		// construct the connection string with escaped username and password
	connStr := fmt.Sprintf("postgres://%s:%s@%s:%d/%s",
		url.QueryEscape(c.Username),
		url.QueryEscape(c.Password),
		c.Host,
		c.Port,
		url.QueryEscape(c.Database))
```

drivers/postgres/internal/config.go
```diff
@@ -0,0 +1,102 @@
...
+		for k, v := range c.JDBCURLParams {
+			params += fmt.Sprintf("%s=%s ", pq.QuoteIdentifier(k), pq.QuoteLiteral(v))
+		}
+
+		query.Add("options", params)
+	}
+
+	// Enable SSL if SSLConfig is provided
+	if c.SSLConfiguration != nil {
+		sslmode := string(c.SSLConfiguration.Mode)
```
shubham19may: shouldn't all of these be in the c. SSLConfiguration.Validate()?

drivers/postgres/internal/config.go
```diff
@@ -0,0 +1,102 @@
...
+		params := ""
+		for k, v := range c.JDBCURLParams {
+			params += fmt.Sprintf("%s=%s ", pq.QuoteIdentifier(k), pq.QuoteLiteral(v))
+		}
+
+		query.Add("options", params)
+	}
+
+	// Enable SSL if SSLConfig is provided
+	if c.SSLConfiguration != nil {
```
shubham19may: No need we are already checking nil condition above

drivers/postgres/internal/config.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+
+		if c.SSLConfiguration.ClientKey != "" {
+			query.Add("sslkey", c.SSLConfiguration.ClientKey)
+		}
+	}
+	parsed.RawQuery = query.Encode()
+	c.Connection = parsed
+
+	return c.SSLConfiguration.Validate()
```
shubham19may: shouldn't we check before applying above settings in query?

drivers/postgres/internal/config.go
```diff
@@ -0,0 +1,102 @@
...
+	connStr := fmt.Sprintf("postgres://%s:%s@%s:%d/%s", c.Username, c.Password, c.Host, c.Port, c.Database)
+	parsed, err := url.Parse(connStr)
+	if err != nil {
+		return err
+	}
+
+	query := parsed.Query()
+
+	// Set additional connection parameters if available
+	if len(c.JDBCURLParams) > 0 {
```
shubham19may, line 66:
```suggestion
	if len(c.JDBCURLParams) > 0 {
		for k, v := range c.JDBCURLParams {
			query.Add(pq.QuoteIdentifier(k), v)
		}
	}
```

drivers/postgres/internal/config.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if c.SSLConfiguration == nil {
+		return fmt.Errorf("ssl config not set")
+	}
+
+	// construct the connection string
+	connStr := fmt.Sprintf("postgres://%s:%s@%s:%d/%s", c.Username, c.Password, c.Host, c.Port, c.Database)
+	parsed, err := url.Parse(connStr)
+	if err != nil {
+		return err
```
shubham19may:
```suggestion
		return fmt.Errorf("failed to parse connection string: %w", err)
```

utils/ssl.go
```diff
@@ -0,0 +1,68 @@
...
+
+// SSLConfig is a dto for deserialized SSL configuration for Postgres
+type SSLConfig struct {
+	// SSL mode
+	//
+	// @jsonschema(
+	// required=true,
+	// enum=["require","disable","verify-ca","verify-full"]
+	// )
+	Mode string `mapstructure:"mode,omitempty" json:"mode,omitempty" yaml:"mode,omitempty"`
```
shubham19may, line 22: there should a simple default mode (Just the way dbeaver has)

utils/ssl.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		if sc.ClientCert == "" {
+			return errors.New("'ssl.client_cert' is required parameter")
+		}
+
+		if sc.ClientKey == "" {
+			return errors.New("'ssl.client_key' is required parameter")
+		}
+	}
```
shubham19may:
```suggestion
// isValidMode checks if the provided mode is one of the valid SSL modes
func isValidMode(mode string) bool {
	return mode == SSLModeRequire ||
		mode == SSLModeDisable ||
		mode == SSLModeVerifyCA ||
		mode == SSLModeVerifyFull
}

// validateFile checks if a file exists and is readable
func validateFile(path string) error {
	if path == "" {
		return errors.New("file path cannot be empty")
	}
	_, err := os.Stat(path)
	return err
}

// Validate returns err if the ssl configuration is invalid
func (sc *SSLConfig) Validate() error {
	if sc == nil {
		return errors.New("'ssl' config is required")
	}

	if sc.Mode == Unknown {
		return errors.New("'ssl.mode' is required parameter")
	}

	if !isValidMode(sc.Mode) {
		return errors.New("'ssl.mode' must be one of: require, disable, verify-ca, verify-full")
	}

	// Skip further validation if SSL is disabled
	if sc.Mode == SSLModeDisable {
		return nil
	}

	// For modes that require certificate validation
	if sc.Mode == SSLModeVerifyCA || sc.Mode == SSLModeVerifyFull {
		if err := validateFile(sc.ServerCA); err != nil {
			return errors.New("'ssl.server_ca' is invalid or inaccessible")
		}

		if err := validateFile(sc.ClientCert); err != nil {
			return errors.New("'ssl.client_cert' is invalid or inaccessible")
		}

		if err := validateFile(sc.ClientKey); err != nil {
			return errors.New("'ssl.client_key' is invalid or inaccessible")
		}
	}

	// For require mode, server CA is recommended for basic verification
	if sc.Mode == SSLModeRequire && sc.ServerCA != "" {
		if err := validateFile(sc.ServerCA); err != nil {
			return errors.New("'ssl.server_ca' is invalid or inaccessible")
		}
	}
```

pkg/waljs/waljs.go
```diff
@@ -0,0 +1,167 @@
...
+
+type Socket struct {
+	pgConn          *pgconn.PgConn
+	clientXLogPos   pglogrepl.LSN
+	idleStartTime   time.Time
+	changeFilter    ChangeFilter
+	RestartLSN      pglogrepl.LSN
+	replicationSlot string
+	initialWaitTime time.Duration
+}
```
shubham19may:
```suggestion
// Socket represents a connection to PostgreSQL's logical replication stream
type Socket struct {
	// pgConn is the underlying PostgreSQL replication connection
	pgConn *pgconn.PgConn
	// clientXLogPos tracks the current position in the Write-Ahead Log (WAL)
	clientXLogPos pglogrepl.LSN
	// idleStartTime tracks when the connection last received data
	idleStartTime time.Time
	// changeFilter filters WAL changes based on configured tables
	changeFilter ChangeFilter
	// RestartLSN is the position from which replication should start/restart
	RestartLSN pglogrepl.LSN
	// replicationSlot is the name of the PostgreSQL replication slot being used
	replicationSlot string
	// initialWaitTime is the duration to wait for initial data before timing out
	initialWaitTime time.Duration
}
```

shubham19may: once check these if correct or not.

pkg/waljs/waljs.go
```diff
@@ -0,0 +1,167 @@
...
+			default:
+				logger.Debugf("Received unhandled message type: %v", copyData.Data[0])
+			}
+		}
+	}
+}
+
+// cleanUpOnFailure drops replication slot and publication if database snapshotting was failed for any reason
+func (s *Socket) Cleanup() {
+	s.pgConn.Close(context.TODO())
```
shubham19may: close the connection?

pkg/waljs/waljs.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("connection config error: %w", err)
+	}
+
+	if config.TLSConfig != nil {
+		cfg.TLSConfig = &tls.Config{InsecureSkipVerify: true}
+	}
+
+	// Establish PostgreSQL connection
+	pgConn, err := pgconn.ConnectConfig(context.Background(), cfg)
```
shubham19may: shouldn't we pass the context of the cdc class?

shubham19may: Also can we not create same method for the any PG connection that is created in the whole code? 

Why separate for waljs.go?

pkg/waljs/waljs.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// Update local pointer and state
+	logger.Debugf("sent standby status message at LSN#%s", s.clientXLogPos.String())
+	return nil
+}
+
+func (s *Socket) StreamMessages(ctx context.Context, callback OnMessage) error {
+	// Start logical replication with wal2json plugin arguments.
+	if err := pglogrepl.StartReplication(
+		context.Background(),
```
shubham19may: can we not use ctx?

pkg/waljs/waljs.go
```diff
@@ -0,0 +1,167 @@
...
+			if time.Since(s.idleStartTime) > s.initialWaitTime {
+				logger.Debug("Idle timeout reached while waiting for new messages")
+				return nil
+			}
+			// Use a context with timeout for receiving a message.
+			msg, err := s.pgConn.ReceiveMessage(ctx)
+			// If the receive timed out, log the idle state and continue waiting.
+			if err != nil {
+				return fmt.Errorf("failed to receive message: %w", err)
+			}
```
shubham19may:
```suggestion
			msg, err := s.pgConn.ReceiveMessage(ctx)
			if err != nil {
                if err == context.DeadlineExceeded {
                    continue  // Just continue on timeout
                }
                return fmt.Errorf("failed to receive message: %w", err)
            }
```

pkg/waljs/waljs.go
```diff
@@ -0,0 +1,167 @@
...
+	q := connURL.Query()
+	q.Set("replication", "database")
+	connURL.RawQuery = q.Encode()
+
+	cfg, err := pgconn.ParseConfig(connURL.String())
+	if err != nil {
+		return nil, fmt.Errorf("connection config error: %w", err)
+	}
+
+	if config.TLSConfig != nil {
```
shubham19may, line 56: it should be == ?

pkg/waljs/waljs.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	// Update local pointer and state
+	logger.Debugf("sent standby status message at LSN#%s", s.clientXLogPos.String())
+	return nil
+}
+
+func (s *Socket) StreamMessages(ctx context.Context, callback OnMessage) error {
+	// Start logical replication with wal2json plugin arguments.
+	if err := pglogrepl.StartReplication(
```
shubham19may, line 110: what if the WAL is too big and it takes a long time to start the sync? What will define the time to wait ?

For example in the Airbyte they have 2400 sec max to wait.

pkg/waljs/filter.go
```diff
@@ -0,0 +1,76 @@
...
+	var changes WALMessage
+	if err := json.NewDecoder(bytes.NewReader(change)).Decode(&changes); err != nil {
+		return fmt.Errorf("cant parse change from database to filter it: %s", err)
+	}
+
+	if len(changes.Change) == 0 {
+		return nil
+	}
+
+	for _, ch := range changes.Change {
```
shubham19may, line 41: what if there are millions of changes? can we parallelise this?

shubham19may: usually if we consider system changes (PG emits many different type of changes)

pkg/waljs/filter.go
```diff
@@ -0,0 +1,76 @@
...
+		stream, exists := c.tables[utils.StreamIdentifier(ch.Table, ch.Schema)]
+		if !exists {
+			continue
+		}
+
+		// builder := array.NewRecordBuilder(memory.DefaultAllocator, schema)
+		changesMap := map[string]any{}
+		if ch.Kind == "delete" {
+			for i, changedValue := range ch.Oldkeys.Keyvalues {
```
shubham19may, line 49: will it always have Oldkeys? what if its null

pkg/waljs/types.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/datazip-inc/olake/typeutils"
+	"github.com/jackc/pglogrepl"
+)
+
+type Config struct {
+	Tables                     *types.Set[protocol.Stream]
+	Connection                 url.URL
+	ReplicationSlotName        string
+	InitialWaitTime            time.Duration
+	SnapshotMemorySafetyFactor float64
```
shubham19may: its never used

drivers/postgres/internal/postgres.go
```diff
@@ -0,0 +1,256 @@
...
+	discoverTime = 5 * time.Minute
+
+	// get all schemas and table
+	getPrivilegedTablesTmpl = `SELECT nspname as table_schema,
+		relname as table_name
+		FROM pg_class c
+		JOIN pg_namespace n ON c.relnamespace = n.oid
+		WHERE has_table_privilege(c.oid, 'SELECT')
+		AND has_schema_privilege(current_user, nspname, 'USAGE')
+		AND relkind IN ('r', 'm', 'v', 't', 'f', 'p')
```
shubham19may: we should not sync views right? 

because they will not generate WALs.

drivers/postgres/internal/postgres.go
```diff
@@ -0,0 +1,260 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to validate config: %s", err)
+	}
+
+	db, err := sqlx.Open("pgx", p.config.Connection.String())
+	if err != nil {
+		return fmt.Errorf("failed to connect database: %s", err)
+	}
+
+	db = db.Unsafe()
```
shubham19may: rename this variable

drivers/postgres/internal/postgres.go
```diff
@@ -0,0 +1,260 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+
+	// force a connection and test that it worked
+	err = db.PingContext(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to ping database: %s", err)
+	}
```
shubham19may: can we not create one single method for check? and call from both main check and setup?

drivers/postgres/internal/postgres.go
```diff
@@ -0,0 +1,260 @@
...
+	p.client = db
+
+	return nil
```
shubham19may: will the connection that we created for testing close?

drivers/postgres/internal/postgres.go
```diff
@@ -0,0 +1,260 @@
...
+		relname as table_name
+		FROM pg_class c
+		JOIN pg_namespace n ON c.relnamespace = n.oid
+		WHERE has_table_privilege(c.oid, 'SELECT')
+		AND has_schema_privilege(current_user, nspname, 'USAGE')
+		AND relkind IN ('r', 'm', 'v', 't', 'f', 'p')
+		AND nspname NOT LIKE 'pg_%'  -- Exclude default system schemas
+		AND nspname != 'information_schema';  -- Exclude information_schema`
+	// get table schema
+	getTableSchemaTmpl = `SELECT column_name, data_type, is_nullable FROM information_schema.columns WHERE table_schema = $1 AND table_name = $2 ORDER BY ordinal_position`
```
shubham19may, line 31: Add TODO: to make these queries version specific, some PG might not support queries exactly.

drivers/postgres/internal/config.go
```diff
@@ -0,0 +1,102 @@
...
+type Config struct {
+	Connection       *url.URL          `json:"-"`
+	Host             string            `json:"host"`
+	Port             int               `json:"port"`
+	Database         string            `json:"database"`
+	Username         string            `json:"username"`
+	Password         string            `json:"password"`
+	JDBCURLParams    map[string]string `json:"jdbc_url_params"`
+	SSLConfiguration *utils.SSLConfig  `json:"ssl"`
+	UpdateMethod     interface{}       `json:"update_method"`
```
shubham19may, line 22: what is this?

drivers/postgres/internal/postgres.go
```diff
@@ -0,0 +1,260 @@
...
+	p.config = &Config{}
+
+	return p.config
+}
+
+func (p *Postgres) Spec() any {
+	return Config{}
+}
+
+func (p *Postgres) Check() error {
```
shubham19may, line 117: check should also check for replicationSlot

drivers/postgres/internal/postgres.go
```diff
@@ -0,0 +1,260 @@
...
+	err = db.PingContext(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to ping database: %s", err)
+	}
+	p.client = db
+
+	return nil
+}
+
+func (p *Postgres) CloseConnection() {
```
shubham19may, line 121: lets close the CDC connection created in Socket?

drivers/postgres/internal/postgres.go
```diff
@@ -0,0 +1,260 @@
...
+		for column, typ := range base.DefaultColumns {
+			stream.UpsertField(column, typ, true)
+		}
+	}
+
+	// TODO: Populate cursor fields
+	if !p.CDCSupport {
+		stream.WithSyncMode(types.FULLREFRESH)
+		// source has cursor fields, hence incremental also supported
+		if stream.SourceDefinedPrimaryKey.Len() > 0 {
```
shubham19may: This is not a cursor right?
AvailableCursorFields this should be the right one?

drivers/postgres/internal/postgres.go
```diff
@@ -0,0 +1,260 @@
...
+	} else {
+		stream.WithSyncMode(types.CDC)
+	}
+
+	// add primary keys for stream
+	for _, column := range primaryKeyOutput {
+		stream.WithPrimaryKey(column.Name)
+	}
+
+	stream.SupportedSyncModes.Insert("cdc")
```
shubham19may: what if its incremental?

drivers/postgres/internal/backfill.go
```diff
@@ -0,0 +1,270 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to query[%s] next chunk end: %s", nextChunkEnd, err)
+	}
+	if chunkEnd == previousChunkEnd {
+		var minVal interface{}
+		minQuery := jdbc.MinQuery(stream, splitColumn, chunkEnd)
+		err := p.client.QueryRow(minQuery).Scan(&minVal)
+		if err != nil {
+			return nil, fmt.Errorf("failed to execute query[%s]: %s", minQuery, err)
+		}
```
shubham19may: why this is needed?

drivers/postgres/internal/backfill.go
```diff
@@ -0,0 +1,270 @@
...
+		var chunkStart interface{}
+		for chunkEnd != nil {
+			if chunkStart == chunkEnd {
+				// no more chunk creation
+				break
+			}
+			splits = append(splits, types.Chunk{Min: chunkStart, Max: chunkEnd})
+			if count%10 == 0 {
+				time.Sleep(1000 * time.Millisecond)
+			}
```
shubham19may: why?

drivers/postgres/internal/backfill.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+		if minValue == maxValue {
+			return []types.Chunk{{Min: minValue, Max: maxValue}}, nil
+		}
+
+		_, contains := utils.ArrayContains(stream.GetStream().SourceDefinedPrimaryKey.Array(), func(element string) bool {
+			return element == splitColumn
+		})
+		if !contains {
+			return nil, fmt.Errorf("provided split column is not a primary key")
```
shubham19may, line 180: why does it matter? if the column is indexed, it should be good to go right?

pkg/jdbc/jdbc.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func MinQuery(stream protocol.Stream, filterColumn string, filterValue interface{}) string {
+	return fmt.Sprintf(`SELECT MIN(%s) FROM "%s"."%s" WHERE %s > %v`, filterColumn, stream.Namespace(), stream.Name(), filterColumn, filterValue)
+}
+
+func BuildSplitScanQuery(stream protocol.Stream, filterColumn string, chunk types.Chunk) string {
+	condition := ""
+	if chunk.Min != nil && chunk.Max != nil {
+		condition = fmt.Sprintf("%s >= %v AND %s <= %v", filterColumn, chunk.Min, filterColumn, chunk.Max)
```
shubham19may, line 44: one of this should be < or >

drivers/postgres/internal/cdc.go
```diff
@@ -0,0 +1,162 @@
...
+		gs.Streams, gs.State.LSN = types.NewSet[string](), currentLSN.String()
+		p.State.SetGlobalState(gs)
+		// reset streams for creating chunks again
+		p.State.ResetStreams()
+	} else {
+		parsed, err := pglogrepl.ParseLSN(gs.State.LSN)
+		if err != nil {
+			return fmt.Errorf("failed to parse stored lsn[%s]: %s", gs.State.LSN, err)
+		}
+		if parsed != currentLSN {
```
shubham19may, line 62: not sure if our assumption here would be correct. basically the parsed LSN will be equal to currentLSN might be wrong in the case of the system moved the LSN forward.

I am saying this because in other scenarios (like debezium), it usually search for LSN rather assuming that whatever currentLSN (restartLSN) will be equal to state LSN.

drivers/postgres/internal/backfill.go
```diff
@@ -0,0 +1,270 @@
...
+			newChunkEnd, err := utils.AddConstantToInterface(chunkEnd, dynamicChunkSize)
+			if err != nil {
+				return nil, fmt.Errorf("failed to split even size chunks: %s", err)
+			}
+			chunkEnd = newChunkEnd
+		}
+		return append(splits, types.Chunk{Min: chunkStart, Max: nil}), nil
+	}
+
+	splitUnevenlySizedChunks := func(stream protocol.Stream, splitColumn string, min, max interface{}) ([]types.Chunk, error) {
```
shubham19may: not making sense of the name. unevently is still evenly as we split it using batch size.

drivers/postgres/internal/cdc.go
```diff
@@ -0,0 +1,162 @@
...
+	for _, s := range streams {
+		if !gs.Streams.Exists(s.ID()) {
+			needsBackfill = append(needsBackfill, s)
+		}
+	}
+	if err = utils.Concurrent(ctx, needsBackfill, len(needsBackfill), func(ctx context.Context, s protocol.Stream, _ int) error {
+		if err := p.backfill(pool, s); err != nil {
+			return fmt.Errorf("failed backfill of stream[%s]: %s", s.ID(), err)
+		}
+		gs.Streams.Insert(s.ID())
```
shubham19may, line 81: why do we need it separately? we can do it in the backfill only right?

drivers/postgres/internal/backfill.go
```diff
@@ -0,0 +1,270 @@
...
+		splitColumn = utils.Ternary(splitColumn == "", "ctid", splitColumn).(string)
+		stmt := jdbc.BuildSplitScanQuery(stream, splitColumn, chunk)
+
+		setter := jdbc.NewReader(backfillCtx, stmt, p.config.BatchSize, func(ctx context.Context, query string, args ...any) (*sql.Rows, error) {
+			return tx.Query(query, args...)
+		})
+		batchStartTime := time.Now()
+		waitChannel := make(chan error, 1)
+		insert, err := pool.NewThread(backfillCtx, stream, protocol.WithErrorChannel(waitChannel))
+		if err != nil {
```
shubham19may, line 63: If this returns, the goroutine sending to waitChannel may be leaked?

drivers/postgres/internal/cdc.go
```diff
@@ -0,0 +1,162 @@
...
+				err = socket.AcknowledgeLSN()
+			}
+		}
+	}()
+
+	// Message processing
+	return socket.StreamMessages(ctx, func(msg waljs.CDCChange) error {
+		pkFields := msg.Stream.GetStream().SourceDefinedPrimaryKey.Array()
+		deleteTS := utils.Ternary(msg.Kind == "delete", msg.Timestamp.UnixMilli(), int64(0)).(int64)
+		return inserters[msg.Stream].Insert(types.CreateRawRecord(
```
shubham19may, line 126: we can save state periodically in order for us to reduce the WAL size?

		// Periodically save state to avoid large reprocessing on crash
		if time.Since(lastStateSave) > stateUpdateInterval {
			gs.State.LSN = socket.ClientXLogPos.String()
			p.State.SetGlobalState(gs)
			if err := socket.AcknowledgeLSN(); err != nil {
				logger.Warnf("Failed to acknowledge LSN: %s", err)
			}
			lastStateSave = time.Now()
		}

drivers/postgres/internal/backfill.go
hash-data: pg_class access is required as we need row count, we can think of it changing in the future if users think it is sensitive.

drivers/postgres/internal/backfill.go
hash-data: The below for loop will cancel out and return 0 chunks

drivers/postgres/internal/backfill.go
hash-data: cool changing it to batch size  (it was hard coded in some other projects)

drivers/postgres/internal/backfill.go
hash-data: yeah duplicate will come, but that will be handled in deduplication part using olake_id, main concern here is not a single record should miss.

drivers/postgres/internal/backfill.go
hash-data: variables are understandable.

utils/utils.go
hash-data: simplified the code for it

drivers/postgres/internal/datatype_conversion.go
hash-data: added todo for it.

drivers/postgres/internal/datatype_conversion.go
hash-data: added one todo for all types

drivers/postgres/internal/config.go
hash-data: options is required will test it

drivers/postgres/internal/config.go
hash-data: added default value

drivers/postgres/internal/config.go
hash-data: we can not do that key required is options

utils/ssl.go
hash-data: added default config

utils/ssl.go
hash-data: adding a todo, for making it for other certifications like tls

pkg/waljs/waljs.go
hash-data: there is no context deadline

pkg/waljs/waljs.go
hash-data: added a todo for it to make it proper with testing

pkg/waljs/waljs.go
hash-data: The initial wait time is that only.

pkg/waljs/filter.go
hash-data: decode will fail

drivers/postgres/internal/postgres.go
hash-data: removed, by the way it is for discover only

drivers/postgres/internal/postgres.go
hash-data: every driver has closeConnection function that close the connection when exited

drivers/postgres/internal/config.go
hash-data: Updated Method checks if it is cdc, or standard. If cdc then there must be initial_wait_time and replication slot

drivers/postgres/internal/postgres.go
hash-data: socket.Cleanup will close the connection

hash-data: will remove redundant functions

hash-data: will change it in refactor

drivers/postgres/internal/postgres.go
hash-data: yes already added a todo for available cursor fields, and this need to be taken care when we start work on incremental

pkg/jdbc/jdbc.go
hash-data: yes it should be < but i have put it just to be on safe side (no data loss)

drivers/postgres/internal/backfill.go
hash-data: missed one statement to make chunkEnd = minVal, basically it is min value that is greater than the chunk end

drivers/postgres/internal/backfill.go
hash-data: flink uses it, so think might be with purpose (will remove it not required)

drivers/postgres/internal/backfill.go
hash-data: currently we are checking the indexation only if it is present in primary key array or not.

drivers/postgres/internal/backfill.go
hash-data: ranamed it

drivers/postgres/internal/cdc.go
hash-data: this needs to be stored only when backfil completed and global state is part of CDC only

drivers/postgres/internal/backfill.go
hash-data: Go's garbage collector will clean up the unused channel once it goes out of scope.

drivers/postgres/internal/cdc.go
hash-data: i am thinking it based on batch size instead of time based which can be added as a feature later on.

drivers/postgres/internal/cdc.go
hash-data: let us discuss it, i may need to research on it little bit.

pkg/waljs/filter.go
hash-data: we can think in this direction, but is it required in v0 ?

pkg/waljs/waljs.go
hash-data: make last one constant

drivers/postgres/internal/backfill.go
shubham19may: Add a TODO, because many time we will only get read access to that specific data base where the table is.

drivers/postgres/internal/backfill.go
hash-data: But if 0 is returned even in the case that there is some data, that case also handled

drivers/postgres/internal/backfill.go
hash-data: added a todo for intersection removal

drivers/postgres/internal/config.go
hash-data: ok, changing it, firstly I thought it is a  comment update

drivers/postgres/internal/postgres.go
hash-data: created it

drivers/postgres/internal/postgres.go
shubham19may: but the point is, there is a standalone check function calls as well. this will not be closed.

drivers/postgres/internal/backfill.go
hash-data: we are also using same table in getPrivilegedTablesTmpl and pg_class. So it is required.

drivers/postgres/internal/cdc.go
```diff
@@ -113,7 +113,8 @@ func (p *Postgres) RunChangeStream(pool *protocol.WriterPool, streams ...protoco
 				gs.State.LSN = socket.ClientXLogPos.String()
 				p.State.SetGlobalState(gs)
 				// mark lsn for wal logs drop
-				err = socket.AcknowledgeLSN()
+				// TODO: acknowledge message should also be send w.r.t batch size
```
shubham19may: // TODO: acknowledge message should be called every 10k records or so to reduce the size of the WAL.

drivers/postgres/internal/backfill.go
```diff
@@ -30,7 +30,8 @@ func (p *Postgres) backfill(pool *protocol.WriterPool, stream protocol.Stream) e
 	var splitChunks []types.Chunk
 	if stateChunks == nil {
 		// check for data distribution
-		splitChunks, err = p.splitTableIntoChunks(stream, approxRowCount)
+		// TODO: remove chunk intersections
```
shubham19may: // TODO: remove chunk intersections where chunks can be {0, 100} {100, 200}. Need to {0, 99} {100, 200}

drivers/postgres/internal/postgres.go
```diff
@@ -16,15 +16,15 @@ import (
 
 const (
 	discoverTime = 5 * time.Minute
-
+	// TODO: make these queries version specific
```
shubham19may: TODO: make these queries Postgres version specific

drivers/postgres/internal/postgres.go
hash-data: only one p.client is there which get closed

drivers/postgres/internal/postgres.go
hash-data: calling setup only, in check

shubham19may (approved):

PR #132, opened 2025-03-07
Chore: Global State Revamp

hash-data:
# Description
Revamping global state for ease of access at multiple places

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


types/state.go
```diff
@@ -29,7 +30,7 @@ type State struct {
 	*sync.Mutex `json:"-"`
 	Type        StateType      `json:"type"`
 	Global      any            `json:"global,omitempty"`
-	Streams     []*StreamState `json:"streams,omitempty"`
+	Streams     []*StreamState `json:"streams,omitempty"` // TODO: make it set
```
shubham19may: why is this named as Streams? it should be StreamStates

types/state.go
```diff
@@ -41,25 +42,113 @@ func (s *State) SetType(typ StateType) {
...
+	} else {
+		newStream := s.InitialState(stream)
+		newStream.State.Store(key, value)
+		newStream.HoldsValue.Store(true)
+		s.Streams = append(s.Streams, newStream)
+	}
+	s.LogState()
+}
+
+func (s *State) GetCursor(stream *ConfiguredStream, key string) any {
```
shubham19may, line 80: which cursor is this? CDC or incremental?

shubham19may: should we also lock this? Any point of concurrency a streams state is modified and returned at the same time?

hash-data: done

shubham19may (approved):

PR #174, opened 2025-03-25
feat: Filter based on _id in mongodb backfill

mrmagicpotato007:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/drivers/mongodb/internal/backfill.go
+++ b/drivers/mongodb/internal/backfill.go
@@ -165,6 +165,7 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 		if err != nil {
 			return nil, fmt.Errorf("failed to get chunk boundaries: %s", err)
 		}
+
 		var chunks []types.Chunk
 		for i := 0; i < len(boundaries)-1; i++ {
 			chunks = append(chunks, types.Chunk{
@@ -208,7 +209,6 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 		if err := cursor.All(ctx, &buckets); err != nil {
 			return nil, fmt.Errorf("failed to decode bucketAuto results: %s", err)
 		}
-
 		var chunks []types.Chunk
 		for _, bucket := range buckets {
 			chunks = append(chunks, types.Chunk{
@@ -238,6 +238,7 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 		if timeDiff < 1 {
 			timeDiff = 1
 		}
+
 		// for every 6hr difference ideal density is 10 Seconds
 		density := time.Duration(timeDiff) * (10 * time.Second)
 		start := first
@@ -263,18 +264,44 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 		return chunks, nil
 	}
 
+	var chunks []types.Chunk
+	var err error
 	switch m.config.PartitionStrategy {
 	case "timestamp":
-		return timestampStrategy()
+		chunks, err = timestampStrategy()
 	default:
-		chunks, err := splitVectorStrategy()
-		// check if authorization error occurs
+		chunks, err = splitVectorStrategy()
 		if err != nil && strings.Contains(err.Error(), "not authorized") {
 			logger.Warnf("failed to get chunks via split vector strategy: %s", err)
-			return bucketAutoStrategy()
+			chunks, err = bucketAutoStrategy()
 		}
-		return chunks, err
 	}
+	if err != nil {
+		return nil, err
+	}
+
+	minThresholdID := stream.GetStream().MinThresholdID
+	var parsedMinThresholdID primitive.ObjectID
+	if minThresholdID == "" {
+		parsedMinThresholdID = primitive.NilObjectID
+	} else {
+		parsedMinThresholdID, err = primitive.ObjectIDFromHex(minThresholdID)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	//filter based on threshold
+	filteredChunks := []types.Chunk{}
+	for _, chunk := range chunks {
+		if parsedMinThresholdID == primitive.NilObjectID || chunk.Max == nil || chunk.Max.(*primitive.ObjectID).Timestamp().After(parsedMinThresholdID.Timestamp()) {
+			filteredChunks = append(filteredChunks, chunk)
+		} else {
+			logger.Infof("Skipping chunk as max object ID [%v] does not exceed the minimum threshold for stream [%s]", chunk.Max, stream.GetStream().Name)
+		}
+	}
+
+	return filteredChunks, nil
 }
 func (m *Mongo) totalCountInCollection(ctx context.Context, collection *mongo.Collection) (int64, error) {
 	var countResult bson.M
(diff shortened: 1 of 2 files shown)


drivers/mongodb/internal/backfill.go
```diff
@@ -123,7 +123,30 @@ func (m *Mongo) backfill(stream protocol.Stream, pool *protocol.WriterPool) erro
...
+
+	isValidChunk := func(maxObjectID *primitive.ObjectID, minThresholdId *primitive.ObjectID) bool {
+		if minThresholdId != &primitive.NilObjectID && (maxObjectID.Timestamp().Before(minThresholdId.Timestamp()) || maxObjectID.Timestamp().Equal(minThresholdId.Timestamp())) {
+			logger.Infof("Skipping chunk as max object ID [%s] does not exceed the minimum threshold for stream [%s]", maxObjectID.String(), stream.GetStream().Name)
+			return false
+		}
+		return true
+	}
+
+	getMinimumThresholdID := func() (primitive.ObjectID, error) {
```
shubham19may: No need to create it as a function as its just used at one place.

drivers/mongodb/internal/backfill.go
```diff
@@ -165,12 +188,16 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
 		}
+
 		var chunks []types.Chunk
 		for i := 0; i < len(boundaries)-1; i++ {
-			chunks = append(chunks, types.Chunk{
-				Min: &boundaries[i],
-				Max: &boundaries[i+1],
-			})
+			maxObjectID := boundaries[i+1]
+			if isValidChunk(maxObjectID, &minThresholdID) {
```
shubham19may: can't we do it at the end only once rather than repeating the code multiple times?

drivers/mongodb/internal/backfill.go
```diff
@@ -123,7 +123,30 @@ func (m *Mongo) backfill(stream protocol.Stream, pool *protocol.WriterPool) erro
 }
 
 func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, stream protocol.Stream) ([]types.Chunk, error) {
+
+	isValidChunk := func(maxObjectID *primitive.ObjectID, minThresholdId *primitive.ObjectID) bool {
```
shubham19may: If this is not going to be used multiple places, lets use it directly the logic rather than function.

drivers/mongodb/internal/backfill.go
mrmagicpotato007: after selecting the strategy is what you are suggesting ? then we need re run a loop on all of our chunks to filter.

```
var chunks []types.Chunk
    var err error
    switch m.config.PartitionStrategy {
    case "timestamp":
        chunks, err = timestampStrategy()
    default:
        chunks, err = splitVectorStrategy()
        if err != nil && strings.Contains(err.Error(), "not authorized") {
            logger.Warnf("failed to get chunks via split vector strategy: %s", err)
            chunks, err = bucketAutoStrategy()
        }
    }
    if err != nil {
        return nil, err
    }

    // Filter chunks
    return filterChunks(chunks, &parsedMinThresholdID), nil
```

drivers/mongodb/internal/backfill.go
mrmagicpotato007: makes sense will remove this

drivers/mongodb/internal/backfill.go
shubham19may: @mrmagicpotato007 this is logically ok right? because anyways we are fetching all the chunks from DB.

mrmagicpotato007: yes logically okay, i wanted to avoid loop again on the chunks array to filter out. so decided to do the filtering during every split strategy.

mrmagicpotato007: hi shubham updated this pr.

hash-data:
@mrmagicpotato007  merge conflicts are there, can you resolve them.

drivers/mongodb/internal/backfill.go
```diff
@@ -263,18 +264,44 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
+	filteredChunks := []types.Chunk{}
+	for _, chunk := range chunks {
+		if parsedMinThresholdID == primitive.NilObjectID || chunk.Max == nil || chunk.Max.(*primitive.ObjectID).Timestamp().After(parsedMinThresholdID.Timestamp()) {
+			filteredChunks = append(filteredChunks, chunk)
+		} else {
+			logger.Infof("Skipping chunk as max object ID [%v] does not exceed the minimum threshold for stream [%s]", chunk.Max, stream.GetStream().Name)
+		}
+	}
+
+	return filteredChunks, nil
```
hash-data, line 304: One idea I have here is to take a filter as a selected streams parameter. The filter will contain something like: updated_at>12-05-2025 or _id>some_hex_id, kind of, and that filter will be used for creating the chunks. How does it sound, @rkhameshra @shubham19may?

shubham19may: Chunking usually happens on the _id column right? usually _id is by default indexed by Mongo other columns are not. Thats why we calculate chunks basis on this.

updated_at column can be a good post-filter (SMT) thing but not a pre-filter. So technically both are processed differently.

hash-data: so now are we just going with _id, as this need to be stream level property not connector level?

shubham19may: Overall as a product feature i am good with this. (We can take these filters right now and depending on the _id filter value we can skip the data reading in chunk)

drivers/mongodb/internal/backfill.go
```diff
@@ -263,18 +264,44 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
+
+	//filter based on threshold
+	filteredChunks := []types.Chunk{}
+	for _, chunk := range chunks {
+		if parsedMinThresholdID == primitive.NilObjectID || chunk.Max == nil || chunk.Max.(*primitive.ObjectID).Timestamp().After(parsedMinThresholdID.Timestamp()) {
+			filteredChunks = append(filteredChunks, chunk)
+		} else {
+			logger.Infof("Skipping chunk as max object ID [%v] does not exceed the minimum threshold for stream [%s]", chunk.Max, stream.GetStream().Name)
+		}
+	}
```
hash-data, line 302: instead of taking out chunks and then filtering it can we just get chunks after the particular id?

types/stream.go
```diff
@@ -27,7 +27,8 @@ type Stream struct {
 	AdditionalProperties string `json:"additional_properties,omitempty"`
 	// Renderable JSON Schema for additional properties supported by respective driver for individual stream
 	AdditionalPropertiesSchema schema.JSONSchema `json:"additional_properties_schema,omitempty"`
-	SyncMode                   SyncMode          `json:"sync_mode,omitempty"` // Mode being used for syncing data
+	SyncMode                   SyncMode          `json:"sync_mode,omitempty"`        // Mode being used for syncing data
+	MinThresholdID             string            `json:"min_threshold_id,omitempty"` // Minimum object ID required to load
```
hash-data, line 31: also it need to be in selected streams option

hash-data:
it is being implemented here: https://github.com/datazip-inc/olake/pull/305
closing it

Issue #120, opened 2025-03-04
[Chore] Configuration for Max chunk size in Split Vector Strategy of mongodb


zriyanshdz:
### Problem

`{Key: "maxChunkSize", Value: 1024},`

Right now the `maxChunkSize` is hardcoded. For data <1GB, it only makes 1 running thread even if the `max_threads` value is passed as 10.

### Solution

We need to take `max_chunk_size` value from the user in the config.json file

```
{
  "hosts": [
    "host1:27017",
    "host2:27017",
    "host3:27017"
  ],
  "username": "test",
  "password": "test",
  "authdb": "admin",
  "replica-set": "rs0",
  "read-preference": "secondaryPreferred",
  "srv": true,
  "server-ram": 16,
  "database": "database",
  "max_threads": 50,
  "default_mode": "cdc",
  "backoff_retry_count": 2,
  "partition_strategy":"",
   "max_chunk_size": 1024 // or any other number, 1024 here is in megabytes (MB).
}
```

## How to test

with `max_chunk_size` set, you can change the values of it and sync data, if the total data to sync is < max_chunk_size value, there will be just 1 chunk, else you can set the value to `1` and see multiple chunks of 1MB getting created.

mrmagicpotato007:
@zriyanshdz i can take this

k0msenapati:
I would like to solve this @zriyanshdz

zriyanshdz:
cc - @shubham19may @hash-data

PR #219, opened 2025-04-09
Variable MaxChunkSize Update

ImDoubD-datazip:
https://github.com/datazip-inc/olake/issues/120

This issue has been resolved from my side. Please do review.
User can now provide variable maxchunksize in the config.json file of MongoDB.
If maxchunksize < 1 or greater than 1024, then by default the chunk size will be set to 1024.


hash-data:
not required as of now.

PR #225, opened 2025-04-10
fix: refactored GetKeysHash to generate olakeID when primary key is absent

### LINKED ISSUE #192: urgent: create hash from all keys if primary key not present
## Problem
Currently, olake_id is being generated using the primary key fields of a record. However, in cases where the primary key is not present, this leads to all records getting the same olake_id, which is clearly undesirable and breaks uniqueness expectations.

## Proposed Solution:
When primary keys are not defined, generate olake_id by hashing all values in the record. This ensures a more consistent and unique identifier per row, even in the absence of a primary key.

This functionality is already available in the utility function utils.GetKeysHash(msg.Data, pkFields...) supports hashing record values. It simply needs to be used with the appropriate data when pkFields are empty or undefined.

## Action Items:

- Check if pkFields is empty

- If so, pass all record values to GetKeysHash

- Otherwise, continue using the primary key-based hashing as usual

https://github.com/datazip-inc/olake/blob/48ef2b72acbf29a3f481d8d76188b85e7661d454/drivers/mysql/internal/backfill.go#L95-L100

---

###### This issue was generated by [todo-issue](https://github.com/DerJuulsn/todo-issue) based on a `TODO` comment in 48ef2b72acbf29a3f481d8d76188b85e7661d454. It&#x27;s been assigned to @hash-data because they committed the code.

vaibhav-datazip:
# Description

If no primary key is present, the entire record is hashed to generate the olakeID. This ensures that all records are uniquely identified.

Fixes #192 

## Type of change


drivers/mysql/internal/backfill.go
```diff
@@ -94,7 +94,13 @@ func (m *MySQL) backfill(pool *protocol.WriterPool, stream protocol.Stream) erro
 				}
 				// TODO : create hash from all keys if primary key not present
 				//genrate olake id
-				olakeID := utils.GetKeysHash(record, stream.GetStream().SourceDefinedPrimaryKey.Array()...)
+				var olakeID string
+				// if no primary key is defined, generate a hash from all values in record
+				if pkColumns.Len() == 0 {
+					olakeID = utils.GetHash(record)
+				} else {
+					olakeID = utils.GetKeysHash(record, pkColumns.Array()...)
```
hash-data: We have to do the same for all drivers

hash-data: also check if we can implement this thing inside get key hash

hash-data (approved):

PR #227, opened 2025-04-13
feat: Added partitioning logic for iceberg writer

shubham19may:
# Description

Partitioning feature for Iceberg writer. We have implemented partitioning as well as the transforms that Iceberg offers. 
What is it : https://iceberg.apache.org/docs/latest/partitioning/#what-is-partitioning
What is transform : https://iceberg.apache.org/spec/#partitioning

Fixes # (issue)

## Type of change

- [X] This change requires a documentation update

# How Has This Been Tested?

- [X] Did partitioning with now() which means the time of ingestion and hourly transform
- [X] Normalisation true, did partitioning with identity transform over a normal column

# Screenshots or Recordings
Screenshot 2025-04-13 at 6 50 02 PM


drivers/base/driver.go
```diff
@@ -19,9 +19,10 @@ type Driver struct {
 }
 
 var DefaultColumns = map[string]types.DataType{
-	constants.CDCDeletedAt:   types.Timestamp,
 	constants.OlakeID:        types.String,
 	constants.OlakeTimestamp: types.Int64,
+	constants.OpType:         types.String,
+	constants.CdcTimestamp:   types.Int64,
```
hash-data, line 25: Is it going to be int column?

types/data_types.go
```diff
@@ -52,13 +52,13 @@ func (r *RawRecord) ToDebeziumFormat(db string, stream string, normalization boo
...
 	// Add olake_id to payload
-	payload["olake_id"] = r.OlakeID
+	payload[constants.OlakeID] = r.OlakeID
 
 	// Handle data based on normalization flag
 	if normalization {
 		// Copy the data fields but remove olake_id if present
 		for key, value := range r.Data {
-			if key != "olake_id" {
+			if key != constants.OlakeID {
```
hash-data: condition not required

types/data_types.go
```diff
@@ -116,14 +116,14 @@ func (r *RawRecord) createDebeziumSchema(db string, stream string, normalization
...
-		"field":    "olake_id",
+		"field":    constants.OlakeID,
 	})
 
 	if normalization {
 		// Add individual data fields
 		for key, value := range r.Data {
 			// Skip olake_id for normalized mode
-			if key == "olake_id" {
+			if key == constants.OlakeID {
```
hash-data: olake id only present in rawrecord not in data so we can remove this

writers/iceberg/debezium-server-iceberg-sink/src/main/java/io/debezium/server/iceberg/IcebergUtil.java
```diff
@@ -118,6 +120,11 @@ public static Table createIcebergTable(Catalog icebergCatalog, TableIdentifier t
 
   public static Table createIcebergTable(Catalog icebergCatalog, TableIdentifier tableIdentifier,
                                          Schema schema, String writeFormat) {
+    return createIcebergTable(icebergCatalog, tableIdentifier, schema, writeFormat, Collections.emptyMap());
+  }
```
hash-data, line 124: is it required to create new function here?

writers/iceberg/iceberg_utils.go
```diff
@@ -156,6 +166,46 @@ func findAvailablePort(serverHost string) (int, error) {
...
+
+		colName := strings.TrimSpace(strings.Trim(match[1], `'"`))
+		defaultValue := strings.TrimSpace(strings.Trim(match[2], `'"`))
+		transform := strings.TrimSpace(strings.Trim(match[3], `'"`))
+
+		// Special handling for now() function
+		if colName == "now()" {
+			// Create a special field for timestamps in Iceberg
+			// We'll use _olake_timestamp which is present in all records
+			field := constants.OlakeTimestamp
```
hash-data: let us make similar in parquet as well or add todo.

writers/iceberg/iceberg_utils.go
```diff
@@ -643,3 +694,31 @@ func sendRecords(records []string, client proto.RecordIngestServiceClient) error
...
+// Returns an error if a required field is missing with no default value
+func applyPartitionDefaults(record *types.RawRecord, partitionInfo map[string]*PartitionField) error {
+	if len(partitionInfo) == 0 {
+		return nil
+	}
+	for field, info := range partitionInfo {
+		// Skip validation for internal fields like _olake_timestamp which are automatically added
+		if strings.HasPrefix(field, "_") {
+			continue
+		}
```
hash-data: not a good idea as client database can also have similar columns

hash-data: we can create an array of system defined columns which i guess we already have map, in which we can check.

writers/iceberg/iceberg_utils.go
```diff
@@ -643,3 +694,31 @@ func sendRecords(records []string, client proto.RecordIngestServiceClient) error
...
+
+		_, exists := record.Data[field]
+		if !exists {
+			// If field doesn't exist, check if we have a default value
+			if info.DefaultValue != "" {
+				// Add the default value to the record
+				record.Data[field] = info.DefaultValue
```
hash-data: Currently, we are updating data for null values with a default value, and on discussion, we are on the same page. Now required decision on else case, whether to fail or not, if it isa  null column and the default value is not provided.

drivers/base/driver.go
shubham19may: No it should be timestamp

shubham19may: It should be handled in the java code as it converts 
```
case "int64": // long 8 bytes
          if (TS_MS_FIELDS.contains(fieldName)) {
            return Types.TimestampType.withZone();
          } else {
            return Types.LongType.get();
          }
```

types/data_types.go
shubham19may: It is required as flatten function also send olake_id right? but we are already adding it here by default separately.

writers/iceberg/debezium-server-iceberg-sink/src/main/java/io/debezium/server/iceberg/IcebergUtil.java
shubham19may: yes because its taking partitionTransforms at the end right? one extra arguement.

drivers/base/driver.go
hash-data: as discussed it must be, timestamp we can add a todo to make it later

drivers/mongodb/internal/cdc.go
```diff
@@ -103,11 +105,18 @@ func (m *Mongo) changeStreamSync(stream protocol.Stream, pool *protocol.WriterPo
 		handleObjectID(record.FullDocument)
 		opType := utils.Ternary(record.OperationType == "update", "u", utils.Ternary(record.OperationType == "delete", "d", "c")).(string)
 
+		var ts time.Time
+		if record.WallTime != 0 {
+			ts = record.WallTime.Time() // millisecond precision
+		} else {
+			ts = time.UnixMilli(int64(record.ClusterTime.T)*1000 + int64(record.ClusterTime.I)) // seconds only
+		}
```
hash-data: use ternary operator

types/data_types.go
```diff
@@ -164,24 +163,24 @@ func (r *RawRecord) createDebeziumSchema(db string, stream string, normalization
...
 		},
 		{
-			"type":     "string",
+			"type":     "int64",
 			"optional": true,
-			"field":    "__db",
+			"field":    constants.CdcTimestamp,
 		},
 		{
 			"type":     "int64",
```
hash-data: here as well

types/data_types.go
```diff
@@ -164,24 +163,24 @@ func (r *RawRecord) createDebeziumSchema(db string, stream string, normalization
...
 		},
 		{
 			"type":     "string",
 			"optional": true,
-			"field":    "__op",
+			"field":    constants.DBName,
 		},
 		{
-			"type":     "string",
+			"type":     "int64",
```
hash-data: for consistency we can make it "timestamptz"

writers/iceberg/README.md
```diff
@@ -143,3 +143,80 @@ Create a json for writer config (writer.json)
...
+/{now(), day}
+```
+
+### Supported Transforms
+
+The Iceberg writer supports the following transforms:
+- `identity`: Use the raw value (good for categorical data)
+- `year`, `month`, `day`, `hour`: Time-based transforms (good for timestamp columns)
+- `bucket[N]`: Hash the value into N buckets (for high-cardinality fields)
+- `truncate[N]`: Truncate the string to N characters (for string fields)
```
hash-data, line 191: can we add link of iceberg page here?

protocol/writers.go
```diff
@@ -187,7 +180,7 @@ func (w *WriterPool) NewThread(parent context.Context, stream Stream, options ..
 							return nil
 						}
 						// add insert time
-						record.OlakeTimestamp = time.Now().UTC().UnixMilli()
+						record.OlakeTimestamp = time.Now().UTC()
```
hash-data, line 183: we need to capture error on thread.Close()

writers/iceberg/iceberg_utils.go
shubham19may: Done

drivers/base/driver.go
shubham19may: Resolving as per our discussion.

protocol/writers.go
```diff
@@ -131,19 +130,13 @@ func (w *WriterPool) NewThread(parent context.Context, stream Stream, options ..
...
-		flattenedData[constants.CDCTimestamp] = rawRecord.CdcTimestamp
-
 		// schema evolution
 		change, typeChange, mutations := fields.Process(flattenedData)
 		if change || typeChange {
 			w.tmu.Lock()
 			stream.Schema().Override(fields.ToProperties()) // update the schema in Stream
 			w.tmu.Unlock()
-			err := thread.EvolveSchema(change, typeChange, mutations.ToProperties(), flattenedData)
+			err := thread.EvolveSchema(change, typeChange, mutations.ToProperties(), flattenedData, rawRecord.OlakeTimestamp)
```
hash-data, line 139: instead of taking type.Record let us take raw record for ease.

writers/iceberg/debezium-server-iceberg-sink/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -64,15 +64,15 @@ public IcebergTableOperator(boolean upsert_records) {
...
-    cdcOpField = "__op";
-    cdcSourceTsMsField = "__source_ts_ms";
+    cdcOpField = "_op_type";
+    cdcSourceTsMsField = "_cdc_timestamp";
   }
 
   static final ImmutableMap<Operation, Integer> CDC_OPERATION_PRIORITY = ImmutableMap.of(Operation.INSERT, 1, Operation.READ, 2, Operation.UPDATE, 3, Operation.DELETE, 4);
   private static final Logger LOGGER = LoggerFactory.getLogger(IcebergTableOperator.class);
-  @ConfigProperty(name = "debezium.sink.iceberg.upsert-dedup-column", defaultValue = "__source_ts_ms")
+  @ConfigProperty(name = "debezium.sink.iceberg.upsert-dedup-column", defaultValue = "_cdc_timestamp")
```
hash-data, line 73: is it used for deduplication ?

writers/iceberg/iceberg.go
```diff
@@ -41,12 +43,39 @@ func (i *Iceberg) Setup(stream protocol.Stream, options *protocol.Options) error
...
 }
 
 func (i *Iceberg) Write(_ context.Context, record types.RawRecord) error {
+	// Set "null" for missing or nil partition fields
+	for field := range i.partitionInfo {
+		// Skip internal Olake fields which are automatically added
+		if field == constants.OlakeID || field == constants.OlakeTimestamp ||
+			field == constants.OpType || field == constants.CdcTimestamp ||
+			field == constants.DBName {
+			continue
```
hash-data: instead of it creating an array or already used map is beneficial as we dont have to maintain it on multiple places, for now we can add todo as well else can fix it.

writers/iceberg/iceberg.go
```diff
@@ -41,12 +43,39 @@ func (i *Iceberg) Setup(stream protocol.Stream, options *protocol.Options) error
...
+		}
+
+		_, exists := record.Data[field]
+		if !exists {
+			record.Data[field] = nil
```
hash-data: is it required to make it nil?

writers/iceberg/iceberg.go
```diff
@@ -83,19 +112,27 @@ func (i *Iceberg) Close() error {
...
+	i.partitionInfo = make(map[string]string)
 
 	// Create a temporary setup for checking
 	err := i.SetupIcebergClient(false)
 	if err != nil {
-		// Restore original stream before returning
+		// Restore original stream and partition info before returning
 		i.stream = originalStream
+		i.partitionInfo = originalPartitionInfo
 		return fmt.Errorf("failed to setup iceberg: %v", err)
```
hash-data, line 111: we can avoid this

writers/iceberg/iceberg_utils.go
```diff
@@ -156,6 +159,39 @@ func findAvailablePort(serverHost string) (int, error) {
 	return 0, fmt.Errorf("no available ports found between 50051 and 59051")
 }
 
+// parsePartitionRegex parses the partition regex and populates the partitionInfo map
+func (i *Iceberg) parsePartitionRegex(pattern string) error {
+	// path pattern example: /{col_name, partition_transform}/{col_name, partition_transform}
+	// This strictly identifies column name and partition transform entries
+	patternRegex := regexp.MustCompile(`\{([^,]+),\s*([^}]+)\}`)
```
hash-data, line 166: can we use common regex for both parquet and iceber

writers/iceberg/iceberg_utils.go
```diff
@@ -156,6 +159,39 @@ func findAvailablePort(serverHost string) (int, error) {
...
+	for _, match := range matches {
+		if len(match) < 3 {
+			continue // We need at least 3 matches: full match, column name, transform
+		}
+
+		colName := strings.TrimSpace(strings.Trim(match[1], `'"`))
+		transform := strings.TrimSpace(strings.Trim(match[2], `'"`))
+
+		// Special handling for now() function
+		if colName == "now()" {
```
hash-data: we can simply say replace now() with constants.OlakeTimestamp, other things as it is.

protocol/writers.go
shubham19may: adding a TODO, not able to understand where to handle it as i have not coded this part.

protocol/writers.go
shubham19may: This means we will process the data twice. 
as we have to execute the same logic again inside the EvolveSchema : 

```
flattenedData, err := thread.Flattener()(rawRecord.Data) // flatten the record first
		if err != nil {
			return nil, err
		}

		// schema evolution
		change, typeChange, mutations := fields.Process(flattenedData)
```

writers/iceberg/debezium-server-iceberg-sink/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
shubham19may: Primarily inside a batch yes. But when record is updated between two different sync then its not used

writers/iceberg/iceberg_utils.go
shubham19may: No better to keep them separate as both has different outputs. 

parquet = /xxx={col,default,format}/
iceberg = /{col,format}/

iceberg can not have middle value and it should fail if given.

writers/iceberg/iceberg.go
shubham19may: great suggestion! its not needed, removed.

writers/iceberg/iceberg.go
shubham19may: how?

protocol/writers.go
hash-data: as if batch. Insert (which happens on close), then the batch must fail. It was not required in Parquet, but required in Iceberg writer, so let us add it, else we will miss records.
here 
```
thread.Close() // close it after closing inserts
```

writers/iceberg/iceberg.go
hash-data: returning error only so olake instance will kill. no need to rechange back

writers/iceberg/iceberg_utils.go
hash-data: i checked both parse the same thing, and both don't have restrictions

hash-data (approved):
thread.close issue need to be worked on

PR #238, opened 2025-04-21
fix: datatypes from postgres 

ImDoubD-datazip:
Converter moved to sql.go. CDC working now.


utils/sql.go
```diff
@@ -1,13 +1,118 @@
...
+		return value, nil
+	}
+
+	// Remove any length specifiers from types (e.g., varchar(50) -> varchar)
+	baseType := strings.Split(columnType, "(")[0]
+	baseType = strings.ToLower(strings.TrimSpace(baseType))
+
+	switch baseType {
+	case "date", "timestamp", "timestamptz", "timestamp without time zone", "timestamp with time zone", "time", "timetz", "time without time zone", "time with time zone":
+		return value, nil
```
hash-data, line 27: we need to add cases here as well, example : I created a column with micro and it is getting as string but we need to typecast it to time.Time. Once check this example

hash-data:
not required closing it

PR #241, opened 2025-04-21
chore: type convertor for postgres

hash-data:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


drivers/logs/sync_2025-04-21_13-49-05/olake.log
shubham19may, line 1: Remove

go.mod
```diff
@@ -85,6 +85,7 @@ require (
 	github.com/mitchellh/hashstructure v1.1.0
 	github.com/oklog/ulid v1.3.1
 	github.com/spf13/pflag v1.0.6 // indirect
+	go.starlark.net v0.0.0-20250417143717-f57e51f710eb
```
shubham19may: need for this?

typeutils/reformat.go
```diff
@@ -245,6 +248,36 @@ func ReformatInt64(v any) (int64, error) {
...
+		return v, nil
+	case uint:
+		//nolint:gosec,G115
+		return int32(v), nil
+	case uint8:
+		return int32(v), nil
+	case uint16:
+		return int32(v), nil
+	case uint64:
+		//nolint:gosec,G115
```
shubham19may, line 281: ? should't this be allowed?

typeutils/reformat.go
```diff
@@ -245,6 +248,36 @@ func ReformatInt64(v any) (int64, error) {
...
+		return int32(v), nil
+	case bool:
+		// Matches ReformatInt64: always returns 1 for bool
+		return 1, nil
+	case *any:
```
shubham19may, line 294: where is string?

typeutils/reformat.go
```diff
@@ -284,6 +317,45 @@ func ReformatFloat64(v interface{}) (interface{}, error) {
...
+		return float32(v), nil
+	case uint8:
+		return float32(v), nil
+	case uint16:
+		return float32(v), nil
+	case uint32:
+		return float32(v), nil
+	case uint64:
+		return float32(v), nil
+	case bool:
```
shubham19may, line 369: could be 0.0 as well right? why only 1.0?

typeutils/reformat.go
hash-data: lint will fail on it

typeutils/reformat.go
hash-data: string to int can't be converted

typeutils/reformat.go
hash-data: if there is some value which is not null, that means it is true only right, i guess duke followed float64 convertor function

PR #248, opened 2025-04-24
fix(mysql): correct datatype handling

ImDoubD-datazip:
# Description

Similar to PostgreSQL, MySQL datatype fixes has been done. MySQL doesn't provides column types directly, instead it provides it in byte format. Conversion from byte format to base string name format was done and the byte data of the types was present in MySQL package in const.go file. Extra cases added in data_conversion.go file in mysql driver.

Fixes #238

## Type of change

- [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?
Testing was done using testingTable present in MySQL's staging database. Columns tested included bigint unsigned, int, decimal, varchar, timestamp, tinyint, date, datetime.

DataTypes (mysql_tests table):

Screenshot 2025-04-25 at 4 43 54 PM

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -33,7 +33,7 @@ jobs:
         run: |
           docker compose -f ./drivers/mysql/docker-compose.yml up -d
           docker compose -f ./drivers/postgres/docker-compose.yml up -d
-          docker compose -f ./writers/iceberg/local-test/docker-compose.yml up -d
+          docker compose -f ./writers/iceberg/local-test/docker-compose.yml up minio mc postgres spark-iceberg -d
 
       - name: Wait for MySQL
         uses: nick-fields/retry@v2
--- a/writers/iceberg/debezium-server-iceberg-sink/pom.xml
+++ b/writers/iceberg/debezium-server-iceberg-sink/pom.xml
@@ -56,6 +56,7 @@
         <version.logback>1.5.13</version.logback>
         <version.kafka>3.7.1</version.kafka>
         <version.httpclient5>5.4.3</version.httpclient5>
+        <version.commons.beanutils>1.11.0</version.commons.beanutils>
     </properties>
 
     <dependencyManagement>
@@ -703,6 +704,13 @@
             <groupId>org.apache.httpcomponents.client5</groupId>
             <artifactId>httpclient5</artifactId>
             <version>${version.httpclient5}</version>
+        </dependency>
+        
+        <!-- Fix for CVE-2025-48734 vulnerability in commons-beanutils -->
+        <dependency>
+            <groupId>commons-beanutils</groupId>
+            <artifactId>commons-beanutils</artifactId>
+            <version>${version.commons.beanutils}</version>
         </dependency>
     </dependencies>
     <build>
--- a/writers/iceberg/local-test/docker-compose.yml
+++ b/writers/iceberg/local-test/docker-compose.yml
@@ -66,7 +66,11 @@ services:
           /bin/sh -c "
           ./entrypoint.sh &&
           echo 'Starting Spark Connect server...' &&
-          /opt/spark/sbin/start-connect-server.sh --packages org.apache.spark:spark-connect_2.12:3.5.2,org.apache.hadoop:hadoop-aws:3.3.4,com.amazonaws:aws-java-sdk-bundle:1.12.262 --conf spark.hadoop.fs.s3a.impl=org.apache.hadoop.fs.s3a.S3AFileSystem &&
+          /opt/spark/sbin/start-connect-server.sh --packages org.apache.spark:spark-connect_2.12:3.5.2,org.apache.hadoop:hadoop-aws:3.3.4,com.amazonaws:aws-java-sdk-bundle:1.12.262 --conf spark.hadoop.fs.s3a.impl=org.apache.hadoop.fs.s3a.S3AFileSystem &
+          echo 'Starting Jupyter Notebook server...' &&
+          export PYSPARK_DRIVER_PYTHON=jupyter-notebook &&
+          export PYSPARK_DRIVER_PYTHON_OPTS=\"--notebook-dir=/home/iceberg/notebooks --ip='*' --NotebookApp.token='' --NotebookApp.password='' --port=8888 --no-browser --allow-root\" &&
+          pyspark &
           while true; do sleep 30; done
           "
 


pkg/waljs/filter.go
```diff
@@ -28,7 +28,6 @@ func NewChangeFilter(typeConverter func(value interface{}, columnType string) (i
 
 	return filter
 }
```
hash-data, line 29: add line back

drivers/mysql/internal/datatype_conversion.go
```diff
@@ -5,19 +5,23 @@ import "github.com/datazip-inc/olake/types"
...
 	// Floating point types
-	"float":   types.Float64,
+	"float":   types.Float32,
 	"double":  types.Float64,
-	"real":    types.Float64,
-	"decimal": types.Float64,
-	"numeric": types.Float64,
+	"real":    types.Float32,
+	"decimal": types.Float32,
+	"numeric": types.Float32,
```
hash-data, line 24: move float32 before float64

drivers/mysql/internal/mysql.go
```diff
@@ -144,6 +144,16 @@ func (m *MySQL) Discover(discoverSchema bool) ([]*types.Stream, error) {
 	return m.GetStreams(), nil
 }
 
+func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interface{}, error) {
+	if value == nil {
+		return nil, typeutils.ErrNullValue
+	}
+	// (e.g., varchar(50) -> varchar)
+	baseType := strings.ToLower(strings.TrimSpace(strings.Split(columnType, "(")[0]))
+	olakeType := mysqlTypeToDataTypes[baseType]
```
hash-data: let us write a util function for it as it is also being used in postgres.

types/data_types.go
```diff
@@ -28,6 +28,48 @@ const (
...
+
+	if !check1 || !check2 {
+		return 0 // If either of the types is not found in weights map, return 0
+	}
+
+	if weight1 < weight2 {
+		return -1
+	} else if weight1 > weight2 {
+		return 1
+	}
```
hash-data: use util.ternary, as well as according to convention when w1 < w2 return 1, w1>w2 return -1.

typeutils/fields.go
```diff
@@ -246,19 +246,26 @@ func (f *Field) Merge(anotherField *Field) {
...
+	t1, t2 types.DataType,
+	compare func(a, b types.DataType) int,
+) types.DataType {
 	// Start from the root node of the tree
 	node := root
 
 	// Traverse the tree
 	for node != nil {
-		if t1 > node.t && t2 > node.t {
+		cmp1 := compare(t1, node.t)
```
hash-data: can't we just get weight and compare weights instead of writing a comparater (seems over engineering)

typeutils/reformat.go
```diff
@@ -30,6 +30,7 @@ var DateTimeFormats = []string{
 	"2020-08-17T05:50:22.895Z",
 	"2006-01-02 15:04:05.999999-07",
 	"2006-01-02 15:04:05.999999+00",
+	"15:04:05",
```
hash-data: It is not a datetime format it is time format

pkg/binlog/filter.go
```diff
@@ -46,6 +49,12 @@ func (f ChangeFilter) FilterRowsEvent(e *replication.RowsEvent, ev *replication.
 		return nil
 	}
 
+	columnTypes := make([]string, len(e.Table.ColumnType))
+	for i, ct := range e.Table.ColumnType {
+		columnTypes[i] = mysqlTypeName(ct)
+	}
+	logger.Debugf("Column types: %v", columnTypes)
```
hash-data, line 57: let us create column and column type map here and pass to convertRowToMap function to make it look cleaner

hash-data: cant we use columnNameString values and use only single mapping that we have mysql/internal/datatype_conversion.go

drivers/mysql/internal/datatype_conversion.go
```diff
@@ -5,19 +5,23 @@ import "github.com/datazip-inc/olake/types"
...
+	"unsigned smallint":  types.Int32,
+	"unsigned tinyint":   types.Int32,
+	"unsigned mediumint": types.Int32,
+	"bigint":             types.Int64,
 
 	// Floating point types
-	"float":   types.Float64,
+	"float":   types.Float32,
+	"real":    types.Float32,
+	"decimal": types.Float32,
```
hash-data, line 23: below time is being converted to timestamp as well let us remove that and make it to string

typeutils/datatype.go
```diff
@@ -117,3 +118,8 @@ func detectTimestampPrecision(t time.Time) types.DataType {
 		return types.TimestampNano // store in nanoseconds
 	}
 }
+
+func ExtractAndMapColumnType(columnType string, typeMapping map[string]types.DataType) types.DataType {
+	baseType := strings.ToLower(strings.TrimSpace(strings.Split(columnType, "(")[0]))
```
hash-data, line 124: add a comment for what we pass and what it return

typeutils/fields.go
```diff
@@ -249,23 +249,32 @@ func GetCommonAncestorType(t1, t2 types.DataType) types.DataType {
...
 	for node != nil {
-		if t1 > node.t && t2 > node.t {
-			// If both t1 and t2 are greater than parent
+		weight1, exists1 := types.TypeWeights[t1]
+		weight2, exists2 := types.TypeWeights[t2]
+		nodeWeight, existsNode := types.TypeWeights[node.t]
+
+		// If any type is not found in weights map, return Unknown
+		if !exists1 || !exists2 || !existsNode {
+			return types.Unknown
```
hash-data, line 264: why? return node.t in this case

typeutils/fields.go
```diff
@@ -249,23 +249,32 @@ func GetCommonAncestorType(t1, t2 types.DataType) types.DataType {
...
+) types.DataType {
 	node := root
 
-	// Traverse the tree
 	for node != nil {
-		if t1 > node.t && t2 > node.t {
-			// If both t1 and t2 are greater than parent
+		weight1, exists1 := types.TypeWeights[t1]
+		weight2, exists2 := types.TypeWeights[t2]
+		nodeWeight, existsNode := types.TypeWeights[node.t]
```
hash-data:
```suggestion
	for node != nil {
		wt1, t1Exist := types.TypeWeight[t1]
		wt2, t2Exist := types.TypeWeight[t2]
		rootW, rootExist := types.TypeWeight[node.t]
```

types/data_types.go
```diff
@@ -28,6 +28,32 @@ const (
 	TimestampNano  DataType = "timestamp_nano"  // storing datetime up to 9 precisions
 )
 
+// Tree Representation of TypeWeights
+//
+//                                5 (String)
+//                               /       	 \
+//                3 (Float64)  /              \ 9 (TimestampNano)
+//                           /  \             /
+//             2 (Int64)   /     \ 4(Float32)/ 8 (TimestampMicro)
```
hash-data, line 38: separate float 32 little bit to show it is in left->right subtree

pkg/binlog/filter.go
```diff
@@ -84,13 +93,88 @@ func (f ChangeFilter) FilterRowsEvent(e *replication.RowsEvent, ev *replication.
...
+
 	record := make(map[string]interface{})
 	for i, val := range row {
-		record[columns[i]] = val
+		convertedVal, err := converter(val, columnTypes[i])
+		if err != nil {
+			record[columns[i]] = val // Fallback to original value
+		} else {
+			record[columns[i]] = convertedVal
+		}
```
hash-data: we must follow postgres implementation for error fallback (we should not do assumptions) :

Check here once: 

```
conv, err := c.converter(val, colType)
if err != nil && err != typeutils.ErrNullValue {
    return nil, err
}
```

pkg/binlog/filter.go
```diff
@@ -84,13 +93,88 @@ func (f ChangeFilter) FilterRowsEvent(e *replication.RowsEvent, ev *replication.
...
+	case mysql.MYSQL_TYPE_TINY_BLOB:
+		return "TINYBLOB"
+	case mysql.MYSQL_TYPE_MEDIUM_BLOB:
+		return "MEDIUMBLOB"
+	case mysql.MYSQL_TYPE_LONG_BLOB:
+		return "LONGBLOB"
+	case mysql.MYSQL_TYPE_BLOB:
+		return "BLOB"
+	case mysql.MYSQL_TYPE_VAR_STRING:
+		return "VAR_STRING"
```
hash-data: this is not available in datatype_convertor file, how is it being used?

pkg/binlog/filter.go
```diff
@@ -84,13 +93,88 @@ func (f ChangeFilter) FilterRowsEvent(e *replication.RowsEvent, ev *replication.
...
+	case mysql.MYSQL_TYPE_NEWDATE:
+		return "NEWDATE"
+	case mysql.MYSQL_TYPE_VARCHAR:
+		return "VARCHAR"
+	case mysql.MYSQL_TYPE_BIT:
+		return "BIT"
+	case mysql.MYSQL_TYPE_TIMESTAMP2:
+		return "TIMESTAMP2"
+	case mysql.MYSQL_TYPE_DATETIME2:
+		return "DATETIME2"
```
hash-data: these are also not available datatype_conversion file.

hash-data: make sure we use similar names that are available in datatype conversion file.

pkg/binlog/filter.go
ImDoubD-datazip: i am removing these conditions for now.

pkg/binlog/filter.go
ImDoubD-datazip: i am removing these conditions for now.

drivers/mysql/internal/datatype_conversion.go
```diff
@@ -37,11 +41,13 @@ var mysqlTypeToDataTypes = map[string]types.DataType{
...
 	// Date and time types
 	"date":      types.Timestamp,
-	"time":      types.Timestamp,
-	"datetime":  types.Timestamp,
 	"timestamp": types.Timestamp,
 	"year":      types.Int64,
 
+	// time and datetime types treated as string for now
+	"time":     types.String,
+	"datetime": types.String,
```
hash-data, line 49: any reason for datetime to be string?

hash-data: any reason for datetime to be string?

ImDoubD-datazip: it is for the postgres compatibility. in postgres, the datetime for now is converted to string only. so to avoid any issue in mysql side, i have put the datetime as string for now. @hash-data

hash-data: @ImDoubD-datazip  as we discussed postgres not provide datetime type, we can handle it as timestamp in mysql

vaibhav-datazip:
field with data type: bit is producing output as string 

Screenshot 2025-05-15 at 6 31 26 PM
Screenshot 2025-05-15 at 6 39 48 PM

Screenshot 2025-05-15 at 6 40 23 PM

ImDoubD-datazip:
@vaibhav-datazip @hash-data 
Issue Fixed
Bit being treated at int32

drivers/postgres/internal/datatype_conversion.go
```diff
@@ -7,6 +7,7 @@ import (
 var pgTypeToDataTypes = map[string]types.DataType{
 	// TODO: add proper types (not only int64)
 	"bigint":      types.Int64,
+	"int8":        types.Int64,
```
hash-data, line 10: are we sure, it is int64?

typeutils/reformat.go
```diff
@@ -265,6 +276,8 @@ func ReformatInt32(v any) (int32, error) {
 		return int32(v), nil
 	case int16:
 		return int32(v), nil
+	case int8:
```
hash-data, line 280: move before int16

hash-data: also add normal int

.gitignore
```diff
@@ -13,4 +13,4 @@ writers/iceberg/local-test/data/
 debezium-server-iceberg-sink.jar
 local-releaser.sh
 writers/iceberg/debezium-server-iceberg-sink/.idea/
-writers/iceberg/debezium-server-iceberg-sink/dependency-reduced-pom.xml
+writers/iceberg/debezium-server-iceberg-sink/dependency-reduced-pom.xml
```
hash-data, line 16: remove change

drivers/postgres/internal/datatype_conversion.go
ImDoubD-datazip: Yes, tested in Postgres
@hash-data

.gitignore
ImDoubD-datazip: This should be removed? @hash-data

hash-data (approved):

PR #253, opened 2025-04-27
Add validation for key field IDs in schema

### LINKED ISSUE #185: validate key fields are correctly set!?
https://github.com/datazip-inc/olake/blob/057741ec925fba783f5e171d69e216c7e4809420/writers/iceberg/debezium-server-iceberg-sink/src/main/java/io/debezium/server/iceberg/RecordConverter.java#L428-L433

---

###### This issue was generated by [todo-issue](https://github.com/DerJuulsn/todo-issue) based on a `TODO` comment in 057741ec925fba783f5e171d69e216c7e4809420. It&#x27;s been assigned to @shubham19may because they committed the code.

Sanket6652:
# Description

This change addresses the validation of key fields within the schema before inserting records into the database. Specifically, the validation checks if the key fields are correctly set in the schema and if they match the current table schema. If a key field is missing, a RuntimeException is thrown to prevent any invalid record insertions.

Fixes: Issue with checking if the key fields are set correctly before insertion into the table schema.

Fixes #185 

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/writers/iceberg/debezium-server-iceberg-sink/src/main/java/io/debezium/server/iceberg/RecordConverter.java
+++ b/writers/iceberg/debezium-server-iceberg-sink/src/main/java/io/debezium/server/iceberg/RecordConverter.java
@@ -425,7 +425,17 @@ public class RecordConverter {
         throw new RuntimeException("Failed to get schema from debezium event, event schema has no fields!");
       }
 
-      // @TODO validate key fields are correctly set!?
+      // Validate key fields
+      if (createIdentifierFields && !schemaData.identifierFieldIds().isEmpty()) {
+        LOGGER.debug("Validating key fields in schema: {}", schemaData.identifierFieldIds());
+        for (int keyFieldId : schemaData.identifierFieldIds()) {
+            boolean fieldExists = schemaData.fields().stream()
+                .anyMatch(field -> field.fieldId() == keyFieldId);
+            if (!fieldExists) {
+                throw new RuntimeException("Key field with ID " + keyFieldId + " is not present in the schema fields!");
+            }
+        }
+      }
       return new Schema(schemaData.fields(), schemaData.identifierFieldIds());
 
     }


writers/iceberg/debezium-server-iceberg-sink/src/main/java/io/debezium/server/iceberg/RecordConverter.java
ImDoubD-datazip, line 1: @shubham19may 
-> if key field is not present in the schema fields, then should it return a Runtime Exception, or do we need to handle this some other way?

@Sanket6652 
-> check whether at any point it is collecting the same id multiple times, because then I believe this validation will pass but there will be redundant entries.
-> Further, try to move this to a private method, so that it can be reused again somewhere. Currently the code block in which it is defined is quite crowded with other functionalities implementation as well. So move it out, call it here.

Sanket6652: @ImDoubD-datazip 

Yes,  there is currently an issue where the code doesn't check for duplicate field IDs in the identifierFieldIds list.
Will add a duplicate check in validateKeyFields using a HashSet and move the logic to a private method.

ImDoubD-datazip: Sure.

ImDoubD-datazip (dismissed):

writers/iceberg/debezium-server-iceberg-sink/src/main/java/io/debezium/server/iceberg/RecordConverter.java
```diff
@@ -425,11 +429,32 @@ private Schema icebergSchema(boolean createIdentifierFields) {
...
+    private void validateKeyFields(RecordSchemaData schemaData) {
+      LOGGER.debug("Validating key fields in schema: {}", schemaData.identifierFieldIds());
+      
+      // Check for duplicate field IDs
+      Set<Integer> seenIds = new HashSet<>();
+      for (int keyFieldId : schemaData.identifierFieldIds()) {
+          if (!seenIds.add(keyFieldId)) {
+              throw new RuntimeException("Duplicate key field ID found: " + keyFieldId + 
+                  ". Each field can only be used as an identifier field once.");
+          }
```
ImDoubD-datazip: @Sanket6652 
Hi, Duke this side.
Actually, the type of identifierFieldIds is itself Set<Integer>, so creating a set again here is redundant. It was wrong on my side as well, should have verified it. Please do remove this check, as its not necessary, this !seenIds one. Being a set, it would only have unique entries from the start.

Sanket6652: @ImDoubD-datazip 
Got it, thanks for the clarification! I’ll remove the redundant duplicate check

ImDoubD-datazip: @Sanket6652 
if (isPkField && !schemaData.identifierFieldIds().contains(field.fieldId())) {
              schemaData.identifierFieldIds().add(field.fieldId());
}

this is needed, as this just finds whether schemaData contains all the relevant fields if missed any.

Sanket6652: @ImDoubD-datazip 
done!

ImDoubD-datazip (approved):

shubham19may:
Not needed.

PR #256, opened 2025-04-28
feat: Append Only Mode Added

### LINKED ISSUE #189: [Feat] Append-Only Write Mode
**Description:**
Implement an "Append-Only" write mode in OLake that appends (inserts) records into the target table instead of modifying or deleting existing records. This mode should introduce a dedicated column, OLAKE_OPERATION_TYPE, to indicate the type of operation performed. Users should have the ability to rename this column based on their preference.

**Current Behavior:**

- The current implementation only supports Soft Delete write mode via equality deletes.

**Proposed Behavior:**

- When enabled, all data modifications (inserts, updates, deletes) are appended as new records.
- A dedicated column, OLAKE_OPERATION_TYPE, is introduced to capture the type of operation.

The possible values for OLAKE_OPERATION_TYPE are:
- Insert: A new record inserted for the first time.
- Update: An update to an existing record, represented as a new row.
- Delete: A record that has been logically deleted, represented as a new row.
- Snapshot: A full snapshot of the table at a point in time.

Users can rename the OLAKE_OPERATION_TYPE column.

**Use Case:**

- Enables audit logging and historical tracking of changes without modifying existing records.
- Supports downstream processing for CDC (Change Data Capture) workloads.
- Helps maintain immutability for compliance or analytical purposes.

**Configuration Options:**
- Add an optional parameter (e.g., write_mode: append_only) to enable this feature.
- Allow customization of the OLAKE_OPERATION_TYPE column name.

ImDoubD-datazip:
# Description

Append Only mode enables all changes (inserts, updates, deletes) to be stored as new, immutable records instead of modifying existing data.

Updated catalog.json :-
```
{
    "selected_streams": {
        "public": [
            {
                "split_column": "",
                "partition_regex": "",
                "stream_name": "postgres_test",
                "append_mode": true
            }
        ]
    },
    "streams: [....]
..
}
```

Fixes #189

## Type of change

# How Has This Been Tested?

- [ ] Initial Scenario
Screenshot 2025-04-28 at 6 07 07 PM

- [ ] Append only mode enabled
Screenshot 2025-04-28 at 6 10 50 PM


types/catalog.go
```diff
@@ -35,6 +35,7 @@ type StreamMetadata struct {
 	SplitColumn    string `json:"split_column"`
 	PartitionRegex string `json:"partition_regex"`
 	StreamName     string `json:"stream_name"`
+	AppendMode     bool   `json:"append_mode"`
```
hash-data: let us update readme of all drivers for it

types/catalog.go
```diff
@@ -57,6 +58,7 @@ func GetWrappedCatalog(streams []*Stream) *Catalog {
 		catalog.SelectedStreams[stream.Namespace] = append(catalog.SelectedStreams[stream.Namespace], StreamMetadata{
 			StreamName:     stream.Name,
 			PartitionRegex: "",
+			AppendMode:     false,
```
hash-data, line 62: reason for hard coding it?, have you tested it?

writers/iceberg/iceberg.go
```diff
@@ -53,7 +53,11 @@ func (i *Iceberg) Setup(stream protocol.Stream, options *protocol.Options) error
 		}
 	}
 
-	return i.SetupIcebergClient(!options.Backfill)
+	upsert := !options.Backfill
+	if i.stream.Self().StreamMetadata.AppendMode {
+		upsert = false
+	}
+	return i.SetupIcebergClient(upsert)
```
hash-data:
```suggestion
	if i.stream.Self().StreamMetadata.AppendMode {
	     // marking upsert mode to false
             return i.SetupIcebergClient(false)
        } 
        return i.SetupIcebergClient(!options.Backfill) 
```

shubham19may:
LGTM, tested

hash-data (approved):

hash-data (approved):

PR #258, opened 2025-04-29
fix: close parquet file

### LINKED ISSUE #122: [Issue] Parquet file not closed when new parquet file being created.
## Description
When a new Parquet writer is created, the previous Parquet file is not closed. This can lead to potential errors if there are open files that are not actively being written.

## Impact
- Unclosed files may lead to resource exhaustion.
- Errors can occur when switching between different Parquet files or partitions.
- Potential data corruption or loss if files remain open unexpectedly.

## Expected Behavior
- When a new Parquet writer is created, the previous one should be properly closed before switching.
- Inactive Parquet files should not cause errors if they are not actively being written.

## Possible Solution
- Ensure that before creating a new writer, the existing one is explicitly closed.

mastersans:
# Description

Fixes #122 

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/writers/parquet/parquet.go
+++ b/writers/parquet/parquet.go
@@ -73,7 +73,43 @@ func (p *Parquet) initS3Writer() error {
 	return nil
 }
 
+// closePartitionFile closes the writer and file for a given partition
+func (p *Parquet) closePartitionFile(basePath string, fileMetadata *FileMetadata) error {
+	if fileMetadata == nil {
+		return nil
+	}
+
+	// Close writer
+	var err error
+	if p.config.Normalization {
+		err = fileMetadata.writer.(*pqgo.GenericWriter[any]).Close()
+	} else {
+		err = fileMetadata.writer.(*pqgo.GenericWriter[types.RawRecord]).Close()
+	}
+	if err != nil {
+		return fmt.Errorf("failed to close writer: %s", err)
+	}
+
+	// Close file
+	if err := fileMetadata.parquetFile.Close(); err != nil {
+		return fmt.Errorf("failed to close file: %s", err)
+	}
+
+	filePath := filepath.Join(p.config.Path, basePath, fileMetadata.fileName)
+	logger.Infof("Closed file [%s] with %d records.", filePath, fileMetadata.recordCount)
+
+	return nil
+}
+
 func (p *Parquet) createNewPartitionFile(basePath string) error {
+	// Close existing file in this partition if it exists
+	if files, exists := p.partitionedFiles[basePath]; exists && len(files) > 0 {
+		lastFile := &files[len(files)-1]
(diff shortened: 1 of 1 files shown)


writers/parquet/parquet.go
```diff
@@ -73,7 +73,43 @@ func (p *Parquet) initS3Writer() error {
...
+}
+
 func (p *Parquet) createNewPartitionFile(basePath string) error {
+	// Close existing file in this partition if it exists
+	if files, exists := p.partitionedFiles[basePath]; exists && len(files) > 0 {
+		lastFile := &files[len(files)-1]
+		if err := p.closePartitionFile(basePath, lastFile); err != nil {
+			return err
+		}
+	}
```
hash-data: after this implementation, seems like we dont require it as partitionedFiles map[string][]FileMetadata 
array of metadata instead it can work simple with  map[string]FileMetadata. can we work in this direction ?
Thanks

zriyanshdz:
hey @mastersans any updates?

hash-data:
hi @mastersans  if you can do commented changes it will be merged

mastersans:
@hash-data I have made required changes can you take a look.

destination/parquet/parquet.go
```diff
@@ -74,7 +74,42 @@ func (p *Parquet) initS3Writer() error {
...
+}
+
 func (p *Parquet) createNewPartitionFile(basePath string) error {
+	// Close existing file in this partition if it exists
+	if files, exists := p.partitionedFiles[basePath]; exists {
+		lastFile := &files
+		if err := p.closePartitionFile(basePath, lastFile); err != nil {
+			return err
+		}
+	}
```
hash-data, line 111:
```suggestion
	if prevFile, exists := p.partitionedFiles[basePath]; exists {
		if err := p.closePartitionFile(basePath, &prevFile); err != nil {
			return err
		}
	}
```

destination/parquet/parquet.go
```diff
@@ -143,12 +178,12 @@ func (p *Parquet) Write(_ context.Context, record types.RawRecord) error {
 		partitionFolder = p.partitionedFiles[partitionedPath]
 	}
 
-	if len(partitionFolder) == 0 {
+	if partitionFolder.recordCount == 0 {
```
hash-data, line 181: i guess record count check does not make sense as partitioned file can be created from schema evolution function

hash-data: also let us rename it from partitionFolder -> partitionedFile

hash-data: check can be of nil

hash-data (commented):
Minor comments

shubham19may:
This is already part of destination refactor PR which is already live.

PR #261, opened 2025-05-05
fix: solves the chunk intersection issue

### LINKED ISSUE #140: remove chunk intersections where chunks can be {0, 100} {100, 200}. Need to {0, 99} {100, 200}
https://github.com/datazip-inc/olake/blob/ed705d25bc0cd1f048f2fa3e6309b3f55fb71b45/drivers/postgres/internal/backfill.go#L33-L38

---

###### This issue was generated by [todo-issue](https://github.com/DerJuulsn/todo-issue) based on a `TODO` comment in ed705d25bc0cd1f048f2fa3e6309b3f55fb71b45. It&#x27;s been assigned to @hash-data because they committed the code.

vaibhav-datazip:
# Description

This PR fixes chunk boundary intersection issues during backfill and sync by making the upper bound of each chunk exclusive. Previously, chunk ranges like `{0, 100}, {100, 200}` caused duplicate processing of boundary values (e.g., `100`), since they appeared in both chunks.

Along with this, the PR also fixes some minor logical flaws in the MySQL and Postgres drivers.

Fixes #140 

## Type of change

# How Has This Been Tested?

The changes were manually tested by running full sync and backfill operations using all chunking strategies across the supported connectors:

- [x] MongoDB
- [x] PostgreSQL
- [x] MySQL

Scenarios covered:
- Verified that chunk boundaries are non-overlapping (i.e., no duplicate records at boundary values)
- Confirmed that all records are included (no missing data)
- Ensured sync and backfill complete without errors


drivers/mysql/internal/backfill.go
```diff
@@ -139,13 +139,13 @@ func (m *MySQL) splitChunks(stream protocol.Stream, chunks *types.Set[types.Chun
 		currentVal := minVal
 		for {
 			var nextValRaw interface{}
-			err := tx.QueryRow(query, currentVal, chunkSize).Scan(&nextValRaw)
+			err := tx.QueryRow(query, currentVal).Scan(&nextValRaw)
```
hash-data, line 142: we are sure chunk size not required? 
Can we add a todo to revamp queries, so that we can take all variables as arguments not embed them using fmt.Sprintf

pkg/jdbc/jdbc.go
```diff
@@ -22,11 +22,11 @@ func NextChunkEndQuery(stream protocol.Stream, column string, chunkSize int) str
 // buildChunkCondition builds the condition for a chunk
 func buildChunkCondition(filterColumn string, chunk types.Chunk) string {
 	if chunk.Min != nil && chunk.Max != nil {
-		return fmt.Sprintf("%s >= %v AND %s <= %v", filterColumn, chunk.Min, filterColumn, chunk.Max)
+		return fmt.Sprintf("%s >= %v AND %s < %v", filterColumn, chunk.Min, filterColumn, chunk.Max)
 	} else if chunk.Min != nil {
 		return fmt.Sprintf("%s >= %v", filterColumn, chunk.Min)
 	}
-	return fmt.Sprintf("%s <= %v", filterColumn, chunk.Max)
+	return fmt.Sprintf("%s < %v", filterColumn, chunk.Max)
```
hash-data, line 29: If the last chunk does not contain max!= nil (let's say in the CTID chunk strategy), we will miss one last record.

hash-data (commented):
have we checked mongodb as well

hash-data (commented):
LGTM

PR #262, opened 2025-05-05
chore: Debug mode for java code & java project name change

shubham19may:
# Description

Currently there is no way to test the iceberg java code with go code. Figured out a way to easily do it using java debugger process attach module

## Type of change

- [X] Chore 
- [X] This change requires a documentation update

# How Has This Been Tested?

- [ ] Started go code with debugger and ./build command as well and added debug point in java code in other VScode window to see if it waits. When attached java process, and it pointed to that specific debug point.

### DIFF
--- a/drivers/postgres/internal/postgres.go
+++ b/drivers/postgres/internal/postgres.go
@@ -77,7 +77,7 @@ func (p *Postgres) Setup(ctx context.Context) error {
 
 		exists, err := doesReplicationSlotExists(pgClient, cdc.ReplicationSlot)
 		if err != nil {
-			return fmt.Errorf("failed to check if provided replication slot %s exists: %s", cdc.ReplicationSlot, err)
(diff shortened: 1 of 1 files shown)


drivers/postgres/internal/postgres.go
```diff
@@ -77,11 +77,11 @@ func (p *Postgres) Setup(ctx context.Context) error {
 
 		exists, err := doesReplicationSlotExists(pgClient, cdc.ReplicationSlot)
 		if err != nil {
-			return fmt.Errorf("failed to check replication slot: %s", err)
+			return fmt.Errorf("failed to check if provided replication slot %s exists: %s", cdc.ReplicationSlot, err)
```
hash-data:
```suggestion
			return fmt.Errorf("failed to check existence of replication slot %s: %s", cdc.ReplicationSlot, err)
```

hash-data (dismissed):
LGTM

hash-data (approved):

PR #268, opened 2025-05-07
[WIP] feat: incremental mongodb sync

### LINKED ISSUE #161: [Feat] Incremental Sync for MongoDB
**Background**
OLake currently supports Change Data Capture (CDC) for MongoDB using Change Streams, which efficiently tracks inserts, updates, and deletes in real time. However, many users—especially those on MongoDB shared clusters (e.g., Atlas M0, M2, M5 tiers) or self-hosted deployments without replica sets—may not have access to Change Streams, as it requires replica set or sharded cluster configurations.

For these users, OLake should support an alternative incremental sync method that does not rely on Change Streams but still efficiently captures newly inserted and updated documents.

**Feature Scope**
This feature will introduce a query-based incremental sync mechanism for MongoDB that does not require Change Streams. Possible approaches include:

**1. Timestamp-based Sync (updatedAt field tracking)**
- Users specify a timestamp field (e.g., updatedAt, lastModified, or a custom field) to track changes.
- OLake will store the last synced timestamp and query documents where updatedAt > last_synced_timestamp.
- This method requires users to ensure that the updatedAt field is updated on every modification (e.g., using MongoDB triggers or application logic).

**2. ObjectId-based Sync (For Insert-Only Collections)**
- MongoDB’s ObjectId contains a timestamp, allowing OLake to track inserts by filtering based on the _id field.
- OLake will query documents where _id > last_synced_id.
- This method only works for append-only collections, as it does not capture updates or deletes.

**3. Soft Delete Handling (Optional)**
- Some collections use a deleted flag or deletedAt timestamp instead of hard deletes. OLake can track deletions using:
- A boolean field (deleted: true)
- A timestamp (deletedAt with non-null values)

**Implementation Details**
**1. User Configuration:**
- Allow users to specify a tracking field (updatedAt, _id, or deletedAt).
- Provide default behavior (e.g., use updatedAt if available).

**2. Efficient Query Execution:**
- Ensure that incremental queries leverage indexes (e.g., index on updatedAt or _id).
- Implement batch processing and pagination to optimize large dataset syncs.
- Use projection to reduce unnecessary data transfer.

**3. Checkpointing & State Management:**
- OLake should maintain a sync state per collection, tracking the last processed timestamp or _id.
- Ensure robustness by handling failures and retries without duplicate processing.

**Fallback Mechanisms:**
If the specified tracking field is missing or unreliable, OLake should:
- Recommend a fallback full sync (configurable).
- Provide a warning and log errors to help users diagnose issues.

**Schema Evolution Handling:**
- If a new field is introduced after initial sync setup, OLake should detect it and allow users to update their sync strategy.
- Handle cases where tracking fields (updatedAt, _id) are removed or modified.

**Deliverables**

- Implementation of query-based incremental sync for MongoDB.
- Automated tests for different sync strategies (timestamp tracking, _id tracking, soft deletes).
- Performance benchmarking to optimize query execution.
- User documentation with setup instructions and best practices.

**Impact**

- Expands OLake’s MongoDB Support: Users on shared MongoDB instances (e.g., Atlas M0, M2, M5) can use incremental sync without needing replica sets.
- Performance Improvements: Avoids full collection scans and reduces database load.
- More Flexible Sync Options: Allows users to configure sync methods based on their collection structure.

Jougan-0:
# Description
Implemented query-based incremental sync for the MongoDB driver so that users without Change Streams (e.g. shared Atlas tiers or standalone servers) can still perform efficient incremental reads. This addresses the requirements in issue #161 by adding three sync strategies—timestamp, ObjectID, and soft-delete—and wiring up checkpointing in state.

Fixes #161 
Thing Implemented : 

- [X] Timestamp-based incremental sync (uses updatedAt > last_ts)
- [ ]  ObjectID-based incremental sync (uses _id > last_id)
- [ ]  Soft-delete support (tracks deleted: true / deletedAt > last_ts)
- [ ]  Configurable strategy, tracking field, and batch size in config.json
- [ ]  Batched, indexed queries with pagination
- [ ]  Checkpointing: persists and hydrates last_ts and last_id in state.json
- [ ]  Fallback full-refresh when tracking field is missing or invalid
- [ ]  Schema evolution detection (new or removed cursor fields)
- [ ]  Automated tests for all sync strategies
- [ ] Documentation

### DIFF
--- a/drivers/mongodb/internal/config.go
+++ b/drivers/mongodb/internal/config.go
@@ -10,19 +10,22 @@ import (
 )
 
 type Config struct {
-	Hosts             []string       `json:"hosts"`
-	Username          string         `json:"username"`
-	Password          string         `json:"password"`
-	AuthDB            string         `json:"authdb"`
-	ReplicaSet        string         `json:"replica_set"`
-	ReadPreference    string         `json:"read_preference"`
-	Srv               bool           `json:"srv"`
-	ServerRAM         uint           `json:"server_ram"`
-	MaxThreads        int            `json:"max_threads"`
-	Database          string         `json:"database"`
-	DefaultMode       types.SyncMode `json:"default_mode"`
-	RetryCount        int            `json:"backoff_retry_count"`
-	PartitionStrategy string         `json:"partition_strategy"`
+	Hosts             []string            `json:"hosts"`
+	Username          string              `json:"username"`
+	Password          string              `json:"password"`
+	AuthDB            string              `json:"authdb"`
+	ReplicaSet        string              `json:"replica_set"`
+	ReadPreference    string              `json:"read_preference"`
+	Srv               bool                `json:"srv"`
+	ServerRAM         uint                `json:"server_ram"`
+	MaxThreads        int                 `json:"max_threads"`
+	Database          string              `json:"database"`
+	DefaultMode       types.SyncMode      `json:"default_mode"`
+	RetryCount        int                 `json:"backoff_retry_count"`
+	PartitionStrategy string              `json:"partition_strategy"`
+	Incremental       IncrementalStrategy `json:"incremental_strategy,omitempty" mapstructure:"incremental_strategy"`
+	TrackingField     string              `json:"tracking_field,omitempty"       mapstructure:"tracking_field"`
+	BatchSize         int32               `json:"batch_size,omitempty"           mapstructure:"batch_size"`
 }
 
 func (c *Config) URI() string {
@@ -45,7 +48,6 @@ func (c *Config) URI() string {
 		}
 		options = fmt.Sprintf("%s&replicaSet=%s&readPreference=%s", options, c.ReplicaSet, c.ReadPreference)
 	}
-
 	//  Handle auth credentials
 	auth := ""
 	if c.Username != "" {
@@ -61,5 +63,11 @@ func (c *Config) URI() string {
 
 // TODO: Add go struct validation in Config
 func (c *Config) Validate() error {
+	if c.Incremental == "" {
+		c.Incremental = StrategyChangeStream // preserve current behaviour
+	}
+	if c.BatchSize == 0 {
+		c.BatchSize = 5000
+	}
 	return utils.Validate(c)
 }
--- a/drivers/mongodb/internal/incremental.go
+++ b/drivers/mongodb/internal/incremental.go
@@ -1,0 +1,156 @@
+package driver
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/datazip-inc/olake/constants"
+	"github.com/datazip-inc/olake/logger"
+	"github.com/datazip-inc/olake/protocol"
+	"github.com/datazip-inc/olake/types"
+	"github.com/datazip-inc/olake/utils"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+func (m *Mongo) incrementalSync(stream protocol.Stream, pool *protocol.WriterPool) error {
+	cfg := m.config
+	db, collName := stream.Namespace(), stream.Name()
+	coll := m.client.Database(db).Collection(collName)
+
+	// ---- hydrated checkpoint from State ------------
+	cstream := stream.Self() // *types.ConfiguredStream
+	var lastTS primitive.DateTime
+	var lastID primitive.ObjectID
+
+	if raw := m.State.GetCursor(cstream, cursorLastTS); raw != nil {
+		switch t := raw.(type) {
+		case primitive.DateTime:
+			lastTS = t
+		case string:
+			if parsed, err := time.Parse(time.RFC3339Nano, t); err == nil {
+				lastTS = primitive.NewDateTimeFromTime(parsed)
+			} else {
+				logger.Warnf("could not parse saved last_ts %q: %s", t, err)
+			}
+		default:
+			logger.Warnf("unexpected type for last_ts in state: %T", raw)
+		}
+	}
+	if raw := m.State.GetCursor(cstream, cursorLastID); raw != nil {
+		switch v := raw.(type) {
+		case primitive.ObjectID:
+			lastID = v
+		case string:
+			if oid, err := primitive.ObjectIDFromHex(v); err == nil {
+				lastID = oid
+			} else {
+				logger.Warnf("could not parse saved last_id %q: %s", v, err)
+			}
+		default:
+			logger.Warnf("unexpected type for last_id in state: %T", raw)
+		}
+	}
+
+	batch := cfg.BatchSize
+	if batch == 0 {
+		batch = 5000
+	}
+	trk := cfg.TrackingField
+	if trk == "" {
+		ac := stream.GetStream().AvailableCursorFields.Array()
+		if len(ac) == 1 {
+			trk = ac[0]
+		} else {
+			trk = "_id"
+		}
+	}
+
+	logger.Infof("incremental sync started on %s.%s (strategy=%s, batch=%d)", db, collName, cfg.Incremental, batch)
+
+	for {
+		var filter bson.M
+		findOpts := options.Find().SetBatchSize(batch)
+
+		switch cfg.Incremental {
+		case StrategyTimestamp:
+			filter = bson.M{trk: bson.M{"$gt": lastTS}}
+			findOpts.SetSort(bson.D{{Key: trk, Value: 1}})
+
+		case StrategyObjectID:
+			filter = bson.M{"_id": bson.M{"$gt": lastID}}
+			findOpts.SetSort(bson.D{{Key: "_id", Value: 1}})
+
+		case StrategySoftDelete:
+			filter = bson.M{
+				"$or": []bson.M{
+					{trk: bson.M{"$gt": lastTS}},
+					{"deleted": true, "deletedAt": bson.M{"$gt": lastTS}},
+				}}
+			findOpts.SetSort(bson.D{{Key: trk, Value: 1}})
+
+		default:
+			return fmt.Errorf("unknown incremental strategy %q", cfg.Incremental)
+		}
+
+		cur, err := coll.Find(context.TODO(), filter, findOpts)
+		if err != nil {
+			return err
+		}
+
+		thread, err := pool.NewThread(context.TODO(), stream)
+		if err != nil {
+			cur.Close(context.TODO())
+			return err
+		}
+
+		var processed int
+		for cur.Next(context.TODO()) {
+			var doc bson.M
+			_ = cur.Decode(&doc)
+			handleMongoObject(doc)
+
+			hash := utils.GetKeysHash(doc, constants.MongoPrimaryID)
+			if err := thread.Insert(types.CreateRawRecord(hash, doc, "r", time.Unix(0, 0))); err != nil {
+				cur.Close(context.TODO())
+				return err
+			}
+
+			// ---- advance in-memory checkpoint safely ----
+			switch cfg.Incremental {
+			case StrategyObjectID:
+				if hex, ok := doc["_id"].(string); ok {
+					if oid, err := primitive.ObjectIDFromHex(hex); err == nil {
+						lastID = oid
+					}
+				}
+			default: // timestamp & soft_delete
+				// handleMongoObject lower-cases keys → use trk in lower case
+				if tsRaw, ok := doc[strings.ToLower(trk)]; ok {
+					switch v := tsRaw.(type) {
+					case primitive.DateTime:
+						lastTS = v
+					case time.Time:
+						lastTS = primitive.NewDateTimeFromTime(v)
+					}
+				}
+			}
+			processed++
+		}
+		cur.Close(context.TODO())
+		thread.Close()
+
+		if processed > 0 {
+			m.State.SetCursor(cstream, cursorLastTS, lastTS)
+			if cfg.Incremental == StrategyObjectID || cfg.Incremental == StrategySoftDelete {
+				m.State.SetCursor(cstream, cursorLastID, lastID)
+			}
+			m.State.LogState()
+		} else {
+			return nil
+		}
+	}
+}
--- a/drivers/mongodb/internal/constants.go
+++ b/drivers/mongodb/internal/constants.go
@@ -1,0 +1,15 @@
+package driver
+
+const (
+	cursorLastTS = "last_ts"
+	cursorLastID = "last_id"
+)
+
+type IncrementalStrategy string
+
+const (
+	StrategyChangeStream IncrementalStrategy = "change_stream" // default – existing behaviour
+	StrategyTimestamp    IncrementalStrategy = "timestamp"     // new
+	StrategyObjectID     IncrementalStrategy = "objectid"      // new
+	StrategySoftDelete   IncrementalStrategy = "soft_delete"   // new
+)
--- a/drivers/mongodb/internal/mon.go
+++ b/drivers/mongodb/internal/mon.go
@@ -3,6 +3,7 @@ package driver
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/datazip-inc/olake/drivers/base"
@@ -142,9 +143,12 @@ func (m *Mongo) Read(pool *protocol.WriterPool, stream protocol.Stream) error {
 		return m.backfill(stream, pool)
 	case types.CDC:
 		return m.changeStreamSync(stream, pool)
+	case types.INCREMENTAL:
+		return m.incrementalSync(stream, pool)
+	default:
+		return fmt.Errorf("unsupported sync mode: %s", stream.GetSyncMode())
 	}
 
-	return nil
 }
 
 // fetch schema types from mongo for streamName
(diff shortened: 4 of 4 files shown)


Jougan-0:
I hope this should be done before this weekend. Once this is GTG I'll move to postgres one.

Jougan-0:
First initial sync of 5000 rows:
image

Second Sync after I added 100 more rows the timestamp from state was used as a Starting point:
image

zriyanshdz:
@shubham19may @hash-data please take a look if the PR (in draft) is going in right direction so it can be merge ready without final moment major changes. 

good work @Jougan-0

Jougan-0:
Last 100 records in new .parquet file
image
First 5000 record in separate .parquet file.
image

hash-data:
hi @zriyanshdz  will check

Jougan-0:
@hash-data I had a doubt regarding the test thing. If you could approve the workflow as well so if something is failing I can fix it too.
I need to write pipeline for it so shall we go CI/CD or there is something existing that I can use?
I just to do some final testing and docs update then this is done. All the testing I did locally is working fine. If you want you can start reviewing. 
I might have not re used a lot of existing functions. Please guide me if there are any redundant code.
PS: I also need to test it rigorously. Some :bug: might be lingering tbh.

drivers/mongodb/README.md
```diff
@@ -115,7 +123,17 @@ Before running the Sync command, the generated `catalog.json` file must be confi
...
                "namespace": "incr",
                ...
-               "sync_mode": "cdc"
+               "sync_mode": "cdc",
+               "supported_sync_modes": [
+                     "full_refresh",
+                     "cdc",
+                     "incremental"
+                  ],
+                  "sync_mode": "incremental",
```
vaibhav-datazip: a stream can have single sync_mode, not both cdc and incremental

drivers/mongodb/README.md
```diff
@@ -115,7 +123,17 @@ Before running the Sync command, the generated `catalog.json` file must be confi
...
+                  "available_cursor_fields": [
+                     "updatedAt",
+                     "_id"
+                  ],
             }
```
vaibhav-datazip, line 140: for incremental sync mode, its better to add cursor_field here

drivers/mongodb/README.md
Jougan-0: My bad, it slipped in Readme by mistake while I was updating the readme. I used singular sync_mode. ~~Updating it~~ Updated.

drivers/mongodb/README.md
Jougan-0: Yes sir, it is there I forgot to add it in docs. :disappointed: 
Screenshot from 2025-05-13 01-12-51

drivers/mongodb/internal/config.go
```diff
@@ -61,5 +63,11 @@ func (c *Config) URI() string {
 
 // TODO: Add go struct validation in Config
 func (c *Config) Validate() error {
+	if c.Incremental == "" {
+		c.Incremental = StrategyChangeStream // preserve current behaviour
+	}
```
vaibhav-datazip: In case of incremental mode, if incremental_strategy is left blank, why is it being assigned StrategyChangeStream as the default?
As this value is not one of the three supported incremental strategies:
- "timestamp" (StrategyTimestamp)
- "objectid" (StrategyObjectID)
- "soft_delete" (StrategySoftDelete)

drivers/mongodb/internal/config.go
```diff
@@ -61,5 +63,11 @@ func (c *Config) URI() string {
 
 // TODO: Add go struct validation in Config
 func (c *Config) Validate() error {
```
vaibhav-datazip: This validate function isn't getting triggered, this function must be called from mon.go Setup function as it is done in Postgres and mysql.
```
	if err := m.config.Validate(); err != nil {
		return fmt.Errorf("failed to validate config: %s", err)
	}
```

drivers/mongodb/internal/incremental.go
```diff
@@ -0,0 +1,162 @@
...
+			findOpts.SetSort(bson.D{{Key: "_id", Value: 1}})
+		case StrategySoftDelete:
+			filter = bson.M{
+				"$or": []bson.M{
+					{trk: bson.M{"$gt": lastTS}},
+					{"deleted": true, "deletedAt": bson.M{"$gt": lastTS}},
+				}}
+			findOpts.SetSort(bson.D{{Key: trk, Value: 1}})
+		default:
+			return fmt.Errorf("unknown incremental strategy %q", cfg.Incremental)
```
vaibhav-datazip: This error message isn't accurate when incremental_strategy is left blank. It currently results in `FATAL error occurred while reading records: unknown incremental strategy "change_stream"` due to the default assignment, which is confusing.

Jougan-0: True, removing the above line in validate.

drivers/mongodb/internal/config.go
Jougan-0: I thought so myself but had a doubt regarding the change of stream in mongo and in backfill or somewhere I read default should be change of stream so I went ahead and added this check. There would an [extra constant](https://github.com/datazip-inc/olake/pull/268#diff-49b59b3099ecf46a04f71101e1799d60205975a3a1bd22cd23f716e421ee40fdR11) as well in the files that I would have to remove I think this check is not required in validate as other than this batchsize is checked that I did in the main function as well. Updating it in a while.

drivers/mongodb/internal/config.go
Jougan-0: It was an existing function I just added in the checks but if it not called, I'll just make a quick enhancement in the code.

drivers/mongodb/internal/incremental.go
Jougan-0: Removed the validate code. This shouldn't be an issue now.

drivers/mongodb/internal/config.go
Jougan-0: There is no need to call this anymore imo . The check was already happening  in the function itself.

drivers/mongodb/internal/config.go
```diff
@@ -10,19 +10,22 @@ import (
...
+	ReadPreference    string              `json:"read_preference"`
+	Srv               bool                `json:"srv"`
+	ServerRAM         uint                `json:"server_ram"`
+	MaxThreads        int                 `json:"max_threads"`
+	Database          string              `json:"database"`
+	DefaultMode       types.SyncMode      `json:"default_mode"`
+	RetryCount        int                 `json:"backoff_retry_count"`
+	PartitionStrategy string              `json:"partition_strategy"`
+	Incremental       IncrementalStrategy `json:"incremental_strategy,omitempty" mapstructure:"incremental_strategy"`
+	TrackingField     string              `json:"tracking_field,omitempty"       mapstructure:"tracking_field"`
```
vaibhav-datazip: These should be stream level properties

Jougan-0: Migrated to stream.

Jougan-0:
For fallback the logic I used is: 
If the tracking field is not present in schema, I just switch the syncmode to FULL_REFRESH.

Jougan-0:
Working on multi thread solution should be done by EOD. Already implemented it, just testing then committing.

drivers/mongodb/README.md
```diff
@@ -111,11 +115,24 @@ Before running the Sync command, the generated `catalog.json` file must be confi
...
+               "supported_sync_modes": [
+                     "full_refresh",
+                     "cdc",
+                     "incremental"
+                  ],
+                  "sync_mode": "incremental",
+                  "available_cursor_fields": [
+                     "updatedAt",
+                     "_id"
+                  ],
```
vaibhav-datazip: available_cursor_fields is not getting populated

drivers/mongodb/README.md
```diff
@@ -111,11 +115,24 @@ Before running the Sync command, the generated `catalog.json` file must be confi
...
+                  // [Incremental Sync configuration]
+                  "incremental_strategy": "timestamp",   // "timestamp" | "objectid" | "soft_delete"
+                  "tracking_field": "updatedAt",        // e.g. "updatedAt", "lastModified"
```
vaibhav-datazip: why do we need tracking_field, shouldn't we use cursor_field for this

drivers/mongodb/README.md
Jougan-0: This is not required.

drivers/mongodb/README.md
Jougan-0: I thought of 2 distinguished identities for the sake of checking. The tracking field is used for checking with schema and the available cursors are populated on the basis of schema. 

We can use the cursor field for it as well so imo we can do with the cursor field itself and remove the tracking field all together. 

But it was just for differentiating and making the thing clear.

drivers/mongodb/README.md
Jougan-0: Updated.

drivers/mongodb/README.md
Jougan-0: Updated.

types/stream_configured.go
```diff
@@ -85,3 +86,24 @@ func (s *ConfiguredStream) Validate(source *Stream) error {
 
 	return nil
 }
+func (s *ConfiguredStream) ValidateTrackingFieldInSchema(source *Stream) error {
```
Jougan-0: Improving this logic it seems to bug out with lower case.

types/stream_configured.go
```diff
@@ -85,3 +86,24 @@ func (s *ConfiguredStream) Validate(source *Stream) error {
...
+			schemaProps[field] = struct{}{}
+			schemaFields = append(schemaFields, field)
+			return true
+		})
+		if _, ok := schemaProps[tf]; !ok {
+			s.Stream.SyncMode = FULLREFRESH
+			source.AvailableCursorFields = NewSet("_id")
+			return fmt.Errorf("invalid tracking field %q: not present in schema fields. "+"Available fields for cursor are: [%s]", tf, strings.Join(schemaFields, ", "))
+		}
+		source.AvailableCursorFields = NewSet(s.CursorField)
```
vaibhav-datazip: it will be better to check this using available_cursor_field instead of just checking the field is present in schema. available_cursor_fields can be populated while discovery command is run and checks can be implemented in that to decide if the schema field is suitable for being used as a cursor field.

Jougan-0: I had a doubt please can we have a chat so this can be clear. 
As I can feed available cursor from streams.json but what if in db one of the field is dropped between 2 syncs in that scenario we need to check the schema it was mention in the issue over here.
Also if we are checking the schema then why populate with available cursor wouldn't that be redundant. If you want me to have a logic that first we check available cursor and if not present then we check schema we can do that as well.
Thoughts on second method?

image

drivers/mongodb/README.md
```diff
@@ -92,7 +96,7 @@ Before running the Sync command, the generated `catalog.json` file must be confi
    For each stream you want to sync:<br>
    - Add the following properties:
       ```json
-      "sync_mode": "cdc",
+      "sync_mode": "incremental",
```
hash-data: let us remove it

drivers/mongodb/README.md
```diff
@@ -111,11 +115,18 @@ Before running the Sync command, the generated `catalog.json` file must be confi
...
                "namespace": "incr",
                ...
-               "sync_mode": "cdc"
+               "supported_sync_modes": [
+                     "full_refresh",
+                     "cdc",
+                     "incremental"
+                  ],
+                  "sync_mode": "incremental",
+                  "incremental_strategy": "timestamp",   // "timestamp" | "objectid" | "soft_delete"
```
hash-data: Do we need to have a strategy in it ?

drivers/mongodb/internal/incremental.go
```diff
@@ -0,0 +1,169 @@
...
+	cfg := m.config
+	db, collName := stream.Namespace(), stream.Name()
+	coll := m.client.Database(db).Collection(collName)
+
+	cstream := stream.Self()
+	mstream := stream.GetStream()
+
+	var lastTS primitive.DateTime
+	var lastID primitive.ObjectID
+	if raw := m.State.GetCursor(cstream, cursorLastTS); raw != nil {
```
hash-data: The steps we have to follow here are :
-> if state not present, do backfill and get the max(cursor) from backfill (via query or using some other strategy)
-> I don't think we need this much strategy in incremental We can directly run parallel sync based on the cursor field and state 

Recommend you visit the Python code here: https://github.com/datazip-inc/dz-tap-mongodb/blob/master/dz_mongodb/sync_strategies/incremental.py

We have been using it for almost 3 years.

Thanks

types/stream_configured.go
hash-data: Let me know if we can have a meet today for it.

drivers/mongodb/README.md
Jougan-0: It is not as a need to have. It's like just good to have to keep things separated.

drivers/mongodb/internal/incremental.go
Jougan-0: Sure I can do that it's just while running the code was hitting panic that's why this check was added. I can go with a simple approach as well.

drivers/mongodb/internal/incremental.go
```diff
@@ -0,0 +1,165 @@
...
+			stateField, cursorField)
+
+		if err := m.backfill(stream, pool); err != nil {
+			return err
+		}
+		if err := snapMaxAsBookmark(coll, cursorField, cstream, m); err != nil {
+			return err
+		}
+		return nil
+	}
```
Jougan-0: This is done because if I had a first sync with the field ```_id``` and now I want to change to ```updatedAt``` it would check the fields and then fallback to backfill and then based on cursor_field it would be repopulated.

drivers/mongodb/internal/incremental.go
Jougan-0: New multi thread logic implemented.

drivers/mongodb/internal/incremental.go
```diff
@@ -0,0 +1,418 @@
...
+			}
+			hash := utils.GetKeysHash(doc, constants.MongoPrimaryID)
+			if e := writer.Insert(types.CreateRawRecord(hash, doc, "r", time.Unix(0, 0))); e != nil {
+				return e
+			}
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
+}
```
vaibhav-datazip: You can follow the same pattern as we have followed in Mongo driver's backfill.go . That will make the code more consistent across the code base.

Jougan-0: Thanks for pointing that out. I agree that consistency is valuable, but in this specific path the extra spawnWriters / writeWorker fan-out is deliberate:
In backfill: I/O wait on writer.Insert is hidden because the other N-1 goroutines are still busy.
In incremental: One goroutine reads the entire range (only way to honour the sort/hint on the incremental cursor).
So we can’t just drop spawnWriters without either
  1.Opening a separate Mongo cursor per worker (Mongo would no longer guarantee order and we’d lose index          usage),
  2. Accepting a big hit in throughput as the single reader blocks on every insert.
  
I hope that clarifies the rationale. I’ve left the back-fill logic untouched because changing proven code would demand extensive retesting, and the small gain in stylistic consistency doesn’t outweigh the added risk and effort.

Jougan-0: @vaibhav-datazip //

Jougan-0: Removed it please have a look when you get time.

drivers/mongodb/internal/incremental.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+		m["$lte"] = rng.Max
+		filter[rng.Field] = m
+	}
+	opts := options.Find().
+		SetSort(bson.D{{Key: rng.Field, Value: 1}}).
+		SetBatchSize(batchSize).
+		SetHint(bson.D{{Key: rng.Field, Value: 1}})
+
+	return base.RetryOnBackoff(3, time.Minute, func() error {
```
vaibhav-datazip, line 155: You can use user defined Retry count here instead of hardcoding

drivers/mongodb/internal/incremental.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+type ScanConfig struct {
+	Coll       *mongo.Collection
+	BatchSize  int32
+	Transform  func(bson.M)
+	WriterPool *protocol.WriterPool
+	MaxThreads int
+	RetryCount int
+}
```
vaibhav-datazip, line 42: This struct is not being used anywhere.

drivers/mongodb/internal/incremental.go
```diff
@@ -0,0 +1,316 @@
...
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+	"go.mongodb.org/mongo-driver/mongo/readconcern"
+)
+
+const (
+	bmkField        = "cursor_field"
+	bmkVal          = "cursor_value"
+	defaultBatch    = 5_000
+	defaultChkEvery = 50_000
```
vaibhav-datazip, line 26: The field `defaultChkEvery` is not being used.

vaibhav-datazip:
Can you also merge the staging branch into your branch

hash-data:
@Jougan-0  I know it will be little pain, but we have revamped the structure of code, so can you take staging pull. Now the concurrency thing will be handled on global level and individual connector have not to take care of it.

Jougan-0:
> @Jougan-0  I know it will be little pain, but we have revamped the structure of code, so can you take staging pull. Now the concurrency thing will be handled on global level and individual connector have not to take care of it.

I'll get time in a few weeks then I'll get back to it.

hash-data:
@Jougan-0  let me know, this week our target is to merge this.

Jougan-0:
@hash-data
I hope you’re well. I wanted to let you know that, unfortunately, I won’t be able to complete this PR, and I will be stepping away from contributing further under the Olake project. As I’ve been contributing without compensation, making numerous changes without pay is no longer sustainable for me in the long run.

Thank you for your time and understanding. Please feel free to reopen the PR and make use of the logic as you see fit.

hash-data:
merged a pr for the same: https://github.com/datazip-inc/olake/pull/385

PR #269, opened 2025-05-08
feat: datatypes integration testing complete

ImDoubD-datazip:
# Description
Integration testing of datatypes of postgres is done in this case.

Fixes # (issue)

## Type of change

# How Has This Been Tested?
This has been tested using different datatypes of postgres in full refresh mode as well as cdc mode.


hash-data:
not required as we are implementing through test containers

PR #67, opened 2025-01-22
feat: Starting With Unit Tests 

hash-data:
# Description
Introducing Basic Unit Tests for mongo connection.

Reason for test.key file 
Screenshot 2025-01-30 at 6 35 20 PM

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

PR #273, opened 2025-05-11
feat: added flatten, resolver and catalog tests

vaibhav-datazip:
# Description

This PR adds unit tests for the flatten, resolver and the catalog file logic. The tests cover various scenarios.

## Type of change

# How Has This Been Tested?

- Running tests added for flatten, resolver and catalog file, using the command 
` go test -v -run <test_name>`
- Verified coverage increase using 
`go test -coverprofile`

### DIFF
--- a/types/catalog_test.go
+++ b/types/catalog_test.go
@@ -1,0 +1,208 @@
+package types
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetWrappedCatalog(t *testing.T) {
+	testCases := []struct {
+		name     string
+		streams  []*Stream
+		expected *Catalog
+	}{
+		{
+			name:    "empty streams",
+			streams: []*Stream{},
+			expected: &Catalog{
+				Streams:         []*ConfiguredStream{},
+				SelectedStreams: make(map[string][]StreamMetadata),
+			},
+		},
+		{
+			name: "single stream",
+			streams: []*Stream{
+				{
+					Name:      "test_stream",
+					Namespace: "test_namespace",
+				},
+			},
+			expected: &Catalog{
+				Streams: []*ConfiguredStream{
+					{
+						Stream: &Stream{
+							Name:      "test_stream",
+							Namespace: "test_namespace",
+						},
+					},
+				},
+				SelectedStreams: map[string][]StreamMetadata{
+					"test_namespace": {
+						{
+							StreamName:     "test_stream",
+							PartitionRegex: "",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "multiple streams same namespace",
+			streams: []*Stream{
+				{
+					Name:      "stream1",
+					Namespace: "namespace1",
+				},
+				{
+					Name:      "stream2",
+					Namespace: "namespace1",
+				},
+			},
+			expected: &Catalog{
+				Streams: []*ConfiguredStream{
+					{
+						Stream: &Stream{
+							Name:      "stream1",
+							Namespace: "namespace1",
+						},
+					},
+					{
+						Stream: &Stream{
+							Name:      "stream2",
+							Namespace: "namespace1",
+						},
+					},
+				},
+				SelectedStreams: map[string][]StreamMetadata{
+					"namespace1": {
+						{
+							StreamName:     "stream1",
+							PartitionRegex: "",
+						},
+						{
+							StreamName:     "stream2",
+							PartitionRegex: "",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "multiple streams different namespaces",
+			streams: []*Stream{
+				{
+					Name:      "stream1",
+					Namespace: "namespace1",
+				},
+				{
+					Name:      "stream2",
+					Namespace: "namespace2",
+				},
+			},
+			expected: &Catalog{
+				Streams: []*ConfiguredStream{
+					{
+						Stream: &Stream{
+							Name:      "stream1",
+							Namespace: "namespace1",
+						},
+					},
+					{
+						Stream: &Stream{
+							Name:      "stream2",
+							Namespace: "namespace2",
+						},
+					},
+				},
+				SelectedStreams: map[string][]StreamMetadata{
+					"namespace1": {
+						{
+							StreamName:     "stream1",
+							PartitionRegex: "",
+						},
+					},
+					"namespace2": {
+						{
+							StreamName:     "stream2",
+							PartitionRegex: "",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "stream with full properties",
+			streams: []*Stream{
+				{
+					Name:      "users",
+					Namespace: "public",
+					Schema: &TypeSchema{
+						Properties: sync.Map{},
+					},
+					SupportedSyncModes:      NewSet[SyncMode]("incremental", "full_refresh"),
+					SourceDefinedPrimaryKey: NewSet("id"),
+					AvailableCursorFields:   NewSet("updated_at"),
+					AdditionalProperties:    `{"replication_method": "incremental"}`,
+					SyncMode:                "incremental",
+				},
+			},
+			expected: &Catalog{
+				Streams: []*ConfiguredStream{
+					{
+						Stream: &Stream{
+							Name:      "users",
+							Namespace: "public",
+							Schema: &TypeSchema{
+								Properties: sync.Map{},
+							},
+							SupportedSyncModes:      NewSet[SyncMode]("incremental", "full_refresh"),
+							SourceDefinedPrimaryKey: NewSet("id"),
+							AvailableCursorFields:   NewSet("updated_at"),
+							AdditionalProperties:    `{"replication_method": "incremental"}`,
+							SyncMode:                "incremental",
+						},
+					},
+				},
+				SelectedStreams: map[string][]StreamMetadata{
+					"public": {
+						{
+							StreamName:     "users",
+							PartitionRegex: "",
+						},
+					},
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			result := GetWrappedCatalog(tc.streams)
+
+			// Verify the number of streams
+			assert.Equal(t, len(tc.expected.Streams), len(result.Streams), "Number of streams should match")
+
+			// Verify each stream
+			for i, expectedStream := range tc.expected.Streams {
+				assert.Equal(t, expectedStream.Stream.Name, result.Streams[i].Stream.Name, "Stream name should match")
+				assert.Equal(t, expectedStream.Stream.Namespace, result.Streams[i].Stream.Namespace, "Stream namespace should match")
+			}
+
+			// Verify selected streams
+			assert.Equal(t, len(tc.expected.SelectedStreams), len(result.SelectedStreams), "Number of namespaces should match")
+
+			for namespace, expectedMetadata := range tc.expected.SelectedStreams {
+				resultMetadata, exists := result.SelectedStreams[namespace]
+				assert.True(t, exists, "Namespace should exist in result")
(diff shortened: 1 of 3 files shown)


shubham19may:
Reference : https://github.com/datazip-inc/olake/pull/67

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: go mod download
+
+      - name: Run Catalog Tests
+        run: go test -v ./types -run "TestCatalog"
+
+      - name: Run Flatten Tests
+        run: go test -v ./utils/typeutils -run "TestFlatten"
+
+      - name: Run Resolve Tests
+        run: go test -v ./utils/typeutils -run "TestResolve"
```
hash-data: if we start running in this way then we have to add a lot of functions, is there any generic way to run all unit tests

types/catalog_test.go
```diff
@@ -0,0 +1,776 @@
...
+			expected: &Catalog{
+				Streams: []*ConfiguredStream{
+					{
+						Stream: &Stream{
+							Name:                  "stream1",
+							Namespace:             "namespace1",
+							Schema:                newSchema(),
+							SupportedSyncModes:    NewSet(SyncMode("incremental"), SyncMode("full_refresh")),
+							SyncMode:              SyncMode("cdc"),
+							AvailableCursorFields: NewSet("created_at"),
```
hash-data, line 340: how supported sync mode not contains cdc and sync mode is set to cdc?

types/catalog_test.go
```diff
@@ -0,0 +1,776 @@
...
+							CursorField:           "created_at",
+							DestinationDatabase:   "db:namespace1",
+							DestinationTable:      "stream1",
+						},
+					},
+				},
+				SelectedStreams: map[string][]StreamMetadata{
+					"namespace1": {
+						{StreamName: "stream1", PartitionRegex: "new_partition"},
```
hash-data: can we add all properties in test, like normalization and all

types/catalog_test.go
```diff
@@ -0,0 +1,776 @@
...
+							DestinationTable:    "stream2",
+						},
+					},
+				},
+				SelectedStreams: map[string][]StreamMetadata{
+					"namespace1": {
+						{StreamName: "stream1", PartitionRegex: "user_partition"},
+					},
+				},
+			},
```
hash-data, line 611: add cases like stream exist in new one but not in old, also stream exist in old but not in new one, same like sync mode cdc configuration were present prev but not now

utils/typeutils/datatype.go
```diff
@@ -19,19 +18,18 @@ func TypeFromValue(v interface{}) types.DataType {
...
 	if valType.Kind() == reflect.Pointer {
-		if valType.Elem() != nil {
-			return TypeFromValue(reflect.ValueOf(v).Elem().Interface())
+		val := reflect.ValueOf(v)
+		if val.IsNil() {
+			return types.Null
 		}
-
-		return types.Null // Handle nil pointers
+		elem := val.Elem()
```
hash-data: any reason?

utils/typeutils/flatten_test.go
```diff
@@ -0,0 +1,507 @@
...
+	testEnabled = true
+	testScore   = 92.52361
+)
+
+var testTimestamp = time.Date(2023, 1, 1, 0, 0, 0, 0, time.UTC)
+
+// TestFlatten tests the public Flatten method of FlattenerImpl
+func TestFlatten(t *testing.T) {
+	tests := []struct {
+		name        string
```
hash-data, line 24: add comments on cases, for what case these are

utils/typeutils/resolver_test.go
```diff
@@ -0,0 +1,293 @@
...
+					"time_nanos":  time.Date(2024, 3, 19, 15, 30, 0, 123456789, time.UTC),
+					"nil_time":    (*time.Time)(nil),
+				},
+			},
+			expected: map[string]struct {
+				dataType types.DataType
+				nullable bool
+			}{
+				"time_sec":    {dataType: types.Timestamp, nullable: false},
+				"time_ms":     {dataType: types.TimestampMilli, nullable: false},
```
hash-data, line 226: have we added multiple types tests?

utils/typeutils/flatten_test.go
```diff
@@ -0,0 +1,507 @@
...
+			key:  "nested_map",
+			value: map[string]interface{}{
+				"level1": map[string]interface{}{
+					"level2": map[string]interface{}{
+						"value": "deep",
+					},
+				},
+			},
+			expected: types.Record{
+				"nested_map": `{"level1":{"level2":{"value":"deep"}}}`,
```
hash-data, line 333: have we added test for level 0 flattening check?

types/catalog_test.go
vaibhav-datazip: as it was discussed earlier 

initially the stream supported all modes and user had selected cdc mode

 then when updated and now it won't support cdc mode, but we won't change the user selected mode

types/catalog_test.go
vaibhav-datazip: Case for addition of new stream is already present, will add case for deletion of older stream.

utils/typeutils/datatype.go
vaibhav-datazip: this was due to a redundant check , have removed it

utils/typeutils/resolver_test.go
vaibhav-datazip: tes test cases have, strings ,arrays,  slices, timestamp types,

utils/typeutils/datatype.go
vaibhav-datazip: this test case will fail in resolver without this change

```
name: "pointer types",
			objects: []map[string]interface{}{
				{
					"int_ptr":     func() *int { v := 42; return &v }(),
					"string_ptr":  func() *string { v := "test"; return &v }(),
					"bool_ptr":    func() *bool { v := true; return &v }(),
					"float64_ptr": func() *float64 { v := 3.14; return &v }(),
					"nil_ptr":     (*int)(nil),
				},
			},
			expected: map[string]struct {
				dataType types.DataType
				nullable bool
			}{
				"int_ptr":     {dataType: types.Int32, nullable: false},
				"string_ptr":  {dataType: types.String, nullable: false},
				"bool_ptr":    {dataType: types.Bool, nullable: false},
				"float64_ptr": {dataType: types.Float64, nullable: false},
				"nil_ptr":     {dataType: types.Null, nullable: true},
			},
```

types/catalog_test.go
```diff
@@ -0,0 +1,956 @@
...
+					},
+				},
+				SelectedStreams: map[string][]StreamMetadata{
+					"namespace1": {
+						{StreamName: "stream1", PartitionRegex: "user_partition", Filter: "test_filter > 10", AppendMode: true, Normalization: true},
+					},
+				},
+			},
+		},
+		// when new stream is introduced, it should be added with new catalog metadata and selected stream data should be preserved
```
hash-data: did not get the comment

types/catalog_test.go
```diff
@@ -0,0 +1,956 @@
...
+					},
+				},
+				SelectedStreams: map[string][]StreamMetadata{
+					"namespace1": {
+						{StreamName: "stream1", PartitionRegex: "old_partition", Filter: "test_filter > 10", AppendMode: true, Normalization: true},
+					},
+				},
+			},
+		},
+		// when old stream is removed, only available streams should be kept, but selected stream data should be preserved
```
hash-data: this is also not understandable

types/catalog_test.go
```diff
@@ -0,0 +1,956 @@
...
+				},
+				SelectedStreams: map[string][]StreamMetadata{
+					"namespace1": {
+						{StreamName: "stream1", PartitionRegex: "user_partition", Filter: "test_filter > 10", AppendMode: true, Normalization: true},
+					},
+				},
+			},
+		},
+		// when destination database is updated, old catalog metadata should be preserved
+		{
```
hash-data, line 522: we still not added case when there is dest db present in two streams which is constant third stream also going to have constant

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,27 @@
+name: Unit Tests
+
+on:
+  push:
+    branches:
+      - staging
```
hash-data: let it run on all branches

utils/typeutils/flatten_test.go
```diff
@@ -0,0 +1,519 @@
...
+	"testing"
+	"time"
+
+	"github.com/datazip-inc/olake/types"
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	testUserID  = 12345
+	testName    = "John Doe"
```
hash-data: use olake as string names

utils/typeutils/flatten_test.go
```diff
@@ -0,0 +1,519 @@
...
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, tc.expected, result)
+			}
+		})
+	}
+}
+
+// TestFlattenInternal tests the private flatten method of FlattenerImpl
+func TestFlattenInternal(t *testing.T) {
```
hash-data, line 248: unable to see this function in main code

utils/typeutils/flatten_test.go
```diff
@@ -0,0 +1,519 @@
...
+			input:       types.Record{},
+			expected:    types.Record{},
+			expectError: false,
+		},
+		// basic values like string, int, bool, float, time, etc.
+		{
+			name: "simple values",
+			input: types.Record{
+				"name":    testName,
+				"age":     testAge,
```
hash-data: add some dates value

utils/typeutils/flatten_test.go
```diff
@@ -0,0 +1,519 @@
...
+				"🚀rocket🎯target":  "🌟success",
+				"key with spaces": "value with\nnewlines\tand\ttabs",
+			},
+			expected: types.Record{
+				"_rocket_target":  "🌟success",
+				"key_with_spaces": "value with\nnewlines\tand\ttabs",
+			},
+			expectError: false,
+		},
+		//  this is to check if lager numbers are loosing precision and scale.
```
hash-data:
```suggestion
		//  this is to check if larger numbers are loosing precision and scale.
```

utils/typeutils/resolver_test.go
```diff
@@ -0,0 +1,304 @@
...
+			dataType types.DataType
+			nullable bool
+		}
+	}{
+		// basic data types like int string float bool
+		{
+			name: "basic types",
+			objects: []map[string]interface{}{
+				{
+					"string_field": "test string",
```
hash-data: add timestamp

utils/typeutils/resolver_test.go
```diff
@@ -0,0 +1,304 @@
...
+						"Field %s should have type %s, got %v",
+						fieldName, expected.dataType, typeSet)
+
+					if expected.nullable {
+						assert.True(t, typeSet.Exists(types.Null),
+							"Field %s should be nullable", fieldName)
+					} else {
+						assert.False(t, typeSet.Exists(types.Null),
+							"Field %s should not be nullable", fieldName)
+					}
```
hash-data: cant we do equal check?

utils/typeutils/flatten_test.go
vaibhav-datazip: this is that function

Screenshot 2025-10-24 at 5 32 13 PM

types/catalog_test.go
vaibhav-datazip: I have added it for empty dest name , if third one has some dest name it will take empty string only.

utils/typeutils/flatten_test.go
hash-data: this is flatten right not flattenInternal

hash-data (approved):
lgtm

PR #280, opened 2025-05-12
fix: support MySQL RDS v8.4 for CDC

### LINKED ISSUE #243: [Feature Request] Support MySQL RDS v8.4 for CDC
### Problem

untill mysql rds v8, `SHOW MASTER STATUS` works, but not in v8.4 - https://dev.mysql.com/doc/refman/8.4/en/show-master-status.html

We need to check the version of mysql and then running the appropriate command in `/olake/pkg/jdbc/jdbc.go` function `func MySQLMasterStatusQuery`. 

The new command supported is `SHOW BINARY LOG STATUS`

This error occurs while performing CDC.

vishalm0509:
# Description

"SHOW MASTER STATUS" mysql command has been deprecated (https://dev.mysql.com/doc/refman/8.4/en/show-master-status.html). Instead "SHOW BINARY LOG STATUS" is used in mysql v8.2.0 and newer versions.

Fixes #243 

## Type of change

# How Has This Been Tested?

This has been tested with CDC and full_refresh modes using MySQL versions both above and below v8.2.0 as the source.

### DIFF
--- a/drivers/mysql/internal/cdc.go
+++ b/drivers/mysql/internal/cdc.go
@@ -192,15 +192,15 @@ func parseMySQLVersion(version string) (int, int, error) {
 	if len(parts) < 2 {
 		return 0, 0, fmt.Errorf("invalid version format")
 	}
-	major, err := strconv.Atoi(parts[0])
+	majorVersion, err := strconv.Atoi(parts[0])
 	if err != nil {
 		return 0, 0, fmt.Errorf("invalid major version: %s", err)
 	}
 
-	minor, err := strconv.Atoi(parts[1])
+	minorVersion, err := strconv.Atoi(parts[1])
 	if err != nil {
 		return 0, 0, fmt.Errorf("invalid minor version: %s", err)
 	}
 
-	return major, minor, nil
+	return majorVersion, minorVersion, nil
 }
--- a/drivers/mysql/internal/cdc.go
+++ b/drivers/mysql/internal/cdc.go
@@ -3,6 +3,7 @@ package driver
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/datazip-inc/olake/logger"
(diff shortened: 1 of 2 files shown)


drivers/mysql/internal/cdc.go
```diff
@@ -149,9 +149,14 @@ func (m *MySQL) RunChangeStream(pool *protocol.WriterPool, streams ...protocol.S
 
 // getCurrentBinlogPosition retrieves the current binlog position from MySQL.
 func (m *MySQL) getCurrentBinlogPosition() (mysql.Position, error) {
+	// MySQL v8.1 and below
 	rows, err := m.client.Query(jdbc.MySQLMasterStatusQuery())
 	if err != nil {
-		return mysql.Position{}, fmt.Errorf("failed to get master status: %s", err)
+		// MySQL v8.2 and above
+		rows, err = m.client.Query(jdbc.MySQLMasterStatusQueryNew())
```
vaibhav-datazip: A check can be Added here to run this fallback only when there is error to check syntax of MASTER STATUS . This will reduce unnecessary calls to this query in case of any other error.

drivers/mysql/internal/cdc.go
```diff
@@ -149,9 +149,14 @@ func (m *MySQL) RunChangeStream(pool *protocol.WriterPool, streams ...protocol.S
...
+		if err != nil {
+			return mysql.Position{}, fmt.Errorf("failed to get master status: %s", err)
```
hash-data: can we use version query here?

drivers/mysql/internal/cdc.go
```diff
@@ -149,9 +150,18 @@ func (m *MySQL) RunChangeStream(pool *protocol.WriterPool, streams ...protocol.S
 
 // getCurrentBinlogPosition retrieves the current binlog position from MySQL.
 func (m *MySQL) getCurrentBinlogPosition() (mysql.Position, error) {
+	// MySQL v8.1 and below
 	rows, err := m.client.Query(jdbc.MySQLMasterStatusQuery())
 	if err != nil {
```
ImDoubD-datazip, line 156: check the version first. 

You can do something like:
var version string
err := m.client.QueryRow("SELECT @@VERSION").Scan(&version)
, after checking this, then if the version is greater than or equal to 8.4, then use query "SHOW BINARY LOG STATUS" else use "SHOW MASTER STATUS". 

Do not use the binlog query in the error of "SHOW MASTER STATUS" query.

vishalm0509: I have made the changes accordingly.

ImDoubD-datazip (dismissed):

drivers/mysql/internal/cdc.go
```diff
@@ -149,7 +151,23 @@ func (m *MySQL) RunChangeStream(pool *protocol.WriterPool, streams ...protocol.S
...
+	// SHOW MASTER STATUS is not supported in MySQL 8.4 and after
+	// So check the version and use the appropriate query
+
+	// Get MySQL version
+	version, err := jdbc.MySQLVersion(m.client)
+	if err != nil {
+		return mysql.Position{}, fmt.Errorf("failed to get MySQL version: %s", err)
+	}
+	// Parse MySQL version to get major and minor version
+	majorVersion, minorVersion, err := parseMySQLVersion(version)
```
hash-data: can we just create one function to get parsed version? 
Like right now 2 different function are there one to get and one to parse it.

vishalm0509: I have made the changes accordingly.

PR #290, opened 2025-05-19
feat: Telemetry Setup Using Segment IO and Mixpanel

ImDoubD-datazip:
# Description
Telemetry setup is done using Segment IO and Mixpanel, where required metrics are sent as events to Segment which directs it to Mixpanel. The metrics covered includes:
discover : stream count total, connector type, duration for which discover is run
sync : stream count total, enabled stream, streams with sync CDC mode, state file provided, sync type ( w.r.t source), source type, destination type , normalised streams , partition streams, sync count - success and fails weekly, sync final status, no of records synced , duration, memory usage in MB, threads used, device , location.

Note: Currently been setup only for CLI V.0, needs to be setup for UI V.0 

## Type of change

# How Has This Been Tested?
This has been tested on multiple syncs where the events are first seen at Segment IO and then at Mixpanel.

# Screenshots or Recordings
Segment IO : \
Screenshot 2025-05-19 at 6 55 07 PM
Mixpanel : \
Screenshot 2025-05-19 at 6 55 40 PM


protocol/sync.go
```diff
@@ -60,8 +69,43 @@ var syncCmd = &cobra.Command{
...
+		telemetryClient := telemetry.GetInstance()
+		configHash := telemetry.ComputeConfigHash(configPath, destinationConfigPath)
+		defer func() {
+			metrics := telemetryClient.TrackSyncResult(configHash, syncMetrics.success)
+			props := map[string]interface{}{
+				"success":         syncMetrics.success,
+				"records_synced":  syncMetrics.records,
+				"duration_sec":    time.Since(startTime).Seconds(),
+				"memory_usage_mb": syncMetrics.memoryUsageMB,
+				"threads_used":    pool.threadCounter.Load(),
```
hash-data: can we move this to telemetry file itself?

hash-data: we already have stats logger that we can use for logging it ?

hash-data: for sync fail and success we can create a telemetry function that will be called via defer and provide info.

protocol/sync.go
```diff
@@ -129,6 +173,34 @@ var syncCmd = &cobra.Command{
...
+			"unique_config_dstination": configHash,
+			"sync_type":                syncType,
+			"source_type":              connector.Type(),
+			"destination_type":         destinationConfig.Type,
+			"catalog_type":             catalogType,
+			"normalized_streams":       countNormalizedStreams(catalog),
+			"partitioned_streams":      countPartitionedStreams(catalog),
+		}); err != nil {
+			fmt.Printf("Failed to send telemetry event SyncStarted: %v\n", err)
+		}
```
hash-data: move to telemetry file and call a function from here.

protocol/sync.go
```diff
@@ -164,17 +236,73 @@ var syncCmd = &cobra.Command{
...
+	count := 0
+	for _, s := range catalog.Streams {
+		if s.StreamMetadata.PartitionRegex != "" {
+			count++
+		}
+	}
+	return count
+}
+
+func getCatalogType(destinationConfig *types.WriterConfig) (string, error) {
```
hash-data: is it required or we can use the unmarshalling of destination config from NewWriter.

telemetry/telemetry.go
```diff
@@ -0,0 +1,382 @@
+package telemetry
```
hash-data: move package to logger

telemetry/telemetry.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+
+	telemetryEnabled = viper.GetBool("telemetry.enabled")
+	segmentAPIKey = viper.GetString("telemetry.segment_api_key")
+	deploymentType = viper.GetString("telemetry.deployment_type")
+	serviceName = viper.GetString("telemetry.service_name")
+}
+
+func createTelemetry() {
+	loadConfig()
```
hash-data: why new function ?

telemetry/telemetry.go
```diff
@@ -0,0 +1,382 @@
...
+
+	ip := getOutboundIP()
+	instance = &Telemetry{
+		client:       client,
+		serviceName:  getServiceName(),
+		enabled:      isTelemetryEnabled(),
+		platform:     getPlatformInfo(),
+		ipAddress:    ip,
+		locationChan: make(chan struct{}), // INITIALIZE THE CHANNEL
+	}
```
hash-data: init a global telemetry client like logger

telemetry/telemetry.go
```diff
@@ -0,0 +1,382 @@
...
+
+	if instance.enabled && instance.client != nil {
+		_ = instance.SendEvent("TelemetryInitialized", map[string]interface{}{
+			"ipAddress": ip,
+			"timestamp": time.Now().String(),
+		})
+		if ip != ipNotFoundPlaceholder {
+			go func() {
```
hash-data: why gorotine here ?

telemetry/telemetry.go
```diff
@@ -0,0 +1,382 @@
...
+				close(instance.locationChan)
+			}
+		}
+	}
+}
+
+func GetInstance() *Telemetry {
+	once.Do(createTelemetry)
+	return instance
+}
```
hash-data: cration of global telemetry object will remove this.

telemetry/telemetry.go
```diff
@@ -0,0 +1,382 @@
...
+		err := t.client.Close()
+		if err != nil {
+			fmt.Printf("Error flushing telemetry: %v\n", err)
+		} else {
+			fmt.Println("Telemetry events flushed successfully")
+		}
+	}
+}
+
+func (t *Telemetry) SendEvent(eventName string, properties map[string]interface{}) error {
```
hash-data: check for nill client here.

telemetry/telemetry.go
```diff
@@ -0,0 +1,382 @@
...
+	if err == nil {
+		ip = ipBody
+	}
+
+	return string(ip)
+}
+
+func getLocationFromIP(ip string) (LocationInfo, error) {
+	client := http.Client{Timeout: 1 * time.Second}
+	resp, err := client.Get(fmt.Sprintf("https://ipinfo.io/%s/json", ip))
```
hash-data: use context while hitting request outside.

telemetry/telemetry.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+
+	return LocationInfo{
+		Country: info.Country,
+		Region:  info.Region,
+		City:    info.City,
+	}, nil
+}
+
+func (t *Telemetry) getLocationWithTimeout() interface{} {
```
hash-data: why two different functions?

telemetry/telemetry.go
```diff
@@ -0,0 +1,382 @@
...
+
+	if err := viper.ReadInConfig(); err != nil {
+		fmt.Printf("Error reading config file: %v\n", err)
+		return
+	}
+
+	telemetryEnabled = viper.GetBool("telemetry.enabled")
+	segmentAPIKey = viper.GetString("telemetry.segment_api_key")
+	deploymentType = viper.GetString("telemetry.deployment_type")
+	serviceName = viper.GetString("telemetry.service_name")
```
hash-data: why we are taking config ? we can hardcode these. As it is not option it is required.

hash-data: also only api key required.

protocol/sync.go
ImDoubD-datazip: Stats logger doesn't has all these things what the product side wants.
But we can create a function and move it to the telemetry file, and provide logic of whatever stats we need in that function itself. all this will then happen via one method present in telemetry.go

telemetry/telemetry.go
ImDoubD-datazip: I believe telemetry should remain in a separate package, because then some new method can come and it will keep everything separated and clean

telemetry/telemetry.go
ImDoubD-datazip: This method kind of provides information that telemetry has been started. So config method and using that config method have been kept separated.

telemetry/telemetry.go
ImDoubD-datazip: it is used to run the location lookup in async way. In some cases, the network itself can be slow. So this will not block rest of the telemetry initialization

telemetry/telemetry.go
ImDoubD-datazip: ok

telemetry/telemetry.go
ImDoubD-datazip: getLocationFromIP is the helper that performs the actual network request to fetch the location based on the IP address. It has its own 1‑second timeout for the HTTP client.

getLocationWithTimeout is used when sending telemetry events. It waits for up to 200 ms for an asynchronous location lookup (triggered via a goroutine that calls getLocationFromIP) to complete. If the location isn’t available by then, it simply returns "NA". This ensures the telemetry event isn’t delayed by a slow location lookup.

telemetry/telemetry.go
ImDoubD-datazip: yes we can hard-code it. As of now also, its been hard-coded but just called via viper.

telemetry/telemetry.go
ImDoubD-datazip: now init is used.

protocol/sync.go
ImDoubD-datazip: What difference will it make, if I change to this way?
I believe it will give the same result.

hash-data: redundant code will be removed

ImDoubD-datazip: done changes

protocol/discover.go
```diff
@@ -25,17 +27,36 @@ var discoverCmd = &cobra.Command{
...
+				discoverError == nil,
+				streamCount,
+				connector.Type(),
+				discoverError,
+			)
+			telemetry.Flush()
+		}()
 		err := connector.Setup(cmd.Context())
 		if err != nil {
+			discoverError = err
```
hash-data: make err as a variable

.gitignore
```diff
@@ -15,4 +15,4 @@ olake-iceberg-java-writer.jar
 local-releaser.sh
 destination/iceberg/olake-iceberg-java-writer/.idea/
 destination/iceberg/olake-iceberg-java-writer/dependency-reduced-pom.xml
-
+config-telemetry.yaml
```
hash-data: remove it

protocol/discover.go
```diff
@@ -37,8 +40,17 @@ var discoverCmd = &cobra.Command{
...
 		types.LogCatalog(streams)
+
+		// Discover Telemetry Tracking
+		defer func() {
+			telemetry.TrackDiscover(
+				time.Since(startTime).Seconds(),
+				len(streams),
+				connector.Type(),
+			)
+			telemetry.Flush()
```
hash-data: track discover can also call flush

protocol/sync.go
```diff
@@ -42,6 +49,8 @@ var syncCmd = &cobra.Command{
 			return err
 		}
 
+		syncTelemetryDetails.configHash = telemetry.ComputeConfigHash(configPath, destinationConfigPath)
```
hash-data: make it sync id

protocol/sync.go
```diff
@@ -4,15 +4,22 @@ import (
...
 	"github.com/datazip-inc/olake/utils/logger"
 	"github.com/spf13/cobra"
 )
 
+var syncTelemetryDetails struct {
+	start      time.Time
+	configHash string
+	status     bool
+}
+
```
hash-data: try removing it

protocol/sync.go
```diff
@@ -60,16 +69,19 @@ var syncCmd = &cobra.Command{
 	RunE: func(cmd *cobra.Command, _ []string) error {
 		pool, err := destination.NewWriter(cmd.Context(), destinationConfig)
 		if err != nil {
+			syncTelemetryDetails.status = false
```
hash-data: why ?

hash-data: check directly err == nil means success or failed

protocol/sync.go
```diff
@@ -142,13 +154,37 @@ var syncCmd = &cobra.Command{
...
 		connector.SetupState(state)
+		syncTelemetryDetails.start = time.Now()
+		// Sync Telemetry tracking
+		telemetry.TrackSyncStarted(
+			streams,
+			selectedStreams,
+			cdcStreams,
+			syncTelemetryDetails.configHash,
+			connector.Type(),
+			string(destinationConfig.Type),
```
hash-data: if you are passing destination config why you cant get its type inside function ?

protocol/sync.go
```diff
@@ -142,13 +154,37 @@ var syncCmd = &cobra.Command{
 
 		// Setup State for Connector
 		connector.SetupState(state)
+		syncTelemetryDetails.start = time.Now()
+		// Sync Telemetry tracking
+		telemetry.TrackSyncStarted(
+			streams,
```
hash-data: always use ctx

protocol/sync.go
```diff
@@ -142,13 +154,37 @@ var syncCmd = &cobra.Command{
...
+			connector.Type(),
+			string(destinationConfig.Type),
+			destinationConfig,
+			catalog,
+		)
+		defer func() {
+			telemetry.TrackSyncCompleted(
+				syncTelemetryDetails.status,
+				pool.SyncedRecords(),
+				time.Since(syncTelemetryDetails.start).Seconds(),
```
hash-data: just store current time, no need to store seconds, start and end is being stored that is fine

telemetry/telemetry.go
```diff
@@ -0,0 +1,377 @@
...
+	viper.SetConfigType("yaml")
+	viper.AddConfigPath("../../")
+
+	if err := viper.ReadInConfig(); err == nil {
+		telemetryEnabled = viper.GetBool("telemetry.enabled")
+		segmentAPIKey = viper.GetString("telemetry.segment_api_key")
+		deploymentType = viper.GetString("telemetry.deployment_type")
+		serviceName = viper.GetString("telemetry.service_name")
+	}
+}
```
hash-data: we not require any configuration for it so we can remove it

telemetry/telemetry.go
```diff
@@ -0,0 +1,377 @@
...
+)
+
+var (
+	client           analytics.Client
+	idLock           sync.Mutex
+	telemetryEnabled bool
+	deploymentType   string
+	serviceName      string
+	segmentAPIKey    string
+	instance         *Telemetry
```
hash-data: rename it to client

telemetry/telemetry.go
```diff
@@ -0,0 +1,377 @@
...
+		segmentAPIKey = viper.GetString("telemetry.segment_api_key")
+		deploymentType = viper.GetString("telemetry.deployment_type")
+		serviceName = viper.GetString("telemetry.service_name")
+	}
+}
+
+func init() {
+	loadConfig()
+	ip := getOutboundIP()
+	enabled := isTelemetryEnabled()
```
hash-data: it is always enabled so need to take it

hash-data: remove enabled check from everywhere

telemetry/telemetry.go
```diff
@@ -0,0 +1,377 @@
...
+			close(instance.locationChan)
+		}
+	}
+}
+
+func TrackDiscover(duration float64, streamCount int, sourceType string) {
+	if instance == nil || !instance.enabled {
+		return
+	}
+	instance.wg.Add(1)
```
hash-data: not required to maintain wait groups

hash-data: if event goes fine if not also fine

hash-data: no wait should happen for it

hash-data (commented):
idea is to remove wait group and add 5 second wait telling user sync completed 5 sec wait for clean up

protocol/sync.go
```diff
@@ -4,15 +4,18 @@ import (
...
 	"github.com/datazip-inc/olake/destination"
-
+	"github.com/datazip-inc/olake/telemetry"
 	"github.com/datazip-inc/olake/types"
 	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/logger"
 	"github.com/spf13/cobra"
 )
 
+var syncID string
```
hash-data: move it to root command

telemetry/telemetry.go
```diff
@@ -0,0 +1,289 @@
...
+	client := analytics.New(segmentAPIKey)
+
+	telemetry = &Telemetry{
+		client:       client,
+		platform:     getPlatformInfo(),
+		ipAddress:    ip,
+		locationChan: make(chan struct{}),
+	}
+
+	if ip != ipNotFoundPlaceholder {
```
hash-data: move to goroutine

telemetry/telemetry.go
```diff
@@ -0,0 +1,289 @@
...
+func getPlatformInfo() platformInfo {
+	return platformInfo{
+		OS:           runtime.GOOS,
+		Arch:         runtime.GOARCH,
+		OlakeVersion: version,
+		DeviceCPU:    fmt.Sprintf("%d cores", runtime.NumCPU()),
+	}
+}
+
+func ComputeConfigHash(srcPath, destPath string) string {
```
hash-data: move to utils

telemetry/telemetry.go
```diff
@@ -0,0 +1,289 @@
...
+		platform:     getPlatformInfo(),
+		ipAddress:    ip,
+		locationChan: make(chan struct{}),
+	}
+
+	if ip != ipNotFoundPlaceholder {
+		go func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+			defer cancel()
+			location, err := getLocationFromIP(ctx, ip)
```
hash-data: let us only do one request if data come fine else store na

telemetry/telemetry.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func countNormalizedStreams(catalog *types.Catalog) int {
+	count := 0
+	for _, s := range catalog.Streams {
+		if s.StreamMetadata.Normalization {
+			count++
+		}
+	}
+	return count
```
hash-data: use utility

telemetry/telemetry.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func countPartitionedStreams(catalog *types.Catalog) int {
+	count := 0
+	for _, s := range catalog.Streams {
+		if s.StreamMetadata.PartitionRegex != "" {
+			count++
+		}
+	}
```
hash-data: utility

telemetry/telemetry.go
```diff
@@ -0,0 +1,289 @@
...
+	locationChan  chan struct{}
+}
+
+var telemetry *Telemetry
+
+const (
+	anonymousIDFile       = "telemetry_id"
+	version               = "0.0.0"
+	ipNotFoundPlaceholder = "NA"
+	segmentAPIKey         = "RFynIrEFUsaTRgKumlSqfme6DRPymYfN" //nolint:gosec
```
hash-data: move to top level constants

telemetry/telemetry.go
```diff
@@ -0,0 +1,289 @@
...
+)
+
+type Telemetry struct {
+	client        analytics.Client
+	serviceName   string
+	platform      platformInfo
+	ipAddress     string
+	locationInfo  *LocationInfo
+	locationMutex sync.Mutex
+	locationChan  chan struct{}
```
hash-data: remove mutex and channels

telemetry/telemetry.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	return t.client.Enqueue(analytics.Track{
+		UserId:     fmt.Sprintf("olake-cli-%s", time.Now()),
+		Event:      eventName,
+		Properties: props,
+	})
+}
+
+func getPlatformInfo() platformInfo {
```
hash-data: remove this

telemetry/telemetry.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/datazip-inc/olake/types"
+	"github.com/datazip-inc/olake/utils"
+	"github.com/datazip-inc/olake/utils/logger"
+	analytics "github.com/segmentio/analytics-go/v3"
+)
+
+const (
+	anonymousIDFile       = "telemetry_id"
+	version               = "0.0.0"
+	ipNotFoundPlaceholder = "NA"
```
hash-data: remove it

hash-data (approved):
LGTM

Issue #160, opened 2025-03-17
Incremental Synch for Postgres


rkhameshra:
**Background**
OLake currently supports Change Data Capture (CDC) for PostgreSQL, enabling efficient incremental data synchronization via logical replication (WAL-based). However, many PostgreSQL users—especially those on managed cloud databases like Amazon RDS, Google Cloud SQL, and Azure Database for PostgreSQL—may not have the necessary privileges to enable logical replication or WAL-based CDC.

For these users, OLake should support an alternative incremental sync method that does not require CDC but still efficiently captures newly inserted and updated records.

**Feature Scope**
This feature will introduce a query-based incremental sync mechanism that allows OLake to track and sync changes without requiring database-level replication privileges. Possible approaches include:

**1. Timestamp-based Sync**
- Users specify a column (e.g., updated_at, last_modified, or similar) that OLake will use to track changes.
- On each sync run, OLake will query for records where the timestamp is greater than the last synced timestamp.
- Requires users to ensure that an appropriate timestamp column is maintained with updates.

**2. Primary Key Auto-Increment Tracking (For Append-Only Tables)**
- For tables where only new rows are inserted and existing records are never updated, OLake can track changes using an auto-incremented primary key (e.g., id).
- On each sync, OLake will fetch rows where the primary key is greater than the last synced value.
- This method is not suitable for tables with updates or deletions.

**3. Soft Delete Handling (Optional)**
- Some tables use a deleted_at column instead of hard deletes. OLake can track deleted rows by detecting non-null values in this column and propagating deletions downstream.
- Alternatively, users can specify a flag column (e.g., is_deleted = true) to indicate deletions.

**Implementation Details**

**1. User Configuration:**
- Allow users to specify tracking columns (e.g., updated_at, id, or deleted_at).
- Provide default behavior where possible but allow customization.

**2. Efficient Query Execution:**
- Ensure that incremental queries leverage indexes (e.g., B-tree indexes on timestamps, primary keys).
- Support pagination (e.g., using LIMIT/OFFSET or cursor-based iteration) to handle large datasets.
- Implement batch processing to optimize performance and reduce database load.

**3. Checkpointing & State Management:**
- OLake should maintain a sync state per table, tracking the last processed timestamp/ID.
- Ensure robustness by handling failures and retries without duplicate processing.

**Fallback Mechanisms:**
- If the specified tracking column is missing or unreliable, OLake should either:
- Recommend a fallback full sync (configurable).
[...]

PR #291, opened 2025-05-20
[WIP] feat: incremental sync for postgres

Jougan-0:
**NOTE : Don't review waiting for MongoDb incremental sync to be merged.**

Implements mixed-cursor incremental sync to resolve Issue #160.

Key updates

- Incremental sync now handles both timestamp/int and UUID cursors
- uses MAX() for aggregatable types, or ORDER BY … DESC LIMIT 1 for UUID/JSON
- auto-selects created_at when present, otherwise respects the catalog’s cursor
- Discover phase now populates available_cursor_fields, enabling smarter cursor selection

### DIFF
--- a/drivers/postgres/internal/incremental.go
+++ b/drivers/postgres/internal/incremental.go
@@ -1,0 +1,171 @@
+// File: driver/incremental.go
+package driver
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/datazip-inc/olake/logger"
+	"github.com/datazip-inc/olake/pkg/jdbc"
+	"github.com/datazip-inc/olake/protocol"
+	"github.com/datazip-inc/olake/types"
+	"github.com/datazip-inc/olake/utils"
+	"github.com/lib/pq"
+)
+
+const (
+	bmkField      = "cursor_field"
+	bmkVal        = "cursor_value"
+	defaultCursor = "created_at"
+)
+
+func (p *Postgres) incremental(pool *protocol.WriterPool, stream protocol.Stream) error {
+	ctx := context.TODO()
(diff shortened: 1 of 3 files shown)


drivers/postgres/internal/incremental.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return err
+	}
+	defer writer.Close()
+
+	lastSeen := bookmark
+	for {
+		q := fmt.Sprintf(
+			`SELECT * FROM %s.%s
+			  WHERE %s > $1
```
loicalleyne, line 65: There is a possibility that you will lose some rows with the same due to the `LIMIT` clause if the cursor column timestamps are the same for multiple rows. 

In my [`leftshove`](https://github.com/loicalleyne/leftshove) incremental sync implementation, I select a time window between a start and end time
(see [here](https://github.com/loicalleyne/leftshove/blob/689db833a10929e07738f5c303b3fb33ca0f5a5a/postgres.go#L43)), and also have optional logic to [adjust the window](https://github.com/loicalleyne/leftshove/blob/master/cdc.go) in case of replication delay (in case the database being queried is a replica).

drivers/postgres/internal/postgres.go
```diff
@@ -223,8 +229,15 @@ func (p *Postgres) populateStream(table Table) (*types.Stream, error) {
 		}
 
 		stream.UpsertField(typeutils.Reformat(column.Name), datatype, strings.EqualFold("yes", *column.IsNullable))
+		if datatype == types.Timestamp {
```
loicalleyne, line 232: It might be interesting to also support doing incremental sync based on an integer cursor (as in the case of tables that do not have a timestamp column and have a sequence as primary key).

PR #292, opened 2025-05-21
chore: refactor v2

hash-data:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/drivers/abstract/abstract.go
+++ b/drivers/abstract/abstract.go
@@ -116,13 +116,17 @@ func (a *AbstractDriver) Read(ctx context.Context, pool *destination.WriterPool,
 		a.GlobalConnGroup = utils.NewCGroupWithLimit(ctx, a.driver.MaxConnections())
 	}
 
-	if a.driver.CDCSupported() {
-		if err := a.RunChangeStream(ctx, pool, cdcStreams...); err != nil {
-			return fmt.Errorf("failed to run change stream: %s", err)
+	// run cdc sync
+	if len(cdcStreams) > 0 {
+		if a.driver.CDCSupported() {
+			if err := a.RunChangeStream(ctx, pool, cdcStreams...); err != nil {
+				return fmt.Errorf("failed to run change stream: %s", err)
+			}
+		} else {
+			return fmt.Errorf("%s cdc configuration not provided, use full refresh for all streams", a.driver.Type())
 		}
-	} else {
-		return fmt.Errorf("CDC configuration is not provided, use full refresh for all streams")
 	}
+
 	// start backfill for standard streams
 	for _, stream := range standardStreams {
 		a.GlobalCtxGroup.Add(func(ctx context.Context) error {
--- a/drivers/abstract/cdc.go
+++ b/drivers/abstract/cdc.go
@@ -11,11 +11,6 @@ import (
 )
 
 func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.WriterPool, streams ...types.StreamInterface) error {
-	// check streams
-	if len(streams) == 0 {
-		return fmt.Errorf("no streams provided for change stream")
-	}
-
 	// run pre cdc of drivers
 	if err := a.driver.PreCDC(ctx, a.state, streams); err != nil {
 		return fmt.Errorf("failed in pre cdc run for driver[%s]: %s", a.driver.Type(), err)
--- a/drivers/postgres/internal/cdc.go
+++ b/drivers/postgres/internal/cdc.go
@@ -60,6 +60,7 @@ func (p *Postgres) PreCDC(ctx context.Context, state *types.State, streams []typ
 			if err != nil {
 				return fmt.Errorf("failed to parse stored lsn[%s]: %s", postgresGlobalState.LSN, err)
 			}
+			// TODO: handle cursor mismatch with user input (Example: user provide if it has to fail or do full load with new resume token)
 			if parsed != currentLSN {
 				logger.Warnf("lsn mismatch, backfill will start again. prev lsn [%s] current lsn [%s]", parsed, currentLSN)
 				state.SetGlobal(waljs.WALState{LSN: currentLSN.String()})


README.md
```diff
@@ -37,7 +37,7 @@ Connector ecosystem for Olake, the key points Olake Connectors focuses on are th
 ### Writers / Destination
 1. [Apache Iceberg Docs](https://olake.io/docs/category/apache-iceberg) 
 2. [AWS S3 Docs](https://olake.io/docs/category/aws-s3) 
-3. [Local FileSystem Docs](https://olake.io/docs/writers/local) 
+3. [Local FileSystem Docs](https://olake.io/docs/destination/local)
```
shubham19may, line 40: Screenshot 2025-05-30 at 10 34 25 PM
some of the places destination is still writers

destination/writers.go
```diff
@@ -202,8 +206,8 @@ func (w *WriterPool) NewThread(parent context.Context, stream Stream, options ..
 							return err
 						}
 						w.recordCount.Add(1) // increase the record count
-						if w.SyncedRecords()%batchSize == 0 {
-							state.LogWithLock()
+						if w.SyncedRecords()%w.batchSize == 0 {
+							w.state.LogWithLock()
```
shubham19may, line 210: benefit of this if we are only changing state when chunk finishes or cdc reaches to last LSN?
lets mark then only right?

drivers/abstract/abstract.go
```diff
@@ -0,0 +1,137 @@
...
+	constants.OlakeTimestamp: types.Int64,
+	constants.OpType:         types.String,
+	constants.CdcTimestamp:   types.Int64,
+}
+
+func NewAbstractDriver(ctx context.Context, driver DriverInterface) *AbstractDriver {
+	return &AbstractDriver{
+		driver:          driver,
+		GlobalCtxGroup:  utils.NewCGroup(ctx),
+		GlobalConnGroup: utils.NewCGroupWithLimit(ctx, 10), // default max connections
```
shubham19may: lets change it to 3 as default

drivers/mongodb/internal/mon.go
```diff
@@ -125,82 +108,72 @@ func (m *Mongo) Discover(ctx context.Context) ([]*types.Stream, error) {
...
 			}
-		}
 
-		return cursor.Err()
-	})
+			return cursor.Err()
+		})
+	}
+	database := m.client.Database(m.config.Database)
+	// Either wait for covering 100k records from both sides for all streams
```
shubham19may, line 169: ? we are not passing any discoverCtx with timeouts

drivers/mysql/internal/mysql.go
```diff
@@ -130,88 +114,75 @@ func (m *MySQL) Discover(ctx context.Context) ([]*types.Stream, error) {
...
+		// TODO: Populate cursor fields for incremental purpose
 
-	if err := rows.Err(); err != nil {
-		return nil, fmt.Errorf("error iterating rows: %w", err)
-	}
-	// TODO: Populate cursor fields for incremental purpose
+		// Add CDC columns if supported
+		if m.CDCSupport {
+			for column, typ := range abstract.DefaultColumns {
+				stream.UpsertField(column, typ, true)
```
shubham19may: ? this is already done in abstract code right?

drivers/abstract/abstract.go
```diff
@@ -0,0 +1,137 @@
...
+	if a.driver.MaxConnections() > 0 {
+		a.GlobalConnGroup = utils.NewCGroupWithLimit(ctx, a.driver.MaxConnections())
+	}
+
+	if a.driver.CDCSupported() {
+		if err := a.RunChangeStream(ctx, pool, cdcStreams...); err != nil {
+			return fmt.Errorf("failed to run change stream: %s", err)
+		}
+	} else {
+		return fmt.Errorf("CDC configuration is not provided, use full refresh for all streams")
```
shubham19may: what if there are no streams marked cdc?

destination/writers.go
hash-data: not removing as of now, to continue with prev arch.

drivers/abstract/abstract.go
hash-data: there is check in runchangestream function

drivers/mysql/internal/mysql.go
hash-data: yeah forgot to remove from here.

drivers/mongodb/internal/mon.go
hash-data: yeah i thought to add in abstract and forgor later, adding it.

README.md
hash-data: currently only changed folder structure later on we can adopt the name all over project.

drivers/abstract/abstract.go
shubham19may: @hash-data  but this shouldn't return error right?

protocol/check.go
```diff
@@ -33,48 +33,9 @@ var checkCmd = &cobra.Command{
...
-				} else if len(missingStreams) > 0 {
-					return fmt.Errorf("found missing streams: %v", missingStreams)
-				}
-			} else {
-				// Only perform checks
-				err := connector.Check(cmd.Context())
-				if err != nil {
-					return err
-				}
+			err := connector.Setup(cmd.Context())
```
shubham19may, line 36: there is a difference between setup and check right (In logical terms based on their working)?

basically if we are removing check and just doing setup then it should also check
1. Connection successful or not
2. read from source is working on couple of tables selected
3. write to destination is happening on respective destination config

drivers/abstract/cdc.go
```diff
@@ -0,0 +1,136 @@
...
+
+	"github.com/datazip-inc/olake/constants"
+	"github.com/datazip-inc/olake/destination"
+	"github.com/datazip-inc/olake/types"
+	"github.com/datazip-inc/olake/utils"
+)
+
+func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.WriterPool, streams ...types.StreamInterface) error {
+	// check streams
+	if len(streams) == 0 {
```
shubham19may: what if no streams was selected in cdc mode? this function will still be called with 0 array.

shubham19may (approved):

PR #293, opened 2025-05-21
fix: bigint precision loss

### LINKED ISSUE #275: BigInt value difference with postgres in Parquet Writer
## Description
Currently value of bigint datatype column is different than that of the present in postgres database value. After debugging we got to the conclusion that this needs to be fixed by confirming with catalog.

Database: 
Image

Parquet Writer:
Image

## Steps to reproduce
Create a table in Postgres db with one of the cols datatype as bigint. Insert value with more that 15-16 digits. Run the full sync on that db on Parquet Writer. Then run cdc sync. Check value in both the cases.

## Expected behavior
The value present in db, exactly same value must be written in Parquet writer

vishalm0509:
# Description

When decoding JSON in Go, numeric values are automatically converted to float64 by default. This causes precision loss for large integers, such as PostgreSQL BIGINT values (e.g., 9999999999999999), because float64 cannot represent all large integers exactly. 

The fix uses decoder.useNumber() to decode numbers as json.Number (string form), allowing safe conversion to int64 or string without rounding errors.

Fixes #275

## Type of change

# How Has This Been Tested?

- [x]  Scenario full_refresh
- [x] Scenario CDC

# Screenshots or Recordings

**CDC (before fix)**
Screenshot 2025-05-23 at 1 58 47 PM

**CDC**
Screenshot 2025-05-23 at 2 03 39 PM

## Documentation

---

> [!NOTE]
> Decode WAL JSON numbers as json.Number and add targeted conversion in type reformatter to avoid float64 precision loss (e.g., BIGINT).
> 
> - **WAL JSON decoding (`pkg/waljs/filter.go`)**:
>   - Use `decoder.UseNumber()` when decoding `WALMessage` to keep numbers as `json.Number` instead of `float64`.
> - **Type reformatting (`utils/typeutils/reformat.go`)**:
>   - Add `json.Number` handling in `ReformatValue` to convert to `Int32`/`Int64`/`Float32`/`Float64`, or fallback to string, preserving precision.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 93347662930b1a3f9f5e594e643cbb1b68b5f037. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/waljs/filter.go
```diff
@@ -31,7 +31,10 @@ func NewChangeFilter(typeConverter func(value interface{}, columnType string) (i
 
 func (c ChangeFilter) FilterChange(lsn pglogrepl.LSN, change []byte, OnFiltered OnMessage) error {
 	var changes WALMessage
-	if err := json.NewDecoder(bytes.NewReader(change)).Decode(&changes); err != nil {
+	decoder := json.NewDecoder(bytes.NewReader(change))
```
ImDoubD-datazip, line 35: error handling needed for decoder

typeutils/reformat.go
```diff
@@ -63,6 +64,37 @@ func ReformatValueOnDataTypes(datatypes []types.DataType, v any) (any, error) {
...
+			return floatVal, nil
+		case types.Float32:
+			floatVal, err := num.Float64()
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse numeric: %v", err)
+			}
+			return float32(floatVal), nil
+		}
+	}
+
```
ImDoubD-datazip, line 103: Here move the Int32 and Float32 to top. First float32 then float64, int32 then int64
check whether the float precision can be solved through this?

pkg/waljs/filter.go
vishalm0509: func json.NewDecoder(r io.Reader) *json.Decoder
NewDecoder returns a new decoder that reads from r.

The decoder introduces its own buffering and may read data from r beyond the JSON values requested.

It does not return an error.

vishalm0509: If any error occurs here, it'll be handled when we decode it

ImDoubD-datazip (dismissed):

pkg/waljs/filter.go
```diff
@@ -31,7 +31,10 @@ func NewChangeFilter(typeConverter func(value interface{}, columnType string) (i
...
+	// Use UseNumber() to convert numbers into json.Number and not float64.
+	decoder.UseNumber()
+	if err := decoder.Decode(&changes); err != nil {
```
hash-data, line 38: we can directly use 
 json.NewDecoder(bytes.NewReader(change)).useNumber().Decode() ???

vishalm0509: json.NewDecoder(bytes.NewReader(change)).UseNumber() (no value) used as valuecompiler[TooManyValues](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#TooManyValues)

vaibhav-datazip:
This issue occurred only when running CDC sync using the Wal2JSON. Since Wal2JSON has now been deprecated in Olake and the system has migrated to pgoutput, this problem no longer exists.

PR #298, opened 2025-05-27
fix(source): full refresh

vikash390:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/protocol/sync.go
+++ b/protocol/sync.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/datazip-inc/olake/logger"
@@ -110,12 +111,20 @@ var syncCmd = &cobra.Command{
 
 			elem.StreamMetadata = sMetadata
 			selectedStreams = append(selectedStreams, elem.ID())
-
+			var cdcSelectedStreams []*types.StreamState
 			if elem.Stream.SyncMode == types.CDC {
 				cdcStreams = append(cdcStreams, elem)
+				cdcSelectedStreams = append(cdcSelectedStreams, &types.StreamState{
+					Namespace:  elem.Namespace(),
+					Stream:     elem.Name(),
+					State:      sync.Map{},
+					HoldsValue: atomic.Bool{},
+					SyncMode:   "cdc",
+				})
 			} else {
 				standardModeStreams = append(standardModeStreams, elem)
 			}
+			state.Streams = cdcSelectedStreams
 
 			return false
 		})
--- a/protocol/sync.go
+++ b/protocol/sync.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/datazip-inc/olake/logger"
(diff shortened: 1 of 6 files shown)


protocol/sync.go
```diff
@@ -110,12 +111,20 @@ var syncCmd = &cobra.Command{
...
+			var cdcSelectedStreams []*types.StreamState
 			if elem.Stream.SyncMode == types.CDC {
 				cdcStreams = append(cdcStreams, elem)
+				cdcSelectedStreams = append(cdcSelectedStreams, &types.StreamState{
+					Namespace:  elem.Namespace(),
+					Stream:     elem.Name(),
+					State:      sync.Map{},
+					HoldsValue: atomic.Bool{},
+					SyncMode:   "cdc",
+				})
```
hash-data: what is benefit of it? you have not used the state that is being passed??

types/state.go
```diff
@@ -112,6 +112,9 @@ func (s *State) GetChunks(stream *ConfiguredStream) *Set[Chunk] {
 
 // set chunks
 func (s *State) SetChunks(stream *ConfiguredStream, chunks *Set[Chunk]) {
+	if stream.Self().Stream.SyncMode != CDC {
```
hash-data: we must add it for full refresh as we may introduce other types in future
```suggestion
	if stream.Self().Stream.SyncMode == FULL_REFRESH {
```

protocol/sync.go
```diff
@@ -87,6 +87,10 @@ var syncCmd = &cobra.Command{
 		selectedStreams := []string{}
 		cdcStreams := []Stream{}
 		standardModeStreams := []Stream{}
+		var stateStreams = make(map[string]*types.StreamState)
```
hash-data:
```suggestion
		var stateStreamMap = make(map[string]*types.StreamState)
```

protocol/sync.go
```diff
@@ -111,11 +115,17 @@ var syncCmd = &cobra.Command{
 			elem.StreamMetadata = sMetadata
 			selectedStreams = append(selectedStreams, elem.ID())
 
+			var cdcSelectedStreams []*types.StreamState
+
 			if elem.Stream.SyncMode == types.CDC {
 				cdcStreams = append(cdcStreams, elem)
+				if streamState, exists := stateStreams[fmt.Sprintf("%s.%s", elem.Namespace(), elem.Name())]; exists {
```
hash-data:
```suggestion
				if streamState, exists := stateStreamMap[fmt.Sprintf("%s.%s", elem.Namespace(), elem.Name())]; exists {
```

types/state.go
```diff
@@ -112,6 +112,9 @@ func (s *State) GetChunks(stream *ConfiguredStream) *Set[Chunk] {
 
 // set chunks
 func (s *State) SetChunks(stream *ConfiguredStream, chunks *Set[Chunk]) {
+	if stream.Self().Stream.SyncMode == FULLREFRESH {
```
hash-data:
```suggestion
	if stream.GetSyncMode() == FULLREFRESH {
```

PR #303, opened 2025-05-29
fix: credential parsing for postgres and mongodb

### LINKED ISSUE #330: Refactor MongoDB and PostgreSQL connection string construction for safety and clarity
## Description
Current connection strings for MongoDB and PostgreSQL are manually constructed using `fmt.Sprintf,` which can lead to issues with special characters in usernames, passwords, or DB names.

## Problem
Manual string formatting does not safely encode credentials or parameters, risking connection errors or injection vulnerabilities.

## Solution
Use Go's package to construct URIs in a safer and more maintainable way for both MongoDB and PostgreSQL.

vaibhav-datazip:
# Description

- Fixed an issue with JDBC URL parameter parsing in the PostgreSQL driver's `Validate` function.  
  Previously, parameters under `jdbc_url_params` were incorrectly formatted, causing issues with connection string generation.  
  This fix ensures that parameters are properly added as individual query parameters rather than being wrapped in a single `options` parameter.

- Fixes #330 
Resolved issues with special character handling in connection strings for both **MongoDB** and **PostgreSQL** drivers.  
  Special characters in usernames, passwords, or database names previously caused parsing failures due to manual string construction with `fmt.Sprintf`.  
  This has been fixed by leveraging Go's `net/url` package to construct connection strings safely, ensuring proper URL encoding for all components.

## Type of change

# How Has This Been Tested?

- [x] Tested with standard PostgreSQL connection:
  - Verified that all `jdbc_url_params` are correctly parsed and reflected in the final URI.

- [x] Tested credentials parsing in both drivers:
  - Passwords with special characters like `test!@#$%^&*()_+-={}|\][:;"'?/> <,.` for PostgreSQL, and without space for MongoDB.
  - Empty (blank) passwords.
  - Database names containing special characters such as `!@#$%^&*`.


hash-data:
let us solve this as well: https://github.com/datazip-inc/olake/issues/330

drivers/mongodb/internal/config.go
```diff
@@ -25,36 +25,39 @@ type Config struct {
...
+		Path:     "/",
+		RawQuery: query.Encode(),
 	}
 
-	// Final MongoDB URI
-	return fmt.Sprintf(
-		"%s://%s%s/%s",
-		connectionPrefix, auth, strings.Join(c.Hosts, ","), options,
-	)
+	return u.String()
```
vikaxsh, line 61: Avoid Hardcoded Defaults 
make them constants
`c.ReadPreference = "secondaryPreferred"`,
`c.MaxThreads = 10`
like this
`const (
	defaultMaxThreads      = 10
	defaultReadPreference  = "secondaryPreferred"
)
`

drivers/mongodb/internal/config.go
```diff
@@ -25,36 +25,39 @@ type Config struct {
...
 func (c *Config) URI() string {
 	connectionPrefix := "mongodb"
-	options := fmt.Sprintf("?authSource=%s", c.AuthDB)
+	if c.Srv {
+		connectionPrefix = "mongodb+srv"
+	}
+
 	if c.MaxThreads == 0 {
 		// set default threads
 		logger.Info("setting max threads to default[10]")
```
vikaxsh: logger formatting
logger.Infof("setting max threads to default [%d]", 10)

drivers/postgres/internal/config.go
```diff
@@ -53,23 +52,21 @@ func (c *Config) Validate() error {
 		c.MaxThreads = 2
```
vikaxsh: consider making it a global constant , can be used for all sources

drivers/mongodb/internal/config.go
```diff
@@ -23,38 +24,45 @@ type Config struct {
 	ChunkingStrategy string   `json:"chunking_strategy"`
 }
 
+const (
+	defaultReadPreference = "secondaryPreferred"
```
ImDoubD-datazip: move this to constant.go

hash-data (approved):

PR #312, opened 2025-06-05
Add Stats Persistence for Resumed Syncs

sambhavnoobcoder:
## Issue Summary
When a sync was resumed after interruption, users had no visibility into progress metrics. Stats like total records to process, progress percentage, and estimated time remaining were missing on resumed syncs. This was particularly frustrating for long-running syncs where monitoring progress is critical.

As noted in issue #110  , the root cause was identified in MongoDB's backfill.go:
```go
// TODO: to get estimated time need to update pool.AddRecordsToSync(totalCount) (Can be done via storing some vars in state)
```

The state object wasn't persisting stats-related information for each stream, resulting in a lack of visibility after resuming a sync.

## Root Cause Analysis
After investigating the codebase, we discovered:

1. The `StreamState` structure didn't store the total number of records to sync
2. When resuming a sync, the `pool.AddRecordsToSync()` method wasn't called with the correct count
3. As a result, metrics like progress percentage and estimated time couldn't be calculated

The `StatsLogger` function relies on the total records count to calculate:
```go
estimatedSeconds := "Not Determined"
if speed > 0 && remainingRecords >= 0 {
    estimatedSeconds = fmt.Sprintf("%.2f s", float64(remainingRecords)/speed)
}
```

Without this value being persisted in state, these calculations weren't possible on resumed syncs.

## Implementation
Our solution implements a comprehensive approach to persist and restore stats across sync sessions:

1. **Extended State Model**:
   - Added a `TotalRecords` field to `StreamState` in `types/state.go`
   - Added methods to get and set total records in the state

2. **Updated Drivers**:
   - Modified MongoDB, MySQL, and Postgres drivers to store total records during initial sync
   - Added code to retrieve and use these counts during resumed syncs

3. **Enhanced Sync Command**:
   - Updated the sync command to restore total records from state on resume
   - Ensured both standard and CDC streams handle the stats correctly

4. **Interface Updates**:
   - Added new methods to the `State` interface in `protocol/interface.go`

## Testing & Verification
We created two comprehensive tests to verify our implementation:

1. **State Serialization Test** (`test_state_records.go`):
   - Creates a state with a `TotalRecords` value
   - Serializes to JSON and deserializes back
   - Verifies the value is correctly preserved

2. **Pool Integration Test** (`test_pool_integration.go`):
   - Creates a mock state with stream record counts
   - Simulates the sync command logic for restoring counts
   - Verifies the pool correctly loads and uses these counts
   - Tests progress calculations with the loaded counts

Both tests pass successfully, confirming our implementation works as expected.

## Screenshots of Passing Tests

image
image


ImDoubD-datazip:
@sambhavnoobcoder Hi, any update on this. also there are lot of merge conflicts which requires resolution. Please move our merge branch to staging. also please sign the CLA.

shubham19may:
Not needed anymore.

PR #322, opened 2025-06-09
fix: pg slot issue

hash-data:
# Description
-> Fix acknowledgement of lsn in pg
-> Check for extended slot

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/destination/iceberg/olake-iceberg-java-writer/pom.xml
+++ b/destination/iceberg/olake-iceberg-java-writer/pom.xml
@@ -501,7 +501,7 @@
         <dependency>
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
-            <version>42.7.4</version>
+            <version>42.7.7</version>
         </dependency>
         <dependency>
             <groupId>io.minio</groupId>
--- a/drivers/mysql/internal/config.go
+++ b/drivers/mysql/internal/config.go
@@ -2,7 +2,7 @@ package driver
 
 import (
 	"fmt"
-	"net/url"
+	"github.com/go-sql-driver/mysql"
 	"strings"
 
 	"github.com/datazip-inc/olake/constants"
@@ -39,15 +39,16 @@ func (c *Config) URI() string {
 		hostStr = "localhost"
 	}
 
-	// Construct full connection string
-	return fmt.Sprintf(
-		"%s:%s@tcp(%s:%d)/%s",
-		url.QueryEscape(c.Username),
-		url.QueryEscape(c.Password),
-		hostStr,
-		c.Port,
-		url.QueryEscape(c.Database),
-	)
+	cfg := mysql.Config{
+		User:                 c.Username,
+		Passwd:               c.Password,
+		Net:                  "tcp",
+		Addr:                 fmt.Sprintf("%s:%d", hostStr, c.Port),
+		DBName:               c.Database,
+		AllowNativePasswords: true,
+	}
+
+	return cfg.FormatDSN()
 }
 
 // Validate checks the configuration for any missing or invalid fields


pkg/waljs/waljs.go
```diff
@@ -105,7 +119,11 @@ func (s *Socket) AcknowledgeLSN(ctx context.Context) error {
 
 func (s *Socket) StreamMessages(ctx context.Context, callback abstract.CDCMsgFn) error {
 	// Start logical replication with wal2json plugin arguments.
-	// TODO: need research on if we need initial wait time or not (currently we are using idle time)
+	var tables []string
+	for key := range s.changeFilter.tables {
+		tables = append(tables, key)
+	}
+	pluginArguments = append(pluginArguments, fmt.Sprintf("\"add-tables\" '%s'", strings.Join(tables, ",")))
```
hash-data, line 134: one thing here is, we will require heartbeat mechanism in order to ack lsn's.

shubham19may (dismissed):

drivers/postgres/internal/cdc.go
```diff
@@ -41,20 +41,25 @@ func (p *Postgres) PreCDC(ctx context.Context, state *types.State, streams []typ
...
 	p.Socket = socket
 	currentLSN := p.Socket.ConfirmedFlushLSN
-	globalState := state.GetGlobal()
+	globalState := p.state.GetGlobal()
 
 	if globalState == nil || globalState.State == nil {
-		state.SetGlobal(waljs.WALState{LSN: currentLSN.String()})
-		state.ResetStreams()
+		p.state.SetGlobal(waljs.WALState{LSN: currentLSN.String()})
+		p.state.ResetStreams()
```
shubham19may, line 48: shouldn't we also mark the current LSN here like we are doing below?

pkg/waljs/filter.go
```diff
@@ -28,15 +29,19 @@ func NewChangeFilter(typeConverter func(value interface{}, columnType string) (i
...
 	}
-	if len(changes.Change) == 0 {
-		return nil
+	nextLSN, err := pglogrepl.ParseLSN(changes.NextLSN)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse received lsn: %s", err)
 	}
 
+	if len(changes.Change) == 0 {
+		return &nextLSN, fmt.Errorf("%s", noRecordErr)
```
shubham19may: why is this the error?

pkg/waljs/waljs.go
```diff
@@ -115,21 +135,25 @@ func (s *Socket) StreamMessages(ctx context.Context, callback abstract.CDCMsgFn)
...
+	messageReceived := false
+	cdcStartTime := time.Now()
 	for {
 		select {
 		case <-ctx.Done():
 			return nil
 		default:
-			if time.Since(startTime) > s.initialWaitTime {
-				logger.Debug("Idle timeout reached while waiting for new messages")
+			if !messageReceived && s.initialWaitTime > 0 && time.Since(cdcStartTime) > s.initialWaitTime {
```
shubham19may, line 152: but this means every sync will take atleast 120s to complete?

pkg/waljs/waljs.go
```diff
@@ -115,21 +135,25 @@ func (s *Socket) StreamMessages(ctx context.Context, callback abstract.CDCMsgFn)
...
+				logger.Warnf("no records found in given initial wait time")
 				return nil
 			}
-			// Use a context with timeout for receiving a message.
+
 			msg, err := s.pgConn.ReceiveMessage(ctx)
 			if err != nil {
-				return fmt.Errorf("failed to receive message from wal: %s", err)
+				if pgconn.Timeout(err) || strings.Contains(err.Error(), "EOF") {
+					return nil
```
shubham19may, line 165: ? this is a real error, we should break the sync.

pkg/waljs/waljs.go
```diff
@@ -141,29 +165,40 @@ func (s *Socket) StreamMessages(ctx context.Context, callback abstract.CDCMsgFn)
 			switch copyData.Data[0] {
 			case pglogrepl.PrimaryKeepaliveMessageByteID:
 				// For keepalive messages, process them (but no ack is sent here).
-				_, err := pglogrepl.ParsePrimaryKeepaliveMessage(copyData.Data[1:])
+				pkm, err := pglogrepl.ParsePrimaryKeepaliveMessage(copyData.Data[1:])
 				if err != nil {
 					return fmt.Errorf("failed to parse primary keepalive message: %s", err)
 				}
-
+				logger.Infof("keep alive message received: %v", pkm)
```
shubham19may: it will be too many logs

pkg/waljs/waljs.go
```diff
@@ -141,29 +165,40 @@ func (s *Socket) StreamMessages(ctx context.Context, callback abstract.CDCMsgFn)
...
+				if pkm.ReplyRequested {
+					s.ClientXLogPos = pkm.ServerWALEnd
+					if !messageReceived {
+						// set state first for fallback
+						state.SetGlobal(WALState{LSN: s.ClientXLogPos.String()})
+						// acknowledge lsn until no message received
+						err := s.AcknowledgeLSN(ctx)
```
shubham19may, line 179: Add todo to commit the transaction in writer side first.

pkg/waljs/waljs.go
hash-data: until no message recieved

pkg/waljs/waljs.go
hash-data: no eof comes once records are not there in buffer

pkg/waljs/waljs.go
hash-data: yeah but we need them as we are marking lsn.

pkg/waljs/waljs.go
```diff
@@ -141,29 +165,40 @@ func (s *Socket) StreamMessages(ctx context.Context, callback abstract.CDCMsgFn)
...
+						if err != nil {
```
shubham19may: we need to commit iceberg writer as well right? 

also why is it important to ack LSN even if no message received. Particularly when receiving the ReplyRequested?

hash-data: for now, we not need to flush batch as we have added message check. Once we have flush thing will remove that.

pkg/waljs/filter.go
hash-data: just to maintain message check

drivers/postgres/internal/cdc.go
```diff
@@ -63,22 +68,27 @@ func (p *Postgres) PreCDC(ctx context.Context, state *types.State, streams []typ
 			// TODO: handle cursor mismatch with user input (Example: user provide if it has to fail or do full load with new resume token)
 			if parsed != currentLSN {
 				logger.Warnf("lsn mismatch, backfill will start again. prev lsn [%s] current lsn [%s]", parsed, currentLSN)
-				state.SetGlobal(waljs.WALState{LSN: currentLSN.String()})
-				state.ResetStreams()
+				p.state.SetGlobal(waljs.WALState{LSN: currentLSN.String()})
+				p.state.ResetStreams()
+				// commit the lsn as we are going to start syncing from this point after backfill
+				if err := p.Socket.AcknowledgeLSN(ctx); err != nil {
```
shubham19may: remove these.

pkg/waljs/filter.go
shubham19may: instead return number of records synced.

pkg/waljs/waljs.go
shubham19may: Ideally if buffer never gets empty, this will run for infinity. 
So we need to use the way of finding the latest LSN when sync starts and only continue sync till that point, then close. Default behaviour. 


Continuous CDC mode is different.

pkg/waljs/waljs.go
shubham19may: lets only print if its true

hash-data: added to debug

PR #329, opened 2025-06-11
fix(mysql): replace queryEscape with FormatDSN 

sclee01:
# Description

This PR fixes an issue where the database connection string was improperly escaping the password using `url.QueryEscape`, which caused connection failures when the password included special characters like `!`.

By removing the unnecessary escaping on the password field in the DSN (`Config.URI()`), the connection now works as expected with all valid MySQL passwords.

Fixes # (no issue number, discovered during local testing)

## Type of change

# How Has This Been Tested?

- [x] Tested connecting to an AWS RDS MySQL instance using a password that includes special characters (e.g., `test1234!`) which previously failed. For example, when the password contains special characters like !, url.QueryEscape encodes it (e.g., ! → %21), which causes authentication to fail with the original password
- [x] Verified that the change does not affect username, host, or database parsing
- [x] Application successfully completes `PingContext()` and proceeds to establish a stable connection after fix.

# Screenshots or Recordings

_N/A_

## Related PR's (If Any):

None


drivers/mysql/internal/config.go
```diff
@@ -43,7 +43,7 @@ func (c *Config) URI() string {
 	return fmt.Sprintf(
 		"%s:%s@tcp(%s:%d)/%s",
 		url.QueryEscape(c.Username),
-		url.QueryEscape(c.Password),
+		c.Password,
```
vikaxsh: Instead of manually building the connection string and escaping special characters, We can use FormatDSN() to safely build the connection string with special characters in credentials.
https://github.com/go-sql-driver/mysql/blob/v1.9.2/dsn.go#L254

sclee01: @vikash390  Updated to use FormatDSN() and checked for backward compatibility.

shubham19may (approved):

sclee01:
@hash-data  All checks have passed and the PR is approved.
Let me know if anything else is needed before merging.

PR #333, opened 2025-06-12
chore: default normalization for relational db's

### LINKED ISSUE #327: task: normalisation true by default for all relational sources
Currently by default normalisation is false for all sources. but we need to make it true for postgres and mysql for all discovered streams.

tanishaAtDatazip:
## Description

This PR updates the default behavior for relational sources (Postgres and MySQL) to have `normalisation: true` for all discovered streams.

So for relational  dbs normalisation default is on state in backend as well as ui
Non relational db defualt normalisation is off both in ui and backend

## Related Issue

Closes #327

### DIFF
--- a/drivers/abstract/incremental.go
+++ b/drivers/abstract/incremental.go
@@ -27,6 +27,8 @@ func (a *AbstractDriver) Incremental(ctx context.Context, pool *destination.Writ
 			backfillWaitChannel <- stream.ID()
 			return nil
 		}
+		// Reset only mentioned cursor state while preserving other state values
+		a.state.ResetCursor(stream.Self())
 		return a.Backfill(ctx, backfillWaitChannel, pool, stream)
 	})
 	if err != nil {
@@ -95,10 +97,6 @@ func (a *AbstractDriver) Incremental(ctx context.Context, pool *destination.Writ
 func (a *AbstractDriver) getIncrementCursorFromState(primaryCursorField string, secondaryCursorField string, stream types.StreamInterface) (any, any, error) {
 	primaryStateCursorValue := a.state.GetCursor(stream.Self(), primaryCursorField)
 	secondaryStateCursorValue := a.state.GetCursor(stream.Self(), secondaryCursorField)
-
-	if primaryStateCursorValue == nil || (secondaryCursorField != "" && secondaryStateCursorValue == nil) {
-		return primaryStateCursorValue, secondaryStateCursorValue, nil
-	}
 
 	getCursorValue := func(cursorField string, cursorValue any) (any, error) {
 		if cursorField == "" {
--- a/drivers/mongodb/README.md
+++ b/drivers/mongodb/README.md
@@ -10,6 +10,8 @@ The MongoDB Driver enables data synchronization from MongoDB to your desired des
    Tracks and syncs incremental changes from MongoDB in real time.
 3. **Strict CDC (Change Data Capture)**
    Tracks only new changes from the current position in the MongoDB change stream, without performing an initial backfill.
+4. **Incremental**
+   Syncs only new or modified records which have cursor value greater than or equal to the saved position.
 
 ---
 
@@ -98,10 +100,6 @@ Before running the Sync command, the generated `streams.json` file must be confi
       ```json
       "sync_mode": "cdc",
       ```
-   - Specify the cursor field (only for incremental syncs):
-      ```json
-      "cursor_field": "<cursor field from available_cursor_fields>"
-      ```
    - To enable `append_only` mode, explicitly set it to `true` in the selected stream configuration.
       ```json
          "selected_streams": {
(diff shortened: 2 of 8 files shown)


types/catalog.go
```diff
@@ -60,6 +62,7 @@ func GetWrappedCatalog(streams []*Stream) *Catalog {
 			StreamName:     stream.Name,
 			PartitionRegex: "",
 			AppendMode:     false,
+			Normalization:  driverType == string(constants.Postgres) || driverType == string(constants.MySQL), // Set normalization based on driver type
```
hash-data: what if user has removed the normalization field what will be value for it for these drivers?

hash-data: Also can we use constant array to check for relational db's.

tanishaAtDatazip: what do you mean by "removing the normalization field"? if the user sets it to false, then that's intentional. but if they omit the field entirely, the logic here defaults it to true for Postgres and MySQL drivers.

yes i will update to constant map/array.

hash-data: When the normalization field itself is not present in the selected stream sections.

constants/constants.go
```diff
@@ -24,3 +25,16 @@ const (
...
 	MySQL    DriverType = "mysql"
 )
+
+var (
+	// RelationalDrivers maps driver types to their relational status
+	RelationalDrivers = map[DriverType]bool{
+		Postgres: true,
+		MySQL:    true,
+		MongoDB:  false,
+	}
```
hash-data: let us create an array of relationalDriver instead of map

destination/iceberg/proto/records_ingest_grpc.pb.go
```diff
@@ -8,6 +8,7 @@ package proto
 
 import (
 	context "context"
+
```
hash-data: remove files which are not relevant for this pr

drivers/abstract/abstract.go
```diff
@@ -96,6 +101,16 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
...
 		}
+
+		if constants.IsRelationalDriver(a.driver.Type()) {
+			props := map[string]interface{}{constants.NormalizationProperty: true}
+			val, err := json.Marshal(props)
+			if err != nil {
+				return false
+			}
+			convStream.AdditionalProperties = string(val)
+		}
```
hash-data: it is not an additional property it is streamMetadata

types/catalog.go
```diff
@@ -58,6 +60,9 @@ func GetWrappedCatalog(streams []*Stream) *Catalog {
 	}
 	// Loop through each stream and populate Streams and SelectedStreams
 	for _, stream := range streams {
+		props := make(map[string]interface{})
+		_ = json.Unmarshal([]byte(stream.AdditionalProperties), &props)
+
```
hash-data: not required

types/catalog.go
```diff
@@ -66,6 +71,7 @@ func GetWrappedCatalog(streams []*Stream) *Catalog {
 			StreamName:     stream.Name,
 			PartitionRegex: "",
 			AppendMode:     false,
+			Normalization:  props[constants.NormalizationProperty].(bool),
```
hash-data: just do utils.ArrayContains a.driverType mark it true else false

destination/iceberg/proto/records_ingest_grpc.pb.go
hash-data, line 1: remove file

destination/iceberg/proto/records_ingest.pb.go
hash-data, line 1: remove file

types/catalog.go
tanishaAtDatazip: removed

drivers/abstract/abstract.go
tanishaAtDatazip: updated

constants/constants.go
```diff
@@ -1,6 +1,8 @@
 package constants
 
-import "time"
+import (
+	"time"
+)
```
vaibhav-datazip, line 5: This change isn't necessary, can be reverted back.

vaibhav-datazip, line 5: This change isn't required, can be reverted back.

types/catalog.go
```diff
@@ -52,7 +53,11 @@ type Catalog struct {
 	Streams         []*ConfiguredStream         `json:"streams,omitempty"`
 }
 
-func GetWrappedCatalog(streams []*Stream) *Catalog {
+func GetWrappedCatalog(streams []*Stream, driver string) *Catalog {
+	// Whether the source is a relational driver to not
```
vaibhav-datazip, line 57:
```suggestion
	// Whether the source is a relational driver or not
```

vaibhav-datazip (changes requested):

constants/constants.go
vaibhav-datazip, line 5: This change isn't required, can be reverted back.

types/catalog.go
vaibhav-datazip:
```suggestion
	// Whether the source is a relational driver or not
```

PR #339, opened 2025-06-13
feat: Kafka As Source V0

### LINKED ISSUE #87: [Feat] Kafka Source
### Problem
Kafka as a source required

### Solution

ImDoubD-datazip:
# Description
Kafka as source has been added as Olake's latest driver. The messages from the topics of Kafka will be synced directly to destination. Currently modes are supported: Incremental. 
- In case of Incremental : 
   - For this sync, cursor field from the available cursor field needs to be provided i.e. `"cursor_field" = "offset"`.
   - when state file not provided, initially backfill will take place and then incremental continuous sync will. This will update the state and commit the last partition and offset of selected topics synced.
   - if any problem occurs in state and state persisted checkpoints can be used then, last commit of given consumer group will be used. If there is no previous commit for that consumer group which is being used, then auto offset reset trigger will be used.
   - if `auto_offset_reset = earliest` then all the unread messages by the consumer will be synced from the start offset, with any of new incoming messages.
   - if `auto_offset_reset = latest` then all the new incoming messages will be synced.

#### TO FIX: 

   - 1st: process the messages 

Fixes #87 

## Type of change

# How Has This Been Tested?
Testing of both the full refresh and incremental sync is done on both parquet and apache iceberg.

- max threads is being utilized when sync-mode set to be full refresh, in case of incremental data is synced as soon as it is streamed into kafka

Incremental Sync:
Screenshot 2025-06-28 at 9 41 47 PM

Full-Load Sync: 
Screenshot 2025-06-28 at 9 43 43 PM

# How to test Kafka as source

Example source.json:
Using PLAINTEXT protocol
```json
{
    "protocol": {
    },
    "max_threads": 3,
    "wait_time": 30
}
```
Using SASL-SSL protocol
```json
{
    "protocol": {
    },
    "max_threads": 3,
    "wait_time": 30
}
```


constants/constants.go
```diff
@@ -23,6 +23,7 @@ const (
 	ConfigFolder           = "CONFIG_FOLDER"
 	// EffectiveParquetSize is the effective size in bytes considering 512MB targeted parquet size, compression ratio as 8 and chunk inflation factor as 2
 	EffectiveParquetSize = int64(512) * 1024 * 1024 * int64(8) * int64(2)
+	DefaultWaitTime      = 30
```
shubham19may: add description for what is it.

shubham19may:
AI review 1 : 

I reviewed the Kafka driver end-to-end (config, setup, discovery, incremental, state, abstract flow) and found several correctness bugs, behavioral gaps, and optimization opportunities. Below are prioritized findings with concrete fixes.

### Critical correctness bugs (fix first)
- Context cancellation passed to writer
  - Issue: You cancel the per-read timeout context and then pass it to `processFn`, so downstream write uses a canceled context and fails.
  - Where:
```182:205:drivers/kafka/internal/incremental.go
fetchCtx, cancel := context.WithTimeout(ctx, remaining)
msg, err := reader.ReadMessage(fetchCtx)
cancel()
// ...
if err := processFn(fetchCtx, data); err != nil { ... }
```
  - Fix: Pass the parent `ctx` (or a fresh context) to `processFn`.
  - Suggested edit:
```go
// after cancel()
if err := processFn(ctx, data); err != nil {
    ...
}
```

- Double-incrementing offsets when resuming
  - Issue: You store offset+1 in state, but on resume you do `startOffset = stateOffset + 1`, skipping one message.
  - Where (read + set):
```103:114:drivers/kafka/internal/incremental.go
if offset, exists := stateOffsets[partition]; exists && offset > 0 {
    startOffset = offset + 1
}
```
```276:283:drivers/kafka/internal/incremental.go
partitionsState[fmt.Sprintf("%d", partition)] = offset + 1
```
  - Fix: Keep state value as “next offset to read”, and set start offset to the stored value (no +1).
  - Suggested edits:
```go
// resume
startOffset = offset
// commit/state
partitionsState[fmt.Sprintf("%d", partition)] = offset + 1 // keep next-to-read in state
```

- Wait-time deadline logic never decreases
  - Issue: `remaining := time.Until(time.Now().Add(wait))` always equals `wait`, so the loop keeps giving a full timeout every iteration.
  - Where:
```175:181:drivers/kafka/internal/incremental.go
remaining := time.Until(time.Now().Add(time.Duration(k.config.WaitTime) * time.Second))
if remaining <= 0 { ... }
fetchCtx, cancel := context.WithTimeout(ctx, remaining)
```
  - Fix: Compute an absolute deadline once and subtract each loop.
  - Suggested edit:
```go
deadline := time.Now().Add(time.Duration(k.config.WaitTime) * time.Second)
for {
    remaining := time.Until(deadline)
    if remaining <= 0 { saveOffset(); return nil }
    fetchCtx, cancel := context.WithTimeout(ctx, remaining)
    ...
}
```

- Wrong error checked for out-of-range offset
  - Issue: You compare against an uninitialized `kerr` instead of `err`.
  - Where:
```136:151:drivers/kafka/internal/incremental.go
var kerr kafka.Error
if errors.Is(kerr, kafka.OffsetOutOfRange) { ... }
```
  - Fix: Check `errors.Is(err, kafka.OffsetOutOfRange)` or type-assert.
  - Suggested edit:
```go
if err := reader.SetOffset(startOffset); err != nil {
    if errors.Is(err, kafka.OffsetOutOfRange) {
        ...
    } else {
        ...
    }
}
```

- TLS/SASL not applied to admin client and consumer group
  - Issue: `adminClient` and `ConsumerGroup` use default transport/dialer; SASL/TLS configs are ignored, so Metadata and group join can fail on secure clusters.
  - Where:
```
74:91:drivers/kafka/internal/kafka.go
adminClient := &kafka.Client{ Addr: kafka.TCP(...), } // no Transport
```
```
69:76:drivers/kafka/internal/incremental.go
consumerGroup, err := kafka.NewConsumerGroup(kafka.ConsumerGroupConfig{ ... }) // no Dialer
```

  - Fix:
    - Set `adminClient.Transport = &kafka.Transport{SASL: dialer.SASLMechanism, TLS: dialer.TLS, Dial: dialer.DialContext}` (exact fields per your kafka-go version).
    - In `ConsumerGroupConfig`, set `Dialer: k.dialer`.

- Single `consumerGen` shared across streams (race/incorrect commits)
  - Issue: `k.consumerGen` is a single pointer overwritten by concurrent streams; commits may target the wrong generation.
  - Where:
```
84:91:drivers/kafka/internal/incremental.go
k.consumerGroups[topic] = consumerGroup
k.consumerGen = consumerGen
```
```
262:271:drivers/kafka/internal/incremental.go
if k.consumerGen == nil { return fmt.Errorf("no active consumer generation") }
offsets, ok := k.offsetMap[topic]
...
if err := k.consumerGen.CommitOffsets(commitOffsets); err != nil { ... }
```
  - Fix: Track generation per topic (e.g., `map[string]*kafka.Generation`) and use the topic-specific generation when committing. Or use `consumerGroup.CommitOffsets` if available on your version.

- Auto offset reset not honored when no committed offsets
  - Issue: For a new/ephemeral group or no state, you end up using `assignment.Offset` or `0`, ignoring `auto_offset_reset`.
  - Fix: Decide startup offset per partition as:
    - If state exists: use stored next offset.
    - Else if group has committed: use that.
    - Else: use `ResolveOffset(auto_offset_reset)` to set earliest/latest.

- Schema/data mismatch
  - Issue: Schema declares `"timestamp"`, but you emit `"kafka_timestamp"`. Schema declares `"message"`, but you don’t populate it.
  - Where:
```
127:132:drivers/kafka/internal/kafka.go    // schema fields
```
```
221:224:drivers/kafka/internal/incremental.go  // data uses kafka_timestamp, not timestamp
```
  - Fix: Use `"timestamp"` consistently and set `"message"` (e.g., raw value as string) alongside parsed JSON fields.

- Using ConsumerGroup but reading with partition readers
  - Issue: You create a consumer group to obtain assignments, then read via standalone partition `Reader`s. This bypasses heartbeats and rebalance handling inherent to the group, and `ErrGenerationEnded` likely won’t be raised by these readers.
  - Fix: Use one of:
    - A `kafka.Reader` with `GroupID` (preferred: it handles assignments, rebalances, and commits), or
    - Use `consumerGen.Claim`/`Release` loop to read from claimed partitions within a live generation and commit accordingly.

### Behavioral inconsistencies / docs
- Full refresh support
  - Docs say full refresh is supported; code explicitly disables backfill for Kafka.
  - Where:
```
17:24:drivers/abstract/backfill.go
if a.driver.Type() == "kafka" { ... skip ... }
```
```
1:12:drivers/kafka/README.md
The Kafka Driver enables ... supports both Full Refresh and Incremental
```
  - Fix: Update README to “Incremental only” and remove full-refresh examples.

- Append flag naming
  - README uses `append_only`; example `streams.json` uses `append_mode`.
  - Fix: Standardize on one key (likely `append_only`) across docs and examples.

- Discover includes internal topics
  - `GetStreamNames` returns everything, including `__consumer_offsets`.
  - Fix: Filter topics starting with `_`/`__` by default, or document that users should remove them.

- Config validation
  - `auto_offset_reset` only validated during runtime in `ResolveOffset`. Add validation in `Config.Validate()` to accept only `"earliest"` or `"latest"` when provided.

### Robustness, UX, and performance improvements
- Commit strategy
  - Consider periodic commits (and state updates) to lower loss on crash versus only committing at the end of a batch.
- Message decoding
  - Fallback if `msg.Value` is not JSON: keep `"message"` as raw string or base64 field, and still populate metadata fields.
- Key encoding
  - Don’t assume UTF-8; either treat as bytes or base64-encode if not valid UTF-8.
- Reader sizing
  - `MaxBytes: 10e6` is okay; consider `10<<20` or `10*1024*1024` for clarity.
- Connectivity check
  - Avoid pinging a non-existent topic `"__test"`. Use a generic metadata call (no topics) or broker-level ping with the configured transport.

### Concrete edit suggestions (sketches)
- Honor auto_offset_reset and fix state vs start offsets:
```go
// After building stateOffsets and getting assignment
var startOffset int64
if v, ok := stateOffsets[partition]; ok && v >= 0 {
    startOffset = v // state holds next offset to read
} else if assignment.Offset > 0 {
    startOffset = assignment.Offset
} else {
    // No state and no committed offset -> use reset policy
    resetOffset, err := ResolveOffset(k.config.AutoOffsetReset)
    if err != nil { return fmt.Errorf("invalid auto_offset_reset: %w", err) }
    startOffset = resetOffset
}
if err := reader.SetOffset(startOffset); err != nil { ... }
```

- Use parent ctx for processing:
```go
// use ctx, not fetchCtx
if err := processFn(ctx, data); err != nil { ... }
```

- Fix wait-time deadline:
```go
deadline := time.Now().Add(time.Duration(k.config.WaitTime) * time.Second)
for {
    remaining := time.Until(deadline)
    if remaining <= 0 { saveOffset(); return nil }
    fetchCtx, cancel := context.WithTimeout(ctx, remaining)
    msg, err := reader.ReadMessage(fetchCtx)
    cancel()
    ...
}
```

- Apply TLS/SASL to admin client and consumer group:
```go
// admin client
k.adminClient = &kafka.Client{
    Addr:      kafka.TCP(strings.Split(k.config.BootstrapServers, ",")...),
    Transport: &kafka.Transport{ TLS: k.dialer.TLS, SASL: k.dialer.SASLMechanism, Dial: k.dialer.DialContext },
}
// consumer group
consumerGroup, err := kafka.NewConsumerGroup(kafka.ConsumerGroupConfig{
    ID: groupID, Brokers: ..., Topics: ...,
    Dialer: k.dialer,
})
```

- Schema/data alignment:
```go
result["timestamp"] = msg.Time.UnixMilli()
result["message"] = string(msg.Value) // always populate
```

- Track generation per topic
```go
// in Kafka struct
consumerGens map[string]*kafka.Generation
// on Next:
k.consumerGens[topic] = consumerGen
// on commit:
gen := k.consumerGens[topic]
if gen == nil { return fmt.Errorf("no active generation for %s", topic) }
if err := gen.CommitOffsets(commitOffsets); err != nil { ... }
```

### Nice-to-have improvements
- Add option to exclude topics by regex in config.
- Increase `MaxRetries` or make it configurable for transient broker hiccups.
- Remove unused `syncedTopics` if not used elsewhere.
- Consider `kafka.Reader` with `GroupID` to simplify group lifecycle, heartbeats, rebalance, and commit handling.

### Quick wins in docs/examples
- Fix README claims about full refresh.
- Align `append_only` vs `append_mode`.
- Mention that if no `consumer_group` is provided, OLake uses `auto_offset_reset` and resumes from the saved state.

Would you like me to implement the high-priority fixes (context, offsets, wait-time, SASL/TLS, per-topic generation, auto_offset_reset) in the code now and run a quick build? I can do them in small, reviewable edits.

shubham19may:
AI review 2 : 

## Code Review Summary for Kafka Driver

After thoroughly reviewing the Kafka driver implementation, I've identified several bugs, optimization opportunities, and areas for improvement:

### 🔴 Critical Issues

#### 1. **Full Refresh Mode Not Implemented**
The README claims support for both Full Refresh and Incremental modes, but the implementation only supports Incremental:
- In `abstract.go` line 91-93, Kafka is hardcoded to only support INCREMENTAL mode
- There's no actual Full Refresh implementation in the Kafka driver
- **Impact**: Documentation misleading users

#### 2. **WaitTime Logic Bug**
In `incremental.go` line 175:
```go
remaining := time.Until(time.Now().Add(time.Duration(k.config.WaitTime) * time.Second))
```
This will always return a negative duration because `time.Until()` calculates time until a future time, but you're adding time to `time.Now()` which creates a future time, then calculating time until that - it's circular logic.

**Should be:**
```go
deadline := time.Now().Add(time.Duration(k.config.WaitTime) * time.Second)
remaining := time.Until(deadline)
```

#### 3. **Offset Commit Race Condition**
In `incremental.go`, the offset commit logic in `PostIncremental` (lines 258-304) has potential race conditions:
- Offsets are stored in `k.offsetMap` during processing but could be overwritten by concurrent partition processing
- No proper synchronization when multiple partitions complete simultaneously

### 🟡 Moderate Issues

#### 4. **Error Handling Inconsistencies**
In `incremental.go` line 139:
```go
var kerr kafka.Error
if errors.Is(kerr, kafka.OffsetOutOfRange) {
```
This won't work - `kerr` is a zero value, not the actual error. Should be:
```go
var kerr kafka.Error
if errors.As(err, &kerr) && errors.Is(kerr, kafka.OffsetOutOfRange) {
```

#### 5. **Consumer Group Management**
- Consumer groups are created but stored in a map (`k.consumerGroups`) that's never properly cleaned up except in `Close()`
- If a sync fails, consumer groups might leak
- Generated consumer group IDs include timestamp, making debugging harder

#### 6. **State Management Issues**
- State persistence for offsets uses string keys for partition IDs (e.g., "0", "1") which could cause type conversion issues
- No validation that restored offsets are valid before using them

### 🟢 Optimization Opportunities

#### 7. **Message Processing Efficiency**
In `incremental.go` lines 210-225, every message is being unmarshaled from JSON:
```go
if err := json.Unmarshal(msg.Value, &result); err != nil {
```
Consider:
- Supporting different message formats (Avro, Protobuf)
- Batch processing messages before sending to processor
- Lazy deserialization if not all fields are needed

#### 8. **Partition Processing**
- Each partition creates its own reader, which is good for parallelism
- However, consider implementing partition rebalancing support properly
- Add metrics/logging for partition lag monitoring

#### 9. **Resource Management**
- Admin client is created but never closed properly
- Consider implementing connection pooling for readers

### 📝 Additional Recommendations

#### 10. **Configuration Validation**
- Add validation for `auto_offset_reset` values during config validation
- MaxThreads should have an upper limit to prevent resource exhaustion

#### 11. **Schema Evolution**
The schema is hardcoded with basic fields. Consider:
- Dynamic schema detection from messages
- Schema registry integration
- Configurable field mappings

#### 12. **Testing Gaps**
No test files found for Kafka driver - critical for a data sync tool

#### 13. **Documentation Improvements**
- README should be updated to reflect actual implementation
- Add examples for different authentication mechanisms
- Document partition assignment behavior

#### 14. **Security Considerations**
- SASL credentials are parsed from config but not validated properly
- No support for certificate-based authentication (SSL client certs)
- Password appears in logs if JAAS config parsing fails

### 🔧 Quick Fixes Needed

1. Fix the WaitTime calculation bug
2. Fix error type checking for offset out of range
3. Update README to reflect actual supported modes
4. Add proper mutex protection for offsetMap updates
5. Implement proper consumer group cleanup

### 💡 Long-term Improvements

1. Implement actual Full Refresh mode if needed, or remove from documentation
2. Add comprehensive testing suite
3. Implement schema registry support
4. Add metrics and monitoring hooks
5. Consider implementing exactly-once semantics with transaction support

This Kafka driver is a good start but needs significant improvements before production use, especially around error handling, state management, and the critical WaitTime bug that will prevent proper timeout behavior.

drivers/kafka/internal/incremental.go
```diff
@@ -0,0 +1,301 @@
...
+		defer func() {
+			if err := reader.Close(); err != nil {
+				logger.Warnf("[KAFKA] failed to close reader for topic %s, partition %d: %v", topic, partition, err)
+			}
+		}()
+
+		// Set offset based on state or from beginning
+		if err := reader.SetOffset(startOffset); err != nil {
+			// If offset is invalid, fallback to configured offset reset policy
+			if errors.Is(err, kafka.OffsetOutOfRange) {
```
ImDoubD-datazip, line 134: when offset goes out of range error comes, only then fall back to auto offset reset is done. in other cases, error is returned

drivers/kafka/internal/incremental.go
```diff
@@ -0,0 +1,301 @@
...
+					result["kafka_timestamp"] = msg.Time.UnixMilli()
+					return result
+				}()
+
+				// Process message with provided function
+				if err := processFn(ctx, data); err != nil {
+					offsetsMu.Lock()
+					lastProcessedOffsets[partition] = lastOffset
+					offsetsMu.Unlock()
+					return fmt.Errorf("[KAFKA] failed to process message at offset %d: %v", msg.Offset, err)
```
ImDoubD-datazip: should we return an error if processing of single message/offset fails for a partition with having many messages or should just give a warning and bypass it? discussion required

drivers/kafka/internal/kafka.go
```diff
@@ -0,0 +1,207 @@
...
+	var topicNames []string
+	for _, topic := range resp.Topics {
+		topicNames = append(topicNames, topic.Name)
+	}
+	return topicNames, nil
+}
+
+func (k *Kafka) ProduceSchema(_ context.Context, streamName string) (*types.Stream, error) {
+	logger.Infof("[KAFKA] producing schema for topic [%s]", streamName)
+	stream := types.NewStream(streamName, "topics", nil)
```
ImDoubD-datazip, line 135: there is no such thing as namespace in kafka. discussion required

drivers/kafka/internal/incremental.go
```diff
@@ -0,0 +1,301 @@
...
+					}
+					// any failure during reading
+					saveOffset()
+					return fmt.Errorf("[KAFKA] error reading message in incremental sync: %v", err)
+				}
+				data := func() map[string]any {
+					var result map[string]any
+					if msg.Value == nil {
+						logger.Warnf("[KAFKA] received nil message value at offset %d for topic %s, partition %d", msg.Offset, topic, partition)
+						return nil
```
ImDoubD-datazip, line 196: here kafka topics can have nil messages, but it gives error during message flattening. discussion required

drivers/kafka/internal/incremental.go
```diff
@@ -0,0 +1,301 @@
...
+		partition := data.Partition
+		startOffset := data.StartOffset
+
+		// Partition specific-reader
+		reader := kafka.NewReader(kafka.ReaderConfig{
+			Brokers:   strings.Split(k.config.BootstrapServers, ","),
+			Topic:     topic,
+			Partition: partition,
+			MinBytes:  1,
+			MaxBytes:  10e6,
```
ImDoubD-datazip, line 122: for now, max bytes set for max batch size that a consumer will accept. so can be altered if message size is quite large

drivers/abstract/incremental.go
```diff
@@ -18,11 +18,27 @@ func (a *AbstractDriver) Incremental(ctx context.Context, pool *destination.Writ
...
+		// For non-Kafka drivers
+		skipBackfill := false
 		primaryCursor, secondaryCursor := stream.Cursor()
 		prevPrimaryCursor := a.state.GetCursor(stream.Self(), primaryCursor)
 		prevSecondaryCursor := a.state.GetCursor(stream.Self(), secondaryCursor)
 		if a.state.HasCompletedBackfill(stream.Self()) && (prevPrimaryCursor != nil && (secondaryCursor == "" || prevSecondaryCursor != nil)) {
 			logger.Infof("Backfill skipped for stream[%s], already completed", stream.ID())
+			skipBackfill = true
+		}
+		if skipBackfill {
```
hash-data: ??

PR #349, opened 2025-06-17
fix: iceberg deduplication fix

shubham19may:
# Description

Currently Iceberg writer can duplicate data as if chunk fails in between it can still take the commit the data. and when chunk gets processed again, then it can write the previously committed data again.

Fixes # (issue)

1. From going forward, we are committing only when the chunk is finished writing. We will not commit in between to avoid duplication.
2. for CDC or delta use case it will have deletes now for inserts as well (Because full-load & cdc running one after another can produce duplicates). 
3. fixed issue of random partitioning order
4. If one threads breaks, others should also fail and should not commit

## Type of change

# How Has This Been Tested?

- [X] Basic full load of postgres (small tables), getting committed successfully
- [X] Basic CDC workflow
- [ ] Failing one of the threads
- [ ] partitioning order test

# Screenshots or Recordings

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -8,7 +8,8 @@ const (
 	DefaultDiscoverTimeout = 5 * time.Minute
 	DefaultRetryTimeout    = 60 * time.Second
 	ParquetFileExt         = "parquet"
-	PartitionRegex         = `\{([^}]+)\}`
+	PartitionRegexIceberg  = `\{([^,]+),\s*([^}]+)\}`
+	PartitionRegexParquet  = `\{([^}]+)\}`
 	MongoPrimaryID         = "_id"
 	OlakeID                = "_olake_id"
 	OlakeTimestamp         = "_olake_timestamp"
--- a/destination/iceberg/iceberg.go
+++ b/destination/iceberg/iceberg.go
@@ -26,8 +26,15 @@ type Iceberg struct {
 	port          int
 	backfill      bool
 	configHash    string
-	partitionInfo []types.PartitionInfo // ordered slice to preserve partition column order
-	localBuffer   *LocalBuffer          // Buffer for this thread
+	partitionInfo []PartitionInfo // ordered slice to preserve partition column order
+	localBuffer   *LocalBuffer    // Buffer for this thread
+	threadID      string
+}
+
+// PartitionInfo represents a Iceberg partition column with its transform, preserving order
+type PartitionInfo struct {
+	Field     string
+	Transform string
 }
 
 func (i *Iceberg) GetConfigRef() destination.Config {
@@ -43,7 +50,8 @@ func (i *Iceberg) Setup(stream types.StreamInterface, options *destination.Optio
 	i.options = options
 	i.stream = stream
 	i.backfill = options.Backfill
-	i.partitionInfo = make([]types.PartitionInfo, 0)
+	i.partitionInfo = make([]PartitionInfo, 0)
+	i.threadID = getGoroutineID()
 
 	// Initialize the local buffer
 	i.localBuffer = &LocalBuffer{
@@ -68,26 +76,15 @@ func (i *Iceberg) Setup(stream types.StreamInterface, options *destination.Optio
 }
 
 func (i *Iceberg) Write(_ context.Context, record types.RawRecord) error {
-	// Get thread ID
-	threadID := getGoroutineID()
-
 	// Convert record to Debezium format with thread ID
-	debeziumRecord, err := record.ToDebeziumFormat(i.config.IcebergDatabase, i.stream.Name(), i.stream.NormalizationEnabled(), threadID)
+	// We are adding the thread ID to process the records from multiple threads in parallel and separately so that we can commit when each thread finishes.
+	debeziumRecord, err := record.ToDebeziumFormat(i.config.IcebergDatabase, i.stream.Name(), i.stream.NormalizationEnabled(), i.threadID)
 	if err != nil {
 		return fmt.Errorf("failed to convert record: %v", err)
 	}
 
-	// Add the record to the local buffer directly
-	if i.localBuffer == nil {
-		i.localBuffer = &LocalBuffer{
-			records: make([]string, 0, 100),
-			size:    0,
-		}
-	}
-
-	recordSize := int64(len(debeziumRecord))
 	i.localBuffer.records = append(i.localBuffer.records, debeziumRecord)
-	i.localBuffer.size += recordSize
+	i.localBuffer.size += int64(len(debeziumRecord))
 
 	// Check if buffer should be flushed
 	if i.localBuffer.size >= maxBatchSize {
@@ -115,7 +112,7 @@ func (i *Iceberg) Close(ctx context.Context) error {
 	}
 
 	// First flush any remaining data in the current thread's buffer
-	if i.localBuffer != nil && len(i.localBuffer.records) > 0 {
+	if len(i.localBuffer.records) > 0 {
 		err := flushLocalBuffer(i.localBuffer, i.client)
 		if err != nil {
 			logger.Errorf("Error flushing buffer on close: %s", err)
@@ -123,15 +120,12 @@ func (i *Iceberg) Close(ctx context.Context) error {
 		}
 	}
 
-	// Get the current thread ID
-	threadID := getGoroutineID()
-
 	// Send commit request for this thread using a special message format
 	ctx, cancel := context.WithTimeout(context.Background(), 1000*time.Second)
 	defer cancel()
 
 	// Create a special commit message
-	commitMessage := fmt.Sprintf(`{"commit": true, "thread_id": "%s"}`, threadID)
+	commitMessage := fmt.Sprintf(`{"commit": true, "thread_id": "%s"}`, i.threadID)
 
 	req := &proto.RecordIngestRequest{
 		Messages: []string{commitMessage},
@@ -139,11 +133,11 @@ func (i *Iceberg) Close(ctx context.Context) error {
 
 	res, err := i.client.SendRecords(ctx, req)
 	if err != nil {
-		logger.Errorf("Error sending commit message for thread %s: %s", threadID, err)
-		return fmt.Errorf("failed to send commit message for thread %s: %s", threadID, err)
+		logger.Errorf("Error sending commit message for thread %s: %s", i.threadID, err)
+		return fmt.Errorf("failed to send commit message for thread %s: %s", i.threadID, err)
 	}
 
-	logger.Infof("Sent commit message for thread %s: %s", threadID, res.GetResult())
+	logger.Infof("Sent commit message for thread %s: %s", i.threadID, res.GetResult())
 
 	return nil
 }
@@ -155,7 +149,7 @@ func (i *Iceberg) Check(ctx context.Context) error {
 
 	// Temporarily set stream to nil and clear partition fields to force a new server for the check
 	i.stream = nil
-	i.partitionInfo = make([]types.PartitionInfo, 0)
+	i.partitionInfo = make([]PartitionInfo, 0)
 
 	// Create a temporary setup for checking
 	err := i.SetupIcebergClient(false)
@@ -175,7 +169,7 @@ func (i *Iceberg) Check(ctx context.Context) error {
 
 	// Try to send a test message
 	req := &proto.RecordIngestRequest{
-		Messages: []string{getTestDebeziumRecord()},
+		Messages: []string{getTestDebeziumRecord(i.threadID)},
 	}
 
 	// Call the remote procedure
--- a/destination/iceberg/iceberg.go
+++ b/destination/iceberg/iceberg.go
@@ -26,7 +26,8 @@ type Iceberg struct {
 	port          int
 	backfill      bool
 	configHash    string
-	partitionInfo map[string]string // map of field names to partition transform
+	partitionInfo []types.PartitionInfo // ordered slice to preserve partition column order
+	localBuffer   *LocalBuffer          // Buffer for this thread
 }
 
 func (i *Iceberg) GetConfigRef() destination.Config {
@@ -42,7 +43,13 @@ func (i *Iceberg) Setup(stream types.StreamInterface, options *destination.Optio
 	i.options = options
 	i.stream = stream
 	i.backfill = options.Backfill
-	i.partitionInfo = make(map[string]string)
+	i.partitionInfo = make([]types.PartitionInfo, 0)
+
+	// Initialize the local buffer
+	i.localBuffer = &LocalBuffer{
+		records: make([]string, 0, 100),
+		size:    0,
+	}
 
 	// Parse partition regex from stream metadata
 	partitionRegex := i.stream.Self().StreamMetadata.PartitionRegex
@@ -61,20 +68,33 @@ func (i *Iceberg) Setup(stream types.StreamInterface, options *destination.Optio
 }
 
 func (i *Iceberg) Write(_ context.Context, record types.RawRecord) error {
-	// Convert record to Debezium format
-	debeziumRecord, err := record.ToDebeziumFormat(i.config.IcebergDatabase, i.stream.Name(), i.stream.NormalizationEnabled())
+	// Get thread ID
+	threadID := getGoroutineID()
 
+	// Convert record to Debezium format with thread ID
+	debeziumRecord, err := record.ToDebeziumFormat(i.config.IcebergDatabase, i.stream.Name(), i.stream.NormalizationEnabled(), threadID)
 	if err != nil {
 		return fmt.Errorf("failed to convert record: %v", err)
 	}
-	// Add the record to the batch
-	flushed, err := addToBatch(i.configHash, debeziumRecord, i.client)
-	if err != nil {
-		return fmt.Errorf("failed to add record to batch: %s", err)
+
+	// Add the record to the local buffer directly
+	if i.localBuffer == nil {
+		i.localBuffer = &LocalBuffer{
+			records: make([]string, 0, 100),
+			size:    0,
+		}
 	}
 
-	// If the batch was flushed, log the event
-	if flushed {
+	recordSize := int64(len(debeziumRecord))
+	i.localBuffer.records = append(i.localBuffer.records, debeziumRecord)
+	i.localBuffer.size += recordSize
+
+	// Check if buffer should be flushed
+	if i.localBuffer.size >= maxBatchSize {
+		err := flushLocalBuffer(i.localBuffer, i.client)
+		if err != nil {
+			return fmt.Errorf("failed to flush buffer: %s", err)
+		}
 		logger.Infof("Batch flushed to Iceberg server for stream %s", i.stream.Name())
 	}
 
@@ -82,17 +102,48 @@ func (i *Iceberg) Write(_ context.Context, record types.RawRecord) error {
 	return nil
 }
 
-func (i *Iceberg) Close(_ context.Context) error {
-	err := flushBatch(i.configHash, i.client)
-	if err != nil {
-		logger.Errorf("Error flushing batch on close: %s", err)
-		return err
+func (i *Iceberg) Close(ctx context.Context) error {
+	// skip flushing on error
+	defer func() {
+		err := i.CloseIcebergClient()
+		if err != nil {
+			logger.Errorf("Error closing Iceberg client: %s", err)
+		}
+	}()
+	if ctx.Err() != nil {
+		return nil
 	}
 
-	err = i.CloseIcebergClient()
-	if err != nil {
-		return fmt.Errorf("error closing Iceberg client: %s", err)
+	// First flush any remaining data in the current thread's buffer
+	if i.localBuffer != nil && len(i.localBuffer.records) > 0 {
+		err := flushLocalBuffer(i.localBuffer, i.client)
+		if err != nil {
+			logger.Errorf("Error flushing buffer on close: %s", err)
+			return err
+		}
 	}
+
+	// Get the current thread ID
+	threadID := getGoroutineID()
+
+	// Send commit request for this thread using a special message format
+	ctx, cancel := context.WithTimeout(context.Background(), 1000*time.Second)
+	defer cancel()
+
+	// Create a special commit message
+	commitMessage := fmt.Sprintf(`{"commit": true, "thread_id": "%s"}`, threadID)
+
+	req := &proto.RecordIngestRequest{
+		Messages: []string{commitMessage},
+	}
+
+	res, err := i.client.SendRecords(ctx, req)
+	if err != nil {
+		logger.Errorf("Error sending commit message for thread %s: %s", threadID, err)
+		return fmt.Errorf("failed to send commit message for thread %s: %s", threadID, err)
+	}
+
+	logger.Infof("Sent commit message for thread %s: %s", threadID, res.GetResult())
 
 	return nil
 }
@@ -104,7 +155,7 @@ func (i *Iceberg) Check(ctx context.Context) error {
 
 	// Temporarily set stream to nil and clear partition fields to force a new server for the check
 	i.stream = nil
-	i.partitionInfo = make(map[string]string)
+	i.partitionInfo = make([]types.PartitionInfo, 0)
 
 	// Create a temporary setup for checking
 	err := i.SetupIcebergClient(false)
--- a/destination/iceberg/iceberg.go
+++ b/destination/iceberg/iceberg.go
@@ -75,7 +75,7 @@ func (i *Iceberg) Setup(stream types.StreamInterface, options *destination.Optio
 	return i.SetupIcebergClient(!options.Backfill)
 }
 
-func (i *Iceberg) Write(_ context.Context, record types.RawRecord) error {
+func (i *Iceberg) Write(ctx context.Context, record types.RawRecord) error {
 	// Convert record to Debezium format with thread ID
 	// We are adding the thread ID to process the records from multiple threads in parallel and separately so that we can commit when each thread finishes.
 	debeziumRecord, err := record.ToDebeziumFormat(i.config.IcebergDatabase, i.stream.Name(), i.stream.NormalizationEnabled(), i.threadID)
@@ -88,11 +88,11 @@ func (i *Iceberg) Write(_ context.Context, record types.RawRecord) error {
 
 	// Check if buffer should be flushed
 	if i.localBuffer.size >= maxBatchSize {
-		err := flushLocalBuffer(i.localBuffer, i.client)
+		err := i.flushLocalBuffer(ctx, i.localBuffer)
 		if err != nil {
-			return fmt.Errorf("failed to flush buffer: %s", err)
+			return fmt.Errorf("thread id %s: failed to flush buffer: %s", i.threadID, err)
 		}
-		logger.Infof("Batch flushed to Iceberg server for stream %s", i.stream.Name())
+		logger.Infof("thread id %s: Batch flushed to Iceberg server for stream %s", i.threadID, i.stream.Name())
 	}
 
 	i.records.Add(1)
@@ -104,7 +104,7 @@ func (i *Iceberg) Close(ctx context.Context) error {
 	defer func() {
 		err := i.CloseIcebergClient()
 		if err != nil {
-			logger.Errorf("Error closing Iceberg client: %s", err)
+			logger.Errorf("thread id %s: Error closing Iceberg client: %s", i.threadID, err)
 		}
 	}()
 	if ctx.Err() != nil {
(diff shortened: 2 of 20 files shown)


constants/constants.go
```diff
@@ -8,7 +8,7 @@ const (
 	DefaultDiscoverTimeout = 5 * time.Minute
 	DefaultRetryTimeout    = 60 * time.Second
 	ParquetFileExt         = "parquet"
-	PartitionRegex         = `\{([^}]+)\}`
+	PartitionRegex         = `\{([^,]+),\s*([^}]+)\}`
```
shubham19may: partitioning was now working with previous regex, reverted it

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeRowsIngester.java
```diff
@@ -60,6 +61,27 @@ public void sendRecords(RecordIngest.RecordIngestRequest request, StreamObserver
...
 
         try {
+            // First, check if this is a commit request
+            if (messages.size() == 1) {
+                String message = messages.get(0);
+                Map<String, Object> messageMap = objectMapper.readValue(message, new TypeReference<Map<String, Object>>() {});
+                if (messageMap.containsKey("commit") && messageMap.get("commit").equals(true) && messageMap.containsKey("thread_id")) {
+                    // This is a commit request
+                    String threadId = (String) messageMap.get("thread_id");
+                    LOGGER.info("{} Received commit request for thread: {}", requestId, threadId);
```
hash-data, line 72: Shouldn't we add an empty thread id check?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -203,73 +210,69 @@ public void addToTable(Table icebergTable, List<RecordConverter> events) {
...
-    } catch (Exception ex) {
-      LOGGER.error("Failed to write data to table: {}", icebergTable.name(), ex);
-      
+      // Commit the files
+      synchronized(commitLock) {
+        try {
+          // Refresh table before committing
+          table.refresh();
+          
+          if (files.deleteFiles().length > 0) {
```
hash-data: A doubt here: are delete files also being created in the case of append mode?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -203,73 +210,69 @@ public void addToTable(Table icebergTable, List<RecordConverter> events) {
...
+            Arrays.stream(files.dataFiles()).forEach(rowDelta::addRows);
+            Arrays.stream(files.deleteFiles()).forEach(rowDelta::addDeletes);
+            rowDelta.commit();
+          } else {
+            AppendFiles appendFiles = table.newAppend();
+            Arrays.stream(files.dataFiles()).forEach(appendFiles::appendFile);
+            appendFiles.commit();
+          }
+          
+          LOGGER.info("Successfully committed {} data files and {} delete files for thread: {}",
```
hash-data: log for append files as well

constants/constants.go
hash-data: it will break for parquet we need to check

destination/iceberg/iceberg.go
```diff
@@ -82,13 +91,44 @@ func (i *Iceberg) Write(_ context.Context, record types.RawRecord) error {
 	return nil
 }
 
-func (i *Iceberg) Close(_ context.Context) error {
+func (i *Iceberg) Close(ctx context.Context) error {
+	// First flush any remaining data in the batch
```
hash-data:
```suggestion
	// skip flushing on error
```

destination/iceberg/iceberg.go
```diff
@@ -82,13 +91,44 @@ func (i *Iceberg) Write(_ context.Context, record types.RawRecord) error {
...
 	if err != nil {
 		logger.Errorf("Error flushing batch on close: %s", err)
 		return err
 	}
 
+	// Get the current thread ID
+	threadID := getGoroutineID()
+
+	// Send commit request for this thread using a special message format
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
hash-data: can we use context from main, and also is it ok to have only 30 seconds for commit?

destination/iceberg/iceberg.go
```diff
@@ -26,7 +32,7 @@ type Iceberg struct {
 	port          int
 	backfill      bool
 	configHash    string
-	partitionInfo map[string]string // map of field names to partition transform
+	partitionInfo []PartitionInfo // ordered slice to preserve partition column order
```
hash-data, line 29: nice one

destination/iceberg/iceberg_utils.go
```diff
@@ -488,7 +493,8 @@ func getOrCreateBatch(configHash string) *recordBatch {
 // getLocalBuffer gets or creates a local buffer for the current goroutine
 func getLocalBuffer(configHash string) *LocalBuffer {
 	// Create a unique key for this goroutine + config hash
-	bufferID := fmt.Sprintf("%s-%s", configHash, getGoroutineID())
+	goroutineID := getGoroutineID()
```
hash-data: assuming it is unique for each chunk right.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -30,6 +30,7 @@
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
```
hash-data: unused

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
hash-data: and also we can add comment that we are writing delete entry for each record first before entering new one so delete file will always be there

destination/iceberg/iceberg.go
```diff
@@ -82,13 +91,44 @@ func (i *Iceberg) Write(_ context.Context, record types.RawRecord) error {
...
 
-func (i *Iceberg) Close(_ context.Context) error {
+func (i *Iceberg) Close(ctx context.Context) error {
+	// First flush any remaining data in the batch
+	if ctx.Err() != nil {
+		err := i.CloseIcebergClient()
+		if err != nil {
+			logger.Errorf("Error closing Iceberg client: %s", err)
+		}
+		return nil
```
hash-data: repeating code below

destination/iceberg/iceberg.go
```diff
@@ -82,13 +91,44 @@ func (i *Iceberg) Write(_ context.Context, record types.RawRecord) error {
 	return nil
 }
 
-func (i *Iceberg) Close(_ context.Context) error {
+func (i *Iceberg) Close(ctx context.Context) error {
+	// First flush any remaining data in the batch
+	if ctx.Err() != nil {
```
hash-data: currently only ctx error is taken care of, but let us say ctx not got cancelled yet but there is error in thread

hash-data:
as we are writitng delete record first for each entry, do we require to clean prev data while retrying on any chunk? currently cleaning of chunk also not happening on error

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeRowsIngester.java
shubham19may: Not needed as its always passed.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
shubham19may: > A doubt here: are delete files also being created in the case of append mode?

NO

> and also we can add comment that we are writing delete entry for each record first before entering new one so delete file will always be there

Not here, as its in another place where we have removed insert operation check. Will add there.

destination/iceberg/iceberg_utils.go
shubham19may: Yes, its always unique per goroutine (which is there for each chunk as a writer thread)

shubham19may:
> as we are writitng delete record first for each entry, do we require to clean prev data while retrying on any chunk? currently cleaning of chunk also not happening on error

This will be handled by iceberg table optimisation which clears orphan files.

destination/iceberg/iceberg.go
shubham19may: lets discuss.

destination/iceberg/iceberg_utils.go
```diff
@@ -651,9 +742,18 @@ func sendRecords(records []string, client proto.RecordIngestServiceClient) error
 		return nil
 	}
 
+	// Sort records based on partition values
+	sortRecordsByPartitionValues(validRecords)
```
hash-data: sort should only happen if partition provided

types/data_types.go
```diff
@@ -65,6 +65,18 @@ type RawRecord struct {
...
+// PartitionInfo represents a Iceberg partition column with its transform, preserving order
+type PartitionInfo struct {
+	Field     string
+	Transform string
+}
+
+// DebeziumRecordWithPartition represents a Debezium record with its partition values
+type DebeziumRecordWithPartition struct {
+	Record          string
+	PartitionValues []any // ordered list of partition values
```
hash-data: did not get point to store partition values for each record?

hash-data: means these are same for each record

destination/iceberg/iceberg_utils.go
```diff
@@ -479,16 +492,94 @@ func (i *Iceberg) CloseIcebergClient() error {
...
+	sort.Slice(records, func(i, j int) bool {
+		// Compare partition values element by element
+		for idx := 0; idx < len(records[i].PartitionValues); idx++ {
+			valI := records[i].PartitionValues[idx]
+			valJ := records[j].PartitionValues[idx]
+
+			// Handle nil values - nil is considered less than any non-nil value
+			if valI == nil && valJ == nil {
+				continue
+			}
```
hash-data: a simplified version could be do fmt.Sprintf and compare them if equal continue else  use utils comparator to compare values.

hash-data: we need to handle internal olake time struct

pkg/waljs/waljs.go
```diff
@@ -189,6 +189,7 @@ func (s *Socket) StreamMessages(ctx context.Context, callback abstract.CDCMsgFn)
 						return fmt.Errorf("failed to ack lsn: %s", err)
 					}
 				}
+				s.ClientXLogPos = pkm.ServerWALEnd
```
hash-data: duplicate

types/data_types.go
```diff
@@ -65,6 +65,12 @@ type RawRecord struct {
 	CdcTimestamp   time.Time      `parquet:"_cdc_timestamp"`
 }
 
+// PartitionInfo represents a Iceberg partition column with its transform, preserving order
```
hash-data: specific to iceberg writer should not it be in iceberg file only

utils/typeutils/compare.go
```diff
@@ -0,0 +1,24 @@
...
+		return 1
+	}
+
+	aTime, aOk := a.(Time)
+	bTime, bOk := b.(Time)
+
+	if aOk && bOk {
+		return aTime.Compare(bTime)
+	}
+	return utils.CompareInterfaceValue(a, b)
```
hash-data, line 23: can we move the whole function here only, it is irrelevant in utils as it is type Comparator and we can move it to rhe file where we do reformat value as well

utils/utils.go
```diff
@@ -288,29 +288,52 @@ func AddConstantToInterface(val interface{}, increment int) (interface{}, error)
...
+	case bool:
+		bBool := b.(bool)
+		// false < true
+		if !aVal && bBool {
+			return -1
+		} else if aVal && !bBool {
+			return 1
 		}
-		return Ternary(a == nil, -1, 1).(int)
+		return 0
```
hash-data, line 338:
```suggestion
        bVal := b.(bool)
		// check for equal 
        if aVal == bVal {
            return 0
        } else if aVal {
            return 1
        } 
		return -1
```

utils/utils.go
```diff
@@ -288,29 +288,52 @@ func AddConstantToInterface(val interface{}, increment int) (interface{}, error)
...
+	// Handle nil cases first
+	if a == nil && b == nil {
+		return 0
+	}
+	if a == nil {
+		return -1
+	}
+	if b == nil {
+		return 1
+	}
```
hash-data, line 308: can use switch case

destination/iceberg/iceberg.go
```diff
@@ -61,39 +68,83 @@ func (i *Iceberg) Setup(stream types.StreamInterface, options *destination.Optio
...
-	if err != nil {
-		return fmt.Errorf("failed to add record to batch: %s", err)
+
+	// Add the record to the local buffer directly
+	if i.localBuffer == nil {
+		i.localBuffer = &LocalBuffer{
+			records: make([]string, 0, 100),
+			size:    0,
+		}
 	}
```
hash-data: why this check ideally it should not be null here.

destination/iceberg/iceberg.go
```diff
@@ -61,39 +68,83 @@ func (i *Iceberg) Setup(stream types.StreamInterface, options *destination.Optio
...
 
-	// If the batch was flushed, log the event
-	if flushed {
+	recordSize := int64(len(debeziumRecord))
+	i.localBuffer.records = append(i.localBuffer.records, debeziumRecord)
+	i.localBuffer.size += recordSize
```
hash-data:
```suggestion
	i.localBuffer.size += int64(len(debeziumRecord))
```

destination/iceberg/iceberg.go
```diff
@@ -61,39 +68,83 @@ func (i *Iceberg) Setup(stream types.StreamInterface, options *destination.Optio
...
+			logger.Errorf("Error closing Iceberg client: %s", err)
+		}
+	}()
+	if ctx.Err() != nil {
+		return nil
 	}
 
-	err = i.CloseIcebergClient()
+	// First flush any remaining data in the current thread's buffer
+	if i.localBuffer != nil && len(i.localBuffer.records) > 0 {
```
hash-data: local buffer should not null here we need to check if it getting null here

destination/iceberg/iceberg_utils.go
```diff
@@ -194,10 +180,16 @@ func (i *Iceberg) getServerConfigJSON(port int, upsert bool) ([]byte, error) {
 		"write.format.default": "parquet",
 	}
 
-	// Add partition fields if defined
-	for field, transform := range i.partitionInfo {
-		partitionKey := fmt.Sprintf("partition.field.%s", field)
-		serverConfig[partitionKey] = transform
+	// Add partition fields as an array to preserve order
+	if len(i.partitionInfo) > 0 {
+		partitionFields := make([]map[string]string, 0, len(i.partitionInfo))
```
hash-data, line 185: is order not matter here?

destination/iceberg/iceberg_utils.go
```diff
@@ -324,9 +318,9 @@ func (i *Iceberg) SetupIcebergClient(upsert bool) error {
 	// Start the Java server process
 	// If debug mode is enabled and stream is available (stream is nil for check operations), start the server with debug options
 	if i.config.DebugMode && i.stream != nil {
```
hash-data: Debug mode should be a env variable not config param

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/RecordConverter.java
```diff
@@ -154,6 +162,33 @@ public String destination() {
...
+    // If not set during construction, try to extract from value data
+    // This is a fallback and shouldn't normally be used since thread_id
+    // is passed at the message level, not in the value data
+    try {
+      JsonNode valueNode = value();
+      if (valueNode != null && valueNode.has("thread_id")) {
+        return valueNode.get("thread_id").asText();
+      }
+    } catch (Exception e) {
+      LOGGER.debug("Failed to extract thread ID from value data", e);
```
hash-data: helper functions must raise exception so that main function can catch it

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
```diff
@@ -18,7 +18,10 @@
 
 import java.util.Collections;
 import java.util.Map;
+import java.util.List;
+import java.util.LinkedHashMap;
```
hash-data: unused import

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
```diff
@@ -95,22 +109,22 @@ public static void main(String[] args) throws Exception {
...
         // Build the server to listen on port 50051
         int port = 50051; // Default port
-        if (configMap.get("port") != null) {
-            port = Integer.parseInt(configMap.get("port"));
+        if (stringConfigMap.get("port") != null) {
+            port = Integer.parseInt(stringConfigMap.get("port"));
         }
         
         // Get max message size from config or use a reasonable default (500MB)
         int maxMessageSize = 2000 * 1024 * 1024; // 2GB default
```
hash-data: according to comment it need to be 500 mb?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
```diff
@@ -48,36 +51,47 @@ public static void main(String[] args) throws Exception {
...
 
-        configMap.forEach(hadoopConf::set);
-        icebergProperties.putAll(configMap);
+        // Convert all config values to strings for hadoopConf
+        Map<String, String> stringConfigMap = new ConcurrentHashMap<>();
+        configMap.forEach((key, value) -> {
+            if (value != null && !"partition-fields".equals(key)) {
+                stringConfigMap.put(key, value.toString());
+            }
+        });
```
hash-data, line 65: cant we directly get the key from config map instead of iterating ?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
```diff
@@ -48,36 +51,47 @@ public static void main(String[] args) throws Exception {
...
-        // Format: "partition.field.<fieldName>=<transform>"
-        // Example: "partition.field.ts_ms=day" or "partition.field.id=identity"
-        for (Map.Entry<String, String> entry : configMap.entrySet()) {
-            if (entry.getKey().startsWith("partition.field.")) {
-                String fieldName = entry.getKey().substring("partition.field.".length());
-                String transform = entry.getValue();
-                partitionTransforms.put(fieldName, transform);
-                LOGGER.info("Adding partition field: {} with transform: {}", fieldName, transform);
+        // Parse partition fields from array to preserve order
+        if (configMap.containsKey("partition-fields")) {
```
hash-data, line 83: we are using stringConfigMap or configMap?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
```diff
@@ -48,36 +51,47 @@ public static void main(String[] args) throws Exception {
...
+            List<Map<String, String>> partitionFieldsList = (List<Map<String, String>>) configMap.get("partition-fields");
+            if (partitionFieldsList != null) {
+                for (Map<String, String> partitionField : partitionFieldsList) {
+                    String field = partitionField.get("field");
+                    String transform = partitionField.get("transform");
+                    if (field != null && transform != null) {
```
hash-data, line 89: can we take default transform?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeRowsIngester.java
```diff
@@ -118,15 +144,17 @@ public void sendRecords(RecordIngest.RecordIngestRequest request, StreamObserver
...
-        return IcebergUtil.loadIcebergTable(icebergCatalog, tableId).orElseGet(() -> {
-            try {
-                return IcebergUtil.createIcebergTable(icebergCatalog, tableId, sampleEvent.icebergSchema(true), "parquet", partitionTransforms);
-            } catch (Exception e) {
-                String errorMessage = String.format("Failed to create table from debezium event schema: %s Error: %s", tableId, e.getMessage());
-                LOGGER.error(errorMessage, e);
-                throw new DebeziumException(errorMessage, e);
-            }
-        });
+        synchronized (tableCreationLock) {
```
hash-data, line 148: can you add a comment for explanation of lock requirement?

hash-data (commented):
we are not sorting records, or are we can you mention ?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
shubham19may: no, iceberg has to have a transform type.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
shubham19may: In the previous line we added all records except `partition-field` in the stringConfigMap thats why we are using that.
Otherwise we are using normal configmap

Line 62

```
        Map<String, String> stringConfigMap = new ConcurrentHashMap<>();
        configMap.forEach((key, value) -> {
            if (value != null && !"partition-fields".equals(key)) {
                stringConfigMap.put(key, value.toString());
            }
        });
```

destination/iceberg/iceberg_utils.go
shubham19may: That was merged as a part of older PR, i will add a todo for it.

destination/iceberg/iceberg_utils.go
shubham19may: Its array of map. so keeping the order maintained right?

hash-data: yeah checked it

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
shubham19may: previously it was string, string map. but now its a string to object map, so we are creating a new string map, this is needed to be passed to the iceberg config.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
shubham19may: Changing it to 1gb as we have reduced the size in the go code to max 800mb.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/RecordConverter.java
shubham19may: Done

types/data_types.go
```diff
@@ -128,6 +128,11 @@ func (r *RawRecord) ToDebeziumFormat(db string, stream string, normalization boo
 		},
 	}
 
+	// Add thread_id if not empty
+	if threadID != "" {
+		debeziumRecord["thread_id"] = threadID
+	}
+
```
hash-data, line 135: should not we return error if thread id is empty ?

pkg/waljs/waljs.go
```diff
@@ -188,6 +188,7 @@ func (s *Socket) StreamMessages(ctx context.Context, callback abstract.CDCMsgFn)
 					if err != nil {
 						return fmt.Errorf("failed to ack lsn: %s", err)
 					}
+					s.ClientXLogPos = pkm.ServerWALEnd
```
hash-data: still duplicate, on line 183 we already have

destination/iceberg/iceberg.go
```diff
@@ -61,39 +76,69 @@ func (i *Iceberg) Setup(stream types.StreamInterface, options *destination.Optio
...
 
-func (i *Iceberg) Close(_ context.Context) error {
-	err := flushBatch(i.configHash, i.client)
-	if err != nil {
-		logger.Errorf("Error flushing batch on close: %s", err)
-		return err
+func (i *Iceberg) Close(ctx context.Context) error {
+	// skip flushing on error
+	defer func() {
+		err := i.CloseIcebergClient()
```
hash-data, line 105: if we check CloseIcebergClient function it never returns err (only nil returned)

destination/iceberg/iceberg_utils.go
```diff
@@ -655,6 +511,7 @@ func sendRecords(records []string, client proto.RecordIngestServiceClient) error
 		return nil
 	}
 
+	logger.Infof("Sending batch to Iceberg server: %d records", len(validRecords))
```
hash-data: can add thread id as well in logs

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
hash-data: no from configMap instead of iterating over it, we can directly get key partition-fields ?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
hash-data: let us discuss

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -203,79 +221,194 @@ public void addToTable(Table icebergTable, List<RecordConverter> events) {
...
+            }
+          }
+          return 0;
+        };
+
+        // Convert and sort in single stream operation
+        // Sorting with partitions values to make sure Iceberg-java writes it without
+        // having too many files open. In a Clustered Writer way.
+        convertedRecords = events.stream()
+            .map(e -> (upsert && !icebergTable.schema().identifierFieldIds().isEmpty())
```
hash-data, line 350: can we parallelize this ?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -203,79 +221,194 @@ public void addToTable(Table icebergTable, List<RecordConverter> events) {
...
+              return cmp;
+            }
+          }
+          return 0;
+        };
+
+        // Convert and sort in single stream operation
+        // Sorting with partitions values to make sure Iceberg-java writes it without
+        // having too many files open. In a Clustered Writer way.
+        convertedRecords = events.stream()
```
hash-data, line 349: later on we can also think of priority queue based on comparator

hash-data:
common ctx

hash-data (approved):

destination/iceberg/iceberg.go
hash-data: we can skip this for now

PR #356, opened 2025-06-20
feat: testcontainers integration

ImDoubD-datazip:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/drivers/abstract/test_schema.go
+++ b/drivers/abstract/test_schema.go
@@ -1,0 +1,201 @@
+package abstract
+
+// GlobalTypeToDataType maps database-specific types (Postgres/MySQL) to internal standard types
+var GlobalTypeMapping = map[string]string{
+	// Integer Types
+	"tinyint":            "int",
+	"smallint":           "int",
+	"mediumint":          "int",
+	"int":                "int",
+	"integer":            "int",
+	"unsigned int":       "int",
+	"unsigned smallint":  "int",
+	"unsigned tinyint":   "int",
+	"unsigned mediumint": "int",
+	"int2":               "int",
+	"int4":               "int",
+	"smallserial":        "int",
+	"serial":             "int",
+	"serial2":            "int",
+	"serial4":            "int",
+
+	"bigint":    "bigint",
+	"int8":      "bigint",
+	"serial8":   "bigint",
+	"bigserial": "bigint",
+	"year":      "bigint",
+
+	// Floating Point Types
+	"float":   "float",
+	"real":    "float",
+	"decimal": "float",
+	"numeric": "float",
+	"float4":  "float",
+	"money":   "float",
+
+	"double":           "double",
+	"float8":           "double",
+	"double precision": "double",
+
+	// Boolean Types
+	"bool":    "boolean",
+	"boolean": "boolean",
+
+	// String Types
+	"char":              "string",
+	"varchar":           "string",
+	"tinytext":          "string",
+	"text":              "string",
+	"mediumtext":        "string",
+	"longtext":          "string",
+	"character":         "string",
+	"character varying": "string",
+	"longvarchar":       "string",
+	"bpchar":            "string",
+	"name":              "string",
+
+	// Binary Types
+	"binary":     "string",
+	"varbinary":  "string",
+	"tinyblob":   "string",
+	"blob":       "string",
+	"mediumblob": "string",
+	"longblob":   "string",
+	"bytea":      "string",
+
+	// JSON and Document Types
+	"json":   "string",
+	"jsonb":  "string",
+	"xml":    "string",
+	"hstore": "string",
+
+	// Network Types
+	"cidr":     "string",
+	"inet":     "string",
+	"macaddr":  "string",
+	"macaddr8": "string",
+
+	// Spatial Types
+	"geometry":           "string",
+	"point":              "string",
+	"linestring":         "string",
+	"polygon":            "string",
+	"multipoint":         "string",
+	"multilinestring":    "string",
+	"multipolygon":       "string",
+	"geometrycollection": "string",
+	"circle":             "string",
+	"path":               "string",
+	"box":                "string",
+	"line":               "string",
+	"lseg":               "string",
+
+	// Full Text Search Types
+	"tsvector": "string",
+	"tsquery":  "string",
+
+	// UUID
+	"uuid": "string",
+
+	// Range Types
+	"tsrange":   "string",
+	"tstzrange": "string",
+	"int4range": "string",
+	"numrange":  "string",
+	"daterange": "string",
+
+	// Array
+	"array":      "string",
+	"ARRAY":      "string",
+	"int2vector": "string",
+
+	// Enum and Set
+	"enum": "string",
+	"set":  "string",
+
+	// Date/Time
+	"date":                        "timestamp",
+	"timestamp":                   "timestamp",
+	"datetime":                    "timestamp",
+	"timestampz":                  "timestamp",
+	"timestamp with time zone":    "timestamp",
+	"timestamp without time zone": "timestamp",
+
+	"time":     "string",
+	"timez":    "string",
+	"interval": "string",
+
+	// Misc
+	"pg_lsn":      "string",
+	"bit varying": "string",
+	"varbit":      "string",
+	"bit(n)":      "string",
+	"varying(n)":  "string",
+}
+
+var PostgresSchema = map[string]string{
+	"col_bigint":            "bigint",
+	"col_bigserial":         "bigserial",
+	"col_bool":              "boolean",
+	"col_char":              "char",
+	"col_character":         "character",
+	"col_character_varying": "varchar",
+	"col_date":              "date",
+	"col_daterange":         "daterange",
+	"col_decimal":           "numeric",
+	"col_double_precision":  "double precision",
+	"col_float4":            "real",
+	"col_int":               "int",
+	"col_int2":              "smallint",
+	"col_integer":           "integer",
+	"col_interval":          "interval",
+	"col_json":              "json",
+	"col_jsonb":             "jsonb",
+	"col_jsonpath":          "jsonpath",
+	"col_name":              "name",
+	"col_numeric":           "numeric",
+	"col_real":              "real",
+	"col_text":              "text",
+	"col_time":              "time",
+	"col_timestamp":         "timestamp",
+	"col_timestamptz":       "timestamptz",
+	"col_timetz":            "timetz",
+	"col_uuid":              "uuid",
+	"col_varbit":            "varbit",
+	"col_xml":               "xml",
+}
+
+var MySQLSchema = map[string]string{
+	"id":                     "unsigned int",
+	"id_bigint":              "bigint",
+	"id_int":                 "int",
+	"id_int_unsigned":        "unsigned int",
+	"id_integer":             "int",
+	"id_integer_unsigned":    "unsigned int",
+	"id_mediumint":           "mediumint",
+	"id_mediumint_unsigned":  "unsigned mediumint",
+	"id_smallint":            "smallint",
+	"id_smallint_unsigned":   "unsigned smallint",
+	"id_tinyint":             "tinyint",
+	"id_tinyint_unsigned":    "unsigned tinyint",
+	"price_decimal":          "decimal",
+	"price_double":           "double",
+	"price_double_precision": "double",
+	"price_float":            "float",
+	"price_numeric":          "decimal",
+	"price_real":             "double",
+	"name_char":              "char",
+	"name_varchar":           "varchar",
+	"name_text":              "text",
+	"name_tinytext":          "tinytext",
+	"name_mediumtext":        "mediumtext",
+	"name_longtext":          "longtext",
+	"name_enum":              "enum",
+	"created_date":           "datetime",
+	"created_time":           "time",
+	"created_timestamp":      "timestamp",
+	"is_active":              "tinyint",
+	"json_data":              "json",
+	"long_varchar":           "mediumtext",
+	"name_bool":              "tinyint",
+}
--- a/drivers/abstract/test_utils.go
+++ b/drivers/abstract/test_utils.go
@@ -12,7 +12,6 @@ import (
 	"github.com/datazip-inc/olake/destination/iceberg"
 	"github.com/datazip-inc/olake/types"
 	"github.com/datazip-inc/olake/utils"
-	"github.com/datazip-inc/olake/utils/logger"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -238,7 +237,6 @@ func verifyIcebergSync(t *testing.T, tableName string, expectedCount string, sch
 	for _, row := range describeRows {
 		colName := row.Value("col_name").(string)
 		dataType := row.Value("data_type").(string)
-		logger.Infof("Column: %s, Data Type: %s", colName, dataType)
 		if !strings.HasPrefix(colName, "#") {
 			icebergSchema[colName] = dataType
 		}
--- a/drivers/abstract/test_utils.go
+++ b/drivers/abstract/test_utils.go
@@ -3,6 +3,7 @@ package abstract
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
@@ -11,6 +12,7 @@ import (
 	"github.com/datazip-inc/olake/destination/iceberg"
 	"github.com/datazip-inc/olake/types"
 	"github.com/datazip-inc/olake/utils"
+	"github.com/datazip-inc/olake/utils/logger"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -60,7 +62,7 @@ func (a *AbstractDriver) TestDiscover(t *testing.T, conn interface{}, execQuery 
 }
 
 // TestRead tests full refresh and CDC read operations
-func (a *AbstractDriver) TestRead(t *testing.T, conn interface{}, execQuery ExecuteQuery) {
+func (a *AbstractDriver) TestRead(t *testing.T, conn interface{}, execQuery ExecuteQuery, schemaMap map[string]string) {
 	t.Helper()
 	ctx := context.Background()
 
@@ -111,7 +113,7 @@ func (a *AbstractDriver) TestRead(t *testing.T, conn interface{}, execQuery Exec
 			testStream := a.getTestStream(t, currentTestTable)
 			configuredStream := &types.ConfiguredStream{Stream: testStream}
 			configuredStream.Stream.SyncMode = tc.syncMode
-
+			configuredStream.StreamMetadata.Normalization = true
 			if tc.syncMode == types.CDC {
 				// Execute the operation for CDC tests
 				execQuery(ctx, t, conn, currentTestTable, tc.operation)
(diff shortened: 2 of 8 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -20,6 +20,9 @@ jobs:
   integration-tests:
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    strategy:
```
hash-data: i think it will not be required, as we have to run tests for all

constants/constants.go
```diff
@@ -18,6 +18,7 @@ const (
 	DBName                 = "_db"
 	EncryptionKey          = "OLAKE_ENCRYPTION_KEY"
 	ConfigFolder           = "CONFIG_FOLDER"
+	IcebergDatabase        = "olake_iceberg"
```
hash-data: remove it (test constants should not be merged)

drivers/abstract/test_schema.go
```diff
@@ -0,0 +1,134 @@
+package abstract
+
+// GlobalTypeToDataType maps database-specific types (Postgres/MySQL) to internal standard types
```
hash-data: are we merging all the types in tests ? also function that does datatype testing could be placed in this file

drivers/abstract/test_utils.go
```diff
@@ -3,197 +3,23 @@ package abstract
...
-	require.Fail(t, "Could not find stream for table %s", tableName)
-	return nil
-}
-
 // verifyIcebergSync verifies that data was correctly synchronized to Iceberg
-func verifyIcebergSync(t *testing.T, tableName string, expectedCount string) {
+func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string]string, schema map[string]interface{}, opSymbol string) {
 	t.Helper()
 	ctx := context.Background()
+	sparkConnectAddress := fmt.Sprintf("sc://%s:15002", sparkConnectHost)
```
hash-data: make port constant as well

drivers/abstract/test_utils.go
```diff
@@ -203,26 +29,68 @@ func verifyIcebergSync(t *testing.T, tableName string, expectedCount string) {
...
+	if opSymbol == "d" {
+		deletedID := selectRows[0].Value("_olake_id")
+		require.Equalf(t, "1", deletedID, "Delete verification failed: expected _olake_id = '1', got %s", deletedID)
+		return
+	}
 
-	actualCount := utils.ConvertToString(countValue)
-	require.Equal(t, expectedCount, actualCount,
-		"Unique olake_id count mismatch in Iceberg")
+	require.NotEmpty(t, selectRows, "No rows returned for _op_type = '%s'", opSymbol)
```
hash-data: it should be before opSymbol check, as you are checking index there which can cause index out of bound

drivers/abstract/test_utils.go
```diff
@@ -203,26 +29,68 @@ func verifyIcebergSync(t *testing.T, tableName string, expectedCount string) {
...
+		for _, col := range row.FieldNames() {
+			icebergMap[col] = row.Value(col)
+		}
+		for key, expected := range schema {
+			icebergValue, ok := icebergMap[key]
+			require.Truef(t, ok, "Row %d: missing column %q in Iceberg result", rowIdx, key)
+			require.Equal(t, icebergValue, expected, "Row %d: mismatch on %q: Iceberg has %#v, expected %#v", rowIdx, key, icebergValue, expected)
+		}
+	}
+	t.Logf("Verified synced data in Iceberg")
```
hash-data: log must be something like: verified iceberg synced data w.r.t data synced from sounce[%s] found equal

drivers/abstract/test_utils.go
```diff
@@ -203,26 +29,68 @@ func verifyIcebergSync(t *testing.T, tableName string, expectedCount string) {
...
+		}
+	}
+
+	for col, dbType := range datatypeSchema {
+		iceType, found := icebergSchema[col]
+		require.True(t, found, "Column %s not found in Iceberg schema", col)
+
+		expectedIceType, mapped := GlobalTypeMapping[dbType]
+		if !mapped {
+			t.Logf("No mapping defined for driver type %s (column %s), skipping check", dbType, col)
```
hash-data: it must fail

drivers/mysql/docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     container_name: olake_mysql-test
     restart: unless-stopped
     environment:
-      MYSQL_DATABASE: mysql
+      MYSQL_DATABASE: olake_mysql_test
```
hash-data, line 9: let us not change it as users might use this for local testing and in docs we have use mysql

drivers/postgres/internal/postgres_test.go
```diff
@@ -1,21 +1,252 @@
...
+							// 4. Verify streams.json file
+							streamsJSON, err := os.ReadFile(dummyStreamFilePath)
+							if err != nil {
+								return fmt.Errorf("failed to read expected streams JSON: %w", err)
+							}
+							testStreamsCmd := fmt.Sprintf("cat %s", streamsPath)
+							_, testStreamJSON, err := utils.ExecCommand(ctx, c, testStreamsCmd)
+							if err != nil {
+								return fmt.Errorf("failed to read actual streams JSON: %w", err)
+							}
```
hash-data: why you cant read this file as well? instead of cat ?

drivers/postgres/internal/postgres_test.go
```diff
@@ -1,21 +1,252 @@
...
+							t.Logf("GeneratedStreamJson: %s", strings.TrimSpace(string(streamsJSON)))
+							expectedStream, err := utils.SortJSONString(strings.TrimSpace(string(streamsJSON)))
+							if err != nil {
+								return fmt.Errorf("failed to sort expected JSON as string: %w", err)
+							}
+							t.Logf("testStreamJson: %s", strings.TrimSpace(string(testStreamJSON)))
```
hash-data: no need to log these

protocol/root.go
```diff
@@ -37,6 +37,11 @@ var RootCmd = &cobra.Command{
 	Use:   "olake",
 	Short: "root command",
 	RunE: func(cmd *cobra.Command, args []string) error {
+		viper.AutomaticEnv()
```
hash-data: is it not working by just viper.GetString() ? if not let us use os.getenv for now

utils/telemetry/telemetry.go
```diff
@@ -144,15 +144,15 @@ func TrackSyncCompleted(status bool, records int64) {
...
+			logger.Debugf("Failed to send SyncCompleted event: %v", err)
 		}
 	}()
 }
 
 func (t *Telemetry) sendEvent(eventName string, properties map[string]interface{}) error {
 	if t.client == nil {
-		logger.Warn("Telemetry client is nil, not sending event:", eventName)
-		return fmt.Errorf("telemetry client is nil")
+		logger.Debug("Telemetry client is nil, not sending event:", eventName)
```
hash-data: remove debug log from here just return err parent function will handle

utils/utils.go
```diff
@@ -2,11 +2,14 @@ package utils
 
 import (
 	//nolint:gosec,G115
+	"context"
+	//nolint:gosec
```
hash-data: why extra ?

utils/utils.go
```diff
@@ -369,3 +373,48 @@ func ComputeConfigHash(srcPath, destPath string) string {
...
+) (int, []byte, error) {
+	code, reader, err := c.Exec(ctx, []string{"/bin/sh", "-c", cmd})
+	if err != nil {
+		return code, nil, err
+	}
+	output, _ := io.ReadAll(reader)
+	return code, output, nil
+}
+
+func SortJSONString(jsonStr string) (string, error) {
```
hash-data: can you check function https://github.com/datazip-inc/olake-ui/compare/d26c9aa6503c2764f2ca899d0f244380ed9caff0..746ced98579b664e7721004034569be0c00f4a6b

go.mod
```diff
@@ -33,7 +33,10 @@ require (
 
 require (
 	cloud.google.com/go/compute/metadata v0.6.0 // indirect
+	dario.cat/mergo v1.0.1 // indirect
```
hash-data: check if these are required?

go.mod
```diff
@@ -110,10 +157,21 @@ require (
...
 )
 
 replace (
+	cloud.google.com/go/compute/metadata => cloud.google.com/go/compute/metadata v0.6.0
 	google.golang.org/genproto => google.golang.org/genproto v0.0.0-20240123012728-ef4313101c80
-	google.golang.org/genproto/googleapis/rpc => google.golang.org/genproto/googleapis/rpc v0.0.0-20240123012728-ef4313101c80
 )
+
+// replace (
+// 	google.golang.org/genproto => google.golang.org/genproto v0.0.0-20240123012728-ef4313101c80
```
hash-data: remove these

hash-data (commented):
Good Progress, Some of functions can be extracted out, if you can do will be good.

utils/utils.go
hash-data: image

.github/workflows/integration-tests.yml
ImDoubD-datazip: ok, removing this

drivers/abstract/test_schema.go
ImDoubD-datazip: this is for data type testing only. the verify iceberg does schema data and datatype verification both. we pass the dummy data schema map, and corresponding driver datatype map as well.

drivers/abstract/test_utils.go
ImDoubD-datazip: it is constant already 

const (
	icebergDatabase  = "olake_iceberg"
	sparkConnectHost = "localhost"
)

drivers/abstract/test_utils.go
ImDoubD-datazip: yes, my bad

drivers/abstract/test_utils.go
ImDoubD-datazip: it is a global datatype map, it will have datatypes of various drivers. so sometime a driver may not have that particular datatype to map with so it should not fail as it does not have it. already it is hard-coded

drivers/mysql/docker-compose.yml
ImDoubD-datazip: there is some problem in keeping mysql as database name, it shows lot of other tables which i don't know from where they are coming. even after dropping and creating new database, those tables come.

drivers/postgres/internal/postgres_test.go
ImDoubD-datazip: exec command takes place inside the container. the os.read will take place in host container

protocol/root.go
ImDoubD-datazip: we needed to add viper.AutomaticEnv() as it was set in env of the container. changed to os.getenv for now

utils/utils.go
ImDoubD-datazip: it was giving lint error
image

utils/utils.go
ImDoubD-datazip: done, using this

drivers/abstract/test_utils.go
hash-data: port: 15002

drivers/abstract/test_utils.go
hash-data: if does not have it then we have to add right

drivers/postgres/internal/postgres_test.go
hash-data: did not get you

utils/utils.go
hash-data: the above one should be used

drivers/abstract/test_utils.go
```diff
@@ -3,198 +3,256 @@ package abstract
...
+					fmt.Sprintf("%s:/test-olake:rw", projectRoot),
+					fmt.Sprintf("%s:/test-olake/drivers/%s/internal/testdata:rw", testdataDir, driver),
+				}
+				hc.ExtraHosts = append(hc.ExtraHosts, "host.docker.internal:host-gateway")
+			},
+			ConfigModifier: func(config *container.Config) {
+				config.WorkingDir = "/test-olake"
+			},
+			Env: map[string]string{
+				"DEBUG": "false",
```
hash-data: debug mode should true here, and on true telemetry should not start

drivers/abstract/test_utils.go
```diff
@@ -3,198 +3,256 @@ package abstract
...
-func (a *AbstractDriver) TestSetup(t *testing.T) {
-	t.Helper()
-	require.NoError(t, a.Setup(context.Background()), "Connection check failed")
-}
+type ExecuteQuery func(ctx context.Context, t *testing.T, db interface{}, tableName, operation string)
 
-// TestDiscover tests the discovery of tables
-func (a *AbstractDriver) TestDiscover(t *testing.T, conn interface{}, execQuery ExecuteQuery) {
-	t.Helper()
+func TestIntegration(t *testing.T, driver, sourceConfigPath, streamsPath, destinationConfigPath, statePath, namespace string, expectedData, expectedUpdatedData map[string]interface{}, datatypeSchema map[string]string, execQuery ExecuteQuery, testSetup func(ctx context.Context, t *testing.T) interface{}) {
```
hash-data: instead of passing these much functions can we do create a interface here that every driver test suite follow. Like one function could be ExecuteQuery and other could be ValidateRecords in test interface that each driver will have. Other then that just loop through driver array and run test container parallel

drivers/abstract/test_utils.go
```diff
@@ -3,198 +3,256 @@ package abstract
...
+							if code, out, err := utils.ExecCommand(ctx, c, discoverCmd); err != nil || code != 0 {
+								return fmt.Errorf("discover failed (%d): %w\n%s", code, err, string(out))
+							}
+
+							// 4. Verify streams.json file
+							streamsJSON, err := os.ReadFile(dummyStreamFilePath)
+							if err != nil {
+								return fmt.Errorf("failed to read expected streams JSON: %w", err)
+							}
+							testStreamsCmd := fmt.Sprintf("cat %s", streamsPath)
```
hash-data: cat is still happening?

drivers/abstract/test_utils.go
```diff
@@ -203,26 +261,65 @@ func verifyIcebergSync(t *testing.T, tableName string, expectedCount string) {
...
+		deletedID := selectRows[0].Value("_olake_id")
+		require.Equalf(t, "1", deletedID, "Delete verification failed: expected _olake_id = '1', got %s", deletedID)
+		return
+	}
 
-	countValue := countRows[0].Value("unique_count")
-	require.NotNil(t, countValue, "Count value is nil")
+	for rowIdx, row := range selectRows {
+		icebergMap := make(map[string]interface{}, len(schema)+1)
+		for _, col := range row.FieldNames() {
```
hash-data: just init map without size

drivers/abstract/test_utils.go
hash-data: you have not updated it

drivers/mysql/internal/mysql_test.go
```diff
@@ -1,22 +1,32 @@
...
+	)
+	require.NoError(t, err, "failed to connect to  mysql")
+	return db
 }
 
-func TestMySQLRead(t *testing.T) {
-	conn, abstractDriver := testAndBuildAbstractDriver(t)
-	abstractDriver.TestRead(t, conn, ExecuteQuery)
+func TestMySQLIntegration(t *testing.T) {
+	abstract.TestIntegration(t, string(constants.MySQL), sourceConfigPath, streamsPath, destinationConfigPath, statePath, namespace, ExpectedMySQLData, ExpectedUpdatedMySQLData, MySQLSchema, ExecuteQuery, testSetup)
```
hash-data: these all be removed only abstract will orchestrate driver will provide two function validateRecords/Schema and ExecQuery  via abstract_test interface

drivers/abstract/test_utils.go
ImDoubD-datazip: will remove this

drivers/abstract/test_utils.go
ImDoubD-datazip: understood

utils/testutils/test_utils.go
```diff
@@ -0,0 +1,340 @@
...
+	"github.com/testcontainers/testcontainers-go"
+)
+
+const (
+	icebergDatabase     = "olake_iceberg"
+	sparkConnectAddress = "sc://localhost:15002"
+	installCmd          = "apt-get update && apt-get install -y openjdk-17-jre-headless maven default-mysql-client postgresql postgresql-client iproute2 dnsutils iputils-ping netcat-openbsd nodejs npm jq && npm install -g chalk-cli"
+)
+
+type TestInterface struct {
```
hash-data:
```suggestion
type IntegrationTest struct {
```

utils/testutils/test_utils.go
```diff
@@ -0,0 +1,340 @@
...
+	Driver             string
+	Namespace          string
+	ExpectedData       map[string]interface{}
+	ExpectedUpdateData map[string]interface{}
+	DataTypeSchema     map[string]string
+	ExecuteQuery       func(ctx context.Context, t *testing.T, db interface{}, tableName, operation string)
+	TestSetup          func(ctx context.Context, t *testing.T) interface{}
```
hash-data: is this required in this struct ?

utils/testutils/test_utils.go
```diff
@@ -0,0 +1,340 @@
...
+
+const (
+	icebergDatabase     = "olake_iceberg"
+	sparkConnectAddress = "sc://localhost:15002"
+	installCmd          = "apt-get update && apt-get install -y openjdk-17-jre-headless maven default-mysql-client postgresql postgresql-client iproute2 dnsutils iputils-ping netcat-openbsd nodejs npm jq && npm install -g chalk-cli"
+)
+
+type TestInterface struct {
+	Driver             string
+	Namespace          string
```
hash-data: this also can be removed

utils/testutils/test_utils.go
```diff
@@ -0,0 +1,340 @@
...
+	ExpectedData       map[string]interface{}
+	ExpectedUpdateData map[string]interface{}
```
hash-data, line 30: like execute query we can have validate data as well?

utils/testutils/test_utils.go
```diff
@@ -0,0 +1,340 @@
...
+					fmt.Sprintf("%s:/test-olake:rw", projectRoot),
+					fmt.Sprintf("%s:/test-olake/drivers/%s/internal/testdata:rw", testdataDir, cfg.Driver),
+				}
+				hc.ExtraHosts = append(hc.ExtraHosts, "host.docker.internal:host-gateway")
+			},
+			ConfigModifier: func(config *container.Config) {
+				config.WorkingDir = "/test-olake"
+			},
+			Env: map[string]string{
+				"DEBUG": "false",
```
hash-data: you can use Telemetry_disabled env var

utils/testutils/test_utils.go
```diff
@@ -0,0 +1,329 @@
...
+			Env: map[string]string{
+				"TELEMETRY_DISABLED": "true",
+			},
+			LifecycleHooks: []testcontainers.ContainerLifecycleHooks{
+				{
+					PostReadies: []testcontainers.ContainerHook{
+						func(ctx context.Context, c testcontainers.Container) error {
+							// 1. Install required tools
+							if code, out, err := utils.ExecCommand(ctx, c, installCmd); err != nil || code != 0 {
+								return fmt.Errorf("install failed (%d): %w\n%s", code, err, out)
```
hash-data: replace %w with %s

hash-data (approved):
LGTM

PR #362, opened 2025-06-23
Clear destination flag

tanishaAtDatazip:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/destination/interface.go
+++ b/destination/interface.go
@@ -35,5 +35,8 @@ type Writer interface {
 	// EvolveSchema updates the schema based on changes.
 	// Need to pass olakeTimestamp as end argument to get the correct partition path based on record ingestion time.
 	EvolveSchema(bool, bool, map[string]*types.Property, types.Record, time.Time) error
+	// ClearDestination is used to clear the destination before re-writing the stream
+	Clear(selectedStream []string) error
+	// Clear(ctx context.Context, stream types.StreamInterface, streamMetadata map[string]types.StreamMetadata) error
 	Close(ctx context.Context) error
 }
--- a/destination/parquet/parquet.go
+++ b/destination/parquet/parquet.go
@@ -367,6 +367,135 @@ func (p *Parquet) getPartitionedFilePath(values map[string]any, olakeTimestamp t
 	return filepath.Join(p.basePath, strings.TrimSuffix(result, "/"))
 }
 
+func (p *Parquet) Clear(selectedStream []string) error {
+	if len(selectedStream) == 0 {
+		logger.Info("No streams selected for clearing, skipping clear operation")
+		return nil
+	}
+
+	logger.Infof("Clearing destination for %d selected streams: %v", len(selectedStream), selectedStream)
+
+	// Clear local files
+	if err := p.clearLocalFiles(selectedStream); err != nil {
+		return fmt.Errorf("failed to clear local files: %s", err)
+	}
+
+	// Clear S3 files if configured
+	if p.s3Client != nil {
+		if err := p.clearS3Files(selectedStream); err != nil {
+			return fmt.Errorf("failed to clear S3 files: %s", err)
+		}
+	}
+
+	logger.Info("Successfully cleared destination for selected streams")
+	return nil
+}
+
+// clearLocalFiles removes local files for the specified streams
+func (p *Parquet) clearLocalFiles(selectedStream []string) error {
+	for _, streamID := range selectedStream {
+		// Parse stream ID to get namespace and stream name
+		// streamID format: "namespace.stream_name"
+		parts := strings.SplitN(streamID, ".", 2)
+		if len(parts) != 2 {
+			logger.Warnf("Invalid stream ID format: %s, skipping", streamID)
+			continue
+		}
+
+		namespace, streamName := parts[0], parts[1]
+		streamPath := filepath.Join(p.config.Path, namespace, streamName)
+
+		logger.Infof("Clearing local path: %s", streamPath)
+
+		// Check if the path exists before attempting to delete
+		if _, err := os.Stat(streamPath); os.IsNotExist(err) {
+			logger.Debugf("Local path does not exist, skipping: %s", streamPath)
+			continue
+		}
+
+		// Remove the entire stream directory and all its contents
+		if err := os.RemoveAll(streamPath); err != nil {
+			return fmt.Errorf("failed to remove local path %s: %s", streamPath, err)
+		}
+
+		logger.Infof("Successfully cleared local path: %s", streamPath)
+	}
+
+	return nil
+}
+
+// clearS3Files removes S3 files for the specified streams
+func (p *Parquet) clearS3Files(selectedStream []string) error {
+	if p.s3Client == nil {
+		return nil
+	}
+
+	for _, streamID := range selectedStream {
+		// Parse stream ID to get namespace and stream name
+		parts := strings.SplitN(streamID, ".", 2)
+		if len(parts) != 2 {
+			logger.Warnf("Invalid stream ID format: %s, skipping", streamID)
+			continue
+		}
+
+		namespace, streamName := parts[0], parts[1]
+
+		// Construct S3 prefix path
+		s3Prefix := filepath.Join(namespace, streamName)
+		if p.config.Prefix != "" {
+			s3Prefix = filepath.Join(p.config.Prefix, s3Prefix)
+		}
+
+		logger.Infof("Clearing S3 prefix: s3://%s/%s", p.config.Bucket, s3Prefix)
+
+		// List objects with the prefix
+		listInput := &s3.ListObjectsV2Input{
+			Bucket: aws.String(p.config.Bucket),
+			Prefix: aws.String(s3Prefix + "/"), // Add trailing slash to ensure we get the directory
+		}
+
+		// Delete objects in batches
+		err := p.s3Client.ListObjectsV2Pages(listInput, func(page *s3.ListObjectsV2Output, lastPage bool) bool {
+			if len(page.Contents) == 0 {
+				return true
+			}
+
+			// Prepare delete objects input
+			var objectsToDelete []*s3.ObjectIdentifier
+			for _, obj := range page.Contents {
+				objectsToDelete = append(objectsToDelete, &s3.ObjectIdentifier{
+					Key: obj.Key,
+				})
+			}
+
+			// Delete objects
+			deleteInput := &s3.DeleteObjectsInput{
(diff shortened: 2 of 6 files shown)


protocol/sync.go
```diff
@@ -140,6 +141,28 @@ var syncCmd = &cobra.Command{
 			return pool.SyncedRecords(), pool.ThreadCounter.Load(), pool.GetRecordsToSync()
 		})
 
+		clearDestinationFlag, _ := cmd.Flags().GetBool("clear-destination")
```
hash-data: it is already set in root, we can directly use that var

protocol/sync.go
```diff
@@ -140,6 +141,28 @@ var syncCmd = &cobra.Command{
...
+			logger.Info("Clearing destination for selected streams...")
+			if err := clearDestination(cmd.Context(), destinationConfig, selectedStreams); err != nil {
+				// This is a critical failure. Stop the sync.
+				return fmt.Errorf("failed to clear destination during reset: %w", err)
+			}
+			logger.Info("Destination cleared successfully.")
+
+			// Step 2: Reset the state to ensure a fresh sync from the beginning.
+			logger.Info("Resetting sync state for a fresh start.")
+			state = &types.State{
```
hash-data: can we use destination.NewWriter function and clear destination there will be good no need to register again.

destination/interface.go
```diff
@@ -35,5 +35,8 @@ type Writer interface {
 	// EvolveSchema updates the schema based on changes.
 	// Need to pass olakeTimestamp as end argument to get the correct partition path based on record ingestion time.
 	EvolveSchema(bool, bool, map[string]*types.Property, types.Record, time.Time) error
+	// ClearDestination is used to clear the destination before re-writing the stream
+	Clear(selectedStream []string) error
```
hash-data:
```suggestion
	DropStreams(selectedStream []string) error
```

destination/parquet/parquet.go
```diff
@@ -367,6 +367,135 @@ func (p *Parquet) getPartitionedFilePath(values map[string]any, olakeTimestamp t
...
+func (p *Parquet) Clear(selectedStream []string) error {
+	if len(selectedStream) == 0 {
+		logger.Info("No streams selected for clearing, skipping clear operation")
+		return nil
+	}
+
+	logger.Infof("Clearing destination for %d selected streams: %v", len(selectedStream), selectedStream)
+
+	// Clear local files
+	if err := p.clearLocalFiles(selectedStream); err != nil {
```
hash-data: it will give file not exist error for s3 credentials, would recommend to only clear local files if s3 client not present

destination/parquet/parquet.go
```diff
@@ -367,6 +367,135 @@ func (p *Parquet) getPartitionedFilePath(values map[string]any, olakeTimestamp t
...
+	}
+
+	return nil
+}
+
+// clearS3Files removes S3 files for the specified streams
+func (p *Parquet) clearS3Files(selectedStream []string) error {
+	if p.s3Client == nil {
+		return nil
+	}
```
hash-data: we are making sure in parent function that s3 client not null

destination/parquet/parquet.go
```diff
@@ -367,6 +367,135 @@ func (p *Parquet) getPartitionedFilePath(values map[string]any, olakeTimestamp t
...
+		logger.Infof("Clearing S3 prefix: s3://%s/%s", p.config.Bucket, s3Prefix)
+
+		// List objects with the prefix
+		listInput := &s3.ListObjectsV2Input{
+			Bucket: aws.String(p.config.Bucket),
+			Prefix: aws.String(s3Prefix + "/"), // Add trailing slash to ensure we get the directory
+		}
+
+		// Delete objects in batches
+		err := p.s3Client.ListObjectsV2Pages(listInput, func(page *s3.ListObjectsV2Output, lastPage bool) bool {
```
hash-data: instead of going one by one can we directly delete the dir ? or via regex?

destination/parquet/parquet.go
tanishaAtDatazip: done

destination/parquet/parquet.go
tanishaAtDatazip: got it, removed

destination/parquet/parquet.go
tanishaAtDatazip: optimised by parallel deletion

hash-data: Yeah that is good, but is it possible to directly delete the directory instead of iterating one by one?

protocol/sync.go
```diff
@@ -135,11 +131,30 @@ var syncCmd = &cobra.Command{
...
 			return pool.SyncedRecords(), pool.ThreadCounter.Load(), pool.GetRecordsToSync()
 		})
 
+		if clearDestinationFlag {
+			logger.Info("Resetting sync state for a fresh start.")
+			state = &types.State{
+				Type:    types.StreamType,
+				RWMutex: &sync.RWMutex{},
+			}
+			logger.Info("Sync state has been reset.")
```
hash-data: we can move this to line 136 as well

destination/writers.go
```diff
@@ -47,19 +67,26 @@ func WithBackfill(backfill bool) ThreadOptions {
...
-	init          NewFunc      // To initialize exclusive destination threads
-	group         *errgroup.Group
-	groupCtx      context.Context
-	tmu           sync.Mutex // Mutex between threads
+func WithClearDestination(selectedStreams []string) WriterOption {
+	return func(adapter Writer) error {
+		if len(selectedStreams) == 0 {
+			logger.Info("No streams selected for clearing, skipping clear operation")
+			return nil
+		}
```
hash-data: checks seems duplicate as we already checking in start of sync for len of selected streams

destination/writers.go
```diff
@@ -47,19 +67,26 @@ func WithBackfill(backfill bool) ThreadOptions {
...
+
+		logger.Info("Clearing destination for selected streams...")
+
+		if err := adapter.DropStreams(selectedStreams); err != nil {
+			return fmt.Errorf("failed to clear destination: %s", err)
+		}
```
hash-data: only this function left which we can directly call from new writer byt just taking flag of clearDestination

protocol/sync.go
```diff
@@ -135,11 +131,30 @@ var syncCmd = &cobra.Command{
 
 		logger.Infof("Valid selected streams are %s", strings.Join(selectedStreams, ", "))
 
+		var writerOpts []destination.WriterOption
+		if clearDestinationFlag {
+			writerOpts = append(writerOpts, destination.WithClearDestination(selectedStreams))
```
hash-data: it is ok if we are doing clear destination on flag, but addition to it we also need to do the same for streams which is having full load, for which we need to iterate in standard mode streams.

destination/parquet/parquet.go
tanishaAtDatazip: actually there's no concept of a real directory in S3. f.e., if all my parquet files are under bucket/public, they’re stored as separate keys like bucket/public/file1.parquet, bucket/public/file2.parquet, etc. S3 uses a flat namespace, so it just treats these as key strings. To `delete a directory`, you basically perform a prefix match to list all keys that start with public/, and then delete them — usually in parallel for performance. AWS CLI provides aws s3 rm --recursive, but under the hood, it also deletes each object one by one

protocol/sync.go
tanishaAtDatazip: done

protocol/sync.go
tanishaAtDatazip: got it

protocol/sync.go
tanishaAtDatazip: my thought for this was whenever someone starts a full refresh, he should pass clear destination flag, wont that be enough?

destination/writers.go
tanishaAtDatazip: got it, removed

protocol/sync.go
tanishaAtDatazip: okay

destination/writers.go
tanishaAtDatazip: moved

protocol/sync.go
hash-data: no that is not true

tanishaAtDatazip: what i am understanfing is 
1. if the clear dest flag is set, ALL selected streams will delete ONLY(no standard mode streams deletion here)
2. if dest flag is not set, delete the standard mode streams
is that true?

hash-data: Yeah, it does not matter if the clear dest flag is passed or not, full refresh sync mode tables are always going to delete. 
One thing standard mode streams can also have streams that have incremental which we don't have to delete.

tanishaAtDatazip: got it

destination/parquet/parquet.go
hash-data: but we can use the recursive thing (to not maintain the current logic), i will research more about it,

destination/parquet/parquet.go
```diff
@@ -412,13 +412,26 @@ func (p *Parquet) clearLocalFiles(selectedStreams []string) error {
...
 	return nil
 }
 
-func (p *Parquet) clearS3Files(selectedStreams []string) error {
+func (p *Parquet) clearS3Files(ctx context.Context, selectedStreams []string) error {
+	deleteS3PrefixStandard := func(prefix string) error {
+		iter := s3manager.NewDeleteListIterator(p.s3Client, &s3.ListObjectsInput{
+			Bucket: aws.String(p.config.Bucket),
+			Prefix: aws.String(prefix + "/"),
+		})
```
ImDoubD-datazip, line 429: need to provide s3 path here if the bucket and path both provided in source

destination/interface.go
```diff
@@ -35,5 +35,8 @@ type Writer interface {
 	// EvolveSchema updates the schema based on changes.
 	// Need to pass olakeTimestamp as end argument to get the correct partition path based on record ingestion time.
 	EvolveSchema(bool, bool, map[string]*types.Property, types.Record, time.Time) error
+	// DropStreams is used to clear the destination before re-writing the stream
+	DropStreams(ctx context.Context, selectedStream []string) error
+	// Clear(ctx context.Context, stream types.StreamInterface, streamMetadata map[string]types.StreamMetadata) error
```
hash-data:
```suggestion
```

PR #395, opened 2025-07-11
feat: incremental for oracle connector

vaibhav-datazip:
# Description

- Extends the functionality of Incremental sync to oracle connector.
- Currently shows and handles the cursor field as it was mentioned in the database be it uppercase, lowercase or mixed case.

## Type of change

# How Has This Been Tested?

- [x] Ran Sync on table with various datatypes
     - Integer type datatype
     - Float/ Number with scale > 0
     - String type

### DIFF
--- a/drivers/mysql/internal/mysql.go
+++ b/drivers/mysql/internal/mysql.go
@@ -151,6 +151,7 @@ func (m *MySQL) ProduceSchema(ctx context.Context, streamName string) (*types.St
 			if err := rows.Scan(&columnName, &columnType, &dataType, &isNullable, &columnKey); err != nil {
 				return nil, fmt.Errorf("failed to scan column: %s", err)
 			}
+			stream.WithCursorField(columnName)
 			datatype := types.Unknown
 
 			if val, found := mysqlTypeToDataTypes[dataType]; found {
--- a/drivers/mysql/internal/mysql.go
+++ b/drivers/mysql/internal/mysql.go
@@ -151,7 +151,9 @@ func (m *MySQL) ProduceSchema(ctx context.Context, streamName string) (*types.St
 			if err := rows.Scan(&columnName, &columnType, &dataType, &isNullable, &columnKey); err != nil {
 				return nil, fmt.Errorf("failed to scan column: %s", err)
 			}
-			stream.WithCursorField(columnName)
+			if strings.EqualFold("no", isNullable) {
+				stream.WithCursorField(columnName)
+			}
 			datatype := types.Unknown
 
 			if val, found := mysqlTypeToDataTypes[dataType]; found {
--- a/drivers/oracle/internal/incremental.go
+++ b/drivers/oracle/internal/incremental.go
@@ -32,9 +32,9 @@ func (o *Oracle) StreamIncrementalChanges(ctx context.Context, stream types.Stre
 	incrementalCondition := ""
 	if isTimestamp {
 		parsedTime := fmt.Sprintf("TO_TIMESTAMP_TZ('%s','YYYY-MM-DD\"T\"HH24:MI:SS.FF9\"Z\"')", lastCursorValue.(time.Time).UTC().Format(time.RFC3339Nano))
-		incrementalCondition = fmt.Sprintf("%q > %s", cursorField, parsedTime)
+		incrementalCondition = fmt.Sprintf("%q >= %s", cursorField, parsedTime)
 	} else {
-		incrementalCondition = fmt.Sprintf("%q > '%v'", cursorField, lastCursorValue)
+		incrementalCondition = fmt.Sprintf("%q >= '%v'", cursorField, lastCursorValue)
 	}
 
 	filter = utils.Ternary(filter != "", fmt.Sprintf("%s AND %s", filter, incrementalCondition), incrementalCondition).(string)
--- a/drivers/oracle/internal/incremental.go
+++ b/drivers/oracle/internal/incremental.go
@@ -3,6 +3,7 @@ package driver
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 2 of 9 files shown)


drivers/oracle/internal/incremental.go
```diff
@@ -5,10 +5,48 @@ import (
...
+	filter, err := jdbc.SQLFilter(stream, o.Type())
+	if err != nil {
+		return fmt.Errorf("failed to create sql filter during incremental sync: %s", err)
+	}
+	incrementalCondition := fmt.Sprintf("%q > '%v'", cursorField, lastCursorValue)
+	if filter != "" {
+		filter = fmt.Sprintf("%s AND %s", filter, incrementalCondition)
+	} else {
+		filter = incrementalCondition
+	}
```
vishalm0509: Can we use utils.Ternary here?

`filter = utils.Ternary(filter != "", fmt.Sprintf("%s AND %s", filter, incrementalCondition), incrementalCondition)`

drivers/oracle/internal/incremental.go
```diff
@@ -5,10 +5,45 @@ import (
...
+// StreamIncrementalChanges implements incremental sync for Oracle
+func (o *Oracle) StreamIncrementalChanges(ctx context.Context, stream types.StreamInterface, processFn abstract.BackfillMsgFn) error {
+	cursorField := stream.Cursor()
+	lastCursorValue := o.state.GetCursor(stream.Self(), cursorField)
+
+	filter, err := jdbc.SQLFilter(stream, o.Type())
+	if err != nil {
+		return fmt.Errorf("failed to create sql filter during incremental sync: %s", err)
+	}
+	incrementalCondition := fmt.Sprintf("%q > '%v'", cursorField, lastCursorValue)
```
tanishaAtDatazip, line 23: `fmt.Sprintf("%q > '%v'", cursorField, lastCursorValue)`
this directly embeds lastCursorValue into the query string. if the value contains quotes or SQL keywords, it can break the query and lead to SQL injection.
f.e. injection vector: `lastCursorValue = "2024-01-01' OR 1=1 --"`
instead use a parameterized query
```
query := "SELECT * FROM ... WHERE created_at > ?"
rows, err := db.QueryContext(ctx, query, lastCursorValue)
```

drivers/mysql/internal/mysql.go
```diff
@@ -151,6 +151,7 @@ func (m *MySQL) ProduceSchema(ctx context.Context, streamName string) (*types.St
 			if err := rows.Scan(&columnName, &columnType, &dataType, &isNullable, &columnKey); err != nil {
 				return nil, fmt.Errorf("failed to scan column: %s", err)
 			}
+			stream.WithCursorField(columnName)
```
hash-data: only add if it is not nullable

drivers/postgres/internal/postgres.go
```diff
@@ -169,6 +169,7 @@ func (p *Postgres) ProduceSchema(ctx context.Context, streamName string) (*types
 		}
 
 		for _, column := range columnSchemaOutput {
+			stream.WithCursorField(column.Name)
```
hash-data: here as well

drivers/oracle/internal/incremental.go
```diff
@@ -5,10 +5,45 @@ import (
...
+	filter = utils.Ternary(filter != "", fmt.Sprintf("%s AND %s", filter, incrementalCondition), incrementalCondition).(string)
+
+	query := fmt.Sprintf("SELECT * FROM %q.%q WHERE %s ORDER BY %q",
+		stream.Namespace(), stream.Name(), filter, cursorField)
```
hash-data: we should also create filter for incremental condition and pass it.

drivers/oracle/internal/incremental.go
```diff
@@ -5,10 +5,45 @@ import (
...
+
+	logger.Infof("Starting incremental sync for stream[%s] with filter: %s", stream.ID(), filter)
+
+	rows, err := o.client.QueryContext(ctx, query)
+	if err != nil {
+		return fmt.Errorf("failed to execute incremental query: %w", err)
```
hash-data: use %s instead of %w

drivers/oracle/internal/oracle.go
```diff
@@ -136,7 +136,7 @@ func (o *Oracle) ProduceSchema(ctx context.Context, streamName string) (*types.S
 		if err := rows.Scan(&columnName, &dataType, &isNullable, &dataPrecision, &dataScale); err != nil {
 			return nil, fmt.Errorf("failed to scan column: %s", err)
 		}
-
+		stream.WithCursorField(columnName)
```
hash-data: here as well add null check

types/type_schema.go
```diff
@@ -86,7 +87,7 @@ func (t *TypeSchema) UnmarshalJSON(data []byte) error {
 }
 
 func (t *TypeSchema) GetType(column string) (DataType, error) {
-	p, found := t.Properties.Load(column)
+	p, found := t.Properties.Load(strings.ToLower(column))
```
hash-data: dont do it here pass from abstract after making lower

drivers/mysql/internal/mysql.go
```diff
@@ -151,6 +151,9 @@ func (m *MySQL) ProduceSchema(ctx context.Context, streamName string) (*types.St
 			if err := rows.Scan(&columnName, &columnType, &dataType, &isNullable, &columnKey); err != nil {
 				return nil, fmt.Errorf("failed to scan column: %s", err)
 			}
+			if strings.EqualFold("no", isNullable) {
```
hash-data, line 154: are you sure here it will come `no` ?

hash-data: as well we can remove it now not required

drivers/oracle/internal/incremental.go
```diff
@@ -3,12 +3,63 @@ package driver
...
+	filter, err := jdbc.SQLFilter(stream, o.Type())
+	if err != nil {
+		return fmt.Errorf("failed to create sql filter during incremental sync: %s", err)
+	}
+
+	datatype, err := stream.Self().Stream.Schema.GetType(strings.ToLower(cursorField))
+	if err != nil {
+		return fmt.Errorf("cursor field %s not found in schema: %s", cursorField, err)
+	}
+	isTimestamp := strings.Contains(string(datatype), "timestamp")
```
hash-data, line 30: can be timestamp mili micro as well

drivers/oracle/internal/incremental.go
```diff
@@ -3,12 +3,63 @@ package driver
...
+
+	incrementalCondition := ""
+	if isTimestamp {
```
hash-data, line 34: add comment why we are doing it

drivers/postgres/internal/postgres.go
```diff
@@ -169,6 +169,9 @@ func (p *Postgres) ProduceSchema(ctx context.Context, streamName string) (*types
 		}
 
 		for _, column := range columnSchemaOutput {
+			if strings.EqualFold("no", *column.IsNullable) {
```
hash-data, line 172: can be removed from all drivers and check for only multiple types

drivers/oracle/internal/incremental.go
```diff
@@ -3,12 +3,65 @@ package driver
...
+	if err != nil {
+		return fmt.Errorf("cursor field %s not found in schema: %s", cursorField, err)
+	}
+	isTimestamp := strings.Contains(string(datatype), "timestamp")
+
+	incrementalCondition := ""
+	
+	// Convert Go time format to Oracle TO_TIMESTAMP_TZ format with timezone support
+	if isTimestamp {
+		parsedTime := fmt.Sprintf("TO_TIMESTAMP_TZ('%s','YYYY-MM-DD\"T\"HH24:MI:SS.FF9\"Z\"')", lastCursorValue.(time.Time).UTC().Format(time.RFC3339Nano))
```
tanishaAtDatazip: lastCursorValue value can be of string type, it can then fail?

hash-data: @vaibhav-datazip  can you use reformatValue function to reformat it in time and use oracle function in query?

drivers/oracle/internal/incremental.go
```diff
@@ -3,12 +3,72 @@ package driver
...
+		if cursorValueType.String() == "string" {
+			parsedTime = fmt.Sprintf("TO_TIMESTAMP_TZ('%s','YYYY-MM-DD\"T\"HH24:MI:SS.FF9\"Z\"')", lastCursorValue)
+		} else {
+			parsedTime = fmt.Sprintf("TO_TIMESTAMP_TZ('%s','YYYY-MM-DD\"T\"HH24:MI:SS.FF9\"Z\"')", lastCursorValue.(time.Time).UTC().Format(time.RFC3339Nano))
+		}
+		incrementalCondition = fmt.Sprintf("%q >= %s", cursorField, parsedTime)
+	} else {
+		incrementalCondition = fmt.Sprintf("%q >= '%v'", cursorField, lastCursorValue)
+	}
+
```
hash-data: would recomend to use reformatValue function and then use TO_TIMESTAMP

tanishaAtDatazip (approved):

PR #399, opened 2025-07-15
feat: incremental mysql

tanishaAtDatazip:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


drivers/mysql/internal/incremental.go
```diff
@@ -5,10 +5,31 @@ import (
...
 )
 
-// IncrementalChanges is not supported for MySQL
-func (m *MySQL) StreamIncrementalChanges(ctx context.Context, stream types.StreamInterface, cb abstract.BackfillMsgFn) error {
-	return fmt.Errorf("incremental sync is not supported for MySQL driver")
+func (m *MySQL) StreamIncrementalChanges(ctx context.Context, stream types.StreamInterface, processFn abstract.BackfillMsgFn) error {
+	cursorField := stream.Cursor()
+	lastCursorValue := m.state.GetCursor(stream.Self(), cursorField)
+	query := fmt.Sprintf(`SELECT * FROM %s WHERE %s >= ? ORDER BY %s`, stream.Name(), cursorField, cursorField)
+
```
vikaxsh, line 28: use `query := fmt.Sprintf(`SELECT * FROM %s.%s WHERE %s >= ? ORDER BY %s`, stream.Namespace(), stream.Name(), cursorField, cursorField)`

vikaxsh, line 28: can we move this to jdbc.go file?

vikaxsh, line 28: can we add a info log here , "Starting incremental sync for stream ......"

tanishaAtDatazip: yes, but the query is mysql specific, simple and isn't reusable, so i guess it makes more sense here?

vikaxsh: if you see all the queries are in that file, for every driver

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,75 @@ package driver
...
+	lastPrimaryCursorValue := m.state.GetCursor(stream.Self(), primaryCursor)
+	lastSecondaryCursorValue := m.state.GetCursor(stream.Self(), secondaryCursor)
+
+	// Get filter string (fully evaluated with values already inserted)
+	staticFilter, err := jdbc.SQLFilter(stream, m.Type())
+	if err != nil {
+		return fmt.Errorf("failed to parse filter during chunk iteration: %s", err)
+	}
+
+	incrementalCondition, err := m.buildIncrementalCondition(primaryCursor, secondaryCursor, lastPrimaryCursorValue)
```
vaibhav-datazip: Here the third argument is lastPrimaryCursorValue. Shouldn't it be lastSecondaryCursorValue according to the function's argument name and logic used

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,75 @@ package driver
...
+	if err != nil {
+		return fmt.Errorf("failed to format cursor condition: %s", err)
+	}
+
+	combinedFilter := incrementalCondition
```
vaibhav-datazip: instead of again defining combined filter the value of incrementalCondition, won't it be better to define incremental condition as combinedFilter like it is in other drivers

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,75 @@ package driver
...
+	if staticFilter != "" {
+		combinedFilter = fmt.Sprintf("(%s) AND (%s)", staticFilter, incrementalCondition)
+	}
```
vaibhav-datazip: utils.ternary can be used here

utils/logger/logger.go
```diff
@@ -98,6 +98,26 @@ func LogRequest(req *http.Request) {
...
+			replacement = "NULL"
+		case string:
+			replacement = fmt.Sprintf("'%s'", strings.ReplaceAll(v, "'", "''"))
+		case time.Time:
+			replacement = fmt.Sprintf("'%s'", v.UTC().Format("2006-01-02 15:04:05.000"))
+		default:
+			replacement = fmt.Sprintf("'%v'", v)
+		}
+		// Replace only the first occurrence of "?"
+		query = strings.Replace(query, "?", replacement, 1)
```
vaibhav-datazip: But this will show incorrect query in the logs

Image

the values saved in state were

Image

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,75 @@ package driver
...
+
+// buildIncrementalCondition generates the incremental condition SQL for MySQL
+func (m *MySQL) buildIncrementalCondition(primaryCursorField string, secondaryCursorField string, lastSecondaryCursorValue any) (string, error) {
+	primaryCondition := fmt.Sprintf("`%s` >= ?", primaryCursorField)
+
+	if secondaryCursorField != "" && lastSecondaryCursorValue != nil {
+		primaryCondition = fmt.Sprintf(" %s OR (`%s` IS NULL AND `%s` >= ?)",
+			primaryCondition, primaryCursorField, secondaryCursorField)
+	}
+	return primaryCondition, nil
```
vaibhav-datazip: we can populate args here and remove if-else block from StreamIncrementalChanges

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,75 @@ package driver
...
 )
 
-// IncrementalChanges is not supported for MySQL
-func (m *MySQL) StreamIncrementalChanges(ctx context.Context, stream types.StreamInterface, cb abstract.BackfillMsgFn) error {
-	return fmt.Errorf("incremental sync is not supported for MySQL driver")
+func (m *MySQL) StreamIncrementalChanges(ctx context.Context, stream types.StreamInterface, processFn abstract.BackfillMsgFn) error {
+	primaryCursor, secondaryCursor := stream.Cursor()
+	lastPrimaryCursorValue := m.state.GetCursor(stream.Self(), primaryCursor)
+	lastSecondaryCursorValue := m.state.GetCursor(stream.Self(), secondaryCursor)
+
```
vaibhav-datazip: A warning log can be added here in case when lastPrimaryCursorValue is null . this warns the user that all the values in the cursor field columns are null.

vaibhav-datazip:
multi cursor incremental sync failing for timestamp values
Screenshot 2025-07-29 at 6 07 53 PM

utils/logger/logger.go
tanishaAtDatazip: right, updated

utils/logger/logger.go
```diff
@@ -98,6 +98,26 @@ func LogRequest(req *http.Request) {
...
+	}
+	return query
+}
+
 // CreateFile creates a new file or overwrites an existing one with the specified filename, path, extension,
 func FileLogger(content any, fileName, fileExtension string) error {
 	// get config folder
```
ImDoubD-datazip: image


how to replicate:
-> Sync without state with timestamp field
-> add a second cursor field, with the source also having an entry where the value for first cursor field is null
-> run sync with state

Then this error must pop.

tanishaAtDatazip: this will be resolved with https://github.com/datazip-inc/olake/pull/418 , you can test by having the changes of that branch in this and let me know if it persist

ImDoubD-datazip: sync without state when cursor field (type timestamp) has null, this will also be solved with #418 , but do check if there any other cases that ill be affected with that.
@tanishaAtDatazip

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,77 @@ package driver
...
+	//TODO:
+	// 1. we need to ensure that state never has nil values for any cursor in all cases
+	if lastPrimaryCursorValue == nil {
+		logger.Warnf("last primary cursor value is nil for stream[%s]", stream.ID())
+	} else if lastSecondaryCursorValue == nil {
+		logger.Warnf("last secondary cursor value is nil for stream[%s]", stream.ID())
+	}
```
vaibhav-datazip: The `else if lastSecondaryCursorValue == nil` condition can trigger a misleading warning even when the user hasn't specified a secondary cursor field. Wouldn't it be better to use separate if blocks for each case, like this:

```
if lastPrimaryCursorValue == nil {
		logger.Warnf("Stored primary cursor value is nil for the stream [%s]", stream.ID())
}
if secondaryCursor != "" && lastSecondaryCursorValue == nil {
		logger.Warnf("Stored secondary cursor value is nil for the stream [%s]", stream.ID())
}
```

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,78 @@ package driver
...
+	lastSecondaryCursorValue := m.state.GetCursor(stream.Self(), secondaryCursor)
+
+	//TODO:
+	// 1. we need to ensure that state never has nil values for any cursor in all cases
+	if lastPrimaryCursorValue == nil {
+		logger.Warnf("last primary cursor value is nil for stream[%s]", stream.ID())
+	}
+	if secondaryCursor != "" &&  lastSecondaryCursorValue == nil {
+		logger.Warnf("last secondary cursor value is nil for stream[%s]", stream.ID())
+	}
```
ImDoubD-datazip: we can remove this, add the same in buildIncrementalCondition. that way we won't need to pass these parameters there, and all checks regarding cursor fields will be done in buildIncrementalCondition()

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,78 @@ package driver
...
+	filter = utils.Ternary(filter != "", fmt.Sprintf("(%s) AND (%s)", filter, incrementalCondition), incrementalCondition).(string)
+	query := fmt.Sprintf("SELECT * FROM `%s`.`%s` WHERE %s", stream.Namespace(), stream.Name(), filter)
+
+	var rows *sql.Rows
+	if secondaryCursor != "" && lastSecondaryCursorValue != nil {
+		rows, err = m.client.QueryContext(ctx, query, lastPrimaryCursorValue, lastSecondaryCursorValue)
+	} else {
+		rows, err = m.client.QueryContext(ctx, query, lastPrimaryCursorValue)
+	}
```
ImDoubD-datazip: i believe we can remove this. You are return queryArgs in buildIncrementalCondition() and doing a check in the same function for secondary cursor. so check in this code block is not needed. we can directly using queryContxt with queryArgs as parameters without any check.

drivers/mysql/internal/mysql.go
```diff
@@ -143,7 +143,7 @@ func (m *MySQL) ProduceSchema(ctx context.Context, streamName string) (*types.St
 			return nil, fmt.Errorf("invalid stream name format: %s", streamName)
 		}
 		schemaName, tableName := parts[0], parts[1]
-		stream := types.NewStream(tableName, schemaName)
+		stream := types.NewStream(tableName, schemaName).WithSyncMode(types.INCREMENTAL)
```
ImDoubD-datazip, line 146: remove this WithSyncMode(types.INCREMENTAL) from here. it will be done abstract/incremental, this is taken care in postgres incremental pr

drivers/mysql/internal/mysql.go
```diff
@@ -179,6 +177,13 @@ func (m *MySQL) ProduceSchema(ctx context.Context, streamName string) (*types.St
 	if err != nil && ctx.Err() == nil {
 		return nil, fmt.Errorf("failed to process table[%s]: %s", streamName, err)
 	}
+
+	stream.Schema.Properties.Range(func(key, value interface{}) bool {
```
ImDoubD-datazip: if possible can you explain why this is required?

utils/logger/logger.go
```diff
@@ -98,6 +98,26 @@ func LogRequest(req *http.Request) {
 	fmt.Println(string(requestDump))
 }
 
+// InterpolateQueryPlaceholders replaces the placeholders in the query with the actual values for logging
+func InterpolateQueryPlaceholders(query string, args []any) string {
```
ImDoubD-datazip: can we use reformat.go (like any existing function) instead of declaring this in logger.go. if it is needed, then we can declare it in reformat.go

drivers/mysql/internal/mysql.go
vaibhav-datazip: mysql will be merged prior to the postgres, it be here, will make these changes after merging with postgres branch.

drivers/mysql/internal/mysql.go
tanishaAtDatazip: to add in available cursor fields in streams

drivers/mysql/internal/mysql.go
ImDoubD-datazip: ok then no need to do this in this PR

tanishaAtDatazip: he can remove there itself, it will fail the integration tests if i remove here right and in case this PR gets merged before postgres then supported sync modes will be incorrect

drivers/mysql/internal/mysql.go
ImDoubD-datazip: without this as well, I am able to see the fields in available cursor fields

utils/logger/logger.go
tanishaAtDatazip: here it is just used for injecting filters in log query and isnt used for any core logic refactoring, so it makes sense to add in logger

vaibhav-datazip:
Readme update is required for mysql incremental

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,73 @@ package driver
...
+
+	//TODO:
+	// 1. we need to ensure that state never has nil values for any cursor in all cases
+	if lastPrimaryCursorValue == nil {
+		logger.Warnf("last primary cursor value is nil for stream[%s]", stream.ID())
+	}
+	if secondaryCursor != "" && lastSecondaryCursorValue == nil {
+		logger.Warnf("last secondary cursor value is nil for stream[%s]", stream.ID())
+	}
+	primaryCondition := fmt.Sprintf("`%s` >= ?", primaryCursor)
```
ImDoubD-datazip: remove this ? from here, we can use %v here, as we don't want any extra interpolate function for showing filter in logger

drivers/mysql/internal/testdata/test_streams.json
```diff
@@ -1 +1,261 @@
...
\ No newline at end of file
+{
```
ImDoubD-datazip: remove formatting

drivers/mysql/internal/incremental.go
tanishaAtDatazip: done with other way as discussed with @hash-data

drivers/mysql/README.md
```diff
@@ -115,6 +117,26 @@ Before running the Sync command, the generated `streams.json` file must be confi
             ]
          },
       ```
+      - Add `cursor_field` from set of `available_cursor_fields` in case of incremental sync. This column will be used to track which rows from the table must be synced. If the primary cursor field is expected to contain `null` values, a fallback cursor field can be specified after the primary cursor field using a colon separator. The system will use the fallback cursor when the primary cursor is `null`.
+   > **Note**: For incremental sync to work correctly, the primary cursor field (and fallback cursor field if defined) must contain at least one non-null value. Defined cursor fields cannot be entirely null.
+      ```json
+         "sync_mode": "incremental",
+         "cursor_field": "UPDATED_AT:CREATED_AT" // UPDATED_AT is the primary cursor field, CREATED_AT is the fallback cursor field (which can be skipped if the primary cursor field does not contain null values)
+      ```
```
vaibhav-datazip: The indentation is making it appear as a sub-point under the previous point, this can be corrected.
you can use online md viewer for preview

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,73 @@ package driver
...
+	}
+
+	incrementalCondition, queryArgs, err := m.buildIncrementalCondition(stream)
+	if err != nil {
+		return fmt.Errorf("failed to format cursor condition: %s", err)
+	}
+	filter = utils.Ternary(filter != "", fmt.Sprintf("(%s) AND (%s)", filter, incrementalCondition), incrementalCondition).(string)
+	query := fmt.Sprintf("SELECT * FROM `%s`.`%s` WHERE %s", stream.Namespace(), stream.Name(), filter)
+
+	logger.Infof("Starting incremental sync for stream[%s] with filter: %s and args: %s", stream.ID(), query, queryArgs)
```
vaibhav-datazip: %v should be used for query args to make the log clear

Image

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,73 @@ package driver
...
+	if err != nil {
+		return fmt.Errorf("failed to parse filter during chunk iteration: %s", err)
+	}
+
+	incrementalCondition, queryArgs, err := m.buildIncrementalCondition(stream)
+	if err != nil {
+		return fmt.Errorf("failed to format cursor condition: %s", err)
+	}
+	filter = utils.Ternary(filter != "", fmt.Sprintf("(%s) AND (%s)", filter, incrementalCondition), incrementalCondition).(string)
+	query := fmt.Sprintf("SELECT * FROM `%s`.`%s` WHERE %s", stream.Namespace(), stream.Name(), filter)
```
hash-data: can we move it to jdbc queries file? so that it will be easy for us to common out queries.

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,73 @@ package driver
...
+}
+
+// buildIncrementalCondition generates the incremental condition SQL for MySQL
+func (m *MySQL) buildIncrementalCondition(stream types.StreamInterface) (string, []any, error) {
+	primaryCursor, secondaryCursor := stream.Cursor()
+	lastPrimaryCursorValue := m.state.GetCursor(stream.Self(), primaryCursor)
+	lastSecondaryCursorValue := m.state.GetCursor(stream.Self(), secondaryCursor)
+
+	//TODO:
+	// 1. we need to ensure that state never has nil values for any cursor in all cases
```
hash-data:
```suggestion
	// TODO: should we fail here or just warning is ok, (to aware user in start that incremental not make sense)
```

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,73 @@ package driver
...
+		logger.Warnf("last primary cursor value is nil for stream[%s]", stream.ID())
+	}
+	if secondaryCursor != "" && lastSecondaryCursorValue == nil {
+		logger.Warnf("last secondary cursor value is nil for stream[%s]", stream.ID())
+	}
+	primaryCondition := fmt.Sprintf("`%s` >= ?", primaryCursor)
+	queryArgs := []any{lastPrimaryCursorValue}
+	if secondaryCursor != "" && lastSecondaryCursorValue != nil {
+		queryArgs = []any{lastPrimaryCursorValue, lastSecondaryCursorValue}
+		primaryCondition = fmt.Sprintf(" %s OR (`%s` IS NULL AND `%s` >= ?)",
```
hash-data: can we common out this kind of logic as it seems to common in all the jdbc drivers ?

hash-data: or can be done in incremental of postgres pr

hash-data: add a todo so that we not forget

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,73 @@ package driver
...
-// IncrementalChanges is not supported for MySQL
-func (m *MySQL) StreamIncrementalChanges(ctx context.Context, stream types.StreamInterface, cb abstract.BackfillMsgFn) error {
-	return fmt.Errorf("incremental sync is not supported for MySQL driver")
+func (m *MySQL) StreamIncrementalChanges(ctx context.Context, stream types.StreamInterface, processFn abstract.BackfillMsgFn) error {
+	filter, err := jdbc.SQLFilter(stream, m.Type())
+	if err != nil {
+		return fmt.Errorf("failed to parse filter during chunk iteration: %s", err)
+	}
+
+	incrementalCondition, queryArgs, err := m.buildIncrementalCondition(stream)
```
hash-data, line 20: have we tested on timestamp, int, float, string?

tanishaAtDatazip: yes

drivers/mysql/internal/incremental.go
tanishaAtDatazip: moved

drivers/mysql/internal/incremental.go
tanishaAtDatazip: added

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,74 @@ package driver
...
+	//TODO:
+	// 1. we need to ensure that state never has nil values for any cursor in all cases
+	// should we fail here or just warning is ok, (to aware user in start that incremental not make sense)  
+	if lastPrimaryCursorValue == nil {
+		logger.Warnf("last primary cursor value is nil for stream[%s]", stream.ID())
+	}
+	if secondaryCursor != "" && lastSecondaryCursorValue == nil {
+		logger.Warnf("last secondary cursor value is nil for stream[%s]", stream.ID())
+	}
+	// TODO: common out incremental condition for all drivers
```
hash-data:
```suggestion
	// TODO: common out incremental condition for all jdbc supported drivers
```

drivers/mysql/internal/incremental.go
```diff
@@ -2,13 +2,74 @@ package driver
...
+	primaryCondition := fmt.Sprintf("`%s` >= ?", primaryCursor)
```
hash-data:
```suggestion
	condition := fmt.Sprintf("`%s` >= ?", primaryCursor)
```

ImDoubD-datazip (approved):

PR #406, opened 2025-07-21
test: performance test for all drivers

vishalm0509:
# Description

Add performance test for all driver which run on push to `staging` branch.

## Type of change

# How Has This Been Tested?

# Screenshots or Recordings


.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,125 @@
...
+    paths:
+    - '**/*.go'
+    - '**/*.java'
+  pull_request:
+    branches:
+    - "*"
+    paths:
+    - '**/*.go'
+    - '**/*.java'
+  workflow_dispatch:
```
ImDoubD-datazip: what is the need of workflow dispatch here? Can we remove this if not needed?

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,125 @@
...
+        echo '${{ secrets.MONGODB_BENCHMARK_JSON }}' > ./drivers/mongodb/internal/testconfig/benchmark.json
+        echo '${{ secrets.POSTGRES_BENCHMARK_CDC_JSON }}' > ./drivers/postgres/internal/testconfig/benchmark_cdc.json
+        echo '${{ secrets.MYSQL_BENCHMARK_CDC_JSON }}' > ./drivers/mysql/internal/testconfig/benchmark_cdc.json
+        echo '${{ secrets.ORACLE_BENCHMARK_CDC_JSON }}' > ./drivers/oracle/internal/testconfig/benchmark_cdc.json
+        echo '${{ secrets.MONGODB_BENCHMARK_CDC_JSON }}' > ./drivers/mongodb/internal/testconfig/benchmark_cdc.json
+
+    - name: Run Performance Tests
+      run: |
+        go test -v -p 2 ./drivers/mysql/internal/... ./drivers/mongodb/internal/... -timeout 0 -run 'Performance'
+        go test -v -p 2 ./drivers/oracle/internal/... ./drivers/postgres/internal/... -timeout 0 -run 'Performance'
```
ImDoubD-datazip: This is okay for now, but we need to somehow parallelise for all driver in near future. Have to explore more on this. Add a todo here.

drivers/mongodb/internal/performance_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"testing"
+
+	"github.com/datazip-inc/olake/utils"
+	"github.com/datazip-inc/olake/utils/testutils"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+const (
+	database       = "mongodb"
```
ImDoubD-datazip: this we have in constant.go, check once

drivers/mongodb/internal/performance_test.go
```diff
@@ -0,0 +1,73 @@
...
+	backfillStream = "users"
+	cdcStream      = "users_cdc"
+	namespace      = "test"
+)
+
+func TestMongodbPerformance(t *testing.T) {
+	config := testutils.PerformanceTestConfig{
+		TestConfig:      testutils.GetTestConfig("mongodb"),
```
ImDoubD-datazip: here also use constants.go

drivers/mongodb/internal/performance_test.go
```diff
@@ -0,0 +1,73 @@
...
+func triggerMongodbCDC(ctx context.Context, conn interface{}) error {
+	db, ok := conn.(*mongo.Client)
+	if !ok {
+		return fmt.Errorf("invalid connection type")
+	}
+
+	// TODO: Insert data from backfill collection to cdc collection
+	db.Database(database).Collection(cdcStream).InsertOne(ctx, bson.M{"name": "test"})
+	return nil
+}
```
ImDoubD-datazip: another way is that create a single function, according to various operation add switch cases which you want to do. Whether you want to connect db, to close, setup db for cdc, trigger mongo cdc etc. implement this function in the mongo_test_utils table, and comment that this is for performance test.

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,203 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
 	}
 	t.Logf("Verified datatypes in Iceberg after sync")
 }
+
```
ImDoubD-datazip, line 430: add a comment that this is now performance test, also add one for integration test above it.

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,203 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+func IsRPSAboveBenchmark(config TestConfig, isBackfill bool) (bool, error) {
+	benchmarkFile := utils.Ternary(isBackfill, "benchmark.json", "benchmark_cdc.json").(string)
+
+	var stats map[string]interface{}
+	if err := utils.UnmarshalFile(filepath.Join(config.HostRoot, fmt.Sprintf("drivers/%s/internal/testconfig/%s", config.Driver, "stats.json")), &stats, false); err != nil {
+		return false, err
+	}
+
+	getRPSFromStats := func(stats map[string]interface{}) (float64, error) {
+		rps, err := strconv.ParseFloat(strings.Split(stats["Speed"].(string), " ")[0], 64)
```
ImDoubD-datazip: we can use our in-built reformatfloat64 function here

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,203 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+	}
+
+	benchmarkRps, err := getRPSFromStats(benchmarkStats)
+	if err != nil {
+		return false, err
+	}
+
+	fmt.Printf("CurrentRPS: %.2f, BenchmarkRPS: %.2f\n", rps, benchmarkRps)
+
+	if rps < 0.9*benchmarkRps {
```
ImDoubD-datazip: how was this calculated? If possible, can we compare it for same number of records and check the avg rps that is printed in stats.json at the end of sync?

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,203 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+		return false, fmt.Errorf("❌ RPS is less than benchmark RPS")
+	}
+
+	return true, nil
+}
+
+func InstallCmd() string {
```
ImDoubD-datazip: check above, if that particular install cmd can be used. If anything extra needed, add in that.

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,203 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+			if i > 0 {
+				conditions += " or "
+			}
+			conditions += fmt.Sprintf(`.stream_name == "%s"`, stream)
+		}
+
+		jqExpr := fmt.Sprintf(
+			`jq '.selected_streams = { "%s": (.selected_streams["%s"] | map(select(%s) | .normalization = true)) }' %s > /tmp/streams.json && mv /tmp/streams.json %s`,
+			namespace,
+			namespace,
```
ImDoubD-datazip: here namespace is not required, you can directly scan the whole selected streams and turn every normalization true

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,203 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+							updateStreamsCmd := updateStreamsCommand(*config.TestConfig, config.Namespace, config.BackfillStreams...)
+							_, _, err = utils.ExecCommand(ctx, c, updateStreamsCmd)
+							require.NoError(t, err, "Failed to update streams")
+
+							syncCmd := syncCommand(*config.TestConfig, true)
+							_, output, err = utils.ExecCommand(ctx, c, syncCmd)
+							require.NoError(t, err, fmt.Sprintf("Failed to perform sync:\n%s", string(output)))
+							t.Log(string(output))
+
+							success, err := IsRPSAboveBenchmark(*config.TestConfig, true)
```
ImDoubD-datazip: change success to checkRPS

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,203 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+
+								err = config.TriggerCDC(ctx, conn)
+								require.NoError(t, err, "Failed to trigger CDC change")
+
+								syncCmd = syncCommand(*config.TestConfig, false)
+								_, output, err = utils.ExecCommand(ctx, c, syncCmd)
+								require.NoError(t, err, fmt.Sprintf("Failed to perform CDC sync:\n%s", string(output)))
+								t.Log(string(output))
+
+								success, err := IsRPSAboveBenchmark(*config.TestConfig, false)
```
ImDoubD-datazip: same here change success name

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,203 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+			Started:          true,
+		})
+		require.NoError(t, err, "Failed to start container")
+		defer func() {
+			if err := container.Terminate(ctx); err != nil {
+				t.Logf("warning: failed to terminate container: %v", err)
+			}
+		}()
+	})
+}
```
ImDoubD-datazip, line 587: move  IsRPSAboveBenchmark below performance test function.

utils/testutils/test_utils.go
```diff
@@ -32,6 +33,28 @@ type IntegrationTest struct {
 	ExecuteQuery       func(ctx context.Context, t *testing.T, tableName, operation string)
 }
 
+type TestConfig struct {
```
ImDoubD-datazip: add comment these are for performance test struct

utils/testutils/test_utils.go
```diff
@@ -32,6 +33,28 @@ type IntegrationTest struct {
...
+	SourcePath      string
+	CatalogPath     string
+	DestinationPath string
+	StatePath       string
+	StatsPath       string
+}
+
+type PerformanceTestConfig struct {
+	TestConfig      *TestConfig
+	Namespace       string
```
ImDoubD-datazip, line 44: check do we really need namespace? we can remove this from the config, if we are only using it for making normalization true.

drivers/mongodb/internal/performance_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/datazip-inc/olake/utils/testutils"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+const (
+	database       = "mongodb"
+	backfillStream = "users"
+	cdcStream      = "users_cdc"
+	namespace      = "test"
```
ImDoubD-datazip: all required constants we can directly pass in below test performance function

drivers/mysql/go.mod
```diff
@@ -10,9 +10,53 @@ require (
...
+	github.com/power-devops/perfstat v0.0.0-20210106213030-5aafc221ea8c // indirect
 	github.com/rs/zerolog v1.15.0 // indirect
+	github.com/shirou/gopsutil/v4 v4.25.1 // indirect
+	github.com/sirupsen/logrus v1.9.3 // indirect
+	github.com/stretchr/testify v1.10.0 // indirect
+	github.com/testcontainers/testcontainers-go v0.37.0 // indirect
+	github.com/tklauser/go-sysconf v0.3.12 // indirect
+	github.com/tklauser/numcpus v0.6.1 // indirect
+	github.com/yusufpapurcu/wmi v1.2.4 // indirect
 	go.mongodb.org/mongo-driver v1.17.3 // indirect
```
ImDoubD-datazip: why there is mongo-driver in mysql go mod?

drivers/mysql/go.mod
```diff
@@ -10,9 +10,53 @@ require (
...
+	go.opentelemetry.io/auto/sdk v1.1.0 // indirect
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.49.0 // indirect
+	go.opentelemetry.io/otel v1.35.0 // indirect
+	go.opentelemetry.io/otel/metric v1.35.0 // indirect
+	go.opentelemetry.io/otel/trace v1.35.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
```
ImDoubD-datazip: if possible, can you check if all these newly added packages are needed

drivers/mysql/internal/performance_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/datazip-inc/olake/utils"
+	"github.com/datazip-inc/olake/utils/testutils"
+	"github.com/jmoiron/sqlx"
+)
+
+const (
+	backfillStream = "test"
+	cdcStream      = "test_cdc"
+	namespace      = "mysql"
```
ImDoubD-datazip: remove all these constants, whatever is needed directly pass

drivers/mysql/internal/performance_test.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+func TestMySQLPerformance(t *testing.T) {
+	config := testutils.PerformanceTestConfig{
+		TestConfig:      testutils.GetTestConfig("mysql"),
```
ImDoubD-datazip: for drivers use constants.go

drivers/mongodb/internal/performance_test.go
```diff
@@ -0,0 +1,73 @@
...
+func TestMongodbPerformance(t *testing.T) {
+	config := testutils.PerformanceTestConfig{
+		TestConfig:      testutils.GetTestConfig("mongodb"),
+		Namespace:       namespace,
+		BackfillStreams: []string{backfillStream},
+		CDCStreams:      []string{cdcStream},
+		ConnectDB:       connectDatabase,
+		CloseDB:         closeDatabase,
+		SetupCDC:        setupDatabaseForCDC,
+		TriggerCDC:      triggerMongodbCDC,
```
ImDoubD-datazip: check the comment below, if that is implemented then all you have to do is pass that one function here and call that function as per operation in performance test utils.

drivers/mysql/internal/performance_test.go
```diff
@@ -0,0 +1,74 @@
...
+		Namespace:       namespace,
+		BackfillStreams: []string{backfillStream},
+		CDCStreams:      []string{cdcStream},
+		ConnectDB:       connectDatabase,
+		CloseDB:         closeDatabase,
+		SetupCDC:        setupDatabaseForCDC,
+		TriggerCDC:      triggerMySQLCDC,
```
ImDoubD-datazip: check mongodb comment for this db operations

drivers/mysql/internal/performance_test.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func triggerMySQLCDC(ctx context.Context, conn interface{}) error {
+	db, ok := conn.(*sqlx.DB)
+	if !ok {
+		return fmt.Errorf("invalid connection type")
+	}
+	_, err := db.ExecContext(ctx, fmt.Sprintf("INSERT INTO %s SELECT * FROM %s", cdcStream, backfillStream))
+	return err
+}
```
ImDoubD-datazip: check mongodb comments

drivers/oracle/internal/performance_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"testing"
+
+	"github.com/datazip-inc/olake/utils"
+	"github.com/datazip-inc/olake/utils/testutils"
+	"github.com/jmoiron/sqlx"
+)
+
+const (
+	backfillStream = "USERS"
+	namespace      = "ADMIN"
```
ImDoubD-datazip: use directly and for drivers use constants.go

drivers/mongodb/internal/performance_test.go
```diff
@@ -0,0 +1,73 @@
...
+	database       = "mongodb"
+	backfillStream = "users"
+	cdcStream      = "users_cdc"
+	namespace      = "test"
+)
+
+func TestMongodbPerformance(t *testing.T) {
+	config := testutils.PerformanceTestConfig{
+		TestConfig:      testutils.GetTestConfig("mongodb"),
+		Namespace:       namespace,
```
ImDoubD-datazip: check for namespace, not needed

drivers/postgres/internal/performance_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/datazip-inc/olake/utils"
+	"github.com/datazip-inc/olake/utils/testutils"
+	"github.com/jmoiron/sqlx"
+	_ "github.com/lib/pq"
+)
+
+const (
+	backfillStream = "test"
+	cdcStream      = "test_cdc"
+	namespace      = "public"
```
ImDoubD-datazip: remove these, use directly and for driver use constants.go

drivers/postgres/internal/performance_test.go
```diff
@@ -0,0 +1,75 @@
...
+func TestPostgresPerformance(t *testing.T) {
+	config := testutils.PerformanceTestConfig{
+		TestConfig:      testutils.GetTestConfig("postgres"),
+		Namespace:       namespace,
+		BackfillStreams: []string{backfillStream},
+		CDCStreams:      []string{cdcStream},
+		ConnectDB:       connectDatabase,
+		CloseDB:         closeDatabase,
+		SetupCDC:        setupDatabaseForCDC,
+		TriggerCDC:      triggerPostgresCDC,
```
ImDoubD-datazip: follow mongodb comment for db operations

.github/workflows/integration-tests.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         run: mvn clean package -DskipTests
 
       - name: Run Integration Tests
-        run: go test -v -p 2 ./drivers/mysql/internal/... ./drivers/postgres/internal/... -tags=integration --timeout 0
+        run: go test -v -p 2 ./drivers/mysql/internal/... ./drivers/postgres/internal/... -timeout 0 -run 'Integration'
```
ImDoubD-datazip, line 90: Nice work

drivers/mysql/internal/performance_test.go
```diff
@@ -0,0 +1,74 @@
...
+const (
+	backfillStream = "test"
+	cdcStream      = "test_cdc"
+	namespace      = "mysql"
+)
+
+func TestMySQLPerformance(t *testing.T) {
+	config := testutils.PerformanceTestConfig{
+		TestConfig:      testutils.GetTestConfig("mysql"),
+		Namespace:       namespace,
```
ImDoubD-datazip: namespace not needed

drivers/oracle/internal/performance_test.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func TestOraclePerformance(t *testing.T) {
+	config := testutils.PerformanceTestConfig{
+		TestConfig:      testutils.GetTestConfig("oracle"),
+		Namespace:       namespace,
```
ImDoubD-datazip: not needed

drivers/mongodb/internal/performance_test.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+const (
+	database       = "mongodb"
+	backfillStream = "users"
+	cdcStream      = "users_cdc"
+	namespace      = "test"
+)
+
+func TestMongodbPerformance(t *testing.T) {
```
ImDoubD-datazip, line 21: do we really need this performance_test.go file separately, can we add this function in the mongo_test.go file?
Because in the test command, you are providing 'Performance' which as you said identifies the test function having this keyword automatically. So i believe there is no need to make a separate file for this. check once. Then do it for other drivers as well. We have a driver_test.go file for each driver.

ImDoubD-datazip: do we really need this performance_test.go file separately, can we add this function in the mongo_test.go file?
Because in the test command, you are providing 'Performance' which as you said identifies the test function having this keyword automatically. So i believe there is no need to make a separate file for this. check once. Then do it for other drivers as well. We have a driver_test.go file for each driver.

utils/testutils/test_utils.go
vishalm0509: We are allowing a delta of 10% in performance as we might be using relatively smaller machine for performance testing purpose.

utils/testutils/test_utils.go
vishalm0509: We are using a hosted database which might have multiple namespaces in selected streams. So providing the namespace is necessary as we will be removing all the namespaces and streams from the selected streams which are not required for the performance test.

utils/testutils/test_utils.go
vishalm0509: we need it to provide which namespace we need in the selected streams.

.github/workflows/performance-test.yml
vishalm0509: it is used so that we can trigger it manually

vishalm0509:
We need to decide which tables are we going to use for this test across all drivers. @hash-data @tanishaAtDatazip

utils/testutils/test_utils.go
```diff
@@ -327,3 +348,183 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+		DestinationPath: fmt.Sprintf("/test-olake/drivers/%s/internal/testconfig/destination.json", driver),
+		StatePath:       fmt.Sprintf("/test-olake/drivers/%s/internal/testconfig/state.json", driver),
+		StatsPath:       fmt.Sprintf("/test-olake/drivers/%s/internal/testconfig/stats.json", driver),
+	}
+}
+
+func (cfg *PerformanceTest) TestPerformance(t *testing.T) {
+	ctx := context.Background()
+
+	isRPSAboveBenchmark := func(config TestConfig, isBackfill bool) (bool, error) {
```
tanishaAtDatazip, line 435: add a comment: `checks if the current RPS (from stats.json) is at least 90% of the benchmark (backfill or CDC)`

utils/testutils/test_utils.go
```diff
@@ -327,3 +348,183 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+								discoverCmd := discoverCommand(*cfg.TestConfig)
+								_, output, err := utils.ExecCommand(ctx, c, discoverCmd)
+								require.NoError(t, err, fmt.Sprintf("Failed to perform discover:\n%s", string(output)))
+								t.Log(string(output))
+
+								updateStreamsCmd := updateStreamsCommand(*cfg.TestConfig, cfg.Namespace, cfg.CDCStream)
+								_, _, err = utils.ExecCommand(ctx, c, updateStreamsCmd)
+								require.NoError(t, err, "Failed to update streams")
+
+								syncCmd := syncCommand(*cfg.TestConfig, true)
```
tanishaAtDatazip: you're already doing a sync above and again in case it supports CDC, wont it lead to doing backfill twice?

utils/testutils/test_utils.go
```diff
@@ -327,3 +348,183 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+		if err := utils.UnmarshalFile(filepath.Join(config.HostRoot, fmt.Sprintf("drivers/%s/internal/testconfig/%s", config.Driver, benchmarkFile)), &benchmarkStats, false); err != nil {
+			return false, err
+		}
+
+		benchmarkRps, err := getRPSFromStats(benchmarkStats)
+		if err != nil {
+			return false, err
+		}
+
+		fmt.Printf("CurrentRPS: %.2f, BenchmarkRPS: %.2f\n", rps, benchmarkRps)
```
tanishaAtDatazip: use: t.Logf("CurrentRPS: %.2f, BenchmarkRPS: %.2f\n", rps, benchmarkRps)

utils/testutils/test_utils.go
```diff
@@ -327,3 +348,183 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+
+	updateStreamsCommand := func(config TestConfig, namespace string, stream string) string {
+		if len(stream) == 0 {
+			return ""
+		}
+
+		condition := fmt.Sprintf(`.stream_name == "%s"`, stream)
+
+		jqExpr := fmt.Sprintf(
+			`jq '.selected_streams = { "%s": (.selected_streams["%s"] | map(select(%s) | .normalization = true)) }' %s > /tmp/streams.json && mv /tmp/streams.json %s`,
```
tanishaAtDatazip: can we do this way?
tmpCatalog := fmt.Sprintf("/tmp/%s_%s_streams.json", config.Driver, utils.Ternary(isBackfill, "backfill", "cdc"))
jqExpr := fmt.Sprintf(`jq '.selected_streams = { "%s": (.selected_streams["%s"] | map(select(.stream_name == "%s") | .normalization = true)) }' %s > %s && mv %s %s`,
	stream, stream, stream, catalogPath, tmpCatalog, tmpCatalog, catalogPath)

this can avoid clashes by ensuring each test run gets its own temp file (and might be easier to debug) like:
/tmp/postgres_backfill_streams.json
/tmp/mysql_cdc_streams.json

utils/testutils/test_utils.go
tanishaAtDatazip: instead will this sync always read 0 records since we're dropping the collection before cdc_trigger?

vishalm0509: Yes it'll have 0 records and then after trigger we will sync again with the state file so CDC is performed for the newly inserted records

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,119 @@
...
+        echo '${{ secrets[format('{0}_BENCHMARK_JSON', matrix.driver_upper)] }}' > ./drivers/${{ matrix.driver }}/internal/testconfig/benchmark.json
+        echo '${{ secrets[format('{0}_BENCHMARK_CDC_JSON', matrix.driver_upper)] }}' > ./drivers/${{ matrix.driver }}/internal/testconfig/benchmark_cdc.json
+
+    - name: Run Performance Tests
+      run: go test -v ./drivers/${{ matrix.driver }}/internal/... -timeout 0 -run 'Performance'
+
+    - name: Cleanup
+      if: always()
+      run: |
+        aws glue delete-database --name performance_${{ matrix.driver }} || echo "failed to delete ${{ matrix.driver }} database"
```
tanishaAtDatazip: do u think below way confirms deletion failure clearly, checks DB existence for debugging, and exits to avoid silent test success? debugging would be easier

aws glue delete-database --name performance_${{ matrix.driver }} || { echo "Failed to delete Glue DB performance_${{ matrix.driver }}"; aws glue get-database --name performance_${{ matrix.driver }} || true; exit 1; }

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,119 @@
...
+    - name: Build Iceberg Sink
+      working-directory: ./destination/iceberg/olake-iceberg-java-writer
+      run: mvn clean package -DskipTests
+
+    - name: Create testconfig directory
+      run: mkdir -p ./drivers/${{ matrix.driver }}/internal/testconfig
+
+    - name: Create source
+      run: |
+        echo '${{ secrets[format('{0}_SOURCE_JSON', matrix.driver_upper)] }}' | base64 --decode > ./drivers/${{ matrix.driver }}/internal/testconfig/source.json
```
tanishaAtDatazip: what happens if the GitHub secrets are missing or malformed?

vishalm0509: Test fails, as discover and sync cannot be performed with malformed config

.github/workflows/performance-test.yml
vishalm0509: ll adopt the robust error handling for the cleanup step, but since it’s focused on the RPS check and reruns are time consuming, I don’t want it to fail the test. Open for discussion.

drivers/mysql/go.mod
vishalm0509: Already present from staging. I did not modify it. Let's check if it's needed and remove if unnecessary.

.github/workflows/performance-test.yml
tanishaAtDatazip: to be safe we can just log clear err for debugging if any and avoid reruns

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,117 @@
...
+    - "master"
+    paths:
+    - '**/*.go'
+    - '**/*.java'
+  pull_request:
+    branches:
+    - "*"
+    paths:
+    - '**/*.go'
+    - '**/*.java'
```
ImDoubD-datazip, line 10: can you once discuss or check that for when exactly this performance test will be running?
I believe it will only be running when any branch will be merged to staging, right? can we change accordingly then. though for debugging purposes we can keep it as it is for now. but during final merge to staging, we need to change it

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,117 @@
...
+        go-version: '1.23.2'
+
+    - name: Set up Java for Maven
+      uses: actions/setup-java@v3
+      with:
+        distribution: 'temurin'
+        java-version: '17'
+
+    - name: Setup VPN Client
+      if: matrix.driver != 'oracle'
```
ImDoubD-datazip, line 51: why this condition? Won'e we setup the VPN client when driver is Oracle. How will we check for Oracle then?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -160,8 +160,17 @@ private void applyFieldAddition(Table icebergTable, Schema newSchema) {
     // @NOTE avoid committing when there is no schema change. commit creates new
     // commit even when there is no change!
     if (!icebergTable.schema().sameSchema(newSchemaCombined)) {
-      LOGGER.warn("Extending schema of {}", icebergTable.name());
-      us.commit();
+      synchronized (commitLock) {
```
ImDoubD-datazip, line 163: can you add a comment, that this block of code does what

drivers/mongodb/internal/mon_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"testing"
+
+	"github.com/datazip-inc/olake/utils/testutils"
+)
+
+func TestMongodbPerformance(t *testing.T) {
+	config := &testutils.PerformanceTest{
+		TestConfig:          testutils.GetTestConfig("mongodb"),
```
ImDoubD-datazip: here write testutils.GetTestConfig(constants.Mongo), do the same for other drivers as well.

.github/workflows/integration-tests.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         run: mvn clean package -DskipTests
```
ImDoubD-datazip, line 87: as of now, false positive is what performance testing is giving. first need to check and debug this, fix this

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,224 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+
+	syncWithTimeout := func(ctx context.Context, c testcontainers.Container, cmd string) ([]byte, error) {
+		timedCtx, cancel := context.WithTimeout(ctx, 10*time.Minute)
+		defer cancel()
+		_, output, err := utils.ExecCommand(timedCtx, c, cmd)
+		// check if sync was canceled due to timeout (expected)
+		if timedCtx.Err() == context.DeadlineExceeded {
+			return output, nil
+		}
+		if err != nil {
```
ImDoubD-datazip: for every utils.ExecCommand() call, do err and code != 0 check both. It should be &&

.github/workflows/performance-test.yml
vishalm0509: Oracle RDS is not private and can be accessed via any IP. So no need to connect to VPN for oracle db.

.github/workflows/integration-tests.yml
vishalm0509: Fixed it by checking the exit code of the process.

shubham19may (dismissed):

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,112 @@
...
+    - name: Create destination
+      run: |
+        echo '${{ secrets[format('{0}_DESTINATION_JSON', matrix.driver_upper)] }}' | base64 --decode > ./drivers/${{ matrix.driver }}/internal/testdata/destination.json
+
+    - name: Run Performance Tests
+      run: go test -v ./drivers/${{ matrix.driver }}/internal/... -timeout 0 -run 'Performance'
+
+    - name: Cleanup
+      if: always()
+      run: |
```
vaibhav-datazip, line 99: shouldn't we use continue_on_error true here for cleanup task, so that performance test dosent fail id there is some error during this

drivers/mysql/internal/testdata/benchmark.json
```diff
@@ -0,0 +1,4 @@
+{
+    "backfill": 16383.31,
+    "cdc": 548.5
+}
```
vaibhav-datazip: formatting can be removed from all hardcoded testdata json files(similar to what is done in integration test files) using tool like [this](https://jsonformatter.org/96ef77)

.github/workflows/performance-test.yml
vishalm0509: Since we are using `|| true` it fails silently without failing the workflow. If it fails it'll exit with status 0 because of the `|| true`. So I think it shouldn't matter.

drivers/mysql/internal/testdata/benchmark.json
```diff
@@ -0,0 +1 @@
+{"backfill":16383.31,"cdc":548.5}
```
hash-data: add a comment that this benchmark is for github runner which runs on small specifications for exact benchmark and configuration check docs.

drivers/mysql/internal/testdata/state.json
```diff
@@ -0,0 +1 @@
...9999"},{"min":"164999999","max":"169999999"},{"min":"259999999","max":"264999999"},{"min":"329999999","max":"334999999"},{"min":"884999999","max":"889999999"},{"min":"1392332121","max":"1397332121"},{"min":"1776596378","max":"1781596378"},{"min":"224999999","max":"229999999"},{"min":"809999999","max":"814999999"},{"min":"1422332121","max":"1427332121"},{"min":"1442332121","max":"1447332121"},{"min":"19999999","max":"24999999"},{"min":"119999999","max":"124999999"},{"min":"664999999","max":"669999999"},{"min":"939999999","max":"944999999"},{"min":"1182332121","max":"1187332121"},{"min":"1187332121","max":"1192332121"},{"min":"1257332121","max":"1262332121"},{"min":"249999999","max":"254999999"},{"min":"589999999","max":"594999999"},{"min":"639999999","max":"644999999"},{"min":"1247332121","max":"1252332121"},{"min":"1477332121","max":"1482332121"},{"min":"1627332121","max":"1632332121"},{"min":"1667332121","max":"1672332121"},{"min":"474999999","max":"479999999"},{"min":"504999999","max":"509999999"},{"min":"1447332121","max":"1452332121"},{"min":"1822913966","max":null},{"min":"999999","max":"5999999"},{"min":"199999999","max":"204999999"},{"min":"549999999","max":"554999999"}]}}]}
```
hash-data, line 1: If you have removed the chunk manually, that is not a good idea. I would suggest using a backfill filter and creating chunks for small data, not removing it manually. (reason: these are not in sorted order)

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,112 @@
...
+    environment: Performance Testing
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+        - driver: mysql
+          driver_upper: MYSQL
+        - driver: postgres
+          driver_upper: POSTGRES
+        # TODO: add these drivers after benchmarking
```
hash-data:
```suggestion
        # TODO: add benchmark tests for the below databases
```

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,112 @@
...
+        distribution: 'temurin'
+        java-version: '17'
+
+    - name: Setup VPN Client
+      if: matrix.driver != 'oracle'
+      run: |
+        sudo apt-get update -qq && sudo apt-get install -y openvpn iproute2 iputils-ping netcat-openbsd telnet
+        sudo mkdir -p /etc/openvpn/client && sudo chmod 700 /etc/openvpn/client
+        echo "${{ secrets[format('{0}_OPENVPN_CONFIG', matrix.driver_upper)] }}" | base64 --decode | sudo tee /etc/openvpn/client/client.ovpn > /dev/null
+        echo "${{ secrets.OPENVPN_USERNAME }}" | sudo tee /etc/openvpn/client/auth.txt > /dev/null
```
hash-data, line 56: let us check if creds can be printed in bash file or not?

destination/iceberg/iceberg_utils.go
```diff
@@ -30,10 +30,8 @@ func determineMaxBatchSize() int64 {
 	var batchSize int64
 
 	switch {
-	case ramGB <= 8:
-		batchSize = 100 * 1024 * 1024 // 100MB
 	case ramGB <= 16:
-		batchSize = 200 * 1024 * 1024 // 200MB
+		batchSize = 100 * 1024 * 1024 // 100MB
```
hash-data, line 34: these kind of changes are not recommended just for tests. Let us sit and check why it fails for 200.

drivers/mongodb/internal/mon_test.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func TestMongodbPerformance(t *testing.T) {
+	config := &testutils.PerformanceTest{
+		TestConfig:          testutils.GetTestConfig(string(constants.MongoDB)),
+		Namespace:           "test",
+		BackfillStreams:     []string{"users"},
+		CDCStreams:          []string{"users_cdc"},
+		ExecuteQuery:        ExecuteQueryPerformance,
+		SupportsCDC:         true,
```
hash-data: is it required, we can also check len of cdcStreams array

drivers/mongodb/internal/mon_test.go
```diff
@@ -0,0 +1,22 @@
...
+		UsesPreChunkedState: false,
```
hash-data: only mysql support it so we can directly add check in test only

drivers/mongodb/internal/mon_test_util.go
```diff
@@ -0,0 +1,41 @@
...
+	require.NoError(t, cfg.Setup(ctx))
+	db := cfg.client
+	defer func() {
+		require.NoError(t, cfg.Close(ctx))
+	}()
+
+	switch op {
+	case "setup_cdc":
+		// truncate the cdc tables
+		for _, stream := range backfillStreams {
```
hash-data: why cleaning backfillStreams, we have to drop only cdc streams right?

hash-data: else change name to streams

drivers/mysql/go.mod
```diff
@@ -10,9 +10,53 @@ require (
 )
 
 require (
+	dario.cat/mergo v1.0.1 // indirect
```
hash-data: Why are there so many changes in this file? I cant see any requirement

hash-data: remove all changes that are not required

drivers/mysql/internal/mysql_test_util.go
```diff
@@ -252,3 +253,36 @@ var MySQLToIcebergSchema = map[string]string{
 	"long_varchar":           "mediumtext",
 	"name_bool":              "tinyint",
 }
+
+// ExecuteQueryPerformance executes MySQL queries for performance testing based on the operation type
```
hash-data: cant you merge function with executeQuery?, we just need to unmarshal file based on flag

drivers/oracle/internal/oracle_test.go
```diff
@@ -0,0 +1,22 @@
+package driver
```
hash-data: remove file as test not supported yet

drivers/postgres/internal/postgres_test_util.go
```diff
@@ -235,3 +236,45 @@ var PostgresToIcebergSchema = map[string]string{
 	"col_varbit":            "varbit",
 	"col_xml":               "xml",
 }
+
+func ExecuteQueryPerformance(ctx context.Context, t *testing.T, op string, backfillStreams []string) {
```
hash-data: same here merge with existing function whose work is to execute queries

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,242 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+
+		streamConditions := make([]string, len(stream))
+		for i, s := range stream {
+			streamConditions[i] = fmt.Sprintf(`.stream_name == "%s"`, s)
+		}
+		condition := strings.Join(streamConditions, " or ")
+		tmpCatalog := fmt.Sprintf("/tmp/%s_%s_streams.json", config.Driver, utils.Ternary(isBackfill, "backfill", "cdc"))
+		jqExpr := fmt.Sprintf(
+			`jq '.selected_streams = { "%s": (.selected_streams["%s"] | map(select(%s) | .normalization = true)) }' %s > %s && mv %s %s`,
+			namespace,
```
hash-data: create a function and use that in both integration and performance test

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,242 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+	hostRoot := filepath.Join(pwd, "../../..")
+
+	return &TestConfig{
+		Driver:          driver,
+		HostRoot:        hostRoot,
+		SourcePath:      fmt.Sprintf("/test-olake/drivers/%s/internal/testdata/source.json", driver),
+		CatalogPath:     fmt.Sprintf("/test-olake/drivers/%s/internal/testdata/streams.json", driver),
+		DestinationPath: fmt.Sprintf("/test-olake/drivers/%s/internal/testdata/destination.json", driver),
+		StatePath:       fmt.Sprintf("/test-olake/drivers/%s/internal/testdata/state.json", driver),
+		StatsPath:       fmt.Sprintf("/test-olake/drivers/%s/internal/testdata/stats.json", driver),
```
hash-data: just make `test-olake/drivers/%s/internal/testdata/%s` as  variable else use filename and driver?

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,242 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
 	t.Logf("Verified datatypes in Iceberg after sync")
 }
+
+// GetTestConfig returns the test config for the given driver for performance tests
+func GetTestConfig(driver string) *TestConfig {
+	pwd, err := os.Getwd()
+	if err != nil {
+		panic(err)
+	}
+	hostRoot := filepath.Join(pwd, "../../..")
```
hash-data: add a comment what is current pwd and what will be after going three levels up

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,242 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+		StatsPath:       fmt.Sprintf("/test-olake/drivers/%s/internal/testdata/stats.json", driver),
+	}
+}
+
+func (cfg *PerformanceTest) TestPerformance(t *testing.T) {
+	ctx := context.Background()
+
+	// checks if the current rps (from stats.json) is at least 90% of the benchmark rps
+	isRPSAboveBenchmark := func(config TestConfig, isBackfill bool) (bool, error) {
+		getRPSFromStats := func(stats map[string]interface{}, isBenchmark bool) (float64, error) {
```
hash-data: why func inside func?

hash-data: let us make it simple

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,242 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+			rps, err := typeutils.ReformatFloat64(strings.Split(stats["Speed"].(string), " ")[0])
+			if err != nil {
+				return 0, err
+			}
+			return rps.(float64), nil
+		}
+
+		// get current RPS
+		var stats map[string]interface{}
+		if err := utils.UnmarshalFile(filepath.Join(config.HostRoot, fmt.Sprintf("drivers/%s/internal/testdata/%s", config.Driver, "stats.json")), &stats, false); err != nil {
```
hash-data: change hostRoot -> rootPath

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,242 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+			return false, err
+		}
+		rps, err := getRPSFromStats(stats, false)
```
hash-data: why to create a function for it just create a internal struct stats and use same for both benchmark and stats file parsing. 
Also for now we dont need file as well just hard code here as constant on top of file. (according to each driver)

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,242 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+		baseCmd := fmt.Sprintf("/test-olake/build.sh driver-%s sync --config %s --catalog %s --destination %s", config.Driver, config.SourcePath, config.CatalogPath, config.DestinationPath)
+		if !isBackfill {
+			baseCmd = fmt.Sprintf("%s --state %s", baseCmd, config.StatePath)
+		}
+		// use state file for backfill if pre-chunked state is used
+		if isBackfill && usesPreChunkedState {
+			baseCmd = fmt.Sprintf("%s --state %s", baseCmd, config.StatePath)
+		}
+		return baseCmd
+	}
```
hash-data: can we extract out same logic from both integration and performance tests and reuse.

utils/testutils/test_utils.go
```diff
@@ -327,3 +350,242 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+		)
+		return jqExpr
+	}
+
+	syncWithTimeout := func(ctx context.Context, c testcontainers.Container, cmd string) ([]byte, error) {
+		timedCtx, cancel := context.WithTimeout(ctx, 10*time.Minute)
+		defer cancel()
+		code, output, err := utils.ExecCommand(timedCtx, c, cmd)
+		// check if sync was canceled due to timeout (expected)
+		if timedCtx.Err() == context.DeadlineExceeded {
```
hash-data: See integration test, runs both discover and sync. There is a toggling thing that may be required, such as which query to execute. So recommended to reuse the available code

drivers/mysql/internal/testdata/state.json
vishalm0509: I have not removed them manually. I ran the sync and exited the code after the chunk creation process is finished.

destination/iceberg/iceberg_utils.go
vishalm0509: it was informed that this will be taken care by the destination refactor.

drivers/mongodb/internal/mon_test.go
vishalm0509: Removed

drivers/mongodb/internal/mon_test.go
vishalm0509: Made changes

drivers/mongodb/internal/mon_test_util.go
vishalm0509: We are cleaning cdc streams only. In Sprintf I am adding "_cdc" at the end.

```go
err := db.Database("mongodb").Collection(fmt.Sprintf("%s_cdc", stream)).Drop(ctx)
```

Since we will be inserting data from backfillStreams to cdcStreams we need to pass backfillStreams as param

utils/testutils/test_utils.go
vishalm0509: Made changes

utils/testutils/test_utils.go
vishalm0509: Done

utils/testutils/test_utils.go
vishalm0509: Removed it outside

drivers/mysql/internal/testdata/benchmark.json
vishalm0509: We cannot add comments in JSON files

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,112 @@
...
+    - "master"
+    paths:
+    - '**/*.go'
+    - '**/*.java'
+  pull_request:
+    branches:
+    - "*"
+    paths:
+    - '**/*.go'
+    - '**/*.java'
```
hash-data: todo: before merging let us make sure it is configured only to run on staging branch after merge

drivers/mongodb/go.mod
```diff
@@ -83,4 +83,4 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
-replace github.com/datazip-inc/olake => ../../
+replace github.com/datazip-inc/olake => ../../
```
hash-data: can you check if we can remove this line as it seems to be unnecessary

drivers/mongodb/internal/mon_test_util.go
```diff
@@ -0,0 +1,52 @@
...
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+func ExecuteQuery(ctx context.Context, t *testing.T, streams []string, operation string, fileConfig bool) {
+	t.Helper()
+
+	var connStr string
+	if fileConfig {
+		var driver Mongo
```
hash-data:
```suggestion
		var config Mongo
```

drivers/mongodb/internal/mon_test_util.go
```diff
@@ -0,0 +1,52 @@
...
+		utils.UnmarshalFile("./testdata/source.json", &driver.config, false)
+		connStr = fmt.Sprintf(
+			"mongodb+srv://%s:%s@%s/%s",
+			driver.config.Username,
+			driver.config.Password,
+			driver.config.Hosts[0],
```
hash-data: is it single host replica? as insert operation only apply on primary replica

drivers/mysql/go.mod
```diff
@@ -85,4 +85,4 @@ require (
 	golang.org/x/text v0.22.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
-)
+)
```
hash-data: remove file changes

drivers/mysql/internal/mysql_test_util.go
```diff
@@ -114,10 +129,38 @@ func ExecuteQuery(ctx context.Context, t *testing.T, tableName string, operation
...
+		for _, stream := range streams {
+			_, err := db.ExecContext(ctx, fmt.Sprintf("TRUNCATE TABLE %s_cdc", stream))
+			require.NoError(t, err, fmt.Sprintf("failed to execute %s operation", operation), err)
+			// mysql chunking strategy does not support 0 record sync
+			_, err = db.ExecContext(ctx, fmt.Sprintf("INSERT INTO %s_cdc SELECT * FROM %s ORDER BY id LIMIT 1", stream, stream))
+			require.NoError(t, err, fmt.Sprintf("failed to execute %s operation", operation), err)
+		}
+		return
+
+	case "trigger_cdc":
```
hash-data:
```suggestion
	case "bulk_cdc_data_insert":
```

drivers/mysql/internal/mysql_test_util.go
```diff
@@ -114,10 +129,38 @@ func ExecuteQuery(ctx context.Context, t *testing.T, tableName string, operation
...
+		// insert the data into the cdc tables concurrently
+		err := utils.Concurrent(ctx, streams, len(streams), func(ctx context.Context, stream string, executionNumber int) error {
+			// truncate the inserted data during setup_cdc step
+			_, err := db.ExecContext(ctx, fmt.Sprintf("TRUNCATE TABLE %s_cdc", stream))
```
hash-data: this will also come in cdc records as lsn already marked, i guess we can skip the truncate part

drivers/oracle/go.mod
```diff
@@ -105,4 +105,4 @@ require (
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
-)
```
hash-data: remove file

utils/testutils/test_utils.go
```diff
@@ -22,42 +25,133 @@ const (
...
+func updateStreamsCommand(config TestConfig, namespace string, stream []string, isBackfill bool) string {
+	if len(stream) == 0 {
+		return ""
+	}
+	streamConditions := make([]string, len(stream))
+	for i, s := range stream {
+		streamConditions[i] = fmt.Sprintf(`.stream_name == "%s"`, s)
+	}
+	condition := strings.Join(streamConditions, " or ")
+	tmpCatalog := fmt.Sprintf("/tmp/%s_%s_streams.json", config.Driver, utils.Ternary(isBackfill, "backfill", "cdc"))
```
hash-data:
```suggestion
	tmpCatalog := fmt.Sprintf("/tmp/%s_%s_streams.json", config.Driver, utils.Ternary(isBackfill, "backfill", "cdc").(string))
```

utils/testutils/test_utils.go
```diff
@@ -22,42 +25,133 @@ const (
...
+func GetTestConfig(driver string) *TestConfig {
+	// pwd is olake/drivers/(driver)/internal
+	pwd, err := os.Getwd()
+	if err != nil {
+		panic(err)
+	}
+	// root path is olake's root path
+	rootPath := filepath.Join(pwd, "../../..")
+
+	containerTestDataPath := fmt.Sprintf("/test-olake/drivers/%s/internal/testdata/%s", driver, "%s")
```
hash-data: you can pass driver below. (just adding to make good readability else fine)

utils/testutils/test_utils.go
```diff
@@ -327,3 +418,161 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
 	t.Logf("Verified datatypes in Iceberg after sync")
 }
+
+func (cfg *PerformanceTest) TestPerformance(t *testing.T) {
+	ctx := context.Background()
+
+	// checks if the current rps (from stats.json) is at least 90% of the benchmark rps
+	isRPSAboveBenchmark := func(config TestConfig, isBackfill bool) (bool, error) {
+		// get current RPS
+		var stats SyncSpeed
```
hash-data, line 437: SyncSpeed struct does not have json tag we must add that

utils/testutils/test_utils.go
```diff
@@ -327,3 +418,161 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+							t.Log("(backfill) sync completed")
+
+							checkRPS, err := isRPSAboveBenchmark(*cfg.TestConfig, true)
+							if err != nil {
+								return fmt.Errorf("failed to check RPS: %s", err)
+							}
+							require.True(t, checkRPS, fmt.Sprintf("%s backfill performance below benchmark", cfg.TestConfig.Driver))
+							t.Logf("✅ SUCCESS: %s backfill", cfg.TestConfig.Driver)
+
+							if len(cfg.CDCStreams) > 0 && strings.TrimSpace(cfg.CDCStreams[0]) != "" {
```
hash-data: any reason for second check ?

utils/testutils/test_utils.go
```diff
@@ -327,3 +418,161 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+									return fmt.Errorf("failed to perform discover:\n%s", string(output))
+								}
+								t.Log("(cdc) discover completed")
+
+								updateStreamsCmd := updateStreamsCommand(*cfg.TestConfig, cfg.Namespace, cfg.CDCStreams, false)
+								if code, _, err := utils.ExecCommand(ctx, c, updateStreamsCmd); err != nil || code != 0 {
+									return fmt.Errorf("failed to update streams: %s", err)
+								}
+
+								t.Log("(cdc) starting initial sync")
```
hash-data:
```suggestion
								t.Log("(cdc) setup started")
```

utils/testutils/test_utils.go
```diff
@@ -327,3 +418,161 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+								syncCmd := syncCommand(*cfg.TestConfig, false)
+								if code, output, err := utils.ExecCommand(ctx, c, syncCmd); err != nil || code != 0 {
+									return fmt.Errorf("failed to perform initial sync:\n%s", string(output))
+								}
+								t.Log("(cdc) initial sync completed")
```
hash-data:
```suggestion
								t.Log("(cdc) setup completed")
```

utils/testutils/test_utils.go
```diff
@@ -327,3 +418,161 @@ func VerifyIcebergSync(t *testing.T, tableName string, datatypeSchema map[string
...
+
+								t.Log("(cdc) starting trigger cdc")
+								cfg.ExecuteQuery(ctx, t, cfg.BackfillStreams, "trigger_cdc", true)
```
hash-data: can we use cdc streams here? as that makes sense. if not then making array of cdc streams in configuration not making sense

hash-data (commented):
Minor comments and questions. 
Really good improvement 🚀🚀

drivers/mysql/internal/mysql_test_util.go
vishalm0509: without `TRUNCATE` the bulk insert results in PRIMARY KEY constraint violation. 
There are 2 ways:
1. We can TRUNCATE
2. We can use the following query
```sql
INSERT INTO %s_cdc SELECT * FROM %s WHERE id > 1 ORDER BY id LIMIT 15000000
```

But sometimes, the first `id` can be different then 1. So I think TRUNCATE is better. What do you think?

.github/workflows/performance-test.yml
vishalm0509: Done

drivers/mongodb/go.mod
vishalm0509: We have that in staging and other drivers as well.

drivers/mongodb/internal/mon_test_util.go
vishalm0509: Done

drivers/mongodb/internal/mon_test_util.go
vishalm0509: No it is not single host. Need to check which is primary.

drivers/mysql/internal/mysql_test_util.go
vishalm0509: Made changes

drivers/oracle/go.mod
vishalm0509: Done

utils/testutils/test_utils.go
vishalm0509: Done

utils/testutils/test_utils.go
vishalm0509: Done

utils/testutils/test_utils.go
vishalm0509: Done

utils/testutils/test_utils.go
vishalm0509: I was passing empty cdc streams like this (my bad)
```go
CDCStreams:      []string{""}, // len -> 1
// now changed to 
CDCStreams:      []string{} // len -> 0,
```

Made changes.

utils/testutils/test_utils.go
vishalm0509:
```go
t.Log("(cdc) starting setup cdc")
cfg.ExecuteQuery(ctx, t, cfg.BackfillStreams, "setup_cdc", true)
t.Log("(cdc) setup cdc completed")
```

this is the setup_cdc step. So that would result in confusion when debuging

utils/testutils/test_utils.go
vishalm0509: Made changes. But can you once check above reply

utils/testutils/test_utils.go
vishalm0509: We need cdc streams, to check if cdc is supported and also to update the streams after discover command. I am passing backfillStreams because it's easy to construct cdc streams than the other way around. 

We can pass the cdc streams and do the following. 
```go
backfillStream := strings.Split(cdcStream, "_'")[0]
```

Can we leave it as it is or pass cdcStream?

drivers/mongodb/internal/mon_test_util.go
vishalm0509: After the changes it's less readable. 
```go
		var config Mongo
		utils.UnmarshalFile("./testdata/source.json", &config.config, false)
		connStr = fmt.Sprintf(
			"mongodb+srv://%s:%s@%s/%s",
			config.config.Username,
			config.config.Password,
			config.config.Hosts[0],
			config.config.Database,
```

I think we can keep the var name as driver itself

drivers/mongodb/internal/mon_test_util.go
hash-data: then we must use all host to construct connection string

drivers/mongodb/go.mod
hash-data: we can remove from all, it exists at all places does not mean it is correct.

drivers/mongodb/internal/mon_test_util.go
hash-data: direct import config

utils/testutils/test_utils.go
vishalm0509: Done

utils/testutils/test_utils.go
vishalm0509: Done

drivers/mongodb/internal/mon_test_util.go
vishalm0509: Done

drivers/mongodb/internal/mon_test_util.go
vishalm0509: Done

drivers/mongodb/go.mod
vishalm0509: Done

hash-data (approved):
LGTM

PR #413, opened 2025-07-24
chore: datatypes across drivers globalised

ImDoubD-datazip:
# Description

Datatype across drivers globalised and shifted to abstract level.

## Type of change

# How Has This Been Tested?
- [ ] Scenario A
- [ ] Scenario B


shubham19may:
Usually people care about the DB that they are syncing, so it should always be separated. which gives better visibility for that database specific types on whats happening there.

PR #416, opened 2025-07-27
feat: mongo filter and multi cursor

tanishaAtDatazip:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/drivers/mongodb/README.md
+++ b/drivers/mongodb/README.md
@@ -117,7 +117,7 @@ Before running the Sync command, the generated `streams.json` file must be confi
    > **Note**: For incremental sync to work correctly, the primary cursor field (and fallback cursor field if defined) must contain at least one non-null value. Defined cursor fields cannot be entirely null.
       ```json
          "sync_mode": "incremental",
-         "cursor_field": "UPDATED_AT:CREATED_AT" // UPDATED_AT is the primary cursor field, CREATED_AT is the fallback cursor field (which can be omitted if primary is not expected to contain null values)
+         "cursor_field": "UPDATED_AT:CREATED_AT" // UPDATED_AT is the primary cursor field, CREATED_AT is the fallback cursor field (which can be skipped if the primary cursor field does not contain null values)
       ```
    - The `filter` mode under selected_streams allows you to define precise   criteria for selectively syncing data from your source.
       ```json
--- a/drivers/mongodb/README.md
+++ b/drivers/mongodb/README.md
@@ -10,6 +10,8 @@ The MongoDB Driver enables data synchronization from MongoDB to your desired des
    Tracks and syncs incremental changes from MongoDB in real time.
 3. **Strict CDC (Change Data Capture)**
    Tracks only new changes from the current position in the MongoDB change stream, without performing an initial backfill.
+2. **Incremental**
+   Syncs only new or modified records which have cursor value greater than or equal to the saved position.
 
 ---
 
@@ -98,10 +100,6 @@ Before running the Sync command, the generated `streams.json` file must be confi
       ```json
       "sync_mode": "cdc",
       ```
-   - Specify the cursor field (only for incremental syncs):
-      ```json
-      "cursor_field": "<cursor field from available_cursor_fields>"
-      ```
    - To enable `append_only` mode, explicitly set it to `true` in the selected stream configuration.
       ```json
          "selected_streams": {
@@ -114,6 +112,12 @@ Before running the Sync command, the generated `streams.json` file must be confi
                   }
             ]
          },
+      ```
+   - Add `cursor_field` from set of `available_cursor_fields` in case of incremental sync. This column will be used to track which rows from the table must be synced. If the primary cursor field is expected to contain `null` values, a fallback cursor field can be specified after the primary cursor field using a colon separator. The system will use the fallback cursor when the primary cursor is `null`.
+   > **Note**: For incremental sync to work correctly, the primary cursor field (and fallback cursor field if defined) must contain at least one non-null value. Defined cursor fields cannot be entirely null.
+      ```json
+         "sync_mode": "incremental",
+         "cursor_field": "UPDATED_AT:CREATED_AT" // UPDATED_AT is the primary cursor field, CREATED_AT is the fallback cursor field (which can be omitted if primary is not expected to contain null values)
       ```
    - The `filter` mode under selected_streams allows you to define precise   criteria for selectively syncing data from your source.
       ```json
--- a/drivers/mongodb/README.md
+++ b/drivers/mongodb/README.md
@@ -10,7 +10,7 @@ The MongoDB Driver enables data synchronization from MongoDB to your desired des
    Tracks and syncs incremental changes from MongoDB in real time.
 3. **Strict CDC (Change Data Capture)**
    Tracks only new changes from the current position in the MongoDB change stream, without performing an initial backfill.
-2. **Incremental**
+4. **Incremental**
    Syncs only new or modified records which have cursor value greater than or equal to the saved position.
 
 ---
--- a/drivers/mongodb/internal/incremental.go
+++ b/drivers/mongodb/internal/incremental.go
@@ -6,16 +6,32 @@ import (
 
 	"github.com/datazip-inc/olake/drivers/abstract"
 	"github.com/datazip-inc/olake/types"
+	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/logger"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
+// StreamIncrementalChanges implements incremental sync for MongoDB
 func (m *Mongo) StreamIncrementalChanges(ctx context.Context, stream types.StreamInterface, processFn abstract.BackfillMsgFn) error {
-	primaryCursor, _ := stream.Cursor()
 	collection := m.client.Database(stream.Namespace()).Collection(stream.Name())
-	lastCursorValue := m.state.GetCursor(stream.Self(), primaryCursor)
-	filter := buildMongoCondition(types.Condition{Column: primaryCursor, Value: fmt.Sprintf("%v", lastCursorValue), Operator: ">="})
+	primaryCursor, secondaryCursor := stream.Cursor()
+	lastPrimaryCursorValue := m.state.GetCursor(stream.Self(), primaryCursor)
+	lastSecondaryCursorValue := m.state.GetCursor(stream.Self(), secondaryCursor)
+
+	filter, err := buildFilter(stream)
+	if err != nil {
+		return fmt.Errorf("failed to build filter: %w", err)
+	}
+
+	incrementalFilter, err := m.buildIncrementalCondition(primaryCursor, secondaryCursor, lastPrimaryCursorValue, lastSecondaryCursorValue)
+	if err != nil {
+		return fmt.Errorf("failed to build incremental condition: %w", err)
+	}
+
+	// Merge cursor filter with stream filter using $and
+	filter = utils.Ternary(len(filter) > 0, bson.D{{Key: "$and", Value: bson.A{incrementalFilter, filter}}}, incrementalFilter).(bson.D)
+
 	// TODO: check performance improvements based on the batch size
 	findOpts := options.Find().SetBatchSize(10000)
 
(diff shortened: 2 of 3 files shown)


drivers/mongodb/internal/incremental.go
```diff
@@ -6,16 +6,32 @@ import (
...
+	lastSecondaryCursorValue := m.state.GetCursor(stream.Self(), secondaryCursor)
+
+	filter, err := buildFilter(stream)
+	if err != nil {
+		return fmt.Errorf("failed to build filter: %w", err)
+	}
+
+	incrementalFilter, err := m.buildIncrementalCondition(primaryCursor, secondaryCursor, lastPrimaryCursorValue, lastSecondaryCursorValue)
+	if err != nil {
+		return fmt.Errorf("failed to build incremental condition: %w", err)
```
vaibhav-datazip: use %s instead of %w for logging

drivers/mongodb/internal/incremental.go
```diff
@@ -40,3 +56,36 @@ func (m *Mongo) StreamIncrementalChanges(ctx context.Context, stream types.Strea
 
 	return cursor.Err()
 }
+
+// buildIncrementalCondition generates the incremental condition BSON for MongoDB based on datatype and cursor value.
+func (m *Mongo) buildIncrementalCondition(primaryCursorField string, secondaryCursorField string, lastPrimaryCursorValue any, lastSecondaryCursorValue any) (bson.D, error) {
+
```
vaibhav-datazip, line 68: This empty newline can be removed

drivers/mongodb/internal/incremental.go
```diff
@@ -40,3 +56,36 @@ func (m *Mongo) StreamIncrementalChanges(ctx context.Context, stream types.Strea
 
 	return cursor.Err()
 }
+
+// buildIncrementalCondition generates the incremental condition BSON for MongoDB based on datatype and cursor value.
+func (m *Mongo) buildIncrementalCondition(primaryCursorField string, secondaryCursorField string, lastPrimaryCursorValue any, lastSecondaryCursorValue any) (bson.D, error) {
+	
+	primaryCondition := buildMongoCondition(types.Condition{
```
vaibhav-datazip: Primary condition can be renamed to IncrementalCondition and used inside the if block, to construct the final incremental condition with secondary condition.

then single return statement will be enough instead of using return inside the if-block as well.

ImDoubD-datazip: @tanishaAtDatazip maybe you understood it wrong, what @vaibhav-datazip  was saying to do something like this

```
incrementalCondition := buildMongoCondition(types.Condition{
		Column:   primaryCursorField,
		Value:    lastPrimaryCursorValue,
		Operator: ">=",
	})

	// If secondary is enabled, build an OR condition with fallback
	if secondary {
		incrementalCondition = bson.D{{
			Key: "$or", Value: bson.A{
				incrementalCondition,
				bson.D{{
					Key: "$and", Value: bson.A{
						bson.D{{Key: primaryCursorField, Value: nil}},
						buildMongoCondition(types.Condition{
							Column:   secondaryCursorField,
							Value:    lastSecondaryCursorValue,
							Operator: ">=",
						}),
					},
				}},
			},
		}}
	}

	return incrementalCondition, nil
```

drivers/mongodb/internal/incremental.go
```diff
@@ -6,16 +6,32 @@ import (
...
 
+// StreamIncrementalChanges implements incremental sync for MongoDB
 func (m *Mongo) StreamIncrementalChanges(ctx context.Context, stream types.StreamInterface, processFn abstract.BackfillMsgFn) error {
-	primaryCursor, _ := stream.Cursor()
 	collection := m.client.Database(stream.Namespace()).Collection(stream.Name())
-	lastCursorValue := m.state.GetCursor(stream.Self(), primaryCursor)
-	filter := buildMongoCondition(types.Condition{Column: primaryCursor, Value: fmt.Sprintf("%v", lastCursorValue), Operator: ">="})
+	primaryCursor, secondaryCursor := stream.Cursor()
+	lastPrimaryCursorValue := m.state.GetCursor(stream.Self(), primaryCursor)
+	lastSecondaryCursorValue := m.state.GetCursor(stream.Self(), secondaryCursor)
```
ImDoubD-datazip: can we create a interface for incremental and pass this interface as parameter to buildIncrementalCondition function? Currently lot of parameters are passed with same confusing names.

drivers/mongodb/internal/incremental.go
```diff
@@ -40,3 +56,38 @@ func (m *Mongo) StreamIncrementalChanges(ctx context.Context, stream types.Strea
 
 	return cursor.Err()
 }
+
+// buildIncrementalCondition generates the incremental condition BSON for MongoDB based on datatype and cursor value.
+func (m *Mongo) buildIncrementalCondition(primaryCursorField string, secondaryCursorField string, lastPrimaryCursorValue any, lastSecondaryCursorValue any) (bson.D, error) {
+	var incrementalCondition bson.D
```
ImDoubD-datazip: @tanishaAtDatazip maybe you understood it wrong, what @vaibhav-datazip  was trying say, to do something like this

```
incrementalCondition := buildMongoCondition(types.Condition{
		Column:   primaryCursorField,
		Value:    fmt.Sprintf("%v", lastPrimaryCursorValue),
		Operator: ">=",
	})

	// If secondary is enabled, build an OR condition with fallback
	if secondary {
		incrementalCondition = bson.D{{
			Key: "$or", Value: bson.A{
				incrementalCondition,
				bson.D{{
					Key: "$and", Value: bson.A{
						bson.D{{Key: primaryCursorField, Value: nil}},
						buildMongoCondition(types.Condition{
							Column:   secondaryCursorField,
							Value:    fmt.Sprintf("%v", lastSecondaryCursorValue),
							Operator: ">=",
						}),
					},
				}},
			},
		}}
	}

	return incrementalCondition, nil
```

drivers/mongodb/internal/incremental.go
tanishaAtDatazip: you mean passing a struct? yes, we could group these into one, but it wouldn't meaningfully improve logic or reduce complexity, since the number of fields (2 cursors) is fixed and the logic remains the same.
so we can skip for now and see later if needed.

ImDoubD-datazip: i mean there are 2 fields and we also pass 2 values. it just looked a bit clumsy. but ok no problem, we can see about it later if needed

drivers/mongodb/internal/incremental.go
```diff
@@ -40,3 +56,33 @@ func (m *Mongo) StreamIncrementalChanges(ctx context.Context, stream types.Strea
 
 	return cursor.Err()
 }
+
+// buildIncrementalCondition generates the incremental condition BSON for MongoDB based on datatype and cursor value.
+func (m *Mongo) buildIncrementalCondition(primaryCursor string, secondaryCursor string, lastPrimaryCursorValue any, lastSecondaryCursorValue any) (bson.D, error) {
```
vaibhav-datazip:
```suggestion
func (m *Mongo) buildIncrementalCondition(primaryCursor, secondaryCursor string, lastPrimaryCursorValue, lastSecondaryCursorValue any) (bson.D, error) {
```

drivers/mongodb/internal/incremental.go
```diff
@@ -6,16 +6,32 @@ import (
...
 	collection := m.client.Database(stream.Namespace()).Collection(stream.Name())
-	lastCursorValue := m.state.GetCursor(stream.Self(), primaryCursor)
-	filter := buildMongoCondition(types.Condition{Column: primaryCursor, Value: fmt.Sprintf("%v", lastCursorValue), Operator: ">="})
+	primaryCursor, secondaryCursor := stream.Cursor()
+	lastPrimaryCursorValue := m.state.GetCursor(stream.Self(), primaryCursor)
+	lastSecondaryCursorValue := m.state.GetCursor(stream.Self(), secondaryCursor)
+
+	filter, err := buildFilter(stream)
+	if err != nil {
+		return fmt.Errorf("failed to build filter: %w", err)
```
vaibhav-datazip: %w can be replaced with %s everywhere

vaibhav-datazip:
error syncing data with incremental when using timestamp as cursor fields
Screenshot 2025-08-01 at 3 38 42 PM

tanishaAtDatazip:
> error syncing data with incremental when using timestamp as cursor fields Screenshot 2025-08-01 at 3 38 42 PM

as discussed only lowercase fields is hadled yet in mongodb, the fields with mixed/uppercase is being handled in refactoring as mentioned by @hash-data

drivers/mongodb/internal/incremental.go
```diff
@@ -6,37 +6,83 @@ import (
...
+	}
+
+	// Merge cursor filter with stream filter using $and
+	filter = utils.Ternary(len(filter) > 0, bson.D{{Key: "$and", Value: bson.A{incrementalFilter, filter}}}, incrementalFilter).(bson.D)
+
 	// TODO: check performance improvements based on the batch size
 	findOpts := options.Find().SetBatchSize(10000)
 
-	logger.Infof("Starting incremental sync for stream[%s] with filter: %v", stream.ID(), filter)
+	logger.Infof("Starting incremental sync for stream[%s] with filter: %s", stream.ID(), filter)
```
vaibhav-datazip:
```suggestion
	logger.Infof("Starting incremental sync for stream[%s] with filter: %v", stream.ID(), filter)
```

Wouldn't it be better to use %v instead of %s here? It tends to produce cleaner logs, when filter is not just a string, eg: for a timestamp or bool.

Image

Image

tanishaAtDatazip: good, right

vaibhav-datazip (approved):
LGTM

ImDoubD-datazip (approved):

drivers/mongodb/internal/incremental.go
hash-data: move these to buildIncrementalCondition

drivers/mongodb/internal/incremental.go
```diff
@@ -6,37 +6,83 @@ import (
...
 	}
 
 	return cursor.Err()
 }
+
+// buildIncrementalCondition generates the incremental condition BSON for MongoDB based on datatype and cursor value.
+func (m *Mongo) buildIncrementalCondition(primaryCursor, secondaryCursor string, lastPrimaryCursorValue, lastSecondaryCursorValue any) (bson.D, error) {
+	incrementalCondition := buildMongoCondition(types.Condition{
+		Column:   primaryCursor,
+		Value:    fmt.Sprintf("%v", lastPrimaryCursorValue),
```
hash-data, line 71: in value we have to pass string only?

drivers/mongodb/internal/incremental.go
```diff
@@ -6,37 +6,83 @@ import (
...
+
+	// If secondary is enabled, build an OR condition with fallback
+	if secondaryCursor != "" {
+		incrementalCondition = bson.D{{
+			Key: "$or", Value: bson.A{
+				incrementalCondition,
+				bson.D{{
+					Key: "$and", Value: bson.A{
+						bson.D{{Key: primaryCursor, Value: nil}},
+						buildMongoCondition(types.Condition{
```
hash-data, line 83: we can add a todo on buildMongoCondition to take the value as any and based on that take decisions.

tanishaAtDatazip: yes, was thinking to update well. will add todo

hash-data:
let us update readme as well

drivers/mongodb/internal/incremental.go
tanishaAtDatazip: buildMongoCondition accepts string only, its parsing there the type

drivers/mongodb/README.md
```diff
@@ -115,6 +113,12 @@ Before running the Sync command, the generated `streams.json` file must be confi
             ]
          },
       ```
+   - Add `cursor_field` from set of `available_cursor_fields` in case of incremental sync. This column will be used to track which rows from the table must be synced. If the primary cursor field is expected to contain `null` values, a fallback cursor field can be specified after the primary cursor field using a colon separator. The system will use the fallback cursor when the primary cursor is `null`.
+   > **Note**: For incremental sync to work correctly, the primary cursor field (and fallback cursor field if defined) must contain at least one non-null value. Defined cursor fields cannot be entirely null.
+      ```json
+         "sync_mode": "incremental",
+         "cursor_field": "UPDATED_AT:CREATED_AT" // UPDATED_AT is the primary cursor field, CREATED_AT is the fallback cursor field (which can be omitted if primary is not expected to contain null values)
```
hash-data:
```suggestion
         "cursor_field": "UPDATED_AT:CREATED_AT" // UPDATED_AT is the primary cursor field, CREATED_AT is the fallback cursor field (which can be skipped if the primary cursor field does not contain null values)
```

PR #418, opened 2025-07-28
fix: oracle incremental fallback cursor deletion handled correctly

vaibhav-datazip:
# Description
Fixed a bug in Oracle incremental sync where removing a fallback cursor field after sync completion would cause errors when the primary cursor field was timestamp-based.

When running Oracle incremental sync with multiple cursor fields (primary + fallback), and then removing the fallback cursor field from the configuration, the system would throw typecasting errors if the primary cursor field contained timestamp data. This occurred because the `getIncrementCursorFromState` function was returning the raw cursor values instead of `nil` when cursor fields were missing, leading to improper type handling.

## Type of change

# How Has This Been Tested?

- [x] Ran Oracle incremental sync with multiple cursor fields (primary timestamp + fallback)
- [x] Verified records sync correctly during full_refresh mode
- [x] Confirmed that after removing the secondary cursor field, the sync automatically switches to full_refresh mode without any typecasting errors
- [x] Tested with timestamp-based primary cursor fields to ensure proper type handling

### DIFF
--- a/drivers/abstract/incremental.go
+++ b/drivers/abstract/incremental.go
@@ -27,6 +27,8 @@ func (a *AbstractDriver) Incremental(ctx context.Context, pool *destination.Writ
 			backfillWaitChannel <- stream.ID()
 			return nil
 		}
+		// Reset only mentioned cursor state while preserving other state values
+		a.state.ResetCursor(stream.Self())
 		return a.Backfill(ctx, backfillWaitChannel, pool, stream)
 	})
 	if err != nil {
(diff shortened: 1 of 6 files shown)


ImDoubD-datazip (approved):

ImDoubD-datazip (dismissed):

tanishaAtDatazip (dismissed):

drivers/abstract/incremental.go
```diff
@@ -97,7 +97,7 @@ func (a *AbstractDriver) getIncrementCursorFromState(primaryCursorField string,
 	secondaryStateCursorValue := a.state.GetCursor(stream.Self(), secondaryCursorField)
 
 	if primaryStateCursorValue == nil || (secondaryCursorField != "" && secondaryStateCursorValue == nil) {
-		return primaryStateCursorValue, secondaryStateCursorValue, nil
```
hash-data, line 100: add a comment why it is required, mention edge case here

hash-data: so that we never forget

types/state.go
```diff
@@ -87,6 +87,28 @@ func (s *State) ResetStreams() {
...
+	index, contains := utils.ArrayContains(s.Streams, func(elem *StreamState) bool {
+		return elem.Namespace == stream.Namespace() && elem.Stream == stream.Name()
+	})
+	if contains {
+		s.Streams[index].State.Range(func(key, _ interface{}) bool {
+			if strKey, ok := key.(string); ok {
+				if strKey == primaryCursor || strKey == secondaryCursor {
+					s.Streams[index].State.Delete(key)
+				}
+			}
```
hash-data: range loop not required

hash-data (approved):

ImDoubD-datazip (approved):

PR #419, opened 2025-07-28
fix: MongoDB _id type string

ImDoubD-datazip:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -19,7 +19,7 @@ env:
 jobs:
   integration-tests:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -73,6 +73,16 @@ jobs:
       - name: Build Project
         run: go build -v ./...
 
+      - name: Cache Maven dependencies
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.m2/repository
+            ./destination/iceberg/olake-iceberg-java-writer/target
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
       - name: Build Iceberg Sink
         working-directory: ./destination/iceberg/olake-iceberg-java-writer
         run: mvn clean package -DskipTests
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -78,7 +78,6 @@ jobs:
         with:
           path: |
             ~/.m2/repository
-            ./destination/iceberg/olake-iceberg-java-writer/target
           key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-maven-
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -19,7 +19,7 @@ env:
 jobs:
   integration-tests:
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 30
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
--- a/drivers/mongodb/internal/backfill.go
+++ b/drivers/mongodb/internal/backfill.go
@@ -81,7 +81,7 @@ func (m *Mongo) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 	if err != nil {
 		return nil, fmt.Errorf("failed to check if _id is ObjectID: %s", err)
 	}
-	logger.Infof("_id is%sObjectID", utils.Ternary(isObjID, "", " not ").(string))
+	logger.Infof("_id is %s ObjectID", utils.Ternary(isObjID, "", "not").(string))
 
 	// Generate and update chunks
 	var retryErr error
--- a/drivers/mongodb/internal/backfill.go
+++ b/drivers/mongodb/internal/backfill.go
@@ -329,6 +329,7 @@ func generatePipeline(start, end any, filter bson.D, isObjID bool) mongo.Pipelin
 	var andCond []bson.D
 
 	if isObjID {
+		// convert to primitive.ObjectID
 		start, _ = primitive.ObjectIDFromHex(start.(string))
 		if end != nil {
 			end, _ = primitive.ObjectIDFromHex(end.(string))
@@ -344,6 +345,7 @@ func generatePipeline(start, end any, filter bson.D, isObjID bool) mongo.Pipelin
 	)
 
 	if end != nil {
+		// Changed from $lte to $lt to remove collision between boundaries
 		andCond = append(andCond,
 			bson.D{{Key: "_id", Value: bson.D{{Key: "$lt", Value: end}}}},
 		)
--- a/drivers/mongodb/internal/backfill.go
+++ b/drivers/mongodb/internal/backfill.go
@@ -193,13 +193,13 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 		var chunks []types.Chunk
 		for _, bucket := range buckets {
 			// converts value according to _id string repr.
-			min, err := toString(bucket.ID.Min)
+			min, err := toIDType(bucket.ID.Min)
 			if err != nil {
-				return nil, fmt.Errorf("failed to convert bucket min value to string: %s", err)
+				return nil, fmt.Errorf("failed to convert bucket min value to required type: %s", err)
 			}
-			max, err := toString(bucket.ID.Max)
+			max, err := toIDType(bucket.ID.Max)
 			if err != nil {
-				return nil, fmt.Errorf("failed to convert bucket max value to string: %s", err)
+				return nil, fmt.Errorf("failed to convert bucket max value to required type: %s", err)
 			}
 			chunks = append(chunks, types.Chunk{
 				Min: min,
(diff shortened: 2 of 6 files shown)


drivers/mongodb/internal/backfill.go
```diff
@@ -294,47 +325,32 @@ func (m *Mongo) fetchExtremes(ctx context.Context, collection *mongo.Collection,
...
 	if end != nil {
-		// Changed from $lte to $lt to remove collision between boundaries
-		andOperation = append(andOperation, bson.D{{
-			Key: "_id",
-			Value: bson.D{{
-				Key:   "$lt",
-				Value: end,
-			}},
-		}})
+		andCond = append(andCond,
```
vaibhav-datazip: The comment can be added back here for clarity

drivers/mongodb/internal/backfill.go
```diff
@@ -179,14 +192,27 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 
 		var chunks []types.Chunk
 		for _, bucket := range buckets {
+			// converts value according to _id string repr.
+			min, err := toString(bucket.ID.Min)
+			if err != nil {
+				return nil, err
```
vaibhav-datazip: Add proper error statement

drivers/mongodb/internal/backfill.go
```diff
@@ -179,14 +192,27 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
+			}
+			max, err := toString(bucket.ID.Max)
+			if err != nil {
+				return nil, err
```
vaibhav-datazip: here as well, add proper error statement

drivers/mongodb/internal/backfill.go
```diff
@@ -179,14 +192,27 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
-				Min: bucket.ID.Min.Hex(),
-				Max: bucket.ID.Max.Hex(),
+				Min: min,
+				Max: max,
 			})
 		}
 		if len(buckets) > 0 {
+			max, err := toString(buckets[len(buckets)-1].ID.Max)
+			if err != nil {
+				return nil, err
```
vaibhav-datazip: Here as well

drivers/mongodb/internal/backfill.go
```diff
@@ -235,14 +261,19 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
+			chunks, err := splitVectorStrategy()
+			// check if authorization error occurs
+			if err != nil && (strings.Contains(err.Error(), "not authorized") ||
+				strings.Contains(err.Error(), "CMD_NOT_ALLOWED")) {
+				logger.Warnf("failed to get chunks via split vector strategy: %s", err)
+				return bucketAutoStrategy()
+			}
+			return chunks, err
+		} else {
+			// fall back to bucket auto startegy
```
vaibhav-datazip: this comment can be removed

drivers/mongodb/internal/backfill.go
ImDoubD-datazip: added

drivers/mongodb/internal/backfill.go
ImDoubD-datazip: this provides fallback clarity, i believe its better if this comment is there

vaibhav-datazip (dismissed):

.github/workflows/integration-tests.yml
```diff
@@ -73,6 +73,16 @@ jobs:
       - name: Build Project
         run: go build -v ./...
 
+      - name: Cache Maven dependencies
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.m2/repository
+            ./destination/iceberg/olake-iceberg-java-writer/target
```
shubham19may: remove target path from here.

.github/workflows/integration-tests.yml
```diff
@@ -73,6 +73,16 @@ jobs:
...
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
```
shubham19may, line 81: what is this? it has to cache the whole folder thats it.

drivers/mongodb/internal/backfill.go
```diff
@@ -426,3 +443,33 @@ func buildFilter(stream types.StreamInterface) (bson.D, error) {
...
+		return t.Hex(), nil
+	case string:
+		return t, nil
+	default:
+		// fallback
+		return utils.ConvertToString(v), nil
+	}
+}
+
+func isObjectID(ctx context.Context, collection *mongo.Collection) (bool, error) {
```
shubham19may, line 470: why can't we directly check with streams.json type only?

ImDoubD-datazip: at streams.json level, the _id we receive is in string type only even though it is a primitive.objectID type

.github/workflows/integration-tests.yml
ImDoubD-datazip: removed

.github/workflows/integration-tests.yml
ImDoubD-datazip: whole folder is getting cached. 
here this is not any path, it is just cache key generated on the basis of pom.xml file. any changes in the pom.xml takes place, a new cache key will be generated

drivers/mongodb/internal/backfill.go
```diff
@@ -179,14 +192,27 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 
 		var chunks []types.Chunk
 		for _, bucket := range buckets {
+			// converts value according to _id string repr.
+			min, err := toString(bucket.ID.Min)
```
vikaxsh: If Min or Max is nil, toString may panic or return undesired values.

drivers/mongodb/internal/backfill.go
```diff
@@ -235,14 +261,19 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
+			chunks, err := splitVectorStrategy()
+			// check if authorization error occurs
+			if err != nil && (strings.Contains(err.Error(), "not authorized") ||
+				strings.Contains(err.Error(), "CMD_NOT_ALLOWED")) {
+				logger.Warnf("failed to get chunks via split vector strategy: %s", err)
+				return bucketAutoStrategy()
+			}
+			return chunks, err
+		} else {
+			// fall back to bucket auto strategy
```
shubham19may: splitvector can still be used in objectId or string right?

why are we restricting it for only object ids?

drivers/mongodb/internal/backfill.go
shubham19may: in case of _id being objectId this will give same value as .Hex() right?

shubham19may: As this is the default strategy for spitvector non-allowed collections

shubham19may: also once test for this scenario.

drivers/mongodb/internal/backfill.go
```diff
@@ -426,3 +443,33 @@ func buildFilter(stream types.StreamInterface) (bson.D, error) {
 		return bson.D{{Key: "$" + filter.LogicalOperator, Value: bson.A{buildMongoCondition(filter.Conditions[0]), buildMongoCondition(filter.Conditions[1])}}}, nil
 	}
 }
+
+func toString(v interface{}) (string, error) {
+	switch t := v.(type) {
```
vikaxsh, line 459: should we handle other possible _id type as well ?, in case of unable to create chunk on that type we can fallback to timestamp with a WARN log

ImDoubD-datazip: for now as per thread, the fix was asked for string type

drivers/mongodb/internal/backfill.go
ImDoubD-datazip: have you tested this?

drivers/mongodb/internal/backfill.go
vikaxsh: we can avoid nested checks simply check in case of non object id and return like this
`if !isObjID {
		return bucketAutoStrategy()
	}`

.github/workflows/integration-tests.yml
```diff
@@ -19,7 +19,7 @@ env:
 jobs:
   integration-tests:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
```
hash-data: revert

drivers/mongodb/internal/backfill.go
```diff
@@ -69,11 +76,17 @@ func (m *Mongo) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 		return nil, fmt.Errorf("failed to parse filter during chunk splitting: %s", err)
 	}
 
+	// check for _id type
+	isObjID, err := isObjectID(ctx, collection)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if _id is ObjectID: %s", err)
+	}
+
```
hash-data, line 85: add a log that if it is object id or not

drivers/mongodb/internal/backfill.go
```diff
@@ -179,14 +192,27 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 
 		var chunks []types.Chunk
 		for _, bucket := range buckets {
+			// converts value according to _id string repr.
+			min, err := toIDType(bucket.ID.Min)
```
```suggestion
			min, err := reformatID(bucket.ID.Min)
```

drivers/mongodb/internal/backfill.go
```diff
@@ -235,14 +261,19 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 	case "timestamp":
 		return timestampStrategy()
 	default:
-		chunks, err := splitVectorStrategy()
-		// check if authorization error occurs
-		if err != nil && (strings.Contains(err.Error(), "not authorized") ||
-			strings.Contains(err.Error(), "CMD_NOT_ALLOWED")) {
-			logger.Warnf("failed to get chunks via split vector strategy: %s", err)
+		if isObjID {
+			chunks, err := splitVectorStrategy()
```
hash-data: if you can add a comment in here why we not using splitvector for non object id's will be good

drivers/postgres/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/datazip-inc/olake/drivers/postgres
 
-go 1.23.0
+go 1.23.2
 
-toolchain go1.23.7
+toolchain go1.24.3
```
hash-data, line 5: why toolchain being updated?

hash-data: also check in mysql as well.

Also in main go.mod toolchain is not changed so i think it might not required

hash-data (approved):

PR #423, opened 2025-07-30
feat: consistent batch size across all drivers

vaibhav-datazip:
# Description

Earlier batch size was different for different drivers, like 10k rows for some and 1gb data for other.
This PR introduces batch size consistency across for all drivers, which will target to generate parquet files of 512MB. 

512 (parquet size) * 8 (considered compression ratio) * 2 (chunk inflation factor)

this has been implemented for following strategies
- mongodb
  - split vector 
  - bucket auto
- mysql
  - primary key chunking
  - limit offset chunking
- postgres
  - ctid strategy  

## Type of change

# How Has This Been Tested?

- [x] Verified consistent chunk sizes across all strategies
- [x] Confirmed no data loss by comparing record counts
- [x] Tested with various collection sizes
- [x] Validated parquet file sizes are closer to 512MB target


drivers/mongodb/internal/backfill.go
```diff
@@ -150,17 +151,16 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
-		for i := 0; i < len(boundaries)-1; i++ {
-			chunks = append(chunks, types.Chunk{
-				Min: boundaries[i].Hex(),
-				Max: boundaries[i+1].Hex(),
-			})
-		}
-		if len(boundaries) > 0 {
+		for idx := 0; idx < len(boundaries)-1; idx += 8 {
+			var maxBoundary any
+			if idx < (len(boundaries) - 9) {
```
ImDoubD-datazip: check if we can use something like a while loop here where the condition is directly using this len(boundaries)-9. this way it wont be re-calculated again and again.

drivers/mongodb/internal/backfill.go
```diff
@@ -172,11 +172,23 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 		if len(filter) > 0 {
 			pipeline = append(pipeline, bson.D{{Key: "$match", Value: filter}})
 		}
+
+		var stats bson.M
+		err := collection.Database().RunCommand(ctx, bson.D{{Key: "collStats", Value: collection.Name()}}).Decode(&stats)
```
ImDoubD-datazip: we need to discuss about cases where there is filter applied, and since this query provides stats for the whole collection, through this lot of near-empty buckets are created, which requires scanning for data but it will be mostly empty due to filter.

pkg/jdbc/jdbc.go
```diff
@@ -111,6 +116,11 @@ func PostgresChunkScanQuery(stream types.StreamInterface, filterColumn string, c
 }
 
 // MySQL-Specific Queries
+// getAvgRowSizeQuery returns the query to fetch the average row size of a table in MySQL
+func AvgRowSizeQuery(stream types.StreamInterface) string {
+	return fmt.Sprintf("SELECT  CEIL(data_length / NULLIF(table_rows, 0)) AS `avg_row_bytes` FROM information_schema.tables WHERE table_schema = %q AND table_name = %q", stream.Namespace(), stream.Name())
```
ImDoubD-datazip: use '%s' instead of %q, if you want to use %q then change every where in mysql queries in the where clause

drivers/mongodb/internal/backfill.go
```diff
@@ -172,11 +172,23 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to get collection stats: %s", err)
+		}
+		storageSize, err := typeutils.ReformatFloat64(stats["storageSize"])
+		if err != nil {
+			return nil, fmt.Errorf("failed to get collection stats: %s", err)
+		}
+
+		numberOfChunks := int(math.Ceil(storageSize.(float64) / float64(constants.EffectiveParquetSize)))
```
ImDoubD-datazip: can we change the name to numberOfBuckets as it is passed to $bucketAuto as buckets

drivers/postgres/internal/backfill.go
```diff
@@ -57,15 +59,21 @@ func (p *Postgres) GetOrSplitChunks(_ context.Context, pool *destination.WriterP
...
 func (p *Postgres) splitTableIntoChunks(stream types.StreamInterface) (*types.Set[types.Chunk], error) {
 	generateCTIDRanges := func(stream types.StreamInterface) (*types.Set[types.Chunk], error) {
-		var relPages uint32
+		var relPages, blockSize uint32
 		relPagesQuery := jdbc.PostgresRelPageCount(stream)
 		err := p.client.QueryRow(relPagesQuery).Scan(&relPages)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get relPages: %s", err)
 		}
+		blockSizeQuery := jdbc.PostgresBlockSizeQuery()
```
ImDoubD-datazip, line 68: here when taking out block_size, can you see if we can hard-code this? by default, it is 8 KBs and can go max upto 32 KBs if changed. Just check it out once.

drivers/mongodb/internal/backfill.go
vaibhav-datazip: I don't think that will make it cleaner than this approach. Can you suggest if you have something in mind ?

ImDoubD-datazip: let's keep it this way for now, can add a todo if necessary

ImDoubD-datazip (dismissed):

drivers/mysql/internal/backfill.go
```diff
@@ -54,19 +54,34 @@ func (m *MySQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
...
 	}
 	pool.AddRecordsToSync(approxRowCount)
 
+	avgRowSizeQuery := jdbc.AvgRowSizeQuery(stream)
+	var avgRowSize int64
+	err = m.client.QueryRow(avgRowSizeQuery).Scan(&avgRowSize)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get rows per chunk: %s", err)
+	}
+
```
vikaxsh: see if we could calculate average row size in MySQL using information_schema table , which we are already using for approx row count

drivers/mongodb/internal/backfill.go
```diff
@@ -172,11 +172,23 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
+		err := collection.Database().RunCommand(ctx, bson.D{{Key: "collStats", Value: collection.Name()}}).Decode(&stats)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get collection stats: %s", err)
+		}
+		storageSize, err := typeutils.ReformatFloat64(stats["storageSize"])
+		if err != nil {
+			return nil, fmt.Errorf("failed to get collection stats: %s", err)
+		}
+
+		numberOfBuckets := int(math.Ceil(storageSize.(float64) / float64(constants.EffectiveParquetSize)))
```
vikaxsh: can we create avg obj size? instead of storage size

pkg/jdbc/jdbc.go
```diff
@@ -246,9 +243,10 @@ func MySQLPrimaryKeyQuery() string {
 }
 
 // MySQLTableRowsQuery returns the query to fetch the estimated row count of a table in MySQL
-func MySQLTableRowsQuery() string {
+func AvgRowSizeAndRowCountQuery() string {
```
vikaxsh: name should be something like that
MysqlTableRowStatsQuery

vikaxsh:
I have tested your PR. Here are the results:

MongoDB (with $bucketAuto chunking):

Generated Parquet file sizes ranged from 4 KB to 4 MB, which is far from the expected 512 MB target size.

PostgreSQL (using ctid chunking):

Generated Parquet file sizes were approximately 2.3 GB (normalization=false) and 1.6 GB (normalization=true).

MySQL:

Tables with primary key (PK): Parquet file size was 3.9 GB.

Tables without primary key (No-PK): Parquet file size was 3.6 GB.


Are these file sizes the expected behavior for each source and configuration?

vaibhav-datazip:
> I have tested your PR. Here are the results:
> 
> MongoDB (with $bucketAuto chunking):
> 
> Generated Parquet file sizes ranged from 4 KB to 4 MB, which is far from the expected 512 MB target size.
> 
> PostgreSQL (using ctid chunking):
> 
> Generated Parquet file sizes were approximately 2.3 GB (normalization=false) and 1.6 GB (normalization=true).
> 
> MySQL:
> 
> Tables with primary key (PK): Parquet file size was 3.9 GB.
> 
> Tables without primary key (No-PK): Parquet file size was 3.6 GB.
> 
> Are these file sizes the expected behavior for each source and configuration?

mongo result is because of datatype change in same named field, if you would've ran the sync for near an hour or two the parquet size would've been consistent (experimented myself) and near 1gb (as the formula has been changed and inflation factor of 2 has been included), rest of the driver parquets are inflated, and will be cutdown to 512mb each from the writer side after the refactoring is completed.

drivers/mongodb/internal/backfill.go
hash-data: Can we precalculate len?

drivers/mongodb/internal/backfill.go
hash-data: any reason why we have not used function totalCountInCollection that is already available, that can return storage size as well.

drivers/mongodb/internal/backfill.go
```diff
@@ -199,31 +211,24 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
 			}
-			max, err := reformatID(bucket.ID.Max)
-			if err != nil {
-				return nil, fmt.Errorf("failed to convert bucket max value to required type: %s", err)
+			var max interface{}
+			if idx != len(buckets)-1 {
+				max, err = reformatID(bucket.ID.Max)
+				if err != nil {
+					return nil, fmt.Errorf("failed to convert bucket max value to required type: %s", err)
+				}
```
hash-data, line 217: add a comment why it was done

drivers/postgres/internal/backfill.go
```diff
@@ -57,15 +59,21 @@ func (p *Postgres) GetOrSplitChunks(_ context.Context, pool *destination.WriterP
...
+		err = p.client.QueryRow(blockSizeQuery).Scan(&blockSize)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get block size: %s", err)
+		}
+		batchSize := uint32(math.Ceil(float64(constants.EffectiveParquetSize) / float64(blockSize)))
```
hash-data, line 73: are we sure the table which is being queried does not require extra permissions, as well any fallback we have to add here?

pkg/jdbc/jdbc.go
```diff
@@ -111,6 +118,11 @@ func PostgresChunkScanQuery(stream types.StreamInterface, filterColumn string, c
 }
 
 // MySQL-Specific Queries
+// AnalyzeTableQuery returns the query to analyze a table in MySQL
+func AnalyzeTableQuery(stream types.StreamInterface) string {
+	return fmt.Sprintf("ANALYZE TABLE %s.%s", stream.Namespace(), stream.Name())
+}
+
```
hash-data: any required permissions for it?

hash-data: check for all queries once. (that are being used in this pr)

drivers/postgres/internal/backfill.go
vaibhav-datazip: this block size query doesn't require any extra permissions, also block size is limited between 1kb to 32kb

pkg/jdbc/jdbc.go
vaibhav-datazip: for the analyze table query , select and insert permissions are required for that table

[refrence](https://dev.mysql.com/doc/refman/8.4/en/analyze-table.html)

drivers/mongodb/internal/backfill.go
```diff
@@ -150,33 +151,41 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
 		pipeline := mongo.Pipeline{}
 		if len(filter) > 0 {
 			pipeline = append(pipeline, bson.D{{Key: "$match", Value: filter}})
 		}
+
+		var stats bson.M
+		err := collection.Database().RunCommand(ctx, bson.D{{Key: "collStats", Value: collection.Name()}}).Decode(&stats)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get collection stats: %s", err)
+		}
```
ImDoubD-datazip: remove this command, it is of no use here

drivers/mongodb/internal/backfill.go
```diff
@@ -199,31 +202,25 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
 			// converts value according to _id string repr.
 			min, err := reformatID(bucket.ID.Min)
 			if err != nil {
 				return nil, fmt.Errorf("failed to convert bucket min value to required type: %s", err)
 			}
-			max, err := reformatID(bucket.ID.Max)
-			if err != nil {
-				return nil, fmt.Errorf("failed to convert bucket max value to required type: %s", err)
+			var max interface{}
+			// for last bucket, max will be nil
```
hash-data, line 212: if we add last chunk max to nil we might end up with one single record which is equal to max only so to avoid such case we are taking prev min to nil

drivers/mongodb/internal/backfill.go
```diff
@@ -282,24 +279,32 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
+			return bucketAutoStrategy(storageSize)
 		}
 		return chunks, err
 	}
 }
 
-func (m *Mongo) totalCountInCollection(ctx context.Context, collection *mongo.Collection) (int64, error) {
-	var countResult bson.M
+func (m *Mongo) totalCountAndStorageSizeInCollection(ctx context.Context, collection *mongo.Collection) (int64, float64, error) {
+	var statResults bson.M
```
```suggestion
	var statsResult bson.M
```

drivers/mongodb/internal/backfill.go
```diff
@@ -282,24 +279,32 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
+	if err != nil {
+		return 0, 0, fmt.Errorf("failed to get total count: %s", err)
+	}
+	storageSize, err := typeutils.ReformatFloat64(statResults["storageSize"])
 	if err != nil {
-		return 0, fmt.Errorf("failed to get total count: %s", err)
+		return 0, 0, fmt.Errorf("failed to get total count: %s", err)
 	}
-	return typeutils.ReformatInt64(countResult["count"])
+	return count, storageSize.(float64), nil
```
hash-data: any reason to typecast storage size here as it already a float64 value

drivers/mysql/internal/backfill.go
```diff
@@ -55,18 +56,25 @@ func (m *MySQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
...
 	var approxRowCount int64
-	approxRowCountQuery := jdbc.MySQLTableRowsQuery()
-	err := m.client.QueryRow(approxRowCountQuery, stream.Name()).Scan(&approxRowCount)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get approx row count: %s", err)
+	var avgRowSize any
+	approxRowCountQuery := jdbc.MySQLTableRowStatsQuery()
+	err := m.client.QueryRow(approxRowCountQuery, stream.Name()).Scan(&approxRowCount, &avgRowSize)
+	if err != nil || avgRowSize == nil {
+		errorMsg := utils.Ternary(err != nil, fmt.Errorf("failed to get approx row count and avg row size: %s", err), fmt.Errorf("stats not populated for [%s] run ANALYZE TABLE query or the table contains 0 records", stream.ID()))
```
hash-data: can you update some thing like this : 
errorMsg := utils.Ternary(err != nil, fmt.Errorf("failed to get approx row count and avg row size: %s", err), fmt.Errorf("either stats not populated for table[%s] or the table contains 0 records. (to populate stats `ANALYZE TABLE` query )", stream.ID()))

drivers/mysql/internal/backfill.go
```diff
@@ -55,18 +56,25 @@ func (m *MySQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to create sql filter during chunk splitting: %s", err)
 	}
-
+	// avgRowSize is returned as []uint8 which is converted to float64
+	avgRowSizeFloat, err := typeutils.ReformatFloat64(avgRowSize)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get avg row size: %s", err)
+	}
+	chunkSize := int64(math.Ceil(float64(constants.EffectiveParquetSize) / avgRowSizeFloat.(float64)))
```
hash-data: avg row size already a float64 value

drivers/mongodb/internal/backfill.go
vaibhav-datazip: in this case as well, we are checking if we are on last bucket then max is set to nil.

And the chunk created becomes min to nil

drivers/mongodb/internal/backfill.go
```diff
@@ -282,24 +279,32 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
+		return 0, 0, fmt.Errorf("failed to get total count: %s", err)
+	}
+	count, err := typeutils.ReformatInt64(statsResult["count"])
+	if err != nil {
+		return 0, 0, fmt.Errorf("failed to get total count: %s", err)
+	}
+	storageSize, err := typeutils.ReformatFloat64(statsResult["storageSize"])
 	if err != nil {
-		return 0, fmt.Errorf("failed to get total count: %s", err)
+		return 0, 0, fmt.Errorf("failed to get total count: %s", err)
```
hash-data: change error properly

vikash390 (approved):

PR #439, opened 2025-08-10
feat: destination refactor

hash-data:
# Description
-> Refactored Java Code 
-> Refactored Go side of writer 
-> Optimized protobuf record structure 
```
{
  "type":"records"
  "metadata": {
    "dest_table_name": "",
    "thread_id": "",
    "primary_key": ""
    "schema": [
      {
        "ice_type": "",
        "key": "key_of_value_0"
      }
    ]
  },
  "records": [
    {
      //The values correspond to the schema array order. For example, 'olake_id' is at position 0 in the schema metadata, so its value should also be at key 0
      "fields" :["value0", "value1"]
      ]
      "record_type": "u/c/r"
    }
  ]
}

{
  "type":"commit"
  "metadata": {
    "dest_table_name": "", 
    "thread_id": "", // required to commit data files 
  }
}


{
  "type":"evolve_schema"
  "metadata": {
    "dest_table_name": "",
    "thread_id": "",
    "primary_key": ""
    "schema": [
      {
        "ice_type": "",
        "key": ""
      }
    ]
  }
}

{
  "type":"drop_table"
  "metadata": {
    "dest_table_name": "",
    "thread_id": ""
  }
}

{
  "type":"get_or_create_table"
  "metadata": {
    "dest_table_name": "",
    "thread_id": "",
    "primary_key": ""
    "schema": [
      {
        "ice_type": "",
        "key": ""
      }
    ]
  }
}
```

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


destination/dest_utils.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+	}
+	return 0, fmt.Errorf("no available ports found between 50051 and 59051")
+}
+
+// GetGoroutineID returns a unique ID for the current goroutine
+// This is a simple implementation that uses the string address of a local variable
+// which will be unique per goroutine
+func GetGoroutineID() string {
+	var buf [64]byte
```
hash-data: Can we add a hash in this goroutine ID, as it is just a number?

destination/iceberg/java_client.go
```diff
@@ -0,0 +1,220 @@
...
+		addMapKeyIfNotEmpty("oauth2-server-uri", config.RestOAuthURI)
+		addMapKeyIfNotEmpty("rest.auth.type", config.RestAuthType)
+		addMapKeyIfNotEmpty("credential", config.RestCredential)
+		addMapKeyIfNotEmpty("scope", config.RestScope)
+	default:
+		return nil, fmt.Errorf("unsupported catalog type: %s", config.CatalogType)
+	}
+
+	// Configure S3 file IO
+	serverConfig["io-impl"] = "org.apache.iceberg.aws.s3.S3FileIO"
```
shubham19may: Remove this, its already defined in the bottom,

destination/writers.go
```diff
@@ -22,31 +20,41 @@ type (
...
-		Insert InsertFunction
+	ThreadOptions   func(opt *Options)
+	StreamArtifacts struct {
+		mutex  sync.RWMutex
+		schema any
 	}
 
+	Stats struct {
+		readCount     atomic.Int64
+		writeCount    atomic.Int64
```
hash-data: this is also not required need to discuss

destination/writers.go
```diff
@@ -22,31 +20,41 @@ type (
...
+		config          any
+		init            NewFunc
+		streamArtifacts sync.Map
+	}
+
+	ThreadEvent struct {
+		buffer         []types.RawRecord
+		writer         Writer
+		stats          *Stats
+		batchSize      int64
```
hash-data: can we do something according to writer ?

destination/parquet/parquet.go
```diff
@@ -215,107 +212,132 @@ func (p *Parquet) Check(_ context.Context) error {
...
-		err := p.createNewPartitionFile(partitionedPath)
-		if err != nil {
-			return err
-		}
+func (p *Parquet) EvolveSchema(ctx context.Context, newSchema any) error {
+	if !p.stream.NormalizationEnabled() {
+		return nil
+	}
+	// override new schema in streams
+	newFieldSchema, ok := newSchema.(typeutils.Fields)
```
hash-data: we might require it to be passed by reference, test case : change in fields

hash-data: need to give thought

destination/dest_utils.go
```diff
@@ -0,0 +1,73 @@
+package destination
```
shubham19may: this file specifically contains functions which are used in Iceberg side, why this file is outside?

also in refactor, please do not change the file names as it makes it more difficult to review. 
Better way is to change files in one more sub-PR which does the file level refactoring.

destination/iceberg/java_client.go
```diff
@@ -0,0 +1,220 @@
...
+	// If this was the last reference, shut down the server
+	logger.Infof("thread id %s: shutting down Iceberg server on port %d", server.serverID, server.port)
+	server.conn.Close()
+	if server.cmd != nil && server.cmd.Process != nil {
+		err := server.cmd.Process.Kill()
+		if err != nil {
+			logger.Errorf("thread id %s: Failed to kill Iceberg server: %s", server.serverID, err)
+		}
+	}
+
```
shubham19may: // Release the port

		portMap.Delete(i.port)

we did this in last implementation. any reason for not to do this?

destination/iceberg/proto/records_ingest.proto
```diff
@@ -4,18 +4,57 @@ package io.debezium.server.iceberg.rpc;
...
+  message IceRecord {
+    // Use oneof for efficient memory usage - only one field type active per record field
+    message FieldValue {
+      oneof value {
+        string string_value = 1;
+        int32 int_value = 2;
+        int64 long_value = 3;
+        float float_value = 4;
+        double double_value = 5;
+        bool bool_value = 6;
```
shubham19may, line 45: what about decimal, date, time, timestamp, timestamptz, uuid, binary?

hash-data:
TODO: docs update if any, readme update, and as well update the batch size

constants/constants.go
```diff
@@ -18,11 +18,12 @@ const (
 	OpType                 = "_op_type"
 	CdcTimestamp           = "_cdc_timestamp"
 	DBName                 = "_db"
+	StringifiedData        = "data"
 	DefaultReadPreference  = "secondaryPreferred"
 	EncryptionKey          = "OLAKE_ENCRYPTION_KEY"
 	ConfigFolder           = "CONFIG_FOLDER"
-	// EffectiveParquetSize is the effective size in bytes considering 512MB targeted parquet size, compression ratio as 8 and chunk inflation factor as 2
-	EffectiveParquetSize = int64(512) * 1024 * 1024 * int64(8) * int64(2)
+	// EffectiveParquetSize is the effective size in bytes considering 256mb targeted parquet size, compression ratio as 4
```
hash-data: update comment properly

utils/typeutils/compare.go
```diff
@@ -1,24 +0,0 @@
...
-	}
-	if b == nil {
-		return 1
-	}
-
-	aTime, aOk := a.(Time)
-	bTime, bOk := b.(Time)
-
-	if aOk && bOk {
-		return aTime.Compare(bTime)
```
shubham19may, line 21: this was added because the CompareInterfaceValue doesn't handle the time.go (struct of time.time) comparision.

shubham19may: needed for sorting partionkey based things

types/data_types.go
```diff
@@ -74,163 +79,6 @@ func CreateRawRecord(olakeID string, data map[string]any, operationType string,
...
-			case int, int8, int16, int32:
-				field["type"] = "int32"
-			case int64:
-				field["type"] = "int64"
-			case float32:
-				field["type"] = "float32"
-			case float64:
-				field["type"] = "float64"
-			case time.Time:
-				field["type"] = "timestamptz" // use with timezone as we use default utc
```
shubham19may, line 177: this is not covered in the new grpc object that we create

types/data_types.go
```diff
@@ -55,17 +52,25 @@ var TypeWeights = map[DataType]int{
...
+	constants.OpType:          String,
+	constants.OlakeID:         String,
+}
+
 type Record map[string]any
 
 type RawRecord struct {
 	Data           map[string]any `parquet:"data,json"`
 	OlakeID        string         `parquet:"_olake_id"`
 	OlakeTimestamp time.Time      `parquet:"_olake_timestamp"`
```
shubham19may, line 68: why not this also pointer?

types/data_types.go
```diff
@@ -74,163 +79,6 @@ func CreateRawRecord(olakeID string, data map[string]any, operationType string,
...
-				field["type"] = "string"
-			}
-
-			dataFields = append(dataFields, field)
-		}
-
-		// Sorting basis on field names is needed because
-		// Iceberg writer detects different schemas for
-		// schema evolution based on order columns passed
-		sort.Slice(dataFields, func(i, j int) bool {
```
shubham19may, line 188: just adding comment for discussion later, this was added because iceberg current writer was detecting different schema everytime basis on column order.

types/adapter.go
```diff
@@ -1,14 +1,16 @@
...
+	Iceberg DestinationType = "ICEBERG"
 )
 
 // TODO: Add validations
 type WriterConfig struct {
-	Type         AdapterType `json:"type"`
-	WriterConfig any         `json:"writer"`
+	Type         DestinationType `json:"type"`
+	BatchSize    int             `json:"batch_size"`
+	MaxThreads   int             `json:"max_threads"`
```
shubham19may: max_threads for destination as well?

types/adapter.go
```diff
@@ -1,14 +1,16 @@
...
+	Parquet DestinationType = "PARQUET"
+	Iceberg DestinationType = "ICEBERG"
 )
 
 // TODO: Add validations
 type WriterConfig struct {
-	Type         AdapterType `json:"type"`
-	WriterConfig any         `json:"writer"`
+	Type         DestinationType `json:"type"`
+	BatchSize    int             `json:"batch_size"`
```
shubham19may: for what?

shubham19may: too many params to be configured.

destination/writers.go
```diff
@@ -66,173 +73,173 @@ func WithBackfill(backfill bool) ThreadOptions {
...
+		return nil
+	}
+
+	// create flush context
+	flushCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	t.streamArtifact.mutex.RLock()
+	cachedSchema := t.streamArtifact.schema
+	t.streamArtifact.mutex.RUnlock()
```
vaibhav-datazip: A comment can be added to explain this.

destination/writers.go
```diff
@@ -27,26 +25,35 @@ type (
...
-		totalRecords  atomic.Int64
-		recordCount   atomic.Int64
-		ThreadCounter atomic.Int64 // Used in naming files in S3 and global count for threads
-		config        any          // respective writer config
-		init          NewFunc      // To initialize exclusive destination threads
-		group         *errgroup.Group
-		groupCtx      context.Context
-		tmu           sync.Mutex // Mutex between threads
+		stats           *Stats
+		totalRecords    atomic.Int64
```
shubham19may: what does this mean? lets add comment.

destination/writers.go
```diff
@@ -27,26 +25,35 @@ type (
...
+	ThreadOptions   func(opt *Options)
+	StreamArtifacts struct {
+		mutex  sync.RWMutex
+		schema any
 	}
 
+	Stats struct {
+		readCount     atomic.Int64
+		writeCount    atomic.Int64
+		writerThreads atomic.Int64
```
shubham19may: add comment on what is this

destination/writers.go
```diff
@@ -27,26 +25,35 @@ type (
...
-		groupCtx      context.Context
-		tmu           sync.Mutex // Mutex between threads
+		stats           *Stats
+		totalRecords    atomic.Int64
+		config          any
+		init            NewFunc
+		streamArtifacts sync.Map
+	}
+
+	ThreadEvent struct {
```
shubham19may: comments.

destination/parquet/parquet.go
hash-data: checked the case as streamArtifact is a pointer so not to worry about

destination/writers.go
shubham19may: name can be changed of this struct.

destination/writers.go
```diff
@@ -66,173 +73,173 @@ func WithBackfill(backfill bool) ThreadOptions {
...
-func NewWriter(ctx context.Context, config *types.WriterConfig, dropStreams []string) (*WriterPool, error) {
-	newfunc, found := RegisteredWriters[config.Type]
-	if !found {
-		return nil, fmt.Errorf("invalid destination type has been passed [%s]", config.Type)
+func (t *ThreadEvent) Push(ctx context.Context, record types.RawRecord) error {
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("context closed")
+	default:
+		t.stats.readCount.Add(1)
```
shubham19may: "Non-blocking" doesn’t mean "no contention." If many goroutines hammer the same atomic variable, the CPU’s cache coherence protocol will serialize updates internally. You won’t see goroutines blocked in Go code, but performance may degrade due to heavy contention on that memory location.

utils/typeutils/reformat.go
```diff
@@ -133,6 +110,33 @@ func ReformatValue(dataType types.DataType, v any) (any, error) {
...
+			return true, nil
+		case "0", "f", "F", "false", "FALSE", "False", "NO", "No", "no":
+			return false, nil
+		}
+	case int, int16, int32, int64, int8:
+		switch booleanValue {
+		case 1:
+			return true, nil
+		case 0:
+			return true, nil
```
vaibhav-datazip: shouldn't this return false

destination/writers.go
shubham19may:
```suggestion
		go t.stats.readCount.Add(1)
```

destination/writers.go
```diff
@@ -66,173 +73,173 @@ func WithBackfill(backfill bool) ThreadOptions {
...
-			}
-			close(errChan)
-		}()
-		// init writer first
-		if err := initNewWriter(); err != nil {
-			return fmt.Errorf("failed to initialize writer: %s", err)
-		}
+	return &ThreadEvent{
+		buffer:         []types.RawRecord{},
+		batchSize:      10000,
```
shubham19may, line 186: have we checked tweaking the batchsize here? (performance impact)

types/adapter.go
vaibhav-datazip: can we mention these as constants ?

destination/writers.go
shubham19may: confusing actually. whats the ideal difference between threadEvent and writer?

drivers/abstract/backfill.go
```diff
@@ -70,23 +69,23 @@ func (a *AbstractDriver) Backfill(ctx context.Context, backfilledStreams chan st
 						err = cursorErr
 						return
 					}
-					if typeutils.Compare(maxPrimaryCursorValue, prevPrimaryCursor) == 1 {
+					if utils.CompareInterfaceValue(maxPrimaryCursorValue, prevPrimaryCursor) == 1 {
```
shubham19may: this doesn't compare time.go struct.

drivers/abstract/abstract.go
```diff
@@ -92,14 +92,20 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
 		convStream.WithSyncMode(types.FULLREFRESH, types.INCREMENTAL)
 		convStream.SyncMode = types.FULLREFRESH
 
+		// add default columns
+		for column, typ := range DefaultColumns {
+			convStream.UpsertField(column, typ, true)
+		}
+
 		// Add CDC columns if supported
```
vaibhav-datazip: This comment can be changed

destination/parquet/parquet.go
```diff
@@ -97,77 +96,75 @@ func (p *Parquet) createNewPartitionFile(basePath string) error {
 
 	writer := func() any {
 		if p.stream.NormalizationEnabled() {
-			return pqgo.NewGenericWriter[any](pqFile, p.stream.Schema().ToParquet(), pqgo.Compression(&pqgo.Snappy))
+			return pqgo.NewGenericWriter[any](pqFile, p.stream.Schema().ToParquet(), pqgo.Compression(&pqgo.Snappy), pqgo.MaxRowsPerRowGroup(10000))
```
shubham19may: any reason?

shubham19may: this will create many small rowgroups

destination/parquet/parquet.go
```diff
@@ -215,107 +212,132 @@ func (p *Parquet) Check(_ context.Context) error {
...
-				// Remove local file after successful upload
-				removeLocalFile(filePath, "uploaded to S3", fileMetadata.recordCount)
-				logger.Infof("Successfully uploaded file to S3: s3://%s/%s", p.config.Bucket, s3KeyPath)
-			}
-			return nil
-		})
+	schemaChange := false
+	for idx, record := range records {
+		// add common fields
+		records[idx].Data[constants.OlakeID] = record.OlakeID
```
shubham19may, line 313: here we are reusing the data field of RawRecords struct. rawrecord itself is being used again to store the flattened values. can we use separate map or something. its just confusing.

destination/parquet/parquet.go
```diff
@@ -215,107 +212,132 @@ func (p *Parquet) Check(_ context.Context) error {
 	return nil
 }
 
-func (p *Parquet) Close(ctx context.Context) error {
-	removeLocalFile := func(filePath, reason string, recordCount int) {
+func (p *Parquet) Close(_ context.Context) error {
+	removeLocalFile := func(filePath, reason string) {
```
shubham19may, line 234: this can upload duplicate data as well. 

ex, if the other threads are closed because of some source read error, this close will be called and files will be uploaded to s3.

destination/parquet/parquet.go
```diff
@@ -97,77 +96,75 @@ func (p *Parquet) createNewPartitionFile(basePath string) error {
...
+			return fmt.Errorf("failed to create partition file for path[%s]", partitionedPath)
 		}
-		partitionFolder = p.partitionedFiles[partitionedPath]
-	}
 
-	if len(partitionFolder) == 0 {
-		return fmt.Errorf("failed to get partitioned files")
+		var err error
+		if p.stream.NormalizationEnabled() {
+			_, err = partitionFile.writer.(*pqgo.GenericWriter[any]).Write([]any{record.Data})
```
shubham19may, line 173: what if the file-size gets > 512?

destination/parquet/parquet.go
```diff
@@ -215,107 +212,132 @@ func (p *Parquet) Check(_ context.Context) error {
...
+	newFieldSchema, ok := newSchema.(typeutils.Fields)
+	if !ok {
+		return fmt.Errorf("failed to typecast new field schema[%T] into (typeutils.Fields)", newFieldSchema)
 	}
+	p.stream.Schema().Override(newFieldSchema.ToProperties())
 
-	return nil
+	// TODO: can we implement something https://github.com/parquet-go/parquet-go?tab=readme-ov-file#evolving-parquet-schemas-parquetconvert
+	// close prev files as change detected (new files will be created with new schema)
+	return p.Close(ctx)
```
shubham19may: ideally close is only called in case of error or finish. this call is against the princinpal.

shubham19may: previous way was better

shubham19may: one behavioural change here is, it will now close all partitions even if there was change detected in one partition.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
```diff
@@ -107,16 +107,16 @@ public static void main(String[] args) throws Exception {
         keySerde.configure(Collections.emptyMap(), true);
         keyDeserializer = keySerde.deserializer();
 
-        OlakeRowsIngester ori = new OlakeRowsIngester(upsert_records, stringConfigMap.get("table-namespace"), icebergCatalog, partitionTransforms, createIdFields);
+        OlakeRowsIngester ori = new OlakeRowsIngester(upsert_records, stringConfigMap.get("table-namespace"), icebergCatalog, partitionTransforms);
```
shubham19may, line 110: commenting just for my own understanding on how createdIdFields flag is passed now?

destination/iceberg/proto/records_ingest.pb.go
```diff
@@ -111,80 +181,473 @@ func (x *RecordIngestResponse) GetResult() string {
...
+	if x != nil {
+		return x.Success
+	}
+	return false
+}
+
+type IcebergPayload_Metadata struct {
+	state         protoimpl.MessageState        `protogen:"open.v1"`
+	DestTableName string                        `protobuf:"bytes,1,opt,name=dest_table_name,json=destTableName,proto3" json:"dest_table_name,omitempty"`
+	ThreadId      string                        `protobuf:"bytes,2,opt,name=thread_id,json=threadId,proto3" json:"thread_id,omitempty"`
```
shubham19may: why do we need this?

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
+			Type: proto.IcebergPayload_GET_OR_CREATE_TABLE,
+			Metadata: &proto.IcebergPayload_Metadata{
+				Schema:        iceSchema,
+				DestTableName: i.stream.Name(),
+				ThreadId:      i.server.serverID,
+				PrimaryKey:    &primaryKey,
+			},
+		}
+
+		resp, err := i.server.sendClientRequest(ctx, &requestPayload)
```
shubham19may: what if two writer threads send this message at once?

types/data_types.go
```diff
@@ -259,3 +107,22 @@ func (d DataType) ToNewParquet() parquet.Node {
...
+		return "boolean"
+	case Int32:
+		return "int"
+	case Int64:
+		return "long"
+	case Float32:
+		return "float"
+	case Float64:
+		return "double"
+	case Timestamp, TimestampMilli, TimestampMicro, TimestampNano:
```
shubham19may, line 123: Screenshot 2025-08-21 at 3 49 14 AM
eventually someone might raise it, iceberg has all data types for timestamp but we are only using timestampz with utc.

we should map date --> date.
time to time
timestamp to timestamp
and only 
timestampz to timestampz

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
+	if createOrLoadSchema {
+		var requestPayload proto.IcebergPayload
+		iceSchema := utils.Ternary(stream.NormalizationEnabled(), stream.Schema().ToIceberg(), icebergRawSchema()).([]*proto.IcebergPayload_SchemaField)
+		requestPayload = proto.IcebergPayload{
+			Type: proto.IcebergPayload_GET_OR_CREATE_TABLE,
+			Metadata: &proto.IcebergPayload_Metadata{
+				Schema:        iceSchema,
+				DestTableName: i.stream.Name(),
+				ThreadId:      i.server.serverID,
+				PrimaryKey:    &primaryKey,
```
shubham19may: commenting for my own understanding, how is it working in java. if its empty string, then are we not creating the identifier key?

destination/iceberg/iceberg.go
```diff
@@ -219,6 +503,76 @@ func (i *Iceberg) DropStreams(_ context.Context, _ []string) error {
...
+		if len(parts) < 3 {
+			continue
+		}
+
+		name := strings.TrimSpace(parts[1])
+		typeInfo := strings.TrimSpace(parts[2])
+
+		// typeInfo will contain `required type (id)` or `optional type`
+		types := strings.Split(typeInfo, " ")
+		fields[name] = types[1]
```
shubham19may, line 603: what about `required type (id)`

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
 
-	// Try to send a test message
-	req := &proto.RecordIngestRequest{
-		Messages: []string{getTestDebeziumRecord(i.threadID)},
+	// try to create table
+	req := &proto.IcebergPayload{
+		Type: proto.IcebergPayload_GET_OR_CREATE_TABLE,
+		Metadata: &proto.IcebergPayload_Metadata{
+			ThreadId:      server.serverID,
+			DestTableName: "test_olake",
```
shubham19may, line 280: can we keep the name and schema of the table same `olake_test_table` as it will create multiple test table for old folks?

destination/iceberg/iceberg.go
```diff
@@ -219,6 +503,76 @@ func (i *Iceberg) DropStreams(_ context.Context, _ []string) error {
...
+		// typeInfo will contain `required type (id)` or `optional type`
+		types := strings.Split(typeInfo, " ")
+		fields[name] = types[1]
+	}
+	return fields, nil
+}
+
+func rawDataColumnBuffer(record types.RawRecord, protoSchema []*proto.IcebergPayload_SchemaField) ([]*proto.IcebergPayload_IceRecord_FieldValue, error) {
+	dataMap := make(map[string]*proto.IcebergPayload_IceRecord_FieldValue)
+	protoColumns := make([]*proto.IcebergPayload_IceRecord_FieldValue, 0, len(protoSchema))
```
shubham19may:
```suggestion
	protoColumnValues := make([]*proto.IcebergPayload_IceRecord_FieldValue, 0, len(protoSchema))
```

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
+				detectedType := typeutils.TypeFromValue(value)
+
+				if detectedType == types.Null {
+					// remove element from data if it is null
+					delete(record.Data, key)
+					continue
+				}
+
+				detecteIceType := detectedType.ToIceberg()
+				if persistedType, exists := newSchema[key]; exists && !isValidIcebergType(persistedType, detecteIceType) {
```
shubham19may: change name of the variable. persistedType means its already present in the iceberg table.

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
+	return nil
 }
 
-func (i *Iceberg) Flattener() destination.FlattenFunction {
-	flattener := typeutils.NewFlattener()
-	return flattener.Flatten
+func isValidIcebergType(oldType, newType string) bool {
+	// Note: not added decimal precision as we not iceberg decimal type support as of now
+	switch oldType {
+	case "int":
```
shubham19may: what if in the same batch int, long, int, long type records come?

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
-		err := i.flushLocalBuffer(ctx, i.localBuffer)
-		if err != nil {
-			return fmt.Errorf("thread id %s: failed to flush buffer: %s", i.threadID, err)
+	if len(i.partitionInfo) > 0 {
+		// sort record based on partition order
+		sort.Slice(records, func(idx, jdx int) bool {
+			iRecord := records[idx].Data
+			jRecord := records[jdx].Data
+
+			for _, partition := range i.partitionInfo {
```
shubham19may: does this work for time.time columns?

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
-		logger.Infof("thread id %s: Batch flushed to Iceberg server for stream %s", i.threadID, i.stream.Name())
+
+		protoColumns := make([]*proto.IcebergPayload_IceRecord_FieldValue, 0, len(protoSchema))
+		if !i.stream.NormalizationEnabled() {
+			protoCols, err := rawDataColumnBuffer(record, protoSchema)
+			if err != nil {
+				return fmt.Errorf("failed to create raw data column buffer")
+			}
+			protoColumns = protoCols
+		} else {
```
shubham19may, line 132: where are the olake_timestamp columns?

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
+						return fmt.Errorf("failed to reformat rawValue[%v] of type[%T] as float32 value: %s", val, val, err)
+					}
+					protoColumns = append(protoColumns, &proto.IcebergPayload_IceRecord_FieldValue{Value: &proto.IcebergPayload_IceRecord_FieldValue_FloatValue{FloatValue: floatValue}})
+				case "double":
+					doubleValue, err := typeutils.ReformatFloat64(val)
+					if err != nil {
+						return fmt.Errorf("failed to reformat rawValue[%v] of type[%T] as float64 value: %s", val, val, err)
+					}
+					protoColumns = append(protoColumns, &proto.IcebergPayload_IceRecord_FieldValue{Value: &proto.IcebergPayload_IceRecord_FieldValue_DoubleValue{DoubleValue: doubleValue}})
+				case "timestamptz":
```
ImDoubD-datazip, line 170: we need to add case for timestamp as well

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
 	// skip flushing on error
 	defer func() {
-		err := i.CloseIcebergClient()
+		if i.server == nil {
+			return
+		}
+		err := i.server.closeIcebergClient(i.server)
 		if err != nil {
-			logger.Errorf("thread id %s: Error closing Iceberg client: %s", i.threadID, err)
+			logger.Errorf("thread id %s: Error closing Iceberg client: %s", i.server.serverID, err)
```
ImDoubD-datazip: why are we not returning this error?

destination/iceberg/iceberg.go
ImDoubD-datazip: if it fails due to some reason for eg. network blips or java out-of-memory, so we can add a retry here as well

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
-	if len(i.localBuffer.records) > 0 {
-		err := i.flushLocalBuffer(ctx, i.localBuffer)
-		if err != nil {
-			logger.Errorf("thread id %s: Error flushing buffer on close: %s", i.threadID, err)
-			return err
-		}
-	}
-
 	// Send commit request for this thread using a special message format
 	ctx, cancel := context.WithTimeout(ctx, 1000*time.Second)
```
ImDoubD-datazip: how we have decided the timeouts anywhere?

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
+	server, err := newIcebergClient(i.config, i.partitionInfo, false, upsertMode)
+	if err != nil {
+		return nil, fmt.Errorf("failed to start iceberg server: %s", err)
+	}
+
+	// persist server details
+	i.server = server
+
+	// check for identifier fields setting
+	primaryKey := utils.Ternary(i.config.NoIdentifierFields, "", constants.OlakeID).(string)
```
vaibhav-datazip: can this be renamed to identifierFields, instead of primaryKey

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/SchemaConvertor.java
```diff
@@ -0,0 +1,178 @@
...
+              case LONG:
+                  return value.getLongValue();
+              case FLOAT:
+                  return value.getFloatValue();
+              case DOUBLE:
+                  return value.getDoubleValue();
+              case BOOLEAN:
+                  return value.getBoolValue();
+              case STRING,UUID:
+                  return value.getStringValue();
```
shubham19may: previous one was 
```
      case UUID:
        val = node.isValueNode() ? UUID.fromString(node.asText(null)) : UUID.fromString(node.toString());
        break;
```

shubham19may: can we use the previous code as is : 
```
private static Object jsonValToIcebergVal(Types.NestedField field, JsonNode node) {
    LOGGER.debug("Processing Field:{} Type:{}", field.name(), field.type());
    final Object val;
    switch (field.type().typeId()) {
      case INTEGER: // int 4 bytes
        val = node.isNull() ? null : node.asInt();
        break;
      case LONG: // long 8 bytes
        val = node.isNull() ? null : node.asLong();
        break;
      case FLOAT: // float is represented in 32 bits,
        val = node.isNull() ? null : node.floatValue();
        break;
      case DOUBLE: // double is represented in 64 bits
        val = node.isNull() ? null : node.asDouble();
        break;
      case BOOLEAN:
        val = node.isNull() ? null : node.asBoolean();
        break;
      case STRING:
        // if the node is not a value node (method isValueNode returns false), convert it to string.
        val = node.isValueNode() ? node.asText(null) : node.toString();
        break;
      case UUID:
        val = node.isValueNode() ? UUID.fromString(node.asText(null)) : UUID.fromString(node.toString());
        break;
      case TIMESTAMP:
        if ((node.isLong() || node.isNumber()) && TS_MS_FIELDS.contains(field.name())) {
          val = OffsetDateTime.ofInstant(Instant.ofEpochMilli(node.longValue()), ZoneOffset.UTC);
        } else if (node.isTextual()) {
          val = OffsetDateTime.parse(node.asText());
        } else {
          throw new RuntimeException("Failed to convert timestamp value, field: " + field.name() + " value: " + node);
        }
        break;
      case BINARY:
        try {
          val = node.isNull() ? null : ByteBuffer.wrap(node.binaryValue());
        } catch (IOException e) {
          throw new RuntimeException("Failed to convert binary value to iceberg value, field: " + field.name(), e);
        }
        break;
      case LIST:
        Types.NestedField listItemsType = field.type().asListType().fields().get(0);
        // recursive value mapping when list elements are nested type
        if (listItemsType.type().isNestedType()) {
          ArrayList<Object> listVal = new ArrayList<>();
          node.elements().forEachRemaining(element -> {
            listVal.add(jsonValToIcebergVal(field.type().asListType().fields().get(0), element));
          });
          val = listVal;
          break;
        }

        val = mapper.convertValue(node, ArrayList.class);
        break;
      case MAP:
        Type keyType = field.type().asMapType().keyType();
        Type valType = field.type().asMapType().valueType();
        if (keyType.isPrimitiveType() && valType.isPrimitiveType()) {
          val = mapper.convertValue(node, Map.class);
          break;
        }
        // convert complex/nested map value with recursion
        HashMap<Object, Object> mapVal = new HashMap<>();
        node.fields().forEachRemaining(f -> {
          if (valType.isStructType()) {
            mapVal.put(f.getKey(), convert(valType.asStructType(), f.getValue()));
          } else {
            mapVal.put(f.getKey(), f.getValue());
          }
        });
        val = mapVal;
        break;
      case STRUCT:
        // create it as struct, nested type
        // recursive call to get nested data/record
        val = convert(field.type().asStructType(), node);
        break;
      default:
        // default to String type
        // if the node is not a value node (method isValueNode returns false), convert it to string.
        val = node.isValueNode() ? node.asText(null) : node.toString();
        break;
    }

    return val;
  }
```

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/SchemaConvertor.java
```diff
@@ -0,0 +1,178 @@
...
+          fieldNameToIndexMap.put(fieldName, index); // Map field name → index
+      }
+      // Pre-size the array list to avoid resizing
+      List<RecordWrapper> result = new ArrayList<>(records.size());
+      
+      StructType tableFields = tableSchema.asStruct();
+      for (IceRecord data : records) {
+          result.add(convertRecord(upsert, fieldNameToIndexMap, data, tableFields));
+      }
+      return result;
```
shubham19may:
```suggestion
      return records.parallelStream().map(data -> convertRecord(upsert, fieldNameToIndexMap, data, tableSchema.asStruct())).toList();
```

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeRowsIngester.java
```diff
@@ -2,130 +2,108 @@
...
-                try {
-                    Table icebergTable = this.loadIcebergTable(TableIdentifier.of(icebergNamespace, tableEvents.getKey()), tableEvents.getValue().get(0));
-                    icebergTableOperator.addToTable(icebergTable, tableEvents.getValue());
-                } catch (Exception e) {
-                    String errorMessage = String.format("%s Failed to process table events for table: %s", requestId, tableEvents.getKey());
-                    LOGGER.error(errorMessage, e);
-                    throw e;
-                }
+            if (this.icebergTable == null) {
+                SchemaConvertor schemaConvertor = new SchemaConvertor(primaryKey, schemaMetadata);
```
shubham19may: race condition based on multiple initialisations.

destination/iceberg/iceberg.go
```diff
@@ -219,6 +503,76 @@ func (i *Iceberg) DropStreams(_ context.Context, _ []string) error {
...
+	return fields, nil
+}
+
+func rawDataColumnBuffer(record types.RawRecord, protoSchema []*proto.IcebergPayload_SchemaField) ([]*proto.IcebergPayload_IceRecord_FieldValue, error) {
+	dataMap := make(map[string]*proto.IcebergPayload_IceRecord_FieldValue)
+	protoColumns := make([]*proto.IcebergPayload_IceRecord_FieldValue, 0, len(protoSchema))
+
+	dataMap[constants.OlakeID] = &proto.IcebergPayload_IceRecord_FieldValue{Value: &proto.IcebergPayload_IceRecord_FieldValue_StringValue{StringValue: record.OlakeID}}
+	dataMap[constants.OpType] = &proto.IcebergPayload_IceRecord_FieldValue{Value: &proto.IcebergPayload_IceRecord_FieldValue_StringValue{StringValue: record.OperationType}}
+	dataMap[constants.OlakeTimestamp] = &proto.IcebergPayload_IceRecord_FieldValue{Value: &proto.IcebergPayload_IceRecord_FieldValue_LongValue{LongValue: time.Now().UTC().UnixMilli()}}
```
ImDoubD-datazip, line 614: can we use UnixMicro over here? Iceberg supports timestamp with micro precision. also in cdcTimestamp

destination/iceberg/java_client.go
```diff
@@ -0,0 +1,220 @@
...
+// getServerConfigJSON generates the JSON configuration for the Iceberg server
+func getServerConfigJSON(config *Config, partitionInfo []PartitionInfo, port int, upsert bool) ([]byte, error) {
+	// Create the server configuration map
+	serverConfig := map[string]interface{}{
+		"port":                     fmt.Sprintf("%d", port),
+		"warehouse":                config.IcebergS3Path,
+		"table-namespace":          config.IcebergDatabase,
+		"catalog-name":             "olake_iceberg",
+		"table-prefix":             "",
+		"create-identifier-fields": !config.NoIdentifierFields,
```
ImDoubD-datazip, line 41: here, if the identifierfields bool is false, then pass true,
or if true then pass false?

Do we really need this in config level, can't we just pass it true for unity case so handling this separately?

destination/iceberg/java_client.go
```diff
@@ -0,0 +1,220 @@
...
+
+	// Connect to gRPC server
+	conn, err := grpc.NewClient(fmt.Sprintf("%s:%s", config.ServerHost, strconv.Itoa(port)),
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithDefaultCallOptions(grpc.WaitForReady(true)))
+
+	if err != nil {
+		// If connection fails, clean up the process
+		if serverCmd != nil && serverCmd.Process != nil {
+			if killErr := serverCmd.Process.Kill(); killErr != nil {
```
ImDoubD-datazip, line 181: instead of killing it, can we retry

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -47,17 +36,9 @@ public class IcebergTableOperator {
...
-
-  // Map to track table references per thread
-  private final Map<String, Table> threadTables = new ConcurrentHashMap<>();
-
-  // Map to store completed WriteResult per thread for later commit
-  private final Map<String, List<WriteResult>> threadWriteResults = new ConcurrentHashMap<>();
+  BaseTaskWriter<Record> writer;
 
-  public IcebergTableOperator(boolean upsert_records, boolean createIdentifierFields) {
-    this.createIdentifierFields = createIdentifierFields;
```
shubham19may, line 60: why ? this we used to send it from the golang right?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/SchemaConvertor.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  public Operation cdcOpValue(String cdcOpField) {
+    return switch (cdcOpField) {
+      case "u" -> Operation.UPDATE;
+      case "d" -> Operation.DELETE;
+      case "r" -> Operation.READ;
+      case "c" -> Operation.INSERT;
+      case "i" -> Operation.INSERT;
+      default ->
```
vaibhav-datazip, line 143: unnecessary cases (i) can be removed from here

destination/iceberg/proto/records_ingest.pb.go
```diff
@@ -111,80 +181,473 @@ func (x *RecordIngestResponse) GetResult() string {
...
+	PrimaryKey    *string                       `protobuf:"bytes,3,opt,name=primary_key,json=primaryKey,proto3,oneof" json:"primary_key,omitempty"`
+	Schema        []*IcebergPayload_SchemaField `protobuf:"bytes,4,rep,name=schema,proto3" json:"schema,omitempty"`
+	unknownFields protoimpl.UnknownFields
+	sizeCache     protoimpl.SizeCache
+}
+
+func (x *IcebergPayload_Metadata) Reset() {
+	*x = IcebergPayload_Metadata{}
+	mi := &file_records_ingest_proto_msgTypes[2]
```
ImDoubD-datazip, line 206: ca we use proper names for this, its a bit hard to understand, ms, mi, x

destination/parquet/parquet.go
```diff
@@ -97,77 +96,75 @@ func (p *Parquet) createNewPartitionFile(basePath string) error {
...
+	return types.RawSchema, nil
 }
 
 // Write writes a record to the Parquet file.
-func (p *Parquet) Write(_ context.Context, record types.RawRecord) error {
-	partitionedPath := p.getPartitionedFilePath(record.Data, record.OlakeTimestamp)
+func (p *Parquet) Write(_ context.Context, _ any, records []types.RawRecord) error {
+	// TODO: use batch writing feature of pq writer
+	for _, record := range records {
+		record.OlakeTimestamp = time.Now().UTC()
```
ImDoubD-datazip, line 156: can we follow same convention for this as iceberg, to use Micro format

destination/parquet/parquet.go
```diff
@@ -215,107 +212,132 @@ func (p *Parquet) Check(_ context.Context) error {
...
+		// Close writers
+		var err error
+		if p.stream.NormalizationEnabled() {
+			err = parquetFile.writer.(*pqgo.GenericWriter[any]).Close()
+		} else {
+			err = parquetFile.writer.(*pqgo.GenericWriter[types.RawRecord]).Close()
+		}
+		if err != nil {
+			return fmt.Errorf("failed to close writer: %s", err)
+		}
```
ImDoubD-datazip, line 256: can we implement this one time and call it both here and during partition as well.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -83,63 +64,6 @@ public IcebergTableOperator(boolean upsert_records, boolean createIdentifierFiel
 
   @ConfigProperty(name = "debezium.sink.iceberg.upsert", defaultValue = "true")
   boolean upsert;
-
-  protected List<RecordConverter> deduplicateBatch(List<RecordConverter> events) {
```
shubham19may, line 87: ideally this should happen before records are added to pom based on cdc_timestamp

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/SchemaConvertor.java
shubham19may: it might not preserve the order which might be necessary for partitioning. But this can improve speed.

Once we will have to check.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
shubham19may: we need to do this in golang and we need to dedup only in case of upsert mode.

destination/iceberg/iceberg.go
```diff
@@ -46,163 +41,452 @@ func (i *Iceberg) Spec() any {
...
+		if record.Data == nil {
+			continue
 		}
-		logger.Infof("thread id %s: Batch flushed to Iceberg server for stream %s", i.threadID, i.stream.Name())
+
+		protoColumns := make([]*proto.IcebergPayload_IceRecord_FieldValue, 0, len(protoSchema))
+		if !i.stream.NormalizationEnabled() {
+			protoCols, err := rawDataColumnBuffer(record, protoSchema)
+			if err != nil {
+				return fmt.Errorf("failed to create raw data column buffer")
```
shubham19may:
```suggestion
				return fmt.Errorf("failed to create raw data column buffer: %s", err)
```

drivers/mongodb/go.mod
```diff
@@ -81,6 +81,4 @@ require (
 	golang.org/x/text v0.24.0 // indirect
 	golang.org/x/tools v0.30.0 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
-)
-
-replace github.com/datazip-inc/olake => ../../
```
shubham19may: this is not remved from Oracle driver

hash-data:
Add a todo for types (referring to current java implementaion)

destination/iceberg/iceberg.go
```diff
@@ -332,10 +348,41 @@ func (i *Iceberg) Type() string {
 }
 
 // validate schema change & evolution and removes null records
-func (i *Iceberg) FlattenAndCleanData(rawOldSchema any, records []types.RawRecord) (bool, any, error) {
+func (i *Iceberg) FlattenAndCleanData(records []types.RawRecord) (bool, []types.RawRecord, any, error) {
 	if !i.stream.NormalizationEnabled() {
```
shubham19may: we will have to dedup in case of normalisation off as well.

destination/writers.go
```diff
@@ -73,105 +77,97 @@ func WithBackfill(backfill bool) ThreadOptions {
...
+			schema: nil,
+		})
+	}
+
+	return pool, nil
 }
 
-func (w *WriterPool) NewWriter(ctx context.Context, stream types.StreamInterface, options ...ThreadOptions) (*ThreadEvent, error) {
-	w.stats.writerThreads.Add(1)
+func (w *WriterPool) AddRecordsToSync(count int64) {
```
shubham19may:
```suggestion
func (w *WriterPool) AddRecordsToSyncStats(count int64) {
```

types/state.go
```diff
@@ -253,13 +253,13 @@ func (s *State) RemoveChunk(stream *ConfiguredStream, chunk Chunk) int {
...
 
 func (s *State) MarshalJSON() ([]byte, error) {
-	if s.isZero() {
-		return json.Marshal(nil)
-	}
-
 	type Alias State
 	p := Alias(*s)
 
+	if s.isZero() {
```
shubham19may, line 259: any reason?

destination/iceberg/iceberg.go
```diff
@@ -123,7 +132,7 @@ func (i *Iceberg) Write(ctx context.Context, schema any, records []types.RawReco
 					return false // i > j
 				}
 
-				if cmp := utils.CompareInterfaceValue(iField, jField); cmp != 0 {
+				if cmp := typeutils.Compare(iField, jField); cmp != 0 {
```
shubham19may: Big bug : we need to convert it into the actual partition value before comparing. 
example : bucket transformation abc--> 5, xyz --> 3
so the order is xyz, abc

destination/iceberg/iceberg.go
```diff
@@ -22,8 +22,9 @@ type Iceberg struct {
 	options       *destination.Options
 	config        *Config
 	stream        types.StreamInterface
-	partitionInfo []PartitionInfo // ordered slice to preserve partition column order
-	server        *serverInstance // java server instance
+	partitionInfo []PartitionInfo   // ordered slice to preserve partition column order
+	server        *serverInstance   // java server instance
+	schema        map[string]string // schema for current thread associated with java writer (col -> type)
```
shubham19may, line 26: we dont need this. lets discuss tomorrow

destination/iceberg/iceberg.go
```diff
@@ -372,97 +421,125 @@ func (i *Iceberg) FlattenAndCleanData(rawOldSchema any, records []types.RawRecor
...
+	promoteEvolution, err := compareSchema(i.schema, mergedSchema)
 
-	return schemaChange, newSchema, nil
+	return promoteEvolution, records, mergedSchema, err
 }
 
-func (i *Iceberg) EvolveSchema(ctx context.Context, newSchema any) error {
-	newSchemaMap, ok := newSchema.(map[string]string)
+// compares with global schema and update schema in destination accordingly
+func (i *Iceberg) EvolveSchema(ctx context.Context, newRawSchema, globalSchema any) (any, error) {
```
shubham19may: i think the way we are writing changes, its pretty easy (Iceberg is designed in a non-failing way to write as long as we are writing compatible changes)

destination/iceberg/iceberg.go
```diff
@@ -46,163 +42,528 @@ func (i *Iceberg) Spec() any {
...
+					return true // i < j
+				} else if !jOk {
+					return false // i > j
+				}
+
+				if cmp := typeutils.Compare(iField, jField); cmp != 0 {
+					return cmp == -1 // i < j
+				}
+			}
+			return true // i == j
```
badalprasadsingh: Here it should be `false`. You should not be returning `true` in both of these cases: `i < j` and `i == j`. This last return is a case of `i>=j` and should return `false`.

it should be:
```
for _, partition := range i.partitionInfo {
iField, iOk := iRecord[partition.field]
jField, jOk := jRecord[partition.field]

if !iOk && !jOk {
	continue // i == j
} else if !iOk {
	return true // i < j
} else if !jOk {
	return false // i > j
}

if cmp := typeutils.Compare(iField, jField); cmp != 0 {
	return cmp == -1 // i < j
}
}
return false // i >= j
```

shubham19may (approved):

PR #452, opened 2025-08-15
feat: Add an optional "target_name" to stream config to use as the Iceberg table name

jabr:
# Description

Iceberg table names have more stringent requirements (lower-case, alphanum and underscore only) than some of our source tables/collections.

To handle this, we've added an optional "target_name" parameter to the streams.json's "selected_streams" entries. This `TargetName` is then used in the Debezium record (with the original stream name as default, if the override target is not set) that is sent to the Iceberg Java code.

## Type of change

# How Has This Been Tested?

Tested on our internal MongoDB to AWS S3Tables replication process.

### DIFF
--- a/destination/iceberg/iceberg.go
+++ b/destination/iceberg/iceberg.go
@@ -78,7 +78,7 @@ func (i *Iceberg) Setup(stream types.StreamInterface, options *destination.Optio
 func (i *Iceberg) Write(ctx context.Context, record types.RawRecord) error {
 	// Convert record to Debezium format with thread ID
 	// We are adding the thread ID to process the records from multiple threads in parallel and separately so that we can commit when each thread finishes.
-	debeziumRecord, err := record.ToDebeziumFormat(i.config.IcebergDatabase, i.stream.Name(), i.stream.NormalizationEnabled(), i.threadID)
+	debeziumRecord, err := record.ToDebeziumFormat(i.config.IcebergDatabase, i.stream.TargetName(), i.stream.NormalizationEnabled(), i.threadID)
 	if err != nil {
 		return fmt.Errorf("failed to convert record: %v", err)
 	}
@@ -92,7 +92,7 @@ func (i *Iceberg) Write(ctx context.Context, record types.RawRecord) error {
 		if err != nil {
 			return fmt.Errorf("thread id %s: failed to flush buffer: %s", i.threadID, err)
 		}
-		logger.Infof("thread id %s: Batch flushed to Iceberg server for stream %s", i.threadID, i.stream.Name())
+		logger.Infof("thread id %s: Batch flushed to Iceberg server for stream %s as %s", i.threadID, i.stream.Name(), i.stream.TargetName())
 	}
 
 	i.records.Add(1)
--- a/types/catalog.go
+++ b/types/catalog.go
@@ -42,6 +42,7 @@ type StreamMetadata struct {
 	ChunkColumn    string `json:"chunk_column,omitempty"`
 	PartitionRegex string `json:"partition_regex"`
 	StreamName     string `json:"stream_name"`
+	TargetName     string `json:"target_name,omitempty"`
 	AppendMode     bool   `json:"append_mode,omitempty"`
 	Normalization  bool   `json:"normalization"`
 	Filter         string `json:"filter,omitempty"`
--- a/types/interface.go
+++ b/types/interface.go
@@ -4,6 +4,7 @@ type StreamInterface interface {
 	ID() string
 	Self() *ConfiguredStream
 	Name() string
+	TargetName() string
 	Namespace() string
 	Schema() *TypeSchema
 	GetStream() *Stream
--- a/types/stream_configured.go
+++ b/types/stream_configured.go
@@ -37,6 +37,13 @@ func (s *ConfiguredStream) Name() string {
 	return s.Stream.Name
 }
 
+func (s *ConfiguredStream) TargetName() string {
+	if s.StreamMetadata.TargetName != "" {
+		return s.StreamMetadata.TargetName
+	}
+	return s.StreamMetadata.StreamName
+}
+
 func (s *ConfiguredStream) GetStream() *Stream {
 	return s.Stream
 }


jabr:
This is my first contribution to olake, and I'm still pretty new to the code. I'd be more than happy to make any requested adjustments to the approach here, if you'd like. This felt like the cleanest, most flexible approach to solving the problem, and we imagine other users have/will encounter this same problem.

I've also submitted my signed CLA.

dschaaff:
When testing this, we used a small script to modify the streams.json file. It lowercases the nameand removes invalid characters like -. With this in placed we were able to successfully replicate mongo collections with names such as `example_Foo-bar` to s3tables successfully.

```python
!/usr/bin/env python3
import json
import re

def normalize_stream_name(stream_name):
    """
    Normalize stream name to be a valid Iceberg table name:
    - Convert to lowercase
    - Replace hyphens with underscores
    - Ensure it starts with letter or underscore
    - Replace any other invalid characters with underscores
    """
    # Convert to lowercase
    normalized = stream_name.lower()

    # Replace hyphens with underscores
    normalized = normalized.replace('-', '_')

    # Replace any characters that aren't alphanumeric or underscore with underscore
    normalized = re.sub(r'[^a-z0-9_]', '_', normalized)

    # Ensure it starts with letter or underscore
    if normalized and not (normalized[0].isalpha() or normalized[0] == '_'):
        normalized = '_' + normalized

    # Remove consecutive underscores
    normalized = re.sub(r'_+', '_', normalized)

    return normalized

def modify_streams_json(input_file, output_file=None):
    """
    Modify streams.json to:
    1. Set normalization to true for all streams
    2. Add target_name field with normalized stream name for Iceberg compatibility
    """
    if output_file is None:
        output_file = input_file

    # Read the JSON file
    with open(input_file, 'r') as f:
        data = json.load(f)

    # Process each account's streams
    if 'selected_streams' in data:
        for account_key, streams in data['selected_streams'].items():
            for stream in streams:
                # Set normalization to true
                stream['normalization'] = True

                # Add normalized stream name as target_name
                if 'stream_name' in stream:
                    original_name = stream['stream_name']
                    normalized_name = normalize_stream_name(original_name)
                    stream['target_name'] = normalized_name
                    print(f"Added target_name: {original_name} -> {normalized_name}")

    # Write the modified JSON back to file
    with open(output_file, 'w') as f:
        json.dump(data, f, indent=2)

    print(f"Modified streams saved to: {output_file}")

if __name__ == "__main__":
    import sys

    if len(sys.argv) < 2:
        print("Usage: python modify_streams.py <input_file> [output_file]")
        sys.exit(1)

    input_file = sys.argv[1]
    output_file = sys.argv[2] if len(sys.argv) > 2 else None

    modify_streams_json(input_file, output_file)%
```

hash-data:
Hi @jabr ,
Thanks for raising the pr for it. Will review it.

rkhameshra:
Hey @jabr @dschaaff , 

Thanks a lot for raising this PR, this is a really useful addition to OLake!  

I had a couple of suggestions that can help us cover more ground and handle edge cases:  

1. **Default Normalisation**  
   Right now, the `TargetName` is set only if the user provides it. To enable a smooth run by default, it’d be great if OLake could automatically create a normalised `TargetName` for every new stream selected (instead of leaving it blank). This way users don’t need to think about it upfront, but can still override if needed.  

2. **Backward Compatibility**  
   For jobs already running in the system, it would help if the system assumes `TargetName = original table name` in case the `TargetName` variable is not found or found `NULL`. This way we don’t break existing syncs.  

Would love to hear your thoughts on this! 

Along with this @vikash390 from our end will be happy to collaborate with you on this and try to close this in highest priority.

jabr:
Apologies for the delay in getting back to this.

> Backward Compatibility
> For jobs already running in the system, it would help if the system assumes TargetName = original table name in case the TargetName variable is not found or found NULL. This way we don’t break existing syncs.

That is the behavior in this MR. If the target name option is not present or is blank, if will use the original name.

> Default Normalisation
> Right now, the TargetName is set only if the user provides it. To enable a smooth run by default, it’d be great if OLake could automatically create a normalised TargetName for every new stream selected (instead of leaving it blank). This way users don’t need to think about it upfront, but can still override if needed.

I definitely agree that would be a good default, but it's not clear what specific change you had in mind here. To maintain backwards compatibility, I think we'd need an explicit override of the behavior like I've done.

However, extending the discover process to generate configs with a normalized target name seems like a good approach for a smoother out-of-the-box experience for future use. I haven't looked at that portion of the code, but I think it would effectively be implementing the logic of Daniel's script above into the "discover" command.

But perhaps we could do that as a follow-up ticket?

jabr:
Also, I can rebase this branch, if you'd like, but I'm not clear if you want it rebased against `master` or `staging` branch.

shubham19may:
Hey @jabr thank you for submitting the PR,

Quick heads-up, we have internally almost finished developing the same feature. @vikash390 will raise a PR by Monday, release will happen by end of the next week. 

Main reason for this is, this will be developed on top of https://github.com/datazip-inc/olake/pull/439/files which is a big change (Refactoring main destination writer code). Also couple of other edge cases has to be handled in merging the streams.json. We can deliver a little faster for your convenience, if we do it internally as a lot of brainstorming has happened on this.

We will be keeping you posted.

jabr:
> Quick heads-up, we have internally almost finished developing the same feature. @vikash390 will raise a PR by Monday, release will happen by end of the next week.

No worries, but can we get some info on how it will work and when it might be expected?

We are about to start production deployment with our patch, but it will be good to get a picture of what changes we'll need to make to our system later to then get back on your upstream releases.

Thanks!

vikaxsh:
> > Quick heads-up, we have internally almost finished developing the same feature. @vikash390 will raise a PR by Monday, release will happen by end of the next week.
> 
> No worries, but can we get some info on how it will work and when it might be expected?
> 
> We are about to start production deployment with our patch, but it will be good to get a picture of what changes we'll need to make to our system later to then get back on your upstream releases.
> 
> Thanks!

Hi,

We've added support for target_database, target_table, and target_column_name in streams.json.  will be normalized by default and won’t be editable.
if a target_table or target_database is defined, we’ll use that; otherwise, we’ll fall back to the source names. If you have any specific naming requirements, happy to discuss!

Thanks!

vikaxsh:
Hi,

We’ve released v0.2.0, which adds automatic normalization for destination names. In `streams.json`, you’ll now see:

- **destination_database** → you can manually configure the database name in `streams.json`.

The destination database name is stored in the format `<driver>_<source_db_name>:<source_namespace>`. (`<driver>:<source_db_name>` if databases don't have a namespace, e.g., MySQL or MongoDB)

By default, the driver name is used as the prefix. You can override it by using the `--destination-database-prefix` flag during the discover command.

During usage, this value is normalized by replacing special characters, including the colon (:), with underscores (_).

**Example:**

  - Source DB → `sales-db`
  - Namespace → `us-east`
  - Driver → `postgres` (default prefix, unless manually changed or overridden)
  - Stored in `streams.json` → `postgres_sales_db:us-east`
  - Final normalized name → `postgres_sales_db_us_east`

- **destination_table** → Normalized table name derived from the source stream name, with special characters replaced by underscores (_).
- **destination_column_name** → Normalized column names derived from source columns, with special characters replaced by underscores (_).

Thanks!

PR #475, opened 2025-09-03
fix(mysql): allow sync of empty tables with 0 records

Sanjay-Kirti:
## Problem
MySQL driver sync fails for tables with 0 records.

## Solution  
- Modified GetOrSplitChunks() to distinguish empty tables from missing stats
- Empty tables now succeed with empty chunk set
- Missing stats still fail with clear error message

## Files Changed
- drivers/mysql/internal/backfill.go

### DIFF
--- a/drivers/mysql/internal/backfill.go
+++ b/drivers/mysql/internal/backfill.go
@@ -59,10 +59,24 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 	var avgRowSize any
 	approxRowCountQuery := jdbc.MySQLTableRowStatsQuery()
 	err := m.client.QueryRow(approxRowCountQuery, stream.Name()).Scan(&approxRowCount, &avgRowSize)
-	if err != nil || avgRowSize == nil {
-		errorMsg := utils.Ternary(err != nil, fmt.Errorf("failed to get approx row count and avg row size: %s", err), fmt.Errorf("either stats not populated for table[%s] or the table contains 0 records. (to populate stats run ANALYZE TABLE query)", stream.ID()))
-		return nil, errorMsg.(error)
+	
+	// Handle different error scenarios
+	if err != nil {
+		return nil, fmt.Errorf("failed to get approx row count and avg row size: %s", err)
 	}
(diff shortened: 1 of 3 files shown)


drivers/mysql/internal/backfill.go
```diff
@@ -59,10 +59,24 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get approx row count and avg row size: %s", err)
+	}
+	
+	// Check if table has 0 records (valid case - should succeed)
+	if approxRowCount == 0 {
+		logger.Infof("Table %s contains 0 records, creating empty chunk set", stream.ID())
+		pool.AddRecordsToSync(0)
+		return types.NewSet[types.Chunk](), nil
+	}
```
vaibhav-datazip: Even when the table statistics are not populated in MySQL, approxRowCount is returned as 0. Therefore, the averageRowSize = nil block will never be triggered.

drivers/mysql/internal/backfill.go
```diff
@@ -59,10 +59,24 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
 	approxRowCountQuery := jdbc.MySQLTableRowStatsQuery()
 	err := m.client.QueryRow(approxRowCountQuery, stream.Name()).Scan(&approxRowCount, &avgRowSize)
-	if err != nil || avgRowSize == nil {
-		errorMsg := utils.Ternary(err != nil, fmt.Errorf("failed to get approx row count and avg row size: %s", err), fmt.Errorf("either stats not populated for table[%s] or the table contains 0 records. (to populate stats run ANALYZE TABLE query)", stream.ID()))
-		return nil, errorMsg.(error)
+	
+	// Handle different error scenarios
+	if err != nil {
+		return nil, fmt.Errorf("failed to get approx row count and avg row size: %s", err)
+	}
```
vaibhav-datazip: You can remove the unnecessary blank line here, since the error handling directly corresponds to the query above.

PR #484, opened 2025-09-04
fix: use cgroup v2 memory limits for batch size calculation

dschaaff:
# Description

Update DetermineSystemMemoryGB to check cgroup v2 memory limits before falling back to system memory detection. This ensures proper memory-aware batch sizing in containerized environments like Docker and Kubernetes. This solves it always sizing based on the underlying host and not the actual memory available to the container.

## Type of change

### DIFF
--- a/utils/memory.go
+++ b/utils/memory.go
@@ -1,6 +1,9 @@
 package utils
 
 import (
+	"errors"
+	"math"
+	"os"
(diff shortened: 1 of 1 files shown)


shubham19may:
After refactoring, this function is removed `DetermineSystemMemoryGB`. So closing this.

PR #488, opened 2025-09-05
fix: environment not passed to icerberg jar breaking IRSA in k8s

dschaaff:
IRSA relies on the environment variables AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE. We should be passing these to the child process to allow working in Kubernetes environments without using hardcoded credentials. This also enables other vars such as JAVA_TOOL_OPTS to be used as needed.

# Description

Without this patch it is impossible to use Olake in Kubernetes with AWS IAM roles for service accounts. Using IAM roles in this fashion is the recommended way to provide temporary AWS credentials inside EKS clusters. 

# How Has This Been Tested?

I am running this patch successfully inside our EKS clusters. The icerberg writer successfully finds and uses the IRSA credentials.

### DIFF
--- a/destination/iceberg/java_client.go
+++ b/destination/iceberg/java_client.go
@@ -151,19 +151,76 @@ func newIcebergClient(config *Config, partitionInfo []PartitionInfo, threadID st
 		serverCmd = exec.Command("java", "-XX:+UseG1GC", "-jar", config.JarPath, string(configJSON))
 	}
 
-	// Get current environment
-	serverCmd.Env = utils.Ternary(serverCmd.Env == nil, []string{}, serverCmd.Env).([]string)
+	// Create a map to hold the final environment variables for the subcommand.
+	envMap := make(map[string]string)
 
-	addEnvIfSet := func(key, value string) {
-		if value != "" {
-			serverCmd.Env = append(serverCmd.Env, fmt.Sprintf("%s=%s", key, value))
+	// Define allowed environment variable prefixes for security
+	allowedEnvPrefixes := []string{
+		"AWS_",        // AWS credentials and configuration
+		"JAVA_",       // Java runtime options
+		"PATH",        // System PATH
+		"HOME",        // User home directory
+		"USER",        // Current user
+		"TZ",          // Timezone
+		"LANG",        // Language settings
+		"LC_",         // Locale settings
+		"SSL_CERT_",   // SSL certificate paths
+		"CA_BUNDLE",   // Certificate authority bundle
+		"HTTPS_PROXY", // Proxy settings
+		"HTTP_PROXY",  // Proxy settings
+		"NO_PROXY",    // Proxy exclusions
+	}
+
+	// Filter and inherit parent environment variables
+	for _, envVar := range os.Environ() {
+		parts := strings.SplitN(envVar, "=", 2)
+		if len(parts) >= 1 {
+			envKey := parts[0]
+			envValue := ""
+			if len(parts) == 2 {
+				envValue = parts[1]
+			}
+
+			// Check if this environment variable is allowed
+			allowed := false
+			for _, prefix := range allowedEnvPrefixes {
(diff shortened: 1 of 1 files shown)


shubham19may:
hey @dschaaff thank you for raising this PR. this is helpful. in OLake project, PRs are raised to staging (then we do a weekly releases after staging to master). The destination refactoring that we talked about in last PR is merged into staging yesterday. Can you please validate the fix there and raise PR to staging rather than master.

https://github.com/datazip-inc/olake/actions/runs/17500329942/job/49712935207?pr=488

dschaaff:
@shubham19may Thanks for the feedback. I rebased off the staging branch and modified the implementation.

destination/iceberg/java_client.go
```diff
@@ -151,19 +151,76 @@ func newIcebergClient(config *Config, partitionInfo []PartitionInfo, threadID st
...
-	// Get current environment
-	serverCmd.Env = utils.Ternary(serverCmd.Env == nil, []string{}, serverCmd.Env).([]string)
+	// Create a map to hold the final environment variables for the subcommand.
+	envMap := make(map[string]string)
 
-	addEnvIfSet := func(key, value string) {
-		if value != "" {
-			serverCmd.Env = append(serverCmd.Env, fmt.Sprintf("%s=%s", key, value))
+	// Define allowed environment variable prefixes for security
+	allowedEnvPrefixes := []string{
```
shubham19may: Better approach : inherit the full parent environment and only override known keys from config when explicitly set. This keeps IRSA working, allows JAVA_TOOL_OPTIONS, and avoids breaking GCS/other providers.

example : 
```
// Inherit full parent environment so IRSA and other vars flow through
serverCmd.Env = os.Environ()

// addEnvIfSet sets or replaces an env var in serverCmd.Env if value is non-empty.
addEnvIfSet := func(key, value string) {
	if value == "" {
		return
	}
	prefix := key + "="
	for i := range serverCmd.Env {
		if strings.HasPrefix(serverCmd.Env[i], prefix) {
			serverCmd.Env[i] = prefix + value
			return
		}
	}
	serverCmd.Env = append(serverCmd.Env, prefix+value)
}

addEnvIfSet("AWS_ACCESS_KEY_ID",    config.AccessKey)
addEnvIfSet("AWS_SECRET_ACCESS_KEY", config.SecretKey)
addEnvIfSet("AWS_REGION",           config.Region)
addEnvIfSet("AWS_SESSION_TOKEN",    config.SessionToken)
addEnvIfSet("AWS_PROFILE",          config.ProfileName)
```

please see if it works and update the PR.

dschaaff: Thanks, I've adjusted the implementation. Its much simpler that way.

destination/iceberg/java_client.go
```diff
@@ -152,7 +152,7 @@ func newIcebergClient(config *Config, partitionInfo []PartitionInfo, threadID st
 	}
 
 	// Get current environment
-	serverCmd.Env = utils.Ternary(serverCmd.Env == nil, []string{}, serverCmd.Env).([]string)
+	serverCmd.Env = os.Environ()
 
 	addEnvIfSet := func(key, value string) {
```
shubham19may, line 157:
```
// addEnvIfSet sets or replaces an env var in serverCmd.Env if value is non-empty.
addEnvIfSet := func(key, value string) {
	if value == "" {
		return
	}
	prefix := key + "="
	for i := range serverCmd.Env {
		if strings.HasPrefix(serverCmd.Env[i], prefix) {
			serverCmd.Env[i] = prefix + value
			return
		}
	}
	serverCmd.Env = append(serverCmd.Env, prefix+value)
}
```

functions has to be changed, otherwise it will add duplicates (from config and if someone has set them in env prop)

shubham19may (approved):

PR #494, opened 2025-09-08
fix: segment removed, telemetry events directly sent to mixpanel

ImDoubD-datazip:
# Description
events sent directly to mixpanel, dependency on segment removed.

## Type of change

- [ X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
Discover and Sync CLI events were sent to mixpanel through cloudfare proxy.

### DIFF
--- a/utils/telemetry/telemetry.go
+++ b/utils/telemetry/telemetry.go
@@ -161,16 +161,22 @@ func (t *Telemetry) sendEvent(eventName string, props map[string]interface{}) er
 	if props == nil {
 		props = make(map[string]interface{})
 	}
-	props["os"] = t.platform.OS
-	props["arch"] = t.platform.Arch
-	props["olake_version"] = t.platform.OlakeVersion
-	props["num_cpu"] = t.platform.DeviceCPU
-	props["service"] = t.serviceName
-	props["ip_address"] = t.ipAddress
-	props["location"] = t.locationInfo
-	props["distinct_id"] = t.userID
-	props["time"] = time.Now().Unix()
-	props["event_original_name"] = eventName
+	properties := map[string]interface{}{
+		"os":                  t.platform.OS,
+		"arch":                t.platform.Arch,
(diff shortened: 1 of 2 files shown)


utils/telemetry/telemetry.go
```diff
@@ -19,19 +19,18 @@ import (
 	"github.com/datazip-inc/olake/types"
 	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/logger"
-	analytics "github.com/segmentio/analytics-go/v3"
```
shubham19may, line 22: lets remove from go.mod

utils/telemetry/telemetry.go
```diff
@@ -99,7 +97,7 @@ func TrackDiscover(streamCount int, sourceType string) {
 		if telemetry == nil {
 			return
 		}
-		defer telemetry.client.Close()
+		// defer telemetry.client.Close()
```
shubham19may: any reason?

utils/telemetry/telemetry.go
```diff
@@ -156,35 +154,55 @@ func TrackSyncCompleted(status bool, records int64) {
...
 		return fmt.Errorf("telemetry client is nil")
 	}
 
 	// Add common properties
-	if properties == nil {
-		properties = make(map[string]interface{})
+	if props == nil {
+		props = make(map[string]interface{})
+	}
+	props["os"] = t.platform.OS
```
shubham19may: once check with merlyn that if the segment properties like event name or id are required or not.

utils/telemetry/telemetry.go
ImDoubD-datazip: this was segment io client, which provided close method

utils/telemetry/telemetry.go
ImDoubD-datazip: it is required, I am adding

shubham19may (dismissed):

PR #496, opened 2025-09-08
respect --state flag when writing olake sync state to file 

### LINKED ISSUE #495: --state flag is not respected and state is always written to config path
## Environment

- **Deploy method**: Docker
- **Olake version**:  0.1.11

## Description

When running olake sync with the --state flag, the state file is not written to the specified path. Instead, it is written to a directory derived from the --config or --destination file paths, with the filename state.json.

## Steps to reproduce

 1. Run the olake sync command with a full path for the state file:
```
olake sync --config /tmp/my-config/source.json \
  --catalog /tmp/my-config/catalog.json \
  --destination /tmp/my-config/destination.json \
   --state /var/lib/olake-state/my-state.json
```
 2. Observe the state file is written to /tmp/my-config/state.json.

## Expected behavior

Olake should write the state to the file specified in the --state flag.

## Actual behavior

Olake always writes the state file with the name state.json in path determined by the config files that are passed.

dschaaff:
# Description

When using `--state /path/to/file.json`, olake reads from the specified path but writes state to `<config_folder>/state.json` instead of the specified path. One question I have is on the intended behavior of the --no-save flag. I implemented this it respects that flag but I'm not completely clear on if I understood the purpose of that flag correctly.

Fixes #495 

## Type of change

# How Has This Been Tested?

I am running this patch in our environment in order to support saving the state file to a persistent volume which is mounted at a different path than the source/destination config files in kubernetes.


types/state.go
```diff
@@ -277,24 +282,61 @@ func (s *State) LogWithLock() {
...
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return fmt.Errorf("failed to create directory %s: %w", dir, err)
+	}
+
+	contentBytes, err := json.Marshal(s)
+	if err != nil {
+		return fmt.Errorf("failed to marshal state: %w", err)
+	}
+
+	file, err := os.OpenFile(filePath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
```
ImDoubD-datazip: instead of this extra step, we can directly use os.WriteFile. which I believe will do the same thing, including writing as well.

protocol/sync.go
```diff
@@ -55,6 +55,11 @@ var syncCmd = &cobra.Command{
 			}
 		}
 
+		state.NoSave = noSave
```
ImDoubD-datazip: we have this no-save flag already which does the same thing. please take a look at root.go. once if possible you can test it, present on current staging if it is working or not.

types/state.go
```diff
@@ -277,24 +282,61 @@ func (s *State) LogWithLock() {
...
+
 	// message := Message{
 	// 	Type:  StateMessage,
 	// 	State: s,
 	// }
 	// TODO: Only Log in logs file, not in CLI
 	// logger.Info(message)
 
-	// log to state file
-	err := logger.FileLogger(s, "state", ".json")
```
shubham19may, line 295: First of all, thanks for reporting this. FileLogger assumes that config folder is by default location to write state.json and streams.json. (if no-save is set, then config-path is tempDir logger.go line 172)

Better approach : we can add extra parameters in the filelogger (as a path), and then pass it from all the places it is getting called (basically from state.json, streams.json which can be provided as a custom path in config), if this is empty string it can by default take it as config path

We can get stream and state path from viper directly (before passing the fileLogger) and also check from viper only that if --no-save is on at any place in the code.

Screenshot 2025-09-09 at 5 14 55 PM

protocol/sync.go
```diff
@@ -55,6 +55,11 @@ var syncCmd = &cobra.Command{
...
+		if !noSave && statePath != "" {
```
shubham19may: no-save is already there.

types/state.go
dschaaff: @shubham19may Thanks for the great feedback. I've reworked the implementation based on your input.

protocol/sync.go
```diff
@@ -55,6 +57,11 @@ var syncCmd = &cobra.Command{
 			}
 		}
 
+		state.NoSave = noSave
```
shubham19may: no need to modify the actual state, as it can be set and accessed by viper in the state.go.

types/state.go
```diff
@@ -62,6 +64,7 @@ func (s *State) initStreamState(stream *ConfiguredStream) *StreamState {
 // State is a dto for airbyte state serialization
 type State struct {
 	*sync.RWMutex `json:"-"`
+	NoSave        bool           `json:"-"`
```
shubham19may: lets remove

types/state.go
```diff
@@ -284,17 +287,22 @@ func (s *State) LogState() {
 		return
 	}
 
+	if s.NoSave {
```
shubham19may: get this directly from viper.

utils/logger/logger.go
```diff
@@ -100,19 +100,34 @@ func LogRequest(req *http.Request) {
...
+		filePath := viper.GetString(constants.ConfigFolder)
+		if filePath == "" {
+			return fmt.Errorf("config folder is not set")
+		}
+		fullPath = filepath.Join(filePath, fileName+fileExtension)
 	}
-	// Construct the full file path
+
+	// Ensure directory exists for custom paths
+	dir := filepath.Dir(fullPath)
```
shubham19may: not needed. we are passing the full-path of file in custom path. so we can remove this.

dschaaff:
@shubham19may I brought the branch up to date with the 0.2.0 release this morning. I also updated my internal build to apply this change on 0.2.0. I can report its working well in my testing and the file path for --state is correclty respected. I appreciate the cleanup on it as well. Let me know if any other changes are required.

shubham19may:
Hey @dschaaff we will be merging and releasing this as a patch along with one another bug fix by tomorrow. 

Thank you for your patience.

PR #497, opened 2025-09-09
fix: type conversion in iceberg types

hash-data:
# Description
-> Adding hierarchy check on iceberg types

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


destination/iceberg/iceberg.go
```diff
@@ -516,6 +516,13 @@ func validIcebergType(oldType, newType string) bool {
 		return true
 	}
 
+	// check according to typecast tree
+	oldDT := types.IcebergTypeToDatatype(oldType)
```
shubham19may: what if the oldType is int and new type is float. THis is the new column and its the first batch. ideally it shouldn't break right?

hash-data: good catch, thanks

destination/iceberg/iceberg.go
```diff
@@ -395,17 +395,23 @@ func (i *Iceberg) FlattenAndCleanData(records []types.RawRecord) (bool, []types.
...
+							)
+						}
+
+						if promotionRequired(typeInNewSchema, detectedIcebergType) {
+							recordsSchema[key] = detectedIcebergType
+							diffThreadSchema = true
+						}
+					} else {
+						// if column not exist in iceberg table
+						recordsSchema[key] = getCommonAncestorType(typeInNewSchema, detectedIcebergType)
```
shubham19may, line 415: what if it got added by some other writer while this was happening? and now the type is float but this has type int?

shubham19may: shouldn't we mark diffThreadSchema = true?

destination/iceberg/iceberg.go
```diff
@@ -395,17 +395,23 @@ func (i *Iceberg) FlattenAndCleanData(records []types.RawRecord) (bool, []types.
...
-						return false, nil, fmt.Errorf(
-							"failed to validate schema for field[%s] (detected two different types in batch), expected type: %s, detected type: %s",
-							key, typeInNewSchema, detectedIcebergType,
-						)
-					}
-
-					if promotionRequired(typeInNewSchema, detectedIcebergType) {
-						recordsSchema[key] = detectedIcebergType
-						diffThreadSchema = true
+					// if column exist in iceberg table validate type
```
shubham19may, line 398: lets add a comment on (now this handles string --> float --> int cases as well, basically we don't break the sync for higher types)

types/data_types.go
```diff
@@ -126,3 +126,22 @@ func (d DataType) ToIceberg() string {
 		return "string"
 	}
 }
+
+func IcebergTypeToDatatype(d string) DataType {
```
shubham19may, line 130: isn't there a function which already does reverse conversion

PR #499, opened 2025-09-10
fix: remove SCN dependency in Oracle and ensure consistent cursor values during backfill

### LINKED ISSUE #450: Backfill chunks use separate transactions, risking inconsistent cursor values during incremental syncs

vaibhav-datazip:
# Description

- Removed SCN dependency from the Oracle driver. SCN is active only for ~15 minutes by default, and extending redo log retention increases database load.
- Fixed cursor value handling to ensure consistency. Previously, the cursor was not guaranteed to be stable, which could cause skipped records during incremental sync. Now, the initial maximum cursor is determined before chunking begins across all drivers.

Fixes #450

## Type of change

# How Has This Been Tested?

- [x] Incremental mode tests using all drivers
- [x] Incremental sync runs with different cursor field types:
  - [x] Timestamp (with and without timezone)
  - [x] String
  - [x] Boolean
  - [x] Int
  - [x] Float

## Documentation

### DIFF
--- a/drivers/mongodb/internal/backfill.go
+++ b/drivers/mongodb/internal/backfill.go
@@ -448,7 +448,7 @@ func (m *Mongo) buildFilter(stream types.StreamInterface) (bson.D, error) {
 
 	switch {
 	case len(filter.Conditions) == 0:
-		return bson.D{{Key: "$and", Value: allConditions}}, nil
+		return utils.Ternary(len(allConditions) == 0, bson.D{}, bson.D{{Key: "$and", Value: allConditions}}).(bson.D), nil
 	case len(filter.Conditions) == 1:
 		allConditions = append(allConditions, buildMongoCondition(filter.Conditions[0]))
 	case len(filter.Conditions) == 2:


drivers/abstract/incremental.go
```diff
@@ -77,8 +76,8 @@ func (a *AbstractDriver) Incremental(ctx context.Context, pool *destination.Writ
 
 					// set state (no comparison)
```
hash-data, line 95: TODO: make incremental streams to run parallel

drivers/abstract/backfill.go
```diff
@@ -65,21 +79,6 @@ func (a *AbstractDriver) Backfill(ctx context.Context, backfilledStreams chan st
...
-						return
-					}
-					if typeutils.Compare(maxPrimaryCursorValue, prevPrimaryCursor) == 1 {
-						a.state.SetCursor(stream.Self(), primaryCursor, a.reformatCursorValue(maxPrimaryCursorValue))
-					}
-					if typeutils.Compare(maxSecondaryCursorValue, prevSecondaryCursor) == 1 {
-						a.state.SetCursor(stream.Self(), secondaryCursor, a.reformatCursorValue(maxSecondaryCursorValue))
-					}
-				}
 			} else {
```
hash-data, line 66: we can remove logic in chunk iterator as well

drivers/abstract/incremental.go
```diff
@@ -135,11 +134,3 @@ func (a *AbstractDriver) getMaxIncrementCursorFromData(primaryCursor, secondaryC
 	}
 	return primaryCursorValue, secondaryCursorValue
 }
-
-// reformatCursorValue is used to make time format consistent in state (Removing timezone info)
```
hash-data, line 139: any reason for moving it ?

drivers/mongodb/internal/backfill.go
```diff
@@ -476,3 +476,55 @@ func isObjectID(ctx context.Context, collection *mongo.Collection) (bool, error)
...
+				return result["maxPrimaryCursor"], maxSecondaryCursor, nil
+			}
+		}
+		return result["maxPrimaryCursor"], nil, nil
+	}
+
+	if err := cursor.Err(); err != nil {
+		return nil, nil, fmt.Errorf("cursor error: %s", err)
+	}
+	return nil, nil, fmt.Errorf("no documents found")
```
hash-data: this will fail the sync on 0 records which is not expected.

pkg/jdbc/jdbc.go
```diff
@@ -626,3 +620,46 @@ func BuildIncrementalQuery(opts IncrementalConditionOptions) (string, []any, err
 
 	return incrementalQuery, queryArgs, nil
 }
+
+func MaxCursorSetter(ctx context.Context, opts IncrementalConditionOptions) (any, any, error) {
```
hash-data:
```suggestion
func GetMaxCursorValue(ctx context.Context, opts IncrementalConditionOptions) (any, any, error) {
```

pkg/jdbc/jdbc.go
```diff
@@ -626,3 +620,46 @@ func BuildIncrementalQuery(opts IncrementalConditionOptions) (string, []any, err
...
+	primaryCursor, secondaryCursor := opts.Stream.Cursor()
+	quotedTable := QuoteTable(opts.Stream.Namespace(), opts.Stream.Name(), opts.Driver)
+
+	filter, err := SQLFilter(opts.Stream, string(opts.Driver))
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to parse stream filter: %s", err)
+	}
+	filterClause := utils.Ternary(filter == "", "", " WHERE ("+filter+")").(string)
```
hash-data: use fmt.Sprintf

pkg/jdbc/jdbc.go
```diff
@@ -626,3 +620,46 @@ func BuildIncrementalQuery(opts IncrementalConditionOptions) (string, []any, err
...
+			return nil, nil, fmt.Errorf("failed to scan the cursor values: %s", err)
+		}
+		maxSecondaryCursorValue = bytesConverter(maxSecondaryCursorValue)
+	} else {
+		err = opts.Client.QueryRowContext(ctx, cursorValueQuery).Scan(&maxPrimaryCursorValue)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to scan primary cursor value: %s", err)
+		}
+	}
+	return bytesConverter(maxPrimaryCursorValue), maxSecondaryCursorValue, nil
```
hash-data, line 664: why byte convertor not used for secondary cursor value?

vaibhav-datazip: byte conversion of secondary cursor value is done inside the if block

drivers/abstract/incremental.go
vaibhav-datazip: reverted back

drivers/mongodb/internal/backfill.go
```diff
@@ -481,6 +481,58 @@ func isObjectID(ctx context.Context, collection *mongo.Collection) (bool, error)
...
+			Value: bson.D{{Key: "$max", Value: "$" + secondaryCursor}},
+		})
+	}
+
+	pipeline := mongo.Pipeline{
+		{{Key: "$match", Value: filter}},
+		{{Key: "$group", Value: groupStage}},
+	}
+
+	cursor, err := collection.Aggregate(ctx, pipeline)
```
hash-data: can we use simple fineOne function ?

drivers/mongodb/internal/backfill.go
```diff
@@ -481,6 +481,58 @@ func isObjectID(ctx context.Context, collection *mongo.Collection) (bool, error)
...
+			}
+		}
+		return result["maxPrimaryCursor"], nil, nil
+	}
+
+	if err := cursor.Err(); err != nil {
+		return nil, nil, fmt.Errorf("cursor error: %s", err)
+	}
+	logger.Warnf("No documents found for stream: %s", stream.ID())
+	return nil, nil, nil
```
```suggestion
	return nil, nil, cursor.Err()
```
add nil check on calling function

drivers/mysql/internal/backfill.go
```diff
@@ -188,3 +188,17 @@ func (m *MySQL) getTableExtremes(stream types.StreamInterface, pkColumns []strin
 	err = tx.QueryRow(query).Scan(&min, &max)
 	return min, max, err
 }
+
+func (m *MySQL) FetchMaxCursorValues(ctx context.Context, stream types.StreamInterface) (any, any, error) {
+	opts := jdbc.IncrementalConditionOptions{
```
hash-data: making incremental condition option is not feasible, as only thing required is driver and stream

pkg/jdbc/jdbc.go
```diff
@@ -626,3 +620,46 @@ func BuildIncrementalQuery(opts IncrementalConditionOptions) (string, []any, err
...
+func GetMaxCursorValues(ctx context.Context, opts IncrementalConditionOptions) (any, any, error) {
+	primaryCursor, secondaryCursor := opts.Stream.Cursor()
+	quotedTable := QuoteTable(opts.Stream.Namespace(), opts.Stream.Name(), opts.Driver)
+
+	filter, err := SQLFilter(opts.Stream, string(opts.Driver))
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to parse stream filter: %s", err)
+	}
+	filterClause := utils.Ternary(filter == "", "", fmt.Sprintf(" WHERE (%s)", filter)).(string)
+
```
hash-data, line 633: we are using '(' any reason?, are we able to add other sql terms in it or not like limit etc?

drivers/mongodb/internal/backfill.go
```diff
@@ -472,6 +472,46 @@ func isObjectID(ctx context.Context, collection *mongo.Collection) (bool, error)
...
+	filter, err := buildFilter(stream)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to build filter: %s", err)
+	}
+
+	findMaxCursor := func(maxCursor string) (any, error) {
+		if maxCursor == "" {
+			return nil, nil
+		}
+		opts := options.FindOne().SetSort(bson.D{{Key: maxCursor, Value: -1}})
```
hash-data: is there no way other then sort of data?

pkg/jdbc/jdbc.go
```diff
@@ -21,7 +21,7 @@ func QuoteIdentifier(identifier string, driver constants.DriverType) string {
 	case constants.MySQL:
 		return fmt.Sprintf("`%s`", identifier)
 	case constants.Postgres:
-		return fmt.Sprintf(`"%s"`, identifier)
+		return fmt.Sprintf("%q", identifier)
```
hash-data, line 24: any reason?

vaibhav-datazip: "%s" is same as %q, and we are doing similar thing in oracle.

this was left out earlier

drivers/mongodb/internal/backfill.go
vaibhav-datazip: can switch to the previous query which was being used in earlier commits .

pkg/jdbc/jdbc.go
hash-data: then in mysql ?

vaibhav-datazip: in mysql parameters need to be put in between `\` \`` whereas for postgres and oracle its " "

drivers/mongodb/internal/backfill.go
```diff
@@ -427,24 +427,41 @@ func buildMongoCondition(cond types.Condition) bson.D {
 }
 
 // buildFilter generates a BSON document for MongoDB
-func buildFilter(stream types.StreamInterface) (bson.D, error) {
+func (m *Mongo) buildFilter(stream types.StreamInterface) (bson.D, error) {
+	primaryCursor, secondaryCursor := stream.Cursor()
+	maxPrimaryCursorValue := m.state.GetCursor(stream.Self(), primaryCursor)
+	maxSecondaryCursorValue := m.state.GetCursor(stream.Self(), secondaryCursor)
```
hash-data: add comments here

drivers/mongodb/internal/backfill.go
```diff
@@ -427,24 +427,41 @@ func buildMongoCondition(cond types.Condition) bson.D {
...
-
-	switch {
-	case len(filter.Conditions) == 0:
+	if len(conditions) == 0 {
 		return bson.D{}, nil
-	case len(filter.Conditions) == 1:
-		return buildMongoCondition(filter.Conditions[0]), nil
-	default:
-		return bson.D{{Key: "$" + filter.LogicalOperator, Value: bson.A{buildMongoCondition(filter.Conditions[0]), buildMongoCondition(filter.Conditions[1])}}}, nil
 	}
```
hash-data: use switch case

drivers/mongodb/internal/backfill.go
```diff
@@ -472,6 +489,57 @@ func isObjectID(ctx context.Context, collection *mongo.Collection) (bool, error)
...
+			logger.Warnf("max secondary cursor value is nil for stream: %s", stream.ID())
+		}
+		return result.MaxPrimaryCursor, result.MaxSecondaryCursor, nil
+	}
+
+	if err := cursor.Err(); err != nil {
+		return nil, nil, fmt.Errorf("cursor error: %s", err)
+	}
+	logger.Warnf("No documents found for stream: %s", stream.ID())
+	return nil, nil, cursor.Err()
```
hash-data: warning you can add in abstract
```suggestion
	return nil, nil, cursor.Err()
```

pkg/jdbc/jdbc.go
hash-data: can you add a comment there

pkg/jdbc/jdbc.go
```diff
@@ -617,12 +615,97 @@ func BuildIncrementalQuery(opts IncrementalConditionOptions) (string, []any, err
...
+	}
+	if maxPrimaryCursorValue == nil {
+		logger.Warnf("max primary cursor value is nil for stream: %s", stream.ID())
+	}
+	return bytesConverter(maxPrimaryCursorValue), maxSecondaryCursorValue, nil
+}
+
+// FilterUpdater is used to update the filter for initial run of incremental sync during backfill.
+// This is to avoid dupliction of records, as cursor value is fetched before the chunk creation.
+func FilterUpdater(opts IncrementalConditionOptions, filter string) (string, []any, error) {
```
hash-data:
```suggestion
func ThresholdFilter(opts IncrementalConditionOptions, filter string) (string, []any, error) {
```

pkg/jdbc/jdbc.go
```diff
@@ -617,12 +615,97 @@ func BuildIncrementalQuery(opts IncrementalConditionOptions) (string, []any, err
...
+	if opts.Stream.GetSyncMode() != types.INCREMENTAL {
+		return filter, nil, nil
+	}
+	primaryCursor, secondaryCursor := opts.Stream.Cursor()
+	primaryCursorValue := opts.State.GetCursor(opts.Stream.Self(), primaryCursor)
+	placeholder := GetPlaceholder(opts.Driver)
+
+	conditionMaker := func(opts IncrementalConditionOptions, argumentPosition int, cursorField string, cursorValue any) (string, any, error) {
```
hash-data:
```suggestion
	createCondition := func(opts IncrementalConditionOptions, argumentPosition int, cursorField string, cursorValue any) (string, any, error) {
```

drivers/mysql/internal/backfill.go
```diff
@@ -29,6 +29,19 @@ func (m *MySQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
...
+			Stream: stream,
+			State:  m.state,
+			Client: m.client,
+		}
+
+		filter, args, err := jdbc.FilterUpdater(opts, filter)
+		if err != nil {
+			return fmt.Errorf("failed to update filter limiting the cursor values: %s", err)
+		}
+		logger.Infof("Starting backfill with filter: %s, args: %v", filter, args)
```
hash-data: let us rename this filter to threshold filter

drivers/oracle/internal/backfill.go
```diff
@@ -135,3 +142,11 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
 	}
 	return splitViaRowId(stream)
 }
+
+func (o *Oracle) FetchMaxCursorValues(ctx context.Context, stream types.StreamInterface) (any, any, error) {
```
hash-data: this function should be part of incremental file

pkg/jdbc/jdbc.go
```diff
@@ -453,9 +461,11 @@ func OracleChunkRetrievalQuery(taskName string) string {
 }
 
 // OracleIncrementalValueFormatter is used to format the value of the cursor field for Oracle incremental sync, mainly because of the various timestamp formats
-func OracleIncrementalValueFormatter(cursorField, argumentPlaceholder string, lastCursorValue any, opts IncrementalConditionOptions) (string, any, error) {
+func OracleIncrementalValueFormatter(cursorField, argumentPlaceholder string, greaterThan bool, lastCursorValue any, opts IncrementalConditionOptions) (string, any, error) {
 	// Get the datatype of the cursor field from streams
 	stream := opts.Stream
+	// in case of incremental sync, during backfill to avoid duplicate records we need to use <= else use >
+	operator := utils.Ternary(greaterThan, ">", "<=").(string)
```
hash-data: let us discuss

pkg/jdbc/jdbc.go
```diff
@@ -617,12 +615,97 @@ func BuildIncrementalQuery(opts IncrementalConditionOptions) (string, []any, err
...
+		maxSecondaryCursorValue = bytesConverter(maxSecondaryCursorValue)
+	} else {
+		err := client.QueryRowContext(ctx, cursorValueQuery).Scan(&maxPrimaryCursorValue)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to scan primary cursor value: %s", err)
+		}
+	}
+	if maxPrimaryCursorValue == nil {
+		logger.Warnf("max primary cursor value is nil for stream: %s", stream.ID())
+	}
```
hash-data: add a check in abstract about nil

pkg/jdbc/jdbc.go
```diff
@@ -617,12 +615,97 @@ func BuildIncrementalQuery(opts IncrementalConditionOptions) (string, []any, err
...
+		secondaryCursorValue := opts.State.GetCursor(opts.Stream.Self(), secondaryCursor)
+		secondaryCondition, argument, err := conditionMaker(opts, 2, secondaryCursor, secondaryCursorValue)
+		if err != nil {
+			return "", nil, fmt.Errorf("failed to format secondary cursor value: %s", err)
+		}
+		incrementalFilter = fmt.Sprintf("%s AND (%s IS NULL OR %s)", incrementalFilter, QuoteIdentifier(secondaryCursor, opts.Driver), secondaryCondition)
+		arguments = append(arguments, argument)
+	}
+
+	filter = utils.Ternary(filter == "", incrementalFilter, fmt.Sprintf("(%s) AND %s", filter, incrementalFilter)).(string)
```
hash-data: let us discuss after you test it

pkg/jdbc/jdbc.go
```diff
@@ -617,12 +615,97 @@ func BuildIncrementalQuery(opts IncrementalConditionOptions) (string, []any, err
...
+
+	conditionMaker := func(opts IncrementalConditionOptions, argumentPosition int, cursorField string, cursorValue any) (string, any, error) {
+		var conditionFilter string
+
+		if opts.Driver == constants.Oracle {
+			return OracleIncrementalValueFormatter(cursorField, placeholder(argumentPosition), false, cursorValue, opts)
+		}
+		conditionFilter = fmt.Sprintf("%s <= %s", QuoteIdentifier(cursorField, opts.Driver), placeholder(argumentPosition))
+		return conditionFilter, cursorValue, nil
+	}
```
hash-data: little bit confusing

drivers/abstract/backfill.go
```diff
@@ -18,6 +18,20 @@ func (a *AbstractDriver) Backfill(ctx context.Context, backfilledStreams chan st
...
+			maxPrimaryCursorValue, maxSecondaryCursorValue, err := a.driver.FetchMaxCursorValues(ctx, stream)
+			if err != nil {
+				return fmt.Errorf("failed to fetch max cursor values: %s", err)
+			}
+			primaryCursor, secondaryCursor := stream.Cursor()
+
+			a.state.SetCursor(stream.Self(), primaryCursor, a.reformatCursorValue(maxPrimaryCursorValue))
+			if secondaryCursor != "" {
+				a.state.SetCursor(stream.Self(), secondaryCursor, a.reformatCursorValue(maxSecondaryCursorValue))
+			}
```
hash-data: you can call this in incremental only

drivers/oracle/internal/backfill.go
vaibhav-datazip: But this function is only used during backfill to fetch the max cursor value.

pkg/jdbc/jdbc.go
vaibhav-datazip: need to discuss about secondary cursor as well

drivers/mongodb/internal/backfill.go
```diff
@@ -427,24 +427,46 @@ func buildMongoCondition(cond types.Condition) bson.D {
...
+	case len(filter.Conditions) == 0 && len(conditions) == 0:
 		return bson.D{}, nil
 	case len(filter.Conditions) == 1:
-		return buildMongoCondition(filter.Conditions[0]), nil
+		conditions = append(conditions, buildMongoCondition(filter.Conditions[0]))
+	case len(filter.Conditions) == 2:
+		conditions = append(conditions, bson.D{{Key: "$" + filter.LogicalOperator, Value: bson.A{buildMongoCondition(filter.Conditions[0]), buildMongoCondition(filter.Conditions[1])}}})
 	default:
-		return bson.D{{Key: "$" + filter.LogicalOperator, Value: bson.A{buildMongoCondition(filter.Conditions[0]), buildMongoCondition(filter.Conditions[1])}}}, nil
+		logger.Warnf("multiple conditions are not supported in filter")
```
hash-data: it should fail will error right?

drivers/mongodb/internal/backfill.go
```diff
@@ -427,24 +427,46 @@ func buildMongoCondition(cond types.Condition) bson.D {
...
+				{Key: primaryCursor, Value: bson.D{{Key: "$lte", Value: maxPrimaryCursorValue}}},
+			})
+		}
+
+		if maxSecondaryCursorValue != nil && secondaryCursor != "" {
+			conditions = append(conditions, bson.D{
+				{Key: secondaryCursor, Value: bson.D{{Key: "$lte", Value: maxSecondaryCursorValue}}},
+			})
+		}
+	}
```
hash-data: let us create a different function named threshold filter

drivers/mysql/internal/backfill.go
```diff
@@ -29,17 +29,30 @@ func (m *MySQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
 		pkColumns := stream.GetStream().SourceDefinedPrimaryKey.Array()
 		chunkColumn := stream.Self().StreamMetadata.ChunkColumn
 		sort.Strings(pkColumns)
+
+		opts := jdbc.IncrementalConditionOptions{
```
hash-data: let us rename this struct to Options

drivers/oracle/internal/backfill.go
```diff
@@ -30,11 +30,25 @@ func (o *Oracle) ChunkIterator(ctx context.Context, stream types.StreamInterface
...
+		State:  o.state,
+		Client: o.client,
+	}
+
+	thresholdFilter, args, err := jdbc.ThresholdFilter(opts, filter)
+	if err != nil {
+		return fmt.Errorf("failed to update filter limiting the cursor values: %s", err)
+	}
+
+	logger.Infof("Starting backfill with filter: %s, args: %v", thresholdFilter, args)
```
hash-data: let us appen chunk min and max in log

pkg/jdbc/jdbc.go
```diff
@@ -453,9 +461,11 @@ func OracleChunkRetrievalQuery(taskName string) string {
 }
 
 // OracleIncrementalValueFormatter is used to format the value of the cursor field for Oracle incremental sync, mainly because of the various timestamp formats
-func OracleIncrementalValueFormatter(cursorField, argumentPlaceholder string, lastCursorValue any, opts IncrementalConditionOptions) (string, any, error) {
+func OracleIncrementalValueFormatter(cursorField, argumentPlaceholder string, isBackfill bool, lastCursorValue any, opts IncrementalConditionOptions) (string, any, error) {
 	// Get the datatype of the cursor field from streams
 	stream := opts.Stream
+	// in case of incremental sync, during backfill to avoid duplicate records we need to use <= else use >
```
hash-data:
```suggestion
	// in case of incremental sync mode, during backfill to avoid duplicate records we need to use '<=', otherwise use '>'
```

PR #506, opened 2025-09-15
feat: preserve custom destination DB for new streams in merge

vikaxsh:
# Description

- Detect user-provided custom database names or prefixes from old catalog
- Apply same custom DB or prefix to newly discovered streams during merge
- Fall back to default naming only if no customization exists
- Flatten nested if/else logic for cleaner, more readable code

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings


types/catalog.go
```diff
@@ -111,21 +112,65 @@ func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
...
 
 	return newCatalog
 }
+func detectDBNamingPattern(streams []*ConfiguredStream) (prefix string, isCustomPrefix, isCustomDB bool) {
+	if len(streams) == 0 {
+		return "", false, false
+	}
+
+	var customDBs = make(map[string]struct{})
+	var prefixes = make(map[string]struct{})
```
shubham19may: set is the right dataset for this.

shubham19may, line 149:
```suggestion
	var dynamicDBs = make(map[string]struct{})
```

types/catalog.go
```diff
@@ -111,21 +112,65 @@ func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
 		}
 		newCatalog.SelectedStreams = selectedStreams
 	}
+	prefix, useCustomPrefix, useCustomDB := detectDBNamingPattern(oldCatalog.Streams)
```
shubham19may: we should only do this if any one of the streams has destination_database. otherwise its older streams.json

types/catalog.go
```diff
@@ -111,21 +112,65 @@ func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
...
+	}
+
+	// Case 2: all use same db (no common prefix)
+	if len(customDBs) == 1 {
+		return "", false, true
+	}
+
+	// fallback: default
+	return "", false, false
+}
```
shubham19may:
```suggestion
func detectDBNamingPattern(streams []*ConfiguredStream) (isSameForAllStreams bool, isCustomDB bool, prefix string) {
	if len(streams) == 0 {
		return false, false, ""
	}
	db := streams[0].Stream.DestinationDatabase

	for _, s := range streams {
		if s.Stream.DestinationDatabase != db {
			return false, false, strings.Split(db, ":")[0]
		}
	}

	return true, strings.Contains(db, ":"), strings.Split(db, ":")[0]
}
```

types/catalog.go
```diff
@@ -111,21 +112,65 @@ func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
...
-		newStream.Stream.CursorField = oldStream.Stream.CursorField
-		newStream.Stream.DestinationDatabase = oldStream.Stream.DestinationDatabase
-		newStream.Stream.DestinationTable = oldStream.Stream.DestinationTable
+		// new stream → apply rule
+		if useCustomPrefix {
+			newStream.Stream.DestinationDatabase = fmt.Sprintf("%s:%s", prefix, utils.Reformat(newStream.Stream.Namespace))
+		} else if useCustomDB {
+			// pick first custom db name (assuming user wants same for all)
+			newStream.Stream.DestinationDatabase = oldCatalog.Streams[0].Stream.DestinationDatabase
+		}
```
shubham19may:
```suggestion
		// new stream → apply rule
		if isSameForAllStreams && !isCustomDB {
			newStream.Stream.DestinationDatabase = fmt.Sprintf("%s:%s", prefix, utils.Reformat(newStream.Stream.Namespace))
		} else if isSameForAllStreams {
			// pick first custom db name (assuming user wants same for all)
			newStream.Stream.DestinationDatabase = oldCatalog.Streams[0].Stream.DestinationDatabase
		}
```

types/catalog.go
```diff
@@ -111,21 +112,65 @@ func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
...
+		// else → leave default generated db
+
 		return nil
 	})
 
 	return newCatalog
 }
+func detectDBNamingPattern(streams []*ConfiguredStream) (prefix string, isCustomPrefix, isCustomDB bool) {
```
shubham19may: add comment on why this is necessary. basically to detect if this is coming from ui or user has manually modified things

hash-data (commented):
LGMT

hash-data (approved):

shubham19may (approved):

PR #507, opened 2025-09-15
feat: adds connection support via SSH tunnel for postgres and mysql

vaibhav-datazip:
# Description

This PR introduces the ability for the PostgreSQL and MySQL source connector to connect through an SSH tunnel. Previously, OLake required a direct connection to the database. This change enables users to securely access their DB instances located in private networks by routing the connection through a bastion host.

## Type of change

# How Has This Been Tested?

- [x] All sync Modes
   Successfully completed syncs for all available modes (e.g., Full Refresh, Incremental, CDC) while connected through the SSH tunnel.
- [x] Sync without SSH credentials
   Verified that the connector successfully establishes a direct database connection when no SSH tunnel configuration is provided, maintaining the original functionality.
- [x] Verify SSH connection is being used to connect to the db
   Confirmed that when SSH details are provided, the database connection is established exclusively through the bastion host and does not bypass the tunnel.


utils/ssh.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if c.Port <= 0 || c.Port > 65535 {
+		return errors.New("invalid ssh port number: must be between 1 and 65535")
+	}
+
+	if c.Username == "" {
+		return errors.New("ssh username is required")
+	}
+
+	if c.PrivateKey == "" && c.Password == "" {
```
shubham19may, line 35: see if we can ssh without any auth/creds

utils/ssh.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+		authMethods = append(authMethods, ssh.PublicKeys(signer))
+	}
+
+	sshCfg := &ssh.ClientConfig{
+		User: c.Username,
+		Auth: authMethods,
+		// Allows everyone to connect to the server without verifying the host key
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(), // #nosec G106
+		Timeout:         10 * time.Second,
```
shubham19may: lets make it 30

shubham19may (dismissed):

utils/utils.go
```diff
@@ -424,3 +425,19 @@ func GenerateDestinationDetails(namespace, name string, sourceDatabase *string)
...
+
+	signer, err := ssh.ParsePrivateKey([]byte(pemText))
+	if err == nil {
+		return signer, nil
+	}
+	if _, ok := err.(*ssh.PassphraseMissingError); ok {
+		return nil, fmt.Errorf("SSH private key appears encrypted, enter the passphrase")
+	}
+	return nil, err
+}
```
vikaxsh: can we move this as well in ssh.go

utils/ssh.go
```diff
@@ -0,0 +1,77 @@
...
+		authMethods = append(authMethods, ssh.PublicKeys(signer))
+	}
+
+	sshCfg := &ssh.ClientConfig{
+		User: c.Username,
+		Auth: authMethods,
+		// Allows everyone to connect to the server without verifying the host key
+		// TODO: Add proper host key verification
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(), // #nosec G106
+		Timeout:         30 * time.Second,
```
vikaxsh, line 67: what if we do not put any timeout?

vikaxsh:
see if we can create an interface
```
type NetworkTunnel interface {
	Start() error
	Stop()
}
```
-letter we can add AWSTunnel, GCPTunnel etc as well by implementing these methods
-also can we create an folder in utils network_tunnel or tunnel

drivers/mysql/internal/mysql.go
```diff
@@ -57,11 +61,40 @@ func (m *MySQL) Setup(ctx context.Context) error {
...
+		cfg.Net = "mysqlTcp"
+		mysql.RegisterDialContext(cfg.Net, func(ctx context.Context, addr string) (net.Conn, error) {
+			return m.sshClient.Dial("tcp", addr)
+		})
+
+		db, err = sql.Open("mysql", cfg.FormatDSN())
+		if err != nil {
+			return fmt.Errorf("failed to open tunneled database connection: %s", err)
+		}
+	} else {
```
vikaxsh, line 90: can we add a info log here as well

drivers/mysql/internal/mysql.go
```diff
@@ -57,11 +61,40 @@ func (m *MySQL) Setup(ctx context.Context) error {
...
+		db, err = sql.Open("mysql", m.config.URI())
```
vikaxsh: why sqlx -> sql?

drivers/mysql/internal/mysql.go
```diff
@@ -57,11 +61,40 @@ func (m *MySQL) Setup(ctx context.Context) error {
...
+	if m.sshClient != nil {
+		logger.Info("Connecting to MySQL via SSH tunnel")
+
+		cfg, err := mysql.ParseDSN(m.config.URI())
+		if err != nil {
+			return fmt.Errorf("failed to parse mysql DSN: %s", err)
+		}
+
+		// Allows mysql driver to use the SSH client to connect to the database
+		cfg.Net = "mysqlTcp"
```
vikaxsh, line 81: what is use of this?

vaibhav-datazip: used to specify custom configuration settings in setup, required for the mysql driver we are using. not required for postgres.

drivers/mysql/internal/mysql.go
vaibhav-datazip: That's for consistency  after if-else block, we are setting up a custom dialer function which won't be supported in sqlx. Hence, first the db library is used and then with the specified connection to `sqlx` wrapper.

drivers/mysql/internal/mysql.go
vaibhav-datazip: Added in case of using ssh tunnel, if connecting via ssh tunnel dosen't appear it will refer to direct connection.

PR #508, opened 2025-09-15
fix: kafka changes 2.0

### LINKED ISSUE #87: [Feat] Kafka Source

ImDoubD-datazip:
# Description
Kafka as source has been added as Olake's latest driver. The messages from the topics of Kafka will be synced directly to destination. Currently modes are supported: Incremental. 
- In case of Incremental : 
   - For this sync, cursor field from the available cursor field needs to be provided i.e. `"cursor_field" = "offset"`.
   - when state file not provided, initially backfill will take place and then incremental continuous sync will. This will update the state and commit the last partition and offset of selected topics synced.
   - if any problem occurs in state and state persisted checkpoints can be used then, last commit of given consumer group will be used. If there is no previous commit for that consumer group which is being used, then auto offset reset trigger will be used.
   - if `auto_offset_reset = earliest` then all the unread messages by the consumer will be synced from the start offset, with any of new incoming messages.
   - if `auto_offset_reset = latest` then all the new incoming messages will be synced.

#### TO FIX: 

   - 1st: process the messages 

Fixes #87 

## Type of change

# How Has This Been Tested?

Incremental Sync:
image

# How to test Kafka as source

Example source.json:
Using PLAINTEXT protocol
```json
{
    "protocol": {
    },
    "max_threads": 3,
    "wait_time": 30
}
```
Using SASL-SSL protocol
```json
{
    "protocol": {
    },
    "max_threads": 3,
    "wait_time": 30
}
```


ImDoubD-datazip:
closing

ImDoubD-datazip:
closing

PR #517, opened 2025-09-17
fix: fallback to bucketAutoStrategy when multiple _id types exist in mongo 

### LINKED ISSUE #514: MongoDB backfill operations fail with a fatal error when processing collections containing documents with different _id field types (e.g., ObjectId and string), even when filters are applied to exclude non-ObjectId documents.
```FATAL error occurred while reading records: failed to run change stream: failed to run backfill: failed to get or split chunks: failed after retry backoff: multiple type _id exist```

The isObjectID() function in /drivers/mongodb/internal/backfill.go:462 only examines the first document in a collection to determine the _id field type. When it finds an ObjectID, it returns true, causing the backfill process to use the
  splitVectorStrategy() for chunking. However, this strategy assumes all documents have ObjectID _id fields and fails when it encounters documents with other types (strings, integers, etc.).

## Environment

- **Olake version**:  for instance `0.2.2`
- Using official docker container

I tried changing the chunking strategy to bucketAuto to avoid this. 

```json
{
  "hosts": [
    "192.168.0.24:27017"
  ],
  "username": "recated",
  "password": "redacted",
  "authdb": "admin",
  "read-preference": "secondaryPreferred",
  "database": "redacted",
  "chunkingStrategy": "bucketAuto"
}
```
I also applied a filter to the selected streams ```"filter": "_id != \"ping\""```, since all these documents that are not ObjectId have an _id of "ping".

## Description

The isObjectID() function in /drivers/mongodb/internal/backfill.go:462 only examines the first document in a collection to determine the _id field type. When it finds an ObjectID, it returns true, causing the backfill process to use the splitVectorStrategy() for chunking. However, this strategy assumes all documents have ObjectID _id fields and fails when it encounters documents with other types (strings, integers, etc.).

## Steps to reproduce

Attempt to sync a mongo collection where the first document returned has a non ObjectID _id field.

## Expected behavior

- Collections with mixed _id types should automatically use bucketAutoStrategy() instead of splitVectorStrategy()
- User filters should be respected regardless of chunking strategy
- Backfill should complete successfully, excluding filtered documents

## Actual behavior

Backfill fails with error ``` FATAL error occurred while reading records: failed to run change stream: failed to run backfill: failed to get or split chunks: failed after retry backoff: multiple type _id exist```

dschaaff:
fixes #514 

When multiple _id types exist in mongo, it was still trying to use the split vector strategy which would fail. This adds a check for the multiple _id type condition so things correctly fall back to the bucket auto strategy. 

I tested by running this patch internally where we have some collections with multiple _id types.

### DIFF
--- a/drivers/mongodb/internal/backfill.go
+++ b/drivers/mongodb/internal/backfill.go
@@ -164,7 +164,7 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 		pipeline := mongo.Pipeline{
 			{{Key: "$sort", Value: bson.D{{Key: "_id", Value: 1}}}},
 		}
-		// Add filter for ObjectID type when multiple types are detected
+		// chunk only ObjectID type _id when multiple types are detected
 		if hasMultipleType(stream) {
 			logger.Warnf("Caution: collection %s contains multiple _id types. Only documents with ObjectID _id will be synced; other types are skipped, which could result in data loss.", stream.ID())
 			pipeline = append(pipeline, bson.D{{Key: "$match", Value: bson.D{{Key: "_id", Value: bson.D{{Key: "$type", Value: 7}}}}}})
@@ -483,11 +483,5 @@ func isObjectID(ctx context.Context, collection *mongo.Collection) (bool, error)
 
 func hasMultipleType(stream types.StreamInterface) bool {
 	_, idProperty := stream.Schema().GetProperty("_id")
-	var validType []types.DataType
-	for _, t := range idProperty.Type.Array() {
-		if t != types.Null {
-			validType = append(validType, t)
-		}
-	}
-	return len(validType) > 1
+	return len(idProperty.Type.Array()) > 1
 }
--- a/drivers/mongodb/internal/backfill.go
+++ b/drivers/mongodb/internal/backfill.go
@@ -268,8 +268,7 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 		chunks, err := splitVectorStrategy()
 		// check if authorization error occurs
 		if err != nil && (strings.Contains(err.Error(), "not authorized") ||
-			strings.Contains(err.Error(), "CMD_NOT_ALLOWED") ||
-			strings.Contains(err.Error(), "multiple type _id exist")) {
+			strings.Contains(err.Error(), "CMD_NOT_ALLOWED")) {
 			logger.Warnf("failed to get chunks via split vector strategy: %s", err)
 			return bucketAutoStrategy(storageSize)
 		}
--- a/drivers/mongodb/internal/backfill.go
+++ b/drivers/mongodb/internal/backfill.go
@@ -263,7 +263,8 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 		chunks, err := splitVectorStrategy()
 		// check if authorization error occurs
 		if err != nil && (strings.Contains(err.Error(), "not authorized") ||
-			strings.Contains(err.Error(), "CMD_NOT_ALLOWED")) {
+			strings.Contains(err.Error(), "CMD_NOT_ALLOWED") ||
+			strings.Contains(err.Error(), "multiple type _id exist")) {
 			logger.Warnf("failed to get chunks via split vector strategy: %s", err)
 			return bucketAutoStrategy(storageSize)
 		}
--- a/drivers/mongodb/internal/backfill.go
+++ b/drivers/mongodb/internal/backfill.go
@@ -31,12 +31,12 @@ func (m *Mongo) ChunkIterator(ctx context.Context, stream types.StreamInterface,
 	}
 
 	// check for _id type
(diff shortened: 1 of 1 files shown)


ImDoubD-datazip:
Hi @dschaaff ,

Currently, the code behaves as follows:
- If the _id field is of type primitiveObjectID, the splitVector strategy is used.
- If the _id field is of type string, the process falls back to the bucketAuto strategy.

The type detection is based on the first record, under the assumption that a collection contains only a single data type for _id. With this PR, the sync will no longer fail when multiple _id types are detected. However, when falling back to the bucketAuto strategy, chunk creation will still depend on the type of the first record. As a result, only records with that specific _id type will be synced, while records with other types _id will be missed.

To better evaluate the right approach, we’d like to understand your specific use case.

In terms of potential solutions (to be tested):
- Single chunk sync: If multiple _id types are detected, create a single chunk and sync it as a whole. This avoids missing records but may result in slower performance.
- String-based chunking: Convert all _id values to strings and create chunks based on the stringified _id. This ensures coverage across types, but feasibility and performance need validation.

dschaaff:
@ImDoubD-datazip 

We have collections where there is a single document with an _id that is the string "ping". All other documents in the collection have a standard mongo ObjectID for the _id. I do not need to sync the "ping" document with olake, only the documents with a standard ObjectID _id. The problem I'm currently facing is that more often than not the "ping" document is the first one returned. In that scenario I hit the multipe _id type error and no data is synced. 

Our scenario may be unique? Hard to say. An ideal setup for me would be a way to completely ignore the "ping" document and sync all other documents without error. 

Let me know if any other info would be helpful.

drivers/mongodb/internal/backfill.go
```diff
@@ -72,7 +72,7 @@ func (m *Mongo) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 	pool.AddRecordsToSyncStats(recordCount)
 
 	// check for _id type
-	isObjID, err := isObjectID(ctx, collection)
+	isObjID, hasMultipleType, err := isObjectID(ctx, collection, stream)
```
hash-data: why we are checking multiple types here, now this variable need to be passed everywhere. 
Create a function that take stream and tell if multiple id exist or not and use that.

ImDoubD-datazip: because already there is a function for objectID. I thought to add it there itself rather than creating a function

drivers/mongodb/internal/backfill.go
```diff
@@ -476,3 +480,14 @@ func isObjectID(ctx context.Context, collection *mongo.Collection) (bool, error)
...
+func hasMultipleType(stream types.StreamInterface) bool {
+	_, idProperty := stream.Schema().GetProperty("_id")
+	var validType []types.DataType
+	for _, t := range idProperty.Type.Array() {
+		if t != types.Null {
+			validType = append(validType, t)
+		}
+	}
+	return len(validType) > 1
+}
```
hash-data, line 490: if len > 2 return true, 
else check if null type is there or not

drivers/mongodb/internal/backfill.go
```diff
@@ -163,11 +163,16 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
...
 		}
+		// Add filter for ObjectID type when multiple types are detected
+		if hasMultipleType(stream) {
+			logger.Warnf("Caution: collection %s contains multiple _id types. Only documents with ObjectID _id will be synced; other types are skipped, which could result in data loss.", stream.ID())
+			pipeline = append(pipeline, bson.D{{Key: "$match", Value: bson.D{{Key: "_id", Value: bson.D{{Key: "$type", Value: 7}}}}}})
+		}
+		pipeline = append(pipeline, bson.D{{Key: "$bucketAuto", Value: bson.D{
+			{Key: "groupBy", Value: "$_id"},
+			{Key: "buckets", Value: numberOfBuckets},
+		}}})
```
hash-data, line 175: let us test this

ImDoubD-datazip: tested

drivers/mongodb/internal/backfill.go
ImDoubD-datazip: null type nit being handled as of now

drivers/mongodb/internal/backfill.go
```diff
@@ -476,3 +480,8 @@ func isObjectID(ctx context.Context, collection *mongo.Collection) (bool, error)
 	_, isObjID := idVal.(primitive.ObjectID)
 	return isObjID, nil
 }
+
+func hasMultipleType(stream types.StreamInterface) bool {
+	_, idProperty := stream.Schema().GetProperty("_id")
+	return len(idProperty.Type.Array()) > 1
```
hash-data: array conversion not required set has len function

ImDoubD-datazip: done

hash-data: can be nil as well

PR #521, opened 2025-09-19
feat: integration test for incremental sync

vikash390:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

### DIFF
--- a/utils/testutils/test_utils.go
+++ b/utils/testutils/test_utils.go
@@ -114,7 +114,7 @@ func discoverCommand(config TestConfig) string {
 	return fmt.Sprintf("/test-olake/build.sh driver-%s discover --config %s", config.Driver, config.SourcePath)
 }
 
-// TODO: check if we can remove namespace from being passed as a parameter and use a common namespace for all drivers
+// update normalization=true for selected streams under selected_streams.<namespace> by name
 func updateStreamsCommand(config TestConfig, namespace string, stream []string, isBackfill bool) string {
 	if len(stream) == 0 {
 		return ""
@@ -136,6 +136,23 @@ func updateStreamsCommand(config TestConfig, namespace string, stream []string, 
 		config.CatalogPath,
 	)
 	return jqExpr
+}
+
+// set sync_mode and cursor_field for a specific stream object in streams[] by namespace+name
+func setStreamSyncModeCommand(config TestConfig, namespace, streamName, syncMode, cursorField string, stateRequired bool) string {
+	tmpCatalog := fmt.Sprintf("/tmp/%s_set_mode_streams.json", config.Driver)
+	// --argjson is used for boolean state; map/select pattern updates nested array members
+	return fmt.Sprintf(
+		`jq --arg ns "%s" --arg name "%s" --arg mode "%s" --arg cursor "%s" --argjson state %t '.streams = (.streams | map(if .stream.namespace == $ns and .stream.name == $name then (.stream.sync_mode = $mode | .stream.cursor_field = $cursor | .stream.state_required = $state) | . else . end))' %s > %s && mv %s %s`,
+		namespace, streamName, syncMode, cursorField, stateRequired,
+		config.CatalogPath, tmpCatalog, tmpCatalog, config.CatalogPath,
+	)
+}
+
+// reset state file so incremental can perform initial load (equivalent to full load on first run)
+func resetStateFileCommand(config TestConfig) string {
+	// Ensure the state is clean irrespective of previous CDC run
+	return fmt.Sprintf(`rm -f %s; echo '{}' > %s`, config.StatePath, config.StatePath)
 }
 
 // to get backfill streams from cdc streams e.g. "demo_cdc" -> "demo"
@@ -270,6 +287,21 @@ func (cfg *IntegrationTest) TestIntegration(t *testing.T) {
 
 							t.Logf("Enabled normalization in %s", cfg.TestConfig.CatalogPath)
 
+							// Helper to run sync and verify
+							runSync := func(c testcontainers.Container, useState bool, operation, opSymbol string, schema map[string]interface{}) error {
+								cmd := syncCommand(*cfg.TestConfig, useState)
+								if useState && operation != "" {
+									cfg.ExecuteQuery(ctx, t, []string{currentTestTable}, operation, false)
+								}
+								if code, out, err := utils.ExecCommand(ctx, c, cmd); err != nil || code != 0 {
+									return fmt.Errorf("sync failed (%d): %s\n%s", code, err, out)
+								}
+								t.Logf("Sync successful for %s driver", cfg.TestConfig.Driver)
+								VerifyIcebergSync(t, currentTestTable, cfg.IcebergDB, cfg.DataTypeSchema, schema, opSymbol, cfg.TestConfig.Driver)
+								return nil
+							}
+
+							// 3. Phase A: Full load + CDC
 							testCases := []struct {
 								syncMode    string
 								operation   string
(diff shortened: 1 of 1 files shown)


utils/testutils/test_utils.go
```diff
@@ -261,15 +282,30 @@ func (cfg *IntegrationTest) TestIntegration(t *testing.T) {
 							// 	`jq '(.selected_streams[][] | .normalization) = true' %s > /tmp/streams.json && mv /tmp/streams.json %s`,
 							// 	cfg.TestConfig.CatalogPath, cfg.TestConfig.CatalogPath,
 							// )
-							streamUpdateCmd := updateStreamsCommand(*cfg.TestConfig, cfg.Namespace, []string{currentTestTable}, true)
+							streamUpdateCmd := updateSelectedStreamsCommand(*cfg.TestConfig, cfg.Namespace, cfg.PartitionRegex, []string{currentTestTable}, true)
 							if code, out, err := utils.ExecCommand(ctx, c, streamUpdateCmd); err != nil || code != 0 {
 								return fmt.Errorf("failed to enable normalization in streams.json (%d): %s\n%s",
```
vaibhav-datazip: Update this error message as well

utils/testutils/test_utils.go
```diff
@@ -306,30 +342,54 @@ func (cfg *IntegrationTest) TestIntegration(t *testing.T) {
...
+							cfg.ExecuteQuery(ctx, t, []string{currentTestTable}, "clean", false)
+							cfg.ExecuteQuery(ctx, t, []string{currentTestTable}, "add", false)
+
+							// Ensure normalization remains on for selected stream
+							streamUpdateCmd = updateSelectedStreamsCommand(*cfg.TestConfig, cfg.Namespace, cfg.PartitionRegex, []string{currentTestTable}, true)
+							if code, out, err := utils.ExecCommand(ctx, c, streamUpdateCmd); err != nil || code != 0 {
+								return fmt.Errorf("failed to enable normalization in streams.json for incremental (%d): %s\n%s",
+									code, err, out,
+								)
+							}
```
vaibhav-datazip: This is getting repeated here.

utils/testutils/test_utils.go
```diff
@@ -306,30 +342,54 @@ func (cfg *IntegrationTest) TestIntegration(t *testing.T) {
...
+							if err := runSync(c, true, "", "r", cfg.ExpectedData); err != nil {
+								return err
+							}
+
+							// Delta incremental: add new rows and sync again
+							t.Log("Running Incremental - inserts")
+							cfg.ExecuteQuery(ctx, t, []string{currentTestTable}, "insert", false)
+							if err := runSync(c, true, "", "u", cfg.ExpectedData); err != nil {
+								return err
+							}
```
vaibhav-datazip: here as well a test cases array can be created and call that same for loop

utils/testutils/test_utils.go
vikaxsh: what needs to update here?

utils/testutils/test_utils.go
vikaxsh: removed

utils/testutils/test_utils.go
vikaxsh: done

utils/testutils/test_utils.go
```diff
@@ -340,7 +360,58 @@ func (cfg *IntegrationTest) TestIntegration(t *testing.T) {
...
+								},
+								{
+									name:       "insert",
+									setupQuery: "insert",
+									useState:   true,
+									opSymbol:   "u",
+									expected:   cfg.ExpectedData,
+								},
+							}
+
```
vaibhav-datazip: Can we not define these test cases along with other once and call it from that single for loop . instead of creating another array for incremental test cases.

drivers/postgres/internal/postgres_test.go
```diff
@@ -18,6 +18,8 @@ func TestPostgresIntegration(t *testing.T) {
 		DataTypeSchema:     PostgresToIcebergSchema,
 		ExecuteQuery:       ExecuteQuery,
 		IcebergDB:          "postgres_postgres_public",
+		CursorField:        "col_bigserial",
+		PartitionRegex:     "/{col_bigserial,identity}",
 	}
```
vaibhav-datazip, line 23: for postgres test, geometrical and array based datatypes should also be added .

vikaxsh: we can create an issue for that , that is not part of this pr

utils/testutils/test_utils.go
vikaxsh: since i have divided test cases into phase, before each phase we need to perform certain ops, for that keeping like that makes it simple, can add parquet test cases similarly for both cdc and incremental

drivers/postgres/internal/postgres_test.go
vaibhav-datazip: adding a few line change for incremental wont need a new PR , it can be done here.

The scope of this PR is to add incremental integration tests, you should add the test covering all the edge cases.

utils/testutils/test_utils.go
vaibhav-datazip: You can check the PR #548 those extra steps are cleanly handled there, you can implement a similar thing in your PR

drivers/postgres/internal/postgres_test.go
vikaxsh: geometrical and array based datatypes are not part of incremental integration test

vaibhav-datazip: don't add them as cursor field just add the datatypes in table which is getting synced.

for now array type will give error, so you may skip that as that is being fixed in PR #646

utils/testutils/test_utils.go
vikaxsh: Adding a reset check inside the test array is incorrect it is out of scope of test case array should only contains test cases info,  that logic needs to be handled separately.

utils/testutils/test_utils.go
```diff
@@ -177,6 +177,187 @@ func GetBackfillStreamsFromCDC(cdcStreams []string) []string {
...
+	testTable string,
+) error {
+	t.Log("Starting Phase B: Full load + Incremental tests")
+
+	// Reset table to isolate incremental scenario
+	cfg.ExecuteQuery(ctx, t, []string{testTable}, "drop", false)
+	cfg.ExecuteQuery(ctx, t, []string{testTable}, "create", false)
+	cfg.ExecuteQuery(ctx, t, []string{testTable}, "clean", false)
+	cfg.ExecuteQuery(ctx, t, []string{testTable}, "add", false)
+
```
vaibhav-datazip, line 307: abstract this out in a function, which will be called from anywhere and will clean the db and re add the data

utils/testutils/test_utils.go
```diff
@@ -177,6 +177,187 @@ func GetBackfillStreamsFromCDC(cdcStreams []string) []string {
...
+			opSymbol:  "r",
+			expected:  cfg.ExpectedData,
+		},
+		{
+			name:      "insert",
+			operation: "insert",
+			useState:  true,
+			opSymbol:  "u",
+			expected:  cfg.ExpectedData,
+		},
```
vaibhav-datazip, line 337: we can also add an update case as well, also incremental should be tested for various types of cursor values

like 
- string 
- time 
- timestamp 
- timestamp with timezone
- datetime 
- float 
- int etc

utils/testutils/test_utils.go
vikaxsh: this is already abstracted, calling functions only

utils/testutils/test_utils.go
vikaxsh: we can create issues for adding multiple cursor fields and test them

vaibhav-datazip: please add a todo for this then

utils/testutils/test_utils.go
vaibhav-datazip: currently its getting called from just this function , will abstract it out later if its called from other functions as well

PR #535, opened 2025-09-23
fix: Added backoff retry mechanism for grpc port bind issue

shubham19may:
# Description

We were facing `java.io.IOException: Failed to bind to address 0.0.0.0/0.0.0.0:50080` while creating the Iceberg writer GRPC server. The cause of the issue is still unknown on why the port is still being used by some other process when dialTimeout in golang goes forward.

Fixes # (issue)

## Type of change

So we added a backoff retry mechanism to make sure that we find any other empty port.

# How Has This Been Tested?

- [X] Happy scenario (No error)
- [X] Used port 5432 which was already been used by postgres (Directly passed while creating java config).
- [X] Thrown any other error apart from binding, it works normally.

# Screenshots or Recordings

### DIFF
--- a/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
+++ b/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
@@ -45,7 +45,6 @@ public class OlakeRpcServer {
             LOGGER.error("Please provide a JSON config as an argument.");
             System.exit(1);
         }
-
         
 
         String jsonConfig = args[0];
--- a/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
+++ b/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
@@ -45,6 +45,7 @@ public class OlakeRpcServer {
             LOGGER.error("Please provide a JSON config as an argument.");
             System.exit(1);
         }
+
         
 
         String jsonConfig = args[0];
--- a/utils/logger/logger.go
+++ b/utils/logger/logger.go
@@ -249,9 +249,21 @@ type ProcessOutputReader struct {
 	readinessPattern *regexp.Regexp
 	readinessCh      chan struct{}
 	readinessOnce    sync.Once
-	// simple bounded buffer for startup diagnostics
-	errBuf   []string
-	errLimit int
+}
+
+var errorLines sync.Map // map[string]string → procKey -> first error line
+
+// RecordErrorLine records the first error line for a process
+func RecordErrorLine(procKey, line string) {
+	errorLines.LoadOrStore(procKey, line)
+}
+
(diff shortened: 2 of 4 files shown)


destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
```diff
@@ -45,7 +45,6 @@ public static void main(String[] args) throws Exception {
             LOGGER.error("Please provide a JSON config as an argument.");
             System.exit(1);
         }
-
```
hash-data, line 48: remove file change

utils/logger/logger.go
```diff
@@ -244,6 +244,26 @@ type ProcessOutputReader struct {
...
 	closeFn   func() error
 	closeOnce sync.Once
+
+	// readiness detection only (do not fail on error logs)
+	readinessPattern *regexp.Regexp
+	readinessCh      chan struct{}
+	readinessOnce    sync.Once
+}
+
+var errorLines sync.Map // map[string]string → procKey -> first error line
```
hash-data: why it is not part of process output reader?

utils/logger/logger.go
```diff
@@ -244,6 +244,26 @@ type ProcessOutputReader struct {
...
+
+// RecordErrorLine records the first error line for a process
+func RecordErrorLine(procKey, line string) {
```
hash-data: why creating extra functions ?

utils/logger/logger.go
```diff
@@ -244,6 +244,26 @@ type ProcessOutputReader struct {
...
+	errorLines.LoadOrStore(procKey, line)
+}
+
+// GetFirstErrorLine returns the first error line for a process required for showing root cause in check command error message
+func GetFirstErrorLine(procKey string) string {
```
hash-data: can be declared inside function only

utils/logger/logger.go
```diff
@@ -345,6 +376,19 @@ func SetupAndStartProcess(processName string, cmd *exec.Cmd) error {
...
+	errCh := make(chan error, 1)
+	readinessPattern := regexp.MustCompile(`(?i)Server started on port`)
+
+	// Configure readers for readiness/error notifications BEFORE starting the process
+	stdoutReader.readinessPattern = readinessPattern
+	stdoutReader.readinessCh = readyCh
+	// do not fail on error lines; only use errCh for early process exit
+	stderrReader.readinessPattern = readinessPattern
+	stderrReader.readinessCh = readyCh
+	// do not fail on error lines; only use errCh for early process exit
```
hash-data: repeated comment

utils/logger/logger.go
vikaxsh: removed

utils/logger/logger.go
```diff
@@ -294,8 +300,19 @@ func (p *ProcessOutputReader) StartReading() {
...
+					select {
+					case p.readinessCh <- struct{}{}:
+					default:
+					}
+				})
+			}
+
 			if isErrorLine || isStackTraceLine || inStackTrace {
 				Error(fmt.Sprintf("%s %s", p.Name, line))
+				p.errorLines.LoadOrStore(p.Name, line)
```
hash-data, line 315: ??

PR #536, opened 2025-09-24
fix: mongodb _id multiple types detection fix

ImDoubD-datazip:
# Description
This PR is the solution for sync when _id is present in multiple types in a collection.

# Fix
For multiple types _id, when one of the type is objectID then only documents having objectID will be synced, else will be ignored. This is handled by split_vector strategy. Split_vector should never fail when objectID type _id exist.

If any thing other than objectID type _id exist, for example int, float, string then strategy used is bucket auto.

## Type of change

# How Has This Been Tested?
Testing done on:

Collection with all type _id, with string and objectID type _id, with all objectID _id, with all string , int, long type _id.

If any more, please do provide to test on what.

### DIFF
--- a/drivers/mongodb/internal/backfill.go
+++ b/drivers/mongodb/internal/backfill.go
@@ -31,12 +31,12 @@ func (m *Mongo) ChunkIterator(ctx context.Context, stream types.StreamInterface,
 	}
 
 	// check for _id type
-	isObjID, err := isObjectID(ctx, collection)
+	ObjectIDPresent, err := isObjectID(ctx, collection)
 	if err != nil {
 		return fmt.Errorf("failed to check if _id is ObjectID: %s", err)
 	}
 
-	cursor, err := collection.Aggregate(ctx, generatePipeline(chunk.Min, chunk.Max, filter, isObjID), opts)
+	cursor, err := collection.Aggregate(ctx, generatePipeline(chunk.Min, chunk.Max, filter, ObjectIDPresent), opts)
 	if err != nil {
 		return fmt.Errorf("failed to create cursor: %s", err)
 	}
@@ -71,18 +71,11 @@ func (m *Mongo) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 	logger.Infof("Total expected count for stream %s: %d", stream.ID(), recordCount)
 	pool.AddRecordsToSyncStats(recordCount)
 
-	// check for _id type
-	isObjID, err := isObjectID(ctx, collection)
-	if err != nil {
-		return nil, fmt.Errorf("failed to check if _id is ObjectID: %s", err)
-	}
-	logger.Infof("_id is %s ObjectID", utils.Ternary(isObjID, "", "not").(string))
-
 	// Generate and update chunks
 	var retryErr error
 	var chunksArray []types.Chunk
 	err = abstract.RetryOnBackoff(m.config.RetryCount, 1*time.Minute, func() error {
-		chunksArray, retryErr = m.splitChunks(ctx, collection, stream, isObjID, storageSize)
+		chunksArray, retryErr = m.splitChunks(ctx, collection, stream, storageSize)
 		return retryErr
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


drivers/mongodb/internal/backfill.go
```diff
@@ -95,18 +95,15 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 	splitVectorStrategy := func() ([]types.Chunk, error) {
 		getID := func(order int) (primitive.ObjectID, error) {
 			var doc bson.M
-			err := collection.FindOne(ctx, bson.D{}, options.FindOne().SetSort(bson.D{{Key: "_id", Value: order}})).Decode(&doc)
+			objectIDBson := bson.D{{Key: "_id", Value: bson.D{{Key: "$type", Value: 7}}}}
```
vikaxsh, line 97: you can create an variable and  objectIDFilter and use every where in file

ImDoubD-datazip: Tried this, was not looking much good. I think it should stay the way it is.

drivers/mongodb/internal/backfill.go
```diff
@@ -256,7 +253,7 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 		return timestampStrategy()
 	default:
 		// Fallback to auto strategy if _id is not ObjectID or has multiple types
-		if hasMultipleType(stream) || !isObjID {
+		if !isObjID {
```
hash-data: add a log here why bucket auto choosen

drivers/mongodb/internal/backfill.go
```diff
@@ -95,18 +95,15 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 	splitVectorStrategy := func() ([]types.Chunk, error) {
 		getID := func(order int) (primitive.ObjectID, error) {
 			var doc bson.M
```
hash-data, line 96: add a comment on it that split vector only syncs objects with objectid types

drivers/mongodb/internal/backfill.go
```diff
@@ -256,7 +253,7 @@ func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, s
 		return timestampStrategy()
 	default:
 		// Fallback to auto strategy if _id is not ObjectID or has multiple types
-		if hasMultipleType(stream) || !isObjID {
+		if !isObjID {
 			return bucketAutoStrategy(storageSize)
 		}
 		// Not using splitVector strategy when _id is not an ObjectID:
```
hash-data: can be simple, if object id present go for split vector else return bucketAuto

drivers/mongodb/internal/backfill.go
```diff
@@ -465,20 +462,16 @@ func reformatID(v interface{}) (interface{}, error) {
 
 func isObjectID(ctx context.Context, collection *mongo.Collection) (bool, error) {
 	var doc bson.M
-	err := collection.FindOne(ctx, bson.D{}).Decode(&doc)
+	objectIDBson := bson.D{{Key: "_id", Value: bson.D{{Key: "$type", Value: 7}}}}
+	err := collection.FindOne(ctx, objectIDBson).Decode(&doc)
```
hash-data: are you sure if no error means some id present?

drivers/mongodb/internal/backfill.go
ImDoubD-datazip: adding debug log

drivers/mongodb/internal/backfill.go
ImDoubD-datazip: that is what is happening by this check. do you want me to use ternary here?

drivers/mongodb/internal/backfill.go
ImDoubD-datazip: No it is doesn't means this. No error doesn't guarantees presence of _id. No error here states there at least exists a _id with type objectID. also after mongodb v4.0, each mongodb collection requires to have unique _id. current version being popularly used is v8.0.

hash-data: then why we returning true? on no err

drivers/mongodb/internal/backfill.go
```diff
@@ -91,22 +84,25 @@ func (m *Mongo) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 	return types.NewSet(chunksArray...), nil
 }
 
-func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, stream types.StreamInterface, isObjID bool, storageSize float64) ([]types.Chunk, error) {
+func (m *Mongo) splitChunks(ctx context.Context, collection *mongo.Collection, stream types.StreamInterface, storageSize float64) ([]types.Chunk, error) {
 	splitVectorStrategy := func() ([]types.Chunk, error) {
+		// NOTE: split-vector only syncs docs with objectID based _id (even when multiple types _id exist)
```
hash-data, line 89: we have tested it that split vector only syncs object id and not any other type?

ImDoubD-datazip: Yes we have tested this. Split vector only syncs objectID, nothing else

hash-data (approved):
LGTM

PR #539, opened 2025-09-26
feat: kafka V0

### LINKED ISSUE #87: [Feat] Kafka Source

ImDoubD-datazip:
# Description
Kafka as source has been added as Olake's latest driver. The messages from the topics of Kafka will be synced directly to destination. Currently modes are supported: streaming cdc. 
- In case of streaming cdc : 

#### TO FIX: 

   - 1st: process the messages 

Fixes #87 

## Type of change

# How Has This Been Tested?

cdc streaming Sync:
image

# How to test Kafka as source

Example source.json:
Using PLAINTEXT protocol
```json
{
    "protocol": {
    },
    "max_threads": 3,
    "wait_time": 30
}
```
Using SASL-SSL protocol
```json
{
    "protocol": {
    },
    "max_threads": 3,
    "wait_time": 30
}
```

## Documentation


go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/jmoiron/sqlx v1.4.0
 	github.com/parquet-go/parquet-go v0.25.0
 	github.com/rs/zerolog v1.34.0
+	github.com/segmentio/kafka-go v0.4.48
```
shubham19may: why is this affecting the main go.mod? ideally dependencies should go into individual projects go.mod right?

drivers/postgres/internal/cdc.go
```diff
@@ -127,3 +127,11 @@ func validateReplicationSlot(conn *sqlx.DB, slotName string) error {
 
 	return nil
 }
+
+func (p *Postgres) PartitionStreamChanges(_ context.Context, _ abstract.PartitionMetaData, _ abstract.CDCMsgFn) error {
```
shubham19may: why does this have to do anything with PG? these are specific to Kafka.

drivers/postgres/internal/testdata/test_streams.json
```diff
@@ -1 +1 @@
...
...number","null"],"destination_column_name":"col_numeric"},"col_real":{"type":["number_small","null"],"destination_column_name":"col_real"},"col_text":{"type":["string","null"],"destination_column_name":"col_text"},"col_timestamp":{"type":["timestamp","null"],"destination_column_name":"col_timestamp"},"col_timestamptz":{"type":["timestamp","null"],"destination_column_name":"col_timestamptz"},"col_uuid":{"type":["string","null"],"destination_column_name":"col_uuid"},"col_varbit":{"type":["string","null"],"destination_column_name":"col_varbit"},"col_xml":{"type":["null","string"],"destination_column_name":"col_xml"}}},"supported_sync_modes":["cdc","strict_cdc"],"source_defined_primary_key":["col_bigserial"],"available_cursor_fields":["col_real","col_timestamp","col_varbit","col_bigint","col_bool","col_double_precision","col_interval","col_numeric","col_char","col_float4","col_int","col_name","col_timestamptz","col_bigserial","col_character_varying","col_date","col_json","col_text","col_uuid","col_xml","col_character","col_decimal","col_int2","col_integer","col_jsonb"],"sync_mode":"cdc","destination_database":"postgres_postgres:public","destination_table":"postgres_test_table_olake"}}]}
```
shubham19may: reason for the change?

constants/constants.go
```diff
@@ -28,6 +28,8 @@ const (
 	DestinationDatabasePrefix = "DESTINATION_DATABASE_PREFIX"
 	// EffectiveParquetSize is the effective size in bytes considering 256mb targeted parquet size, compression ratio as 8
 	EffectiveParquetSize = int64(256) * 1024 * 1024 * int64(8)
+	// Default wait time of kafka per partition
+	DefaultWaitTime = 30
```
shubham19may: this will work just like cdc works in postgres WAL. we will find the latest offset and sync till that and finish the job.

many kafka topics have 100 RPS+ this sync will never finish.

drivers/abstract/abstract.go
```diff
@@ -86,8 +86,10 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
 	var finalStreams []*types.Stream
 	streamMap.Range(func(_, value any) bool {
 		convStream, _ := value.(*types.Stream)
-		convStream.WithSyncMode(types.FULLREFRESH, types.INCREMENTAL)
-		convStream.SyncMode = types.FULLREFRESH
+		if isKafkaStreaming(a.driver.Type()) {
```
shubham19may:
```suggestion
		if isKafkaDriver(a.driver.Type()) {
```

drivers/abstract/abstract.go
```diff
@@ -86,8 +86,10 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
...
+			convStream.WithSyncMode(types.FULLREFRESH, types.INCREMENTAL)
+			convStream.SyncMode = types.FULLREFRESH
```
shubham19may, line 97: why not incremental?

drivers/abstract/backfill.go
```diff
@@ -15,6 +15,13 @@ import (
 )
 
 func (a *AbstractDriver) Backfill(ctx context.Context, backfilledStreams chan string, pool *destination.WriterPool, stream types.StreamInterface) error {
+	if a.driver.Type() == string(constants.Kafka) {
```
shubham19may, line 18:
```suggestion
	if isKafkaDriver(a.driver.Type()) {
```

drivers/abstract/backfill.go
```diff
@@ -68,16 +75,18 @@ func (a *AbstractDriver) Backfill(ctx context.Context, backfilledStreams chan st
 
 				// if it is incremental update the max cursor value received in chunk
 				if stream.GetSyncMode() == types.INCREMENTAL && (maxPrimaryCursorValue != nil || maxSecondaryCursorValue != nil) {
```
shubham19may:
```suggestion
				if stream.GetSyncMode() == types.INCREMENTAL && !isKafkaDriver(a.driver.Type()) && (maxPrimaryCursorValue != nil || maxSecondaryCursorValue != nil) {
```

drivers/abstract/cdc.go
```diff
@@ -51,10 +51,49 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
...
+			switch {
+			case isKafkaStreaming(a.driver.Type()):
+				index, _ := utils.ArrayContains(streams, func(s types.StreamInterface) bool { return s.ID() == streamID })
+				utils.ConcurrentInGroup(a.GlobalConnGroup, a.driver.GetPartitions()[streams[index].ID()], func(ctx context.Context, data PartitionMetaData) error {
+					inserter, err := pool.NewWriter(ctx, data.Stream, destination.WithThreadID(data.ReaderID))
+					if err != nil {
+						return fmt.Errorf("failed to create new thread in pool, error: %s", err)
+					}
+					logger.Infof("Thread[%s]: created cdc writer for stream %s", data.ReaderID, data.Stream.ID())
+					defer func() {
```
shubham19may, line 76: where are we closing Kafka client or driver?

drivers/abstract/cdc.go
```diff
@@ -51,10 +51,49 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
...
+							err = fmt.Errorf("post cdc error: %s, cdc insert thread error: %s", postCDCErr, err)
+						}
 
+						if err != nil {
+							err = fmt.Errorf("thread[%s]: %s", data.ReaderID, err)
+						}
+					}()
+					return RetryOnBackoff(a.driver.MaxRetries(), constants.DefaultRetryTimeout, func() error {
+						return a.driver.PartitionStreamChanges(ctx, data, func(ctx context.Context, message CDCChange) error {
+							pkFields := message.Stream.GetStream().SourceDefinedPrimaryKey.Array()
```
shubham19may, line 98: why kafka will have this?

drivers/abstract/interface.go
```diff
@@ -9,6 +9,12 @@ import (
 
 type BackfillMsgFn func(ctx context.Context, message map[string]any) error
 type CDCMsgFn func(ctx context.Context, message CDCChange) error
+type PartitionMetaData struct {
```
shubham19may: this is for every driver out there, why should it have PartitionMetaData?

drivers/abstract/abstract.go
shubham19may: Also we have written logic in cdc, then why are we not showing CDC here?
Have we self-reviewed the PR fully?

drivers/kafka/internal/config.go
```diff
@@ -0,0 +1,60 @@
...
+	if c.WaitTime <= 0 {
+		c.WaitTime = int(constants.DefaultWaitTime)
+	}
+
+	if c.AutoOffsetReset != "" && c.AutoOffsetReset != "earliest" && c.AutoOffsetReset != "latest" {
+		return fmt.Errorf("auto_offset_reset must be either 'earliest' or 'latest'")
+	}
+
+	if c.MaxThreads <= 0 {
+		c.MaxThreads = constants.DefaultThreadCount
```
shubham19may, line 43: for Kafka as we discussed it should be equal to number of total partition (we can define it as 0 and infer it while creating the executor GlobalConnGroup)

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,255 @@
...
+		assignments, ok := gen.Assignments[stream.Name()]
+		if !ok {
+			return fmt.Errorf("no assignments found for topic %s", stream.Name())
+		}
+		for _, assignment := range assignments {
+			startOffset := assignment.Offset
+			partition := assignment.ID
+			readerID := fmt.Sprintf("%s_%s", stream.ID(), utils.ULID())
+			// creating reader
+			reader := kafka.NewReader(kafka.ReaderConfig{
```
shubham19may, line 49: You create per-partition readers without GroupID; they won’t rebalance or emit ErrGenerationEnded, so generation handling is ineffective and reads are not tied to group membership.

* Fix (preferred): create max_threads readers using kafka.Reader with shared GroupID and subscribe to all selected topics (multi-topic via Reader’s GroupTopics if supported, or one reader per topic). Commits then use the reader’s commit API and remain in-group.

* Fix (if staying with ConsumerGroup API): run a generation loop:
    * for { gen, _ := cg.Next(ctx); spawn workers per gen.Assignments; each worker reads only until gen.Done(); commit via gen.CommitOffsets; }
    * Recreate readers per generation; remove reliance on ErrGenerationEnded from non-group readers.

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,255 @@
...
+			return fmt.Errorf("failed to set offset: %v", err)
+		}
+	}
+	// wait time
+	partitionIdleWait := time.Duration(k.config.WaitTime) * time.Second
+	for {
+		deadline := time.Now().Add(partitionIdleWait)
+		fetchCtx, cancel := context.WithDeadline(partitionCtx, deadline)
+		msg, err := readerInstance.FetchMessage(fetchCtx)
+		cancel()
```
shubham19may: ?

drivers/kafka/internal/kafka.go
```diff
@@ -0,0 +1,195 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to ping Kafka brokers: %v", err)
+	}
+	k.dialer = dialer
+	k.adminClient = adminClient
+	return nil
+}
+
+func (k *Kafka) Close() error {
```
shubham19may, line 90: we should also close consumergroup here as well

drivers/kafka/internal/kafka.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	var topicNames []string
+	for _, topic := range resp.Topics {
+		topicNames = append(topicNames, topic.Name)
+	}
+	return topicNames, nil
+}
+
+func (k *Kafka) ProduceSchema(_ context.Context, streamName string) (*types.Stream, error) {
```
shubham19may, line 125: shouldn't we fetch at max 10000 messages to tell people the normalised schema as well as sample?

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,255 @@
...
+			return fmt.Errorf("failed to reset offset: %v", setErr)
+		}
+	} else {
+		if err := readerInstance.SetOffset(data.StartOffset); err != nil {
+			return fmt.Errorf("failed to set offset: %v", err)
+		}
+	}
+	// wait time
+	partitionIdleWait := time.Duration(k.config.WaitTime) * time.Second
+	for {
```
shubham19may, line 78: need to change it to fetch till the latest offer for that partition discovered at the time of the sync start is achieved.

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,255 @@
...
+				logger.Infof("generation ended (rebalance) for partition %d, continuing", data.PartitionID)
+				time.Sleep(100 * time.Millisecond)
+				continue
+			}
+			if errors.Is(err, context.DeadlineExceeded) {
+				logger.Infof("no messages in topic %s within wait time for reader %s", data.Stream.Name(), data.ReaderID)
+				return nil
+			}
+			if errors.Is(err, context.Canceled) {
+				logger.Infof("context canceled for reader %s", data.ReaderID)
```
shubham19may, line 87: this can be because of the main code has been broken because of other error.

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,255 @@
...
+	for {
+		deadline := time.Now().Add(partitionIdleWait)
+		fetchCtx, cancel := context.WithDeadline(partitionCtx, deadline)
+		msg, err := readerInstance.FetchMessage(fetchCtx)
+		cancel()
+
+		if err != nil {
+			if errors.Is(err, kafka.ErrGenerationEnded) {
+				// rebalance / generation end - keep reader alive and let membership reassign
+				logger.Infof("generation ended (rebalance) for partition %d, continuing", data.PartitionID)
```
shubham19may: how will it be done? don't you have to re-assign partitions in the map that you created above as well?

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+}
+
+func (k *Kafka) PostCDC(_ context.Context, stream types.StreamInterface, noErr bool, readerID string) error {
+	if !noErr {
+		logger.Warnf("skipping commit for reader %s due to error", readerID)
+		return nil
+	}
+
+	load := func(m *sync.Map, key string) (any, bool) {
```
shubham19may: need?

drivers/abstract/abstract.go
ImDoubD-datazip: this should be ! isKafkaStreaming(a.driver.Type()) for full refresh and incremental. will do check again

ImDoubD-datazip: > Also we have written logic in cdc, then why are we not showing CDC here? Have we self-reviewed the PR fully?

we are not showing, just had removing the full streams part

ImDoubD-datazip: > Also we have written logic in cdc, then why are we not showing CDC here? Have we self-reviewed the PR fully?

we are not showing, just had removing the full streams / incremental part

drivers/abstract/cdc.go
ImDoubD-datazip: we are closing the reader instances in postcdc function, and we are closing the kafka client in close function which is called in main.go of kafka after registeration of driver is done and all it activities are done.

drivers/abstract/cdc.go
ImDoubD-datazip: since I removed the incremental logic, similar to fivetran showing offset and partition as primary key, I also included this.

drivers/kafka/internal/config.go
ImDoubD-datazip: I needed one discussion with Ankit bhaiya on how to do it more cleanly, rather just custom setting it for kafka only

drivers/kafka/internal/cdc.go
ImDoubD-datazip: > You create per-partition readers without GroupID; they won’t rebalance or emit ErrGenerationEnded, so generation handling is ineffective and reads are not tied to group membership.
> 
> * Fix (preferred): create max_threads readers using kafka.Reader with shared GroupID and subscribe to all selected topics (multi-topic via Reader’s GroupTopics if supported, or one reader per topic). Commits then use the reader’s commit API and remain in-group.
> * Fix (if staying with ConsumerGroup API): run a generation loop:
>   
>   * for { gen, _ := cg.Next(ctx); spawn workers per gen.Assignments; each worker reads only until gen.Done(); commit via gen.CommitOffsets; }
>   * Recreate readers per generation; remove reliance on ErrGenerationEnded from non-group readers.

I did the preferred fix, but then during max threads readers when were less than the partitions of selected topics, the data of some partitions were read again due to rebalance.

go.mod
```diff
@@ -16,23 +16,31 @@ require (
...
 	github.com/spf13/viper v1.20.1
-	github.com/stretchr/testify v1.10.0
+	github.com/stretchr/testify v1.11.1
 	github.com/xitongsys/parquet-go v1.6.2
 	github.com/xitongsys/parquet-go-source v0.0.0-20241021075129-b732d2ac9c9b
 	go.mongodb.org/mongo-driver v1.17.3
-	golang.org/x/tools v0.30.0
-	google.golang.org/grpc v1.71.3
-	google.golang.org/protobuf v1.36.6
+	google.golang.org/grpc v1.74.2
```
shubham19may: remove these changes.

drivers/abstract/interface.go
ImDoubD-datazip: have removed this for others

drivers/kafka/internal/cdc.go
ImDoubD-datazip: removed this

drivers/kafka/internal/kafka.go
ImDoubD-datazip: consumer grp closes when reader closes as reader is consumer group based

drivers/kafka/internal/kafka.go
ImDoubD-datazip: need to discuss with product

drivers/kafka/internal/cdc.go
ImDoubD-datazip: have changed. till last offset or highwatermark -1 fetch takes place

drivers/kafka/internal/cdc.go
ImDoubD-datazip: since reader are consumer group based, reassignment will be handled by it

drivers/kafka/internal/cdc.go
ImDoubD-datazip: have removed this

go.mod
ImDoubD-datazip: yes

drivers/abstract/abstract.go
```diff
@@ -21,6 +21,7 @@ type CDCChange struct {
 
 type AbstractDriver struct { //nolint:gosec,revive
 	driver          DriverInterface
+	kafkaDriver     KafkaInterface
```
shubham19may: we should avoid it, abstract has to be away from any connector specific details

shubham19may:
AI review : 

### High-impact issues
- Admin client ignores TLS/SASL; metadata/list-offsets/offset-fetch will fail on secured brokers
```75:89:drivers/kafka/internal/kafka.go
// Create admin client for metadata and offset operations
adminClient := &kafka.Client{
	Addr: kafka.TCP(strings.Split(k.config.BootstrapServers, ",")...),
}
```
  - Fix: create a `kafka.Transport` that uses the same TLS/SASL as the `Dialer` and assign it to the client (otherwise SASL_SSL/SASL_PLAINTEXT will break).

- auto_offset_reset and wait_time are documented but not implemented; config fields missing
```10:16:drivers/kafka/internal/config.go
type Config struct {
	BootstrapServers string         `json:"bootstrap_servers"`
	Protocol         ProtocolConfig `json:"protocol"`
	ConsumerGroupID  string         `json:"consumer_group_id,omitempty"`
	MaxThreads       int            `json:"max_threads"`
	RetryCount       int            `json:"backoff_retry_count"`
}
```
```20:33:drivers/kafka/README.md
        "consumer_group": "test-consumer",
        "max_threads": 3,
        "wait_time": 30
```
  - Fix: add `AutoOffsetReset ("earliest"|"latest")` and `WaitTime` to `Config`, validate them, use them in reader setup. Today behavior depends on broker defaults and is unpredictable.

- Reader-per-partition assumption is brittle; commit only one partition’s last message per reader
```72:135:drivers/kafka/internal/cdc.go
// one reader created per 'partition' entry, but the group may assign multiple partitions per reader
msg, err := readerInstance.FetchMessage(partitionCtx)
// ...
k.lastMessages.Store(data.ReaderID, msg)
// ...
if msg.Offset == lastOffsetCheckPoint {
	return nil
}
```
```137:166:drivers/kafka/internal/cdc.go
lastMsg, ok := value.(kafka.Message)
// commit only last message's partition
if commitErr := r.CommitMessages(ctx, lastMsg); commitErr != nil { ... }
```
  - Risk: in group rebalance or uneven assignment, a single reader can receive multiple partitions; committing only the last message commits just one partition’s offset, risking duplicates on restart for other partitions.
  - Options:
    - Use per-partition readers without a consumer group, and manage explicit offsets (then you lose group commits).
    - Or keep the group but collect and commit offsets for all partitions processed by that reader (e.g., batch commit messages per partition periodically and on shutdown).
    - Alternatively use a single reader per topic and parallelize writes only, committing offsets per message/partition with a buffer and periodic commit.

- Streaming drains to current high watermark then stops; doesn’t continuously tail unless messages keep arriving
```113:134:drivers/kafka/internal/cdc.go
lastOffsetCheckPoint := msg.HighWaterMark - 1
// ...
if msg.Offset == lastOffsetCheckPoint {
	return nil
}
```
  - This implements “drain backlog then exit” semantics. If you want long-running streaming, add a tailing loop or a configurable idle timeout (`WaitTime`) before exiting.

- Admin API calls won’t work on secured brokers (again) for offsets and metadata
```211:240:drivers/kafka/internal/kafka.go
metadataResp, err := k.adminClient.Metadata(ctx, metadataReq)
// ...
topicDetail, err := k.GetTopicMetadata(ctx, stream.Name())
```
```242:267:drivers/kafka/internal/cdc.go
committedOffsetsResp, err := k.adminClient.OffsetFetch(ctx, fetchOffsetReq)
```
  - Same as above: ensure `Transport` carries TLS/SASL on the client.

- Non-JSON payloads are dropped (contradicts README note “returned as null”)
```97:118:drivers/kafka/internal/cdc.go
if msg.Value == nil { ...; return nil }
if err := json.Unmarshal(msg.Value, &result); err != nil {
	logger.Errorf("failed to unmarshal message value: %v", err)
	return nil
}
```
  - This silently discards messages. Consider:
    - Always include a `message` (raw string) or `data` column fallback when JSON unmarshal fails (you already have `constants.StringifiedData`).
    - Add a config flag, e.g., `drop_non_json: bool`.

### Medium-priority issues and inconsistencies
- Consumer group ID read from only the first stream
```23:33:drivers/kafka/internal/cdc.go
if cursor := k.state.GetCursor(streams[0].Self(), "consumer_group_id"); ...
k.consumerGroupID = groupID
```
  - If state differs per stream, this chooses one arbitrarily. Either store the group at global state or validate consistency across streams.

- Topic discovery lists internal topics
```113:125:drivers/kafka/internal/kafka.go
resp, err := k.adminClient.Metadata(ctx, &kafka.MetadataRequest{})
for _, topic := range resp.Topics {
	topicNames = append(topicNames, topic.Name)
}
```
  - Filter topics starting with `"__"` and/or leave it to user by intersecting with selected streams during sync.

- Schema/content mismatch: schema declares `message`, but it’s never set
```127:138:drivers/kafka/internal/kafka.go
schema.AddTypes("message", types.String)
// ...
schema.AddTypes("key", types.String)
```
```97:112:drivers/kafka/internal/cdc.go
// result derived from JSON, but no "message" field added
```
  - Either populate `message` with raw payload (always), or remove it from schema and add `data` only on non-JSON.

- Docs mismatch: `consumer_group` vs `consumer_group_id`, `auto_offset_reset`, `wait_time`
```20:33:drivers/kafka/README.md
"consumer_group": "test-consumer", "auto_offset_reset": "latest", "wait_time": 30
```
  - Align README with code and config tags; ensure end-to-end examples work.

- ErrGenerationEnded handling
```83:94:drivers/kafka/internal/cdc.go
if errors.Is(err, kafka.ErrGenerationEnded) {
	logger.Infof("generation ended (rebalance) ... continuing")
	continue
}
```
  - Consider reconnect/refresh logic or surfacing a backoff; a tight loop on repeated generation changes can spin.

- Partition metadata `EndOffset` unused in the read loop
```212:219:drivers/kafka/internal/cdc.go
EndOffset: lastOffset,
```
  - You rely on `HighWaterMark`. Either remove `EndOffset` or use it as a cutoff when not in continuous mode.

- Regex in `parseSASLPlain` refers to PLAIN in error message but used for SCRAM too
```198:209:drivers/kafka/internal/kafka.go
return "", "", fmt.Errorf("invalid sasl_jaas_config for PLAIN")
```
  - Adjust error message to be mechanism-agnostic.

### Performance and resilience
- Commit granularity is “commit once per reader at the end”
```137:166:drivers/kafka/internal/cdc.go
if noErr { ... CommitMessages(ctx, lastMsg) }
```
  - For large backlogs this risks long replays on failure. Prefer periodic commits (e.g., every N messages or T seconds) while still gating final commit after destination ack.

- Backpressure
  - Consider configurable `MinBytes`, `MaxBytes`, `ReadLag` checks, and a per-reader `MaxWait`/`WaitTime`.

### Suggested improvements (implementable steps)
- Wire security into admin client:
  - Create `transport := &kafka.Transport{TLS: dialer.TLS, SASL: dialer.SASLMechanism}` and set `Client.Transport = transport`.
- Add config fields and honor them:
  - `AutoOffsetReset` ("earliest"|"latest") and `WaitTime` (seconds). For groups with no committed offsets, emulate behavior by peeking first/last offsets and pausing/looping until `WaitTime` idle; or document reliance on broker `auto.offset.reset`.
- Continuous streaming mode:
  - If `WaitTime > 0`, don’t exit at `HighWaterMark-1`; instead idle-wait (poll with timeout) and exit after idle threshold.
- Precise commits:
  - Buffer per-partition last processed offsets per reader and commit them together in `PostCDC` and periodically during fetch. Alternatively, move to a single reader per topic and accept group assignment semantics.
- Non-JSON handling:
  - On unmarshal error, set `message` or `data` with raw string and proceed.
- Discovery/topic filtering:
  - Exclude `"__consumer_offsets"` and other internal topics by default (configurable).
- State cohesion:
  - Save/read `consumer_group_id` in global state; validate on PreCDC if mixing previously persisted group IDs across streams.

### Notable code references
```65:90:drivers/kafka/internal/kafka.go
func (k *Kafka) Setup(ctx context.Context) error {
	if err := k.config.Validate(); err != nil { ... }
	dialer, err := k.createDialer() { ... }

	adminClient := &kafka.Client{
		Addr: kafka.TCP(strings.Split(k.config.BootstrapServers, ",")...),
	}
	_, err = adminClient.Metadata(ctx, &kafka.MetadataRequest{ Topics: []string{"__test"} })
	...
}
```
```52:66:drivers/kafka/internal/cdc.go
for idx, p := range partitions {
	readerID := fmt.Sprintf("%s_%s", streamID, utils.ULID())
	reader := kafka.NewReader(kafka.ReaderConfig{
		Brokers:  strings.Split(k.config.BootstrapServers, ","),
		Topic:    p.Stream.Name(),
		GroupID:  k.consumerGroupID,
		MinBytes: 1,
		MaxBytes: 10e6,
		Dialer:   k.dialer,
	})
	// Save reader and update partition metadata
	k.readers.Store(readerID, reader)
	partitions[idx].ReaderID = readerID
}
```
```97:112:drivers/kafka/internal/cdc.go
if err := json.Unmarshal(msg.Value, &result); err != nil {
	logger.Errorf("failed to unmarshal message value: %v", err)
	return nil
}
result["partition"] = msg.Partition
result["offset"] = msg.Offset
result["key"] = string(msg.Key)
result["kafka_timestamp"] = msg.Time.UnixMilli()
```
```137:166:drivers/kafka/internal/cdc.go
if noErr {
	value, ok := k.lastMessages.Load(readerID)
	...
	if commitErr := r.CommitMessages(ctx, lastMsg); commitErr != nil { ... }
	k.state.SetCursor(stream.Self(), "consumer_group_id", k.consumerGroupID)
}
```

- If you’d like, I can implement: admin client transport wiring, config additions (`auto_offset_reset`, `wait_time`), raw fallback handling, periodic commits, and topic filtering.

- In the abstract layer, the Kafka integration is neat and minimally invasive: dynamic MaxConnections by partition count and skipping backfill work as intended.

- One more minor: filter discover’d streams for Kafka to default `SyncMode=CDC` (already done) and maybe default `append_only=true` at selection time, as per README.

- Tests: consider adding a small fake-broker/integration suite for offset reset behaviors and non-JSON payloads.


- I completed a targeted review of Kafka driver, abstract CDC, config, and state interactions. Key bugs are security on admin client, missing `auto_offset_reset`/`wait_time`, and commit/partition assumptions. I also suggested concrete changes to make streaming continuous, robust, and aligned with docs.

drivers/abstract/abstract.go
```diff
@@ -58,10 +64,14 @@ func (a *AbstractDriver) Type() string {
...
 
+func (a *AbstractDriver) IsKafkaDriver() bool {
+	return a.kafkaDriver != nil
+}
+
 func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error) {
 	// set max connections
-	if a.driver.MaxConnections() > 0 {
-		a.GlobalConnGroup = utils.NewCGroupWithLimit(ctx, a.driver.MaxConnections())
+	if a.driver.MaxConnections(ctx, []types.StreamInterface{}) > 0 {
```
shubham19may: in case of every driver it will pass empty array? even for Kafka?

drivers/abstract/abstract.go
```diff
@@ -58,10 +64,14 @@ func (a *AbstractDriver) Type() string {
 	return a.driver.Type()
 }
 
+func (a *AbstractDriver) IsKafkaDriver() bool {
+	return a.kafkaDriver != nil
```
shubham19may:
```suggestion
	_, ok := a.driver.(KafkaInterface)
	return ok
```

drivers/abstract/abstract.go
shubham19may: we can fully avoid passing streams array in this function. logically no need for it.

drivers/kafka/internal/kafka.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	// Create admin client for metadata and offset operations
+	adminClient := &kafka.Client{
+		Addr: kafka.TCP(strings.Split(k.config.BootstrapServers, ",")...),
+	}
+
+	// Test connectivity by fetching metadata for a dummy topic
+	_, err = adminClient.Metadata(ctx, &kafka.MetadataRequest{
+		Topics: []string{"__test"},
```
shubham19may: is this always present?

drivers/abstract/interface.go
```diff
@@ -36,5 +36,16 @@ type DriverInterface interface {
 	CDCSupported() bool
 	PreCDC(ctx context.Context, streams []types.StreamInterface) error // to init state
 	StreamChanges(ctx context.Context, stream types.StreamInterface, processFn CDCMsgFn) error
-	PostCDC(ctx context.Context, stream types.StreamInterface, success bool) error // to save state
+	PostCDC(ctx context.Context, stream types.StreamInterface, success bool, readerID string) error // to save state
+}
+
+type PartitionInterface interface {
```
shubham19may: Not needed to be an interface. This is only used for Kafka which has concept of partition. Interfaces are used when multiple objects need to have different implementations of same methods

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,271 @@
...
+				logger.Errorf("failed to unmarshal message value: %v", err)
+				return nil
+			}
+			result["partition"] = msg.Partition
+			result["offset"] = msg.Offset
+			result["key"] = string(msg.Key)
+			result["kafka_timestamp"] = msg.Time.UnixMilli()
+			return result
+		}()
+		lastOffsetCheckPoint := msg.HighWaterMark - 1
```
shubham19may: why -1?

drivers/kafka/internal/cdc.go
shubham19may: dont think so, you are creating readers, how it will be automatically assigned to new partition created? simpler approach right not would be the break the sync.

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,271 @@
...
+			reader := kafka.NewReader(kafka.ReaderConfig{
+				Brokers:  strings.Split(k.config.BootstrapServers, ","),
+				Topic:    p.Stream.Name(),
+				GroupID:  k.consumerGroupID,
+				MinBytes: 1,    // can be taken from user
+				MaxBytes: 10e6, // can be taken from user
+				Dialer:   k.dialer,
+			})
+			// Save reader and update partition metadata
+			k.readers.Store(readerID, reader)
```
shubham19may: this is creating as many readers as partitions. whats the point of max_threads?

drivers/kafka/internal/cdc.go
shubham19may: also what is highwatermark?

ImDoubD-datazip: highwatermark is the value of next upcoming offset. so let's say there are 5 message in a partition so offset count is 5. the highwatermark will be 6. so when sync reaches the last offset which is highwatermark-1, then stop it.

ImDoubD-datazip:
> dont think so, you are creating readers, how it will be automatically assigned to new partition created? simpler approach right not would be the break the sync.

for this, there is an option in consumer group id based readers to `WatchPartitionChanges : true`. this will look up for new partitions if any and then do a rebalance if necessary. 
Now if you want then I canbreak the sync as well. I believe breaking the sync makes much more sense.

drivers/kafka/internal/cdc.go
ImDoubD-datazip: separate readers are created for each partition pre-hand. the work of max_threads come during there processing. for example, if there are 6 partitions in total, then in precdc 6 reader instances will be created. But let's say the user has provided 4 as max_threads, then message processing will be done for the first 4 reader of the 4 partitions. the other 2 reader will wait until any slot is empty. so purpose of max threads is during processing of message concurrently, not during reader instance creation.

drivers/abstract/interface.go
ImDoubD-datazip: I have removed partition interface, but kafka related 2 functions are to be kept in interface, which we don't want for other drivers.

drivers/kafka/internal/kafka.go
ImDoubD-datazip: no it is not. I have removed this. for ping, I just call above .metadata function now.

drivers/abstract/abstract.go
ImDoubD-datazip: yes I have removed this. there is a discrepancy in case of kafka, that is when defaultcount for thread is used(when user doesn't gives max thread), then number of partitions will be taken as max threads. but if the user also gives same number of max thread as default max thread, then also total number of max threads will be spawned as per total partitions, which it should not as it is given by the user

drivers/abstract/cdc.go
```diff
@@ -51,10 +51,63 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
...
+				if a.driver.MaxConnections() == constants.DefaultThreadCount {
+					totalPartitions := 0
+					_ = utils.ForEach(streams, func(stream types.StreamInterface) error {
+						if partitions, ok := partitionData[stream.ID()]; ok {
+							totalPartitions += len(partitions) // for max connection equal to number of partitions
+						}
+						return nil
+					})
+					a.GlobalConnGroup = utils.Ternary(totalPartitions > 0, utils.NewCGroupWithLimit(ctx, totalPartitions), a.GlobalConnGroup).(*utils.CxGroup)
+				}
```
shubham19may, line 69: what about the else part here? shouldn't we think about what happens if user set max_threads?

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,282 @@
...
+				return nil
+			}
+			return fmt.Errorf("error reading message in incremental sync: %v", err)
+		}
+
+		messageData := func() map[string]any {
+			var result map[string]any
+			if msg.Value == nil {
+				logger.Warnf("received nil message value at offset %d for topic %s, partition %d", msg.Offset, data.Stream.Name(), msg.Partition)
+				return nil
```
shubham19may, line 113: Recommendation: honor the discovered EndOffset for each partition (stop when msg.Offset >= data.EndOffset-1). That matches your “bounded run until latest discovered offset” goal and avoids race with newly arriving messages.

shubham19may:
Old logic which was not using consumer groups properly

PR #556, opened 2025-10-02
Fix recursive handling of primitive types in nested MongoDB documents…

### LINKED ISSUE #422: Fix `filterMongoObject` is not recursive, causing crashes on nested `DateTime` fields
## Environment

- **Deploy method**: local docker latest
- **Olake version**:  0.0.17
- **OS**: Mac OS 15.5
- **Cloud Provider**: None
- **Docker Params (if deployed with docker)**: None

## Description
The MongoDB connector was experiencing a fatal crash during the data sync process with the following error:
`Time.MarshalJSON: year outside of range [0,9999]`

This error occurred when the application attempted to serialize a Go `time.Time` object that contained an invalid year (e.g., year `0000`), a value that the standard `encoding/json` library cannot handle.

## Steps to reproduce

1. Ingest a MongoDB document that contains a `primitive.DateTime` with a malformed or out-of-range year (e.g., year 0).
2. Ensure this malformed `DateTime` is located in a nested position within the document, such as inside a nested map or an array of documents.
3. Run the Olake sync process against the collection containing this document.

## Expected behavior

The Olake connector should successfully identify and clean all `primitive.DateTime` fields, regardless of their position within the document. The malformed date should be converted to a valid `time.Time` object, and the sync process should complete without errors.

## Actual behavior

The sync process crashes with a fatal `Time.MarshalJSON: year outside of range [0,9999]` error. The non-recursive `filterMongoObject` function fails to find and clean the nested `DateTime`, leading to a JSON serialization failure downstream.

#### Suspected Root Cause

A thorough investigation revealed that the root cause was the non-recursive implementation of the `filterMongoObject` function in `drivers/mongodb/internal/mon.go`.

This function was responsible for cleaning and reformatting BSON-specific types (like `primitive.DateTime`) into standard Go types before the data was sent to the destination writer. However, its implementation only processed the top-level fields of a document.

If a malformed `primitive.DateTime` value was present in a nested document (e.g., a map within a map) or within an array of documents, the function would not traverse deep enough to find and clean it. This unhandled `DateTime` was then passed to the writer, where the JSON marshaler would attempt to serialize it and crash.

#### Proposed Solution

The fix was to rewrite the `filterMongoObject` function to be fully recursive. The new implementation now correctly handles:
- Nested documents (`map[string]any`)
- BSON-specific document types (`primitive.M`, `primitive.D`)
- Arrays containing nested documents (`[]any`)

yashranaway:
# Description

Fixed the MongoDB connector crash caused by malformed `primitive.DateTime` values (e.g., year 0) in nested documents and arrays. The [filterMongoObject](cci:1://file:///Users/adityagarud/olake/drivers/mongodb/internal/mon.go:193:0-311:1) function was non-recursive and only processed top-level fields, causing `Time.MarshalJSON: year outside of range [0,9999]` errors when malformed dates existed in nested structures.

**Changes:**
- Made [filterMongoObject](cci:1://file:///Users/adityagarud/olake/drivers/mongodb/internal/mon.go:193:0-311:1) fully recursive to handle nested documents (`bson.M`, `primitive.D`)
- Added handling for primitive types directly within arrays (`bson.A` and generic slices)
- Ensured all MongoDB primitive types are converted at any nesting level

Fixes #422 

## Type of change

# How Has This Been Tested?

- [x] **Nested documents with DateTime** - Verified DateTime conversion in documents nested 3+ levels deep
- [x] **Arrays with nested documents** - Tested arrays containing documents with malformed DateTimes
- [x] **primitive.D document types** - Validated conversion of BSON ordered documents with nested DateTimes
- [x] **Arrays with primitive types directly** - Tested arrays containing DateTime values not wrapped in documents
- [x] **Mixed types with NaN/Inf handling** - Verified special float value handling in nested structures
- [x] **All MongoDB primitive types** - Tested Timestamp, Binary, Decimal128, ObjectID, Null conversions
- [x] **Deeply nested structures** - Validated 6+ level nesting with multiple primitive types
- [x] **JSON marshal safety** - Confirmed filtered documents can be marshaled to JSON without errors

All tests pass: `go test -v -run TestFilterMongoObject ./drivers/mongodb/internal/`

# Screenshots or Recordings
image

Test output showing all recursive filtering tests passing:

### DIFF
--- a/drivers/mongodb/internal/mon.go
+++ b/drivers/mongodb/internal/mon.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"reflect"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


yashranaway:
@shubham19may hey, please let me know if you'd like anything changed or updated

vaibhav-datazip:
@yashranaway thanks for your contribution , but the pr needs to be raised with staging as base, also please follow the convention while naming your branch 
eg. fix/primitive-type-handling

yashranaway:
sorry about that, moving to #561

PR #561, opened 2025-10-03
Fix recursive handling of primitive types in nested MongoDB documents…

### LINKED ISSUE #422: Fix `filterMongoObject` is not recursive, causing crashes on nested `DateTime` fields

yashranaway:
# Description

Fixed the MongoDB connector crash caused by malformed `primitive.DateTime` values (e.g., year 0) in nested documents and arrays. The [filterMongoObject](cci:1://file:///Users/adityagarud/olake/drivers/mongodb/internal/mon.go:193:0-311:1) function was non-recursive and only processed top-level fields, causing `Time.MarshalJSON: year outside of range [0,9999]` errors when malformed dates existed in nested structures.

**Changes:**
- Made [filterMongoObject](cci:1://file:///Users/adityagarud/olake/drivers/mongodb/internal/mon.go:193:0-311:1) fully recursive to handle nested documents (`bson.M`, `primitive.D`)
- Added handling for primitive types directly within arrays (`bson.A` and generic slices)
- Ensured all MongoDB primitive types are converted at any nesting level

Fixes #422 

## Type of change

# How Has This Been Tested?

- [x] **Nested documents with DateTime** - Verified DateTime conversion in documents nested 3+ levels deep
- [x] **Arrays with nested documents** - Tested arrays containing documents with malformed DateTimes
- [x] **primitive.D document types** - Validated conversion of BSON ordered documents with nested DateTimes
- [x] **Arrays with primitive types directly** - Tested arrays containing DateTime values not wrapped in documents
- [x] **Mixed types with NaN/Inf handling** - Verified special float value handling in nested structures
- [x] **All MongoDB primitive types** - Tested Timestamp, Binary, Decimal128, ObjectID, Null conversions
- [x] **Deeply nested structures** - Validated 6+ level nesting with multiple primitive types
- [x] **JSON marshal safety** - Confirmed filtered documents can be marshaled to JSON without errors

All tests pass: `go test -v -run TestFilterMongoObject ./drivers/mongodb/internal/`

# Screenshots or Recordings
image

Test output showing all recursive filtering tests passing:

### DIFF
--- a/drivers/mongodb/internal/mon.go
+++ b/drivers/mongodb/internal/mon.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"reflect"
 	"sync"
 	"time"
 
@@ -213,8 +214,99 @@ func filterMongoObject(doc bson.M) {
 			} else {
 				doc[key] = value
 			}
-		default:
+		case bson.M:
+			// Recursively process nested documents
+			filterMongoObject(value)
 			doc[key] = value
+		case bson.A:
+			// Process arrays containing nested objects and primitive types
+			for i, item := range value {
+				switch itemVal := item.(type) {
+				case bson.M:
+					filterMongoObject(itemVal)
+					value[i] = itemVal
+				case primitive.D:
+					// Handle primitive.D documents within arrays
+					tempDoc := bson.M{}
+					for _, elem := range itemVal {
+						tempDoc[elem.Key] = elem.Value
+					}
+					filterMongoObject(tempDoc)
+					value[i] = tempDoc
+				case primitive.DateTime:
+					value[i] = itemVal.Time()
+				case primitive.Timestamp:
+					value[i] = itemVal.T
+				case primitive.Binary:
+					value[i] = fmt.Sprintf("%x", itemVal.Data)
+				case primitive.Decimal128:
+					value[i] = itemVal.String()
+				case primitive.ObjectID:
+					value[i] = itemVal.Hex()
+				case primitive.Null:
+					value[i] = nil
+				case float64:
+					if math.IsNaN(itemVal) || math.IsInf(itemVal, 0) {
+						value[i] = nil
+					}
+				}
+			}
+			doc[key] = value
+		case primitive.D:
+			// Handle BSON document type (primitive ordered document)
+			// Convert to regular bson.M for recursive processing
+			tempDoc := bson.M{}
+			for _, elem := range value {
+				tempDoc[elem.Key] = elem.Value
+			}
+			filterMongoObject(tempDoc)
+			
+			// Keep as bson.M (converted from primitive.D for simplicity)
+			doc[key] = tempDoc
+		default:
+			// Check for slice types that might contain nested structures or primitive types
+			v := reflect.ValueOf(value)
+			if v.Kind() == reflect.Slice {
+				newSlice := make([]interface{}, v.Len())
+				for i := 0; i < v.Len(); i++ {
+					elem := v.Index(i).Interface()
+					switch elemVal := elem.(type) {
+					case bson.M:
+						filterMongoObject(elemVal)
+						newSlice[i] = elemVal
+					case primitive.D:
+						tempDoc := bson.M{}
+						for _, e := range elemVal {
+							tempDoc[e.Key] = e.Value
+						}
+						filterMongoObject(tempDoc)
+						newSlice[i] = tempDoc
+					case primitive.DateTime:
+						newSlice[i] = elemVal.Time()
+					case primitive.Timestamp:
+						newSlice[i] = elemVal.T
+					case primitive.Binary:
+						newSlice[i] = fmt.Sprintf("%x", elemVal.Data)
+					case primitive.Decimal128:
+						newSlice[i] = elemVal.String()
+					case primitive.ObjectID:
+						newSlice[i] = elemVal.Hex()
+					case primitive.Null:
+						newSlice[i] = nil
+					case float64:
+						if math.IsNaN(elemVal) || math.IsInf(elemVal, 0) {
+							newSlice[i] = nil
+						} else {
+							newSlice[i] = elemVal
+						}
+					default:
+						newSlice[i] = elem
+					}
+				}
+				doc[key] = newSlice
+			} else {
+				doc[key] = value
+			}
 		}
 	}
 }
--- a/drivers/mongodb/internal/mon_edge_case_test.go
+++ b/drivers/mongodb/internal/mon_edge_case_test.go
@@ -1,0 +1,56 @@
+package driver
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+func TestFilterMongoObject_ArrayWithPrimitiveDateTime(t *testing.T) {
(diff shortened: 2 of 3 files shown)


yashranaway:
@ImDoubD-datazip hey please let me know if you'd like anything changed

drivers/mongodb/internal/mon.go
```diff
@@ -213,8 +214,99 @@ func filterMongoObject(doc bson.M) {
...
+					filterMongoObject(tempDoc)
+					value[i] = tempDoc
+				case primitive.DateTime:
+					value[i] = itemVal.Time()
+				case primitive.Timestamp:
+					value[i] = itemVal.T
+				case primitive.Binary:
+					value[i] = fmt.Sprintf("%x", itemVal.Data)
+				case primitive.Decimal128:
+					value[i] = itemVal.String()
```
ImDoubD-datazip: if decimal128 is expected to have return type as string (i believe due to precision loss), then please mention in comment why it is being done. please mention above as well.

drivers/mongodb/internal/mon.go
ImDoubD-datazip, line 1: for now no need for unit tests. but please mention in description of PR what all cases have you tested.

ImDoubD-datazip: once check all the test case pass for code snippet below. If yes, then please use this.
```go
func filterMongoObject(value interface{}) interface{} {
	if value == nil {
		return nil
	}

	switch v := value.(type) {
	case primitive.Timestamp:
		return v.T

	case primitive.DateTime:
		return v.Time()

	case primitive.Null:
		return nil

	case primitive.Binary:
		return fmt.Sprintf("%x", v.Data)

	case primitive.Decimal128:
		return v.String()

	case primitive.ObjectID:
		return v.Hex()

	case float64:
		if math.IsNaN(v) || math.IsInf(v, 0) {
			return nil
		}
		return v

	case bson.M:
		for key, val := range v {
			v[key] = filterMongoObject(val)
		}
		return v

	case primitive.D:
		// Convert primitive.D to bson.M
		m := make(bson.M, len(v))
		for _, elem := range v {
			m[elem.Key] = filterMongoObject(elem.Value)
		}
		return m

	case bson.A:
		for i, item := range v {
			v[i] = filterMongoObject(item)
		}
		return v

	default:
		// For other slice types using reflection
		val := reflect.ValueOf(value)
		if val.Kind() == reflect.Slice {
			newSlice := make([]interface{}, val.Len())
			for i := 0; i < val.Len(); i++ {
				newSlice[i] = filterMongoObject(val.Index(i).Interface())
			}
			return newSlice
		}
		return value
	}
}
```

ImDoubD-datazip:
> @ImDoubD-datazip hey please let me know if you'd like anything changed

Hi I have given the first set of review. Please check. Thanks

drivers/mongodb/internal/mon.go
yashranaway: added comment

drivers/mongodb/internal/mon.go
yashranaway: do i revert them back? and yes all tests pass and updated in description

yashranaway: Implemented cleaner version

ImDoubD-datazip:
@yashranaway ,
Hey I have just tested a nested mongodb document with year as 0000 on original staging, and the sync is not failing or giving any error. It is working fine. Can you provide any example to test, where you have seen it fail? In normalization true, level 1 flattening must happen. There the current staging code is not failing or breaking.

I used this:
```bash
db.bad_dates_test.insertOne({
  _id: "root_1",
  level1: {
    name: "Nested bad date example",
    // Malformed date directly in nested object
    malformed_direct: ISODate("0000-01-01T00:00:00Z"),

    level2: {
      description: "Deeply nested malformed date",
      createdAt: ISODate("0000-12-31T23:59:59Z"),
      events: [
        {
          eventName: "bad_event_1",
          eventTime: ISODate("0000-02-29T00:00:00Z") // leap year invalid
        },
        {
          eventName: "good_event",
          eventTime: ISODate("2024-03-15T10:00:00Z")
        },
        {
          eventName: "nested_array_event",
          details: {
            subArray: [
              { subEvent: "ok", ts: ISODate("2022-10-10T00:00:00Z") },
              { subEvent: "bad", ts: ISODate("0000-05-05T00:00:00Z") } // deeply nested invalid date
            ]
          }
        }
      ]
    }
  },
  meta: {
    source: "unit_test",
    processedAt: ISODate("0000-07-07T00:00:00Z") // another invalid field
  }
});
```

yashranaway:
@ImDoubD-datazip you're right - the old code already handled nested structures properly. This refactor is about code simplification, not bug fixing specifically

yashranaway:
ahh, i should update description sorry about the confusion

ImDoubD-datazip:
> @ImDoubD-datazip you're right - the old code already handled nested structures properly. This refactor is about code simplification, not bug fixing specifically

isn' the code itself simple. There were only certain cases we required to handle, and I believe we handled it well.

ImDoubD-datazip:
@yashranaway , I believe this PR might have served its purpose already, I mean the staging is good in itself I believe. When you get a chance, could you please take a look and close it if it’s no longer needed?

yashranaway:
yes

ImDoubD-datazip:
@yashranaway Thanks for you effort. Great Job 🙌.

PR #563, opened 2025-10-04
fix(mysql): enforce non-zero initial wait time for CDC setup

### LINKED ISSUE #492: fix: fail cdc sync if initial_wait_time provided is zero
Currently, the sync runs even when `initial_wait_time` is set to `0`, which leads to CDC sync failures.
This happens because the database requires some time to read logs and capture CDC events.

Add a validation to fail the sync when `initial_wait_time` is set to `0`.

> https://github.com/datazip-inc/olake/blob/791271226219df1c58abc433d0ad9e517a3fcf50/drivers/mysql/internal/mysql.go#L73-L77

Kamlesh72:
# Description

Fail the setup if initial_wait_time is 0 or negative

Fixes #492 

# Screenshots or Recordings

**UI**
Screenshot 2025-10-05 at 1 34 01 AM

**CLI**
Screenshot 2025-10-05 at 1 38 52 AM

### DIFF
--- a/drivers/mysql/internal/mysql.go
+++ b/drivers/mysql/internal/mysql.go
@@ -110,9 +110,9 @@ func (m *MySQL) Setup(ctx context.Context) error {
 		if err := utils.Unmarshal(m.config.UpdateMethod, cdc); err != nil {
 			return err
 		}
-		if cdc.InitialWaitTime == 0 {
-			// default set 10 sec
-			cdc.InitialWaitTime = 10
+		if cdc.InitialWaitTime <= 0 {
+			// Fail the setup if initial_wait_time is 0 or negative
+			return fmt.Errorf("Failed to setup CDC: Initial wait time must be greater than 0")
 		}
 		m.cdcConfig = *cdc
 	}


vaibhav-datazip:
similar check needs to be added in Postgres as well

Kamlesh72:
@vaibhav-datazip Currently in postgres if User provide `initial_wait_time` to 0, we set it to default value ( 1200 ).

Instead of that, If User provide `initial_wait_time` we validate it [ 120, 2400 ].
If User doesn't provide `initial_wait_time` then we set to 1200.

Does that works?

https://github.com/datazip-inc/olake/blob/1001235c0bd288c3e3420c8b23aa85124df7451a/drivers/postgres/internal/postgres.go#L101-L114

vaibhav-datazip:
@Kamlesh72 , one more check needs to be added here , if cdc config is selected and [Olake mysql cdc setup permissions](https://olake.io/docs/connectors/mysql/setup/generic) are missing or not granted, then too the check should fail when clicked next

Kamlesh72:
@vaibhav-datazip Shouldn't we check cdcSupported only if user has provided `initial_wait_time` and fail setup if cdc is not supported ?

https://github.com/datazip-inc/olake/blob/a79f88af07fb97e3b9181d906585b7965e9172cf/drivers/mysql/internal/mysql.go#L121-130

vaibhav-datazip:
> @vaibhav-datazip Shouldn't we check cdcSupported only if user has provided `initial_wait_time` and fail setup if cdc is not supported ?
> 
> https://github.com/datazip-inc/olake/blob/a79f88af07fb97e3b9181d906585b7965e9172cf/drivers/mysql/internal/mysql.go#L121-130

exactly, if user has selected `cdc` mode instead of `standalone` mode (speaking in terms of UI), we should add a check that cdc conditions are met. If these condition checks fail then we should throw error, like binlog permissions alongwith correct cdc initial wait time.

Kamlesh72:
@vaibhav-datazip Added mysql permission check and postgres checks.

drivers/mysql/internal/mysql.go
```diff
@@ -254,8 +275,7 @@ func (m *MySQL) IsCDCSupported(ctx context.Context) (bool, error) {
 		}
 
 		if strings.ToUpper(value) != expectedValue {
-			logger.Warnf(warnMessage)
-			return false, nil
+			return false, fmt.Errorf(warnMessage)
```
vaibhav-datazip, line 278: why Is this change done ??

drivers/postgres/internal/postgres.go
```diff
@@ -105,8 +105,13 @@ func (p *Postgres) Setup(ctx context.Context) error {
...
 		}
-		// set default value
-		cdc.InitialWaitTime = utils.Ternary(cdc.InitialWaitTime == 0, 1200, cdc.InitialWaitTime).(int)
+
+		// Set initial_wait_time to 1200 if not provided by user
+		userProvided, _ := utils.IsOfType(p.config.UpdateMethod, "initial_wait_time")
+		if !userProvided {
+			cdc.InitialWaitTime = 1200
+			logger.Infof("CDC initial wait time not provided by user, defaulting to %d", cdc.InitialWaitTime)
+		}
```
vaibhav-datazip, line 114: any reason for this change ?

drivers/postgres/internal/postgres.go
```diff
@@ -105,8 +105,13 @@ func (p *Postgres) Setup(ctx context.Context) error {
 		if err := utils.Unmarshal(p.config.UpdateMethod, cdc); err != nil {
 			return err
 		}
-		// set default value
-		cdc.InitialWaitTime = utils.Ternary(cdc.InitialWaitTime == 0, 1200, cdc.InitialWaitTime).(int)
+
+		// Set initial_wait_time to 1200 if not provided by user
+		userProvided, _ := utils.IsOfType(p.config.UpdateMethod, "initial_wait_time")
```
vaibhav-datazip, line 110: also why no query based check is added in postgres side.

you can visit the [olake documentation on how to verify cdc configurations](https://olake.io/docs/connectors/postgres/setup/generic)


```
SELECT name, setting FROM pg_settings 
WHERE name IN ('wal_level', 'max_replication_slots', 'max_wal_senders');
```

drivers/mysql/internal/mysql.go
Kamlesh72: https://github.com/datazip-inc/olake/blob/93a6da44205cf3f3008595c9e7f8c6e9e020669f/drivers/mysql/internal/mysql.go#L135-L139

So if error occurs, user can see error like `binlog_format is not set to ROW`

drivers/postgres/internal/postgres.go
Kamlesh72: If user leave `Initial wait time` field empty, we set to 1200.
If user enters any value ( including 0 ), then only we validate.

So if user enters 0, I thought we shouldn't silently change that to default value.

drivers/postgres/internal/postgres.go
Kamlesh72: My mistake I didnt checked that. I will include it in code.

Kamlesh72: Will update this by tomorrow ( saturday ).

vaibhav-datazip: any updates on this @Kamlesh72 ?

vaibhav-datazip:
hi @Kamlesh72 , will be closing this PR if you are no longer working on this

Kamlesh72:
@vaibhav-datazip I will surely merge it this week. Sorry for the delay.

vaibhav-datazip:
> @vaibhav-datazip I will surely merge it this week. Sorry for the delay.

okay then, no problem . please resolve the merge conflicts and acknowledge the previous comments

vaibhav-datazip:
closing this PR as the requested changes were not made.

PR #587, opened 2025-10-13
fix: duplicate table creation by threads

vaibhav-datazip:
# Description

Earlier when multiple tables with same destination were being triggered. Each thread tried to create a table , sometimes this table already existed, due to which error was being thrown.

Now the try catch block will handle the already exists exception.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings
[
image (1)
](url)

## Documentation

## Related PR's (If Any):

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -21,11 +21,15 @@ This directory contains self-contained, end-to-end demo stacks for OLake. Each e
 # 1) Start base Olake stack
 curl -sSL https://raw.githubusercontent.com/datazip-inc/olake-ui/master/docker-compose.yml | docker compose -f - up -d
 
-# 2) Start an example
+# 2) Clone the repository and navigate to root directory
+git clone https://github.com/datazip-inc/olake.git
+cd olake
+
+# 3) Start an example
 cd examples/presto-tabularest-minio-mysql
 docker compose up -d
 
-# 3) Follow suggested steps in README.md for the example
+# 4) Follow suggested steps in README.md for the example
 ```
 
 Each example’s `README.md` includes:
--- a/examples/presto-tabularest-minio-mysql/README.md
+++ b/examples/presto-tabularest-minio-mysql/README.md
@@ -23,13 +23,7 @@ This example demonstrates a complete data pipeline using:
 
 ## Quick Start
 
-### 1. Start the Base Olake Stack
-
-```bash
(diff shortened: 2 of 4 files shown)


destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/IcebergUtil.java
```diff
@@ -113,8 +114,14 @@ public static <T> T selectInstance(Instance<T> instances, String name) {
   public static Table createIcebergTable(Catalog icebergCatalog, TableIdentifier tableIdentifier, Schema schema) {
 
     if (!((SupportsNamespaces) icebergCatalog).namespaceExists(tableIdentifier.namespace())) {
-      ((SupportsNamespaces) icebergCatalog).createNamespace(tableIdentifier.namespace());
-      LOGGER.warn("Created namespace:'{}'", tableIdentifier.namespace());
+      // multiple threads can try to create the namespace concurrently
+      // if table was already created, this will avoid the error of already exists
+      try {
+        ((SupportsNamespaces) icebergCatalog).createNamespace(tableIdentifier.namespace());
+        LOGGER.warn("Created namespace:'{}'", tableIdentifier.namespace());
```
vaibhav-datazip, line 121: shouldn't this be Info instead of warn

drivers/abstract/incremental.go
```diff
@@ -25,23 +25,27 @@ func (a *AbstractDriver) Incremental(ctx context.Context, pool *destination.Writ
 			logger.Infof("Backfill skipped for stream[%s], already completed", stream.ID())
 			backfillWaitChannel <- stream.ID()
 			return nil
-		}
-		// Reset only mentioned cursor state while preserving other state values
-		a.state.ResetCursor(stream.Self())
+		} else if chunks := a.state.GetChunks(stream.Self()); chunks == nil || chunks.Len() == 0 {
```
hash-data, line 28: method, has completed back fil checks the same

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/IcebergUtil.java
hash-data: Let us not add try catch on table creation, as we need to be solid, that table is not being created asynchronously

Issue #421, opened 2025-07-29
bug: MongoDB timestamps with year 0 cause JSON marshaling error


qjflores:
## Environment

- **Deploy method**:  latest docker image from docker hub running on local
- **Olake version**:  0.0.17
- **OS**: for instance Mac OS 15.5
- **Cloud Provider**:  none
- **Docker Params (if deployed with docker)**: a list of env variables with values (please, hide sensitive credentials!), mapped volumes and ports

## Description
The Olake sync process is failing with a fatal error when processing records from MongoDB. The error is `json: error calling MarshalJSON for type time.Time: Time.MarshalJSON: year outside of range [0,9999]`.

This error occurs when a `time.Time` field with a year of 0 is passed to Go's `encoding/json` library for marshaling.

## Steps to reproduce

1.  **Set up a MongoDB instance:** Ensure you have access to a MongoDB database (local or remote).
2.  **Insert a document with a malformed timestamp:** In a collection within your MongoDB instance, insert a document containing a timestamp field with a year of 0. For example:
    ```javascript
    db.your_collection_name.insertOne({ "timestampField": ISODate("00-1-11-30T00:00:00Z") });
    ```
    Replace `your_collection_name` with the actual name of your collection and `timestampField` with the name of your timestamp field.
3.  **Configure Olake to sync from this MongoDB collection:** Ensure your Olake configuration is set up to read from the MongoDB collection created in the previous step.
4.  **Run the Olake sync process:** Execute the command to start the data synchronization with Olake.
5.  **Observe the error:** The Olake sync process should terminate with a fatal error similar to:
    ```
    json: error calling MarshalJSON for type time.Time: Time.MarshalJSON: year outside of range [0,9999]
    ```

## Expected behavior

The Olake pipeline should gracefully handle timestamps with a year of 0. Instead of failing, the `ReformatDate` function should correctly adjust the year (e.g., to 1) to ensure the timestamp can be marshaled to JSON without error. The record should then be processed and sent to the destination (e.g., Iceberg) with the corrected timestamp.

## Actual behavior

When the Olake sync process encounters a MongoDB record with a timestamp field where the year is 0 (e.g., `ISODate("00-1-11-30T00:00:00Z")`), the pipeline fails with the following error:

```
json: error calling MarshalJSON for type time.Time: Time.MarshalJSON: year outside of range [0,9999]
```
This indicates that the `ReformatDate` function's current logic does not correctly handle year 0, leading to a marshaling failure.

[...]

PR #592, opened 2025-10-15
fix: Mysql invalid date breaks sync

### LINKED ISSUE #570: Sync failed with 0000-00-00 date values in MySQL db
## Environment

- Latest OLake v 0.2.6 + MySQL Connector

## Description
MySQL allows to store ZERO date/datetime like 0000-00-00 ( 0000-00-00 00:00:00)  or even more strange combinations like 2025-00-15.
See details here
https://dev.mysql.com/doc/refman/8.4/en/sql-mode.html#sqlmode_no_zero_in_date

It breaks sync process for entire table with error

error occurred while reading records: 
error occurred while waiting for connections: thread [ TABLE_NAME ]: 
failed to scan record data as map: failed to parse datetime from available formats: parsing time "0000-00-00 00:00:00": month out of range
 
## Expected behavior

It would be good to replace incorrect values to NULL or start of  epoch (January 1, 1970, at 00:00:00 UTC)

vaibhav-datazip:
# Description

Mysql supports invalid date addition in case invalid dates is allowed in db. Due to this when OLake detected a invalid date like
- 0000-00-00 (all 0)
- 2025-00-12 (month 0)
- 2025-07-00 (date 0)
- 2025-02-31 (invalid date of month)

the sync used to fail.

This PR fixes that issue and changes such invalid dates to start of epoch i.e. January 1, 1970, at 00:00:00 UTC

Fixes #570 

## Type of change

# How Has This Been Tested?

- [x] Postgres CDC, Full refresh, incremental
- [x] Mysql CDC, Full refresh, incremental
- [x] MongoDB CDC, Full refresh

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/drivers/mysql/internal/mysql.go
+++ b/drivers/mysql/internal/mysql.go
@@ -218,6 +218,15 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
 	if value == nil {
 		return nil, typeutils.ErrNullValue
 	}
+
+	// for special geospatial type, mysql returns non-utf8 binary data
+	for _, geoType := range typeutils.GeospatialTypes {
+		if strings.Contains(strings.ToLower(columnType), geoType) {
+			// conversion to wkt from non-utf8 binary wkb
+			return typeutils.ReformatGeoType(value)
+		}
+	}
+
 	olakeType := typeutils.ExtractAndMapColumnType(columnType, mysqlTypeToDataTypes)
 	return typeutils.ReformatValue(olakeType, value)
 }
--- a/utils/typeutils/reformat.go
+++ b/utils/typeutils/reformat.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/datazip-inc/olake/types"
 	"github.com/datazip-inc/olake/utils/logger"
+	"github.com/paulmach/orb/encoding/wkb"
+	"github.com/paulmach/orb/encoding/wkt"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 )
 
@@ -33,6 +35,8 @@ var DateTimeFormats = []string{
 	"2006-01-02 15:04:05.999999-07",
 	"2006-01-02 15:04:05.999999+00",
 }
+
+var GeospatialTypes = []string{"geometry", "point", "polygon", "linestring", "multi"}
 
 func getFirstNotNullType(datatypes []types.DataType) types.DataType {
 	for _, datatype := range datatypes {
@@ -497,4 +501,39 @@ func ReformatByteArraysToString(data map[string]any) map[string]any {
 		}
 	}
 	return data
+}
+
+func ReformatGeoType(v any) (any, error) {
+	if v == nil {
+		return nil, ErrNullValue
+	}
+
+	geoValue := func(b []byte) (any, error) {
+		// skipping 4-byte SRID prefix (mysql stores 25-byte wkb including SRID)
+		if len(b) > 4 {
+			// Well Known Binary (WKB) unmarshal -> Well Known Text (WKT)
+			if geom, err := wkb.Unmarshal(b[4:]); err == nil {
+				if s := wkt.MarshalString(geom); s != "" {
+					return s, nil
+				}
+			}
+		}
+
+		return fmt.Sprintf("%x", b), nil
+	}
+
+	switch vv := v.(type) {
+	case string:
+		// already textual WKT or similar
+		return vv, nil
+	case []uint8:
+		return geoValue([]byte(vv))
+	case *any:
+		if vv == nil {
+			return nil, ErrNullValue
+		}
+		return ReformatGeoType(*vv)
+	default:
+		return fmt.Sprintf("%v", v), nil
+	}
 }
--- a/go.mod
+++ b/go.mod
@@ -31,6 +31,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require github.com/paulmach/orb v0.12.0 // indirect
+
 require (
 	cloud.google.com/go/compute/metadata v0.6.0 // indirect
 	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect


ImDoubD-datazip:
haven't you added the fix for mongodb as well?

drivers/mysql/internal/mysql.go
ImDoubD-datazip, line 1: there are various supported date formats in reformat.go. why haven't you done the changes there itself? there the issue of mongodb year 0 could also have been fixed.

drivers/mysql/internal/mysql.go
```diff
@@ -219,9 +219,56 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
...
+	if isDateTimeColumn(columnType) {
+		value = handleInvalidMySQLDates(value)
+	}
+
 	return typeutils.ReformatValue(olakeType, value)
 }
 
+// isDateTimeColumn checks if the column type is a date/datetime/timestamp type
+func isDateTimeColumn(columnType string) bool {
+	dateTimeTypes := []string{"date", "timestamp"}
```
ImDoubD-datazip: there are types like timestamp_micro, timestamp_nano, are they not considered?

drivers/mysql/internal/mysql.go
```diff
@@ -219,9 +219,56 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
...
+	strValue := utils.ConvertToString(value)
+
+	isInvalidDate := func(value string) bool {
+		dateFormats := []string{
+			"2006-01-02",
+			"2006-01-02 15:04:05",
+			"2006-01-02T15:04:05",
+			"2006-01-02T15:04:05Z",
+			"2006-01-02T15:04:05.000000",
+			"2006-01-02T15:04:05.000000Z",
```
ImDoubD-datazip: also if you have specified the formats here, then where are nano precisions. also this is redundant code.

drivers/mysql/internal/mysql.go
```diff
@@ -219,9 +219,56 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
...
+			if _, err := time.Parse(layout, value); err == nil {
+				return false
+			}
+		}
+		return true
+	}
+
+	if isInvalidDate(strValue) {
+		return time.Unix(0, 0).UTC()
+	}
```
ImDoubD-datazip: what is the use of this function?

vaibhav-datazip: return time of epoch start that is 1970

drivers/mysql/internal/mysql.go
vaibhav-datazip: if you will read the code below this you will find .

```
strings.Contains(strings.ToLower(columnType), dtType)
```

so if the string will contain timestamp or date , in all those cases this should work

drivers/mysql/internal/mysql.go
vaibhav-datazip: so If you would see the previous commit, you will be able to see the changes were done in reformat.go only, but this caused an issue in oracle driver , also this change was only required in mysql hence moved it here.

apart from this , it was discussed on the Friday call that entering 0000 in monogdb and syncing is not causing any error.

vaibhav-datazip:
> haven't you added the fix for mongodb as well?

as discussed on the Friday call ,
the test case reported by you, when we enter year as 0000 in mongodb and try to sync no error occurs and sync is completing successfully

similar closing comment is put by you in Issue #421 
Screenshot 2025-10-20 at 3 27 53 PM

drivers/mysql/internal/mysql.go
ImDoubD-datazip: Arey where it is getting called i mean

vaibhav-datazip: no, I think you might have selected wrong lines in the comment or might have got confused.

 this is a if statement , that function is getting called in this block , here is that function which checks invalid date

```
isInvalidDate := func(value string) bool {
		dateFormats := []string{
			"2006-01-02",
			"2006-01-02 15:04:05",
			"2006-01-02T15:04:05",
			"2006-01-02T15:04:05Z",
			"2006-01-02T15:04:05.000000",
			"2006-01-02T15:04:05.000000Z",
		}
```

ImDoubD-datazip:
> > haven't you added the fix for mongodb as well?
> 
> as discussed on the Friday call , the test case reported by you, when we enter year as 0000 in mongodb and try to sync no error occurs and sync is completing successfully
> 
> similar closing comment is put by you in Issue #421 Screenshot 2025-10-20 at 3 27 53 PM

this gives error during query. the sync is successful, but querying on it gives error. i though you were fixing that as well.

vaibhav-datazip:
> this gives error during query. the sync is successful, but querying on it gives error. i though you were fixing that as well.

I think this is spark side error, because when we see the data in parquet it is fine.

drivers/mysql/internal/mysql.go
ImDoubD-datazip: ok. also can you use arrayContains function from utils.

drivers/mysql/internal/mysql.go
ImDoubD-datazip: oh it is inside the same function. i thought it was out. my bad.

drivers/mysql/internal/mysql.go
vaibhav-datazip: array contains can't be used here, its for strict comparison

drivers/mysql/internal/mysql.go
```diff
@@ -219,9 +219,45 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
 		return nil, typeutils.ErrNullValue
 	}
 	olakeType := typeutils.ExtractAndMapColumnType(columnType, mysqlTypeToDataTypes)
+
+	// Only apply invalid date handling to actual datetime/timestamp columns
+	if isDateTimeColumn(columnType) {
```
hash-data: is it failing here?

hash-data:
if we carefully check error is coming from the below  function, we can add a check there to verify the invalid date. It will be useful for all olake drivers
image

utils/typeutils/reformat.go
```diff
@@ -205,16 +207,49 @@ func ReformatDate(v interface{}) (time.Time, error) {
...
+		logger.Warnf("Detected invalid year %d (year 0000 or negative). Converting to epoch start time (1970-01-01 00:00:00 UTC)", parsed.Year())
+		parsed = time.Unix(0, 0).UTC()
 	} else if parsed.Year() > 9999 {
 		parsed = parsed.AddDate(-(parsed.Year() - 9999), 0, 0)
 	}
 
 	return parsed, nil
 }
 
+func startsWithDatePattern(value string) bool {
```
ImDoubD-datazip: please make this inline function in the parseStringTimestamp.

utils/typeutils/reformat.go
```diff
@@ -205,8 +207,11 @@ func ReformatDate(v interface{}) (time.Time, error) {
...
 	// manage year limit
 	// even after data being parsed if year doesn't lie in range [0,9999] it failed to get marshaled
-	if parsed.Year() < 0 {
-		parsed = parsed.AddDate(0-parsed.Year(), 0, 0)
+	// Check if year is 0000 (not supported by Spark)
+	// Spark only supports years from 1 to 9999, we are converting year 0000 to epoch start time
+	if parsed.Year() < 1 {
+		logger.Warnf("Detected invalid year %d (year 0000 or negative). Converting to epoch start time (1970-01-01 00:00:00 UTC)", parsed.Year())
+		parsed = time.Unix(0, 0).UTC()
 	} else if parsed.Year() > 9999 {
```
hash-data, line 219: Is it still required?

vaibhav-datazip: yes, because go parses it correctly , that's why sync was also getting completed successfully. but spark was unable to query it as it wasn't supported from there.

hash-data: can you add a comment for the same

PR #594, opened 2025-10-17
fix: Integration test spark error fixed

vaibhav-datazip:
# Description

While running the integration tests, sometimes this error used to appear.
```
Received unexpected error:
        	            	[sparkerror] Error Type: execution error
        	            	Error Cause: failed to execute sql: SELECT * FROM olake_iceberg.postgres_postgres_public.postgres_test_table_olake WHERE _op_type = 'r': execution error: [Internal] Cannot check and eventually update SQL schema
        	            	/home/runner/go/pkg/mod/github.com/apache/spark-connect-go/v35@v35.0.0-20250317154112-ffd832059443/spark/sql/sparksession.go:141 (0x125c503)
        	            		(*sparkSessionImpl).Sql: return nil, sparkerrors.WithType(fmt.Errorf("failed to execute sql: %s: %w", query, err), sparkerrors.ExecutionError)
        	            	/home/runner/work/olake/olake/utils/testutils/test_utils.go:374 (0x127d213)
        	            		VerifyIcebergSync: selectQueryDf, err := spark.Sql(ctx, selectQuery)
        	            	/home/runner/work/olake/olake/utils/testutils/test_utils.go:320 (0x127bccc)
        	            		(*IntegrationTest).TestIntegration.func2.3.1: VerifyIcebergSync(t, currentTestTable, cfg.IcebergDB, cfg.DataTypeSchema, schema, opSymbol, cfg.TestConfig.Driver)
```

The error occurred due to a race condition, where the data was not properly updated in the catalog side, and before that only the spark query was fired. Which eventually threw this error.

Now, this  is handled by attempting this query ( if it fails) for `maxRetryCounts` (for now 5) after every 2 seconds.

## Type of change

# Screenshots or Recordings

Previously
Screenshot 2025-10-18 at 3 40 16 AM

After the change
Screenshot 2025-10-19 at 12 15 37 PM

## Documentation

### DIFF
--- a/utils/testutils/test_utils.go
+++ b/utils/testutils/test_utils.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/apache/spark-connect-go/v35/spark/sql"
+	"github.com/apache/spark-connect-go/v35/spark/sql/types"
 	"github.com/datazip-inc/olake/constants"
 	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/typeutils"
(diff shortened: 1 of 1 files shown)


utils/testutils/test_utils.go
```diff
@@ -375,14 +375,37 @@ func VerifyIcebergSync(t *testing.T, tableName, icebergDB string, datatypeSchema
...
+			t.Logf("Query attempt %d failed: %v", attempt+1, queryErr)
+			if attempt == maxRetries-1 {
+				require.NoError(t, queryErr, "Failed to query table after %d attempts", maxRetries)
+			}
+			continue
+		}
+
+		selectQueryDf = df
+		break
+	}
```
vaibhav-datazip, line 417: This will detect and throw error for the first time , when we try to create a new table and the data in catalog is not even updated. But what about when a query runs to update the data. In that case stale data will appear when queried by spark.

vaibhav-datazip: Resolved this issue, by adding a check for number of rows with the op symbol , as op symbol is unique for every test case.

If no rows for op symbol exists then retry will be done.

shubham19may (approved):

PR #602, opened 2025-10-25
fix: new writer instance on retry & refactor interface

hash-data:
# Description
Previously, the writer instance was not created anew, which tends to fail, and in some cases, it pushes failure data that is not expected. 

Fixes # (issue)

In this pr, we have made sure that on retry, all the artifacts are new and are not being affected by previous failures.
## Type of change

# How Has This Been Tested?

- [x] Failing source for parquet writer (delete prev created file and on retry create a clean one)
- [x] Failing source on iceberg writer (does not commit data files, only commits after retry and success)

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

---

> [!NOTE]
> Reworks retry flows to recreate fresh writer instances with per-attempt contexts, adds error-aware Close(closeOnError) to writers, and centralizes backoff logic.
> 
> - **Core/Retry**:
>   - Centralize retry helper to `utils.RetryOnBackoff(attempt int)`; remove `drivers/abstract/utils.go` and update all callers.
>   - On retries, cancel prior contexts, close writers with `closeOnError=true`, and recreate writers with new `threadID`s (Backfill, CDC, Incremental, Kafka paths).
> - **Destination API**:
>   - Change `Writer.Close(ctx)` → `Close(ctx, closeOnError bool)` in `destination.Writer` and implementations.
>   - Parquet: `closePqFiles(closeOnError)` deletes local files on retry; S3 upload only on non-error close.
>   - Iceberg: `Close(ctx, closeOnError)` skips commit on error and always closes client.
>   - Update all call sites (`Check`, `DropStreams`, `WriterThread.Close/Push`).
> - **Drivers**:
>   - Use `mainCtx`; create per-operation contexts for retries (backfill/incremental/cdc), re-init writers and CDC sessions on retry; improved PostCDC/PreCDC handling.
>   - Incremental: standardize cursor persistence via `formatTimestampToUTC`.
>   - Mongo: use new retry helper when splitting chunks.
> - **Utils/Typeutils**:
>   - Date parsing hardens: normalize years to supported range; fallback returns epoch without warn.
> - **Protocol**:
>   - Log note that read counts include retried records.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0c95744c8379652747f446c3aa7e5d39e08d7b39. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/drivers/abstract/abstract.go
+++ b/drivers/abstract/abstract.go
@@ -89,7 +89,7 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error) 
 
 		// add default columns
 		for column, typ := range DefaultColumns {
-			if column == constants.CdcTimestamp && !a.driver.CDCSupported() {
+			if column == constants.CdcTimestamp && !a.supportsCdcColumn() {
 				continue
 			}
 			convStream.UpsertField(column, typ, true)
@@ -222,8 +222,10 @@ func (a *AbstractDriver) waitForBackfillCompletion(mainCtx context.Context, back
 }
 
 // generateThreadID creates a unique thread ID for a stream
-func generateThreadID(streamID string) string {
-	return fmt.Sprintf("%s_%s", streamID, utils.ULID())
+func generateThreadID(streamID string, suffix string) string {
+	withSuffix := fmt.Sprintf("%s_%s_%s", streamID, utils.ULID(), suffix)
+	withoutSuffix := fmt.Sprintf("%s_%s", streamID, utils.ULID())
+	return utils.Ternary(suffix != "", withSuffix, withoutSuffix).(string)
 }
 
 // handleWriterCleanup is a helper that creates a defer function for common writer cleanup operations
@@ -285,4 +287,12 @@ func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *er
 			*err = fmt.Errorf("thread[%s]: %s", threadID, *err)
 		}
 	}
+}
+
+func (a *AbstractDriver) supportsCdcColumn() bool {
+	if a.driver.CDCSupported() && a.driver.Type() != string(constants.Kafka) {
+		// kafka driver does not support cdc column
+		return true
+	}
+	return false
 }
--- a/drivers/abstract/backfill.go
+++ b/drivers/abstract/backfill.go
@@ -44,7 +44,7 @@ func (a *AbstractDriver) Backfill(mainCtx context.Context, backfilledStreams cha
 		backfillCtx, backfillCtxCancel := context.WithCancel(gCtx)
 		defer backfillCtxCancel()
 
-		threadID := generateThreadID(stream.ID())
+		threadID := generateThreadID(stream.ID(), fmt.Sprintf("min[%v]-max[%v]", chunk.Min, chunk.Max))
 		inserter, err := pool.NewWriter(backfillCtx, stream, destination.WithBackfill(true), destination.WithThreadID(threadID))
 		if err != nil {
 			return fmt.Errorf("failed to create new writer thread: %s", err)
--- a/drivers/abstract/cdc.go
+++ b/drivers/abstract/cdc.go
@@ -114,7 +114,7 @@ func (a *AbstractDriver) streamChanges(mainCtx context.Context, pool *destinatio
 	return a.driver.StreamChanges(cdcCtx, streamIndex, func(ctx context.Context, change CDCChange) error {
 		writer := writers[change.Stream.ID()]
 		if writer == nil {
-			threadID := fmt.Sprintf("%d_%s", streamIndex, change.Stream.ID())
+			threadID := generateThreadID(change.Stream.ID(), "")
 			writer, err = pool.NewWriter(ctx, change.Stream, destination.WithThreadID(threadID))
 			if err != nil {
 				return fmt.Errorf("failed to create writer for stream %s: %s", change.Stream.ID(), err)
--- a/drivers/abstract/incremental.go
+++ b/drivers/abstract/incremental.go
@@ -73,7 +73,7 @@ func (a *AbstractDriver) Incremental(mainCtx context.Context, pool *destination.
 			incrementalCtx, incrementalCtxCancel := context.WithCancel(gCtx)
 			defer incrementalCtxCancel()
 
-			threadID := generateThreadID(stream.ID())
+			threadID := generateThreadID(stream.ID(), "")
 			inserter, err := pool.NewWriter(incrementalCtx, stream, destination.WithThreadID(threadID))
 			if err != nil {
 				return fmt.Errorf("failed to create new writer thread: %s", err)


destination/iceberg/iceberg.go
```diff
@@ -221,7 +221,7 @@ func (i *Iceberg) Write(ctx context.Context, records []types.RawRecord) error {
 	return nil
 }
 
-func (i *Iceberg) Close(ctx context.Context) error {
```
shubham19may, line 151: why need of the extra bool, can't we detect it from ctx directly as it will get cancelled for sure.

destination/parquet/parquet.go
```diff
@@ -245,6 +245,9 @@ func (p *Parquet) closePqFiles() error {
 	for basePath, parquetFile := range p.partitionedFiles {
 		// construct full file path
 		filePath := filepath.Join(p.config.Path, basePath, parquetFile.fileName)
+		if closeOnError {
+			removeLocalFile(filePath, "closing parquet files due to retry attempt")
```
shubham19may: return?

shubham19may:
Potential context leak on retries: previous attempt contexts aren’t canceled before reassign

In Backfill, CDC (both parallel and non-parallel paths), and Incremental, each retry creates a new per-attempt context but the previous attempt’s cancel func isn’t called before reassigning. The initial `defer cancel()` only cancels the first attempt, so subsequent attempt contexts (with timers/goroutines) can leak.

Why this matters:
- Each attempt sets up new goroutines and timers (e.g., stream readers, flush contexts). Not canceling the previous context before creating the new one can keep them alive unnecessarily until they finish or the parent context is canceled.

Fix pattern:
- Before reassigning `{backfill,cdc,incremental}Ctx, {backfill,cdc,incremental}CtxCancel = context.WithCancel(ctx)`, invoke the previous cancel: e.g.,
  - `cdcCtxCancel()`; then `cdcCtx, cdcCtxCancel = context.WithCancel(ctx)`

Relevant spots (apply the same pattern):
- drivers/abstract/backfill.go
  - In the retry block (`attempt > 0`), before reassigning `backfillCtx, backfillCtxCancel = context.WithCancel(ctx)`, call `backfillCtxCancel()`.
- drivers/abstract/cdc.go (parallel per-stream path)
  - In the retry block, before `cdcCtx, cdcCtxCancel = context.WithCancel(ctx)`, call `cdcCtxCancel()`.
- drivers/abstract/cdc.go (single-stream aggregator path)
  - In the retry block, before `cdcCtx, cdcCtxCancel = context.WithCancel(ctx)`, call `cdcCtxCancel()`.
- drivers/abstract/incremental.go
  - In the retry block, before `incrementalCtx, incrementalCtxCancel = context.WithCancel(ctx)`, call `incrementalCtxCancel()`.

Optional enhancements:
- Consider a small helper that returns a closure to cancel previous attempt before creating the next, to avoid repetition and mistakes.

shubham19may:
Possible resource leak in WriterThread.Close on flush error

In `destination/writers.go` → `WriterThread.Close`, if `wt.group.Block()` returns an error, the function returns immediately and does not call `writer.Close(...)`. This can leak destination-side resources (e.g., open files, S3 handles, gRPC client connections) when a close is attempted after a failed flush.

Current flow (simplified):
- schedule final flush
- `if err := wt.group.Block(); err != nil { return err }`
- only on success → lock and `return wt.writer.Close(ctx, closeOnError)`

Suggestion:
- Ensure the underlying writer is always closed even when the flush fails, using `closeOnError=true` in the error path. A simple, robust approach is to use a `defer` that calls `writer.Close(ctx, true)` and then, on the success path, do a final `writer.Close(ctx, closeOnError)` (or gate the second call) to keep semantics intact.

Example pattern:
```go
// always attempt to free destination resources, even on error
var finalCloseOnce sync.Once
defer func() { finalCloseOnce.Do(func() { _ = wt.writer.Close(ctx, true) }) }()

wt.group.Add(func(ctx context.Context) error { return wt.flush(ctx, wt.buffer) })
if err := wt.group.Block(); err != nil {
    // early return, but defer will still close with closeOnError=true
    return fmt.Errorf("failed to flush data while closing: %w", err)
}

// success path: close normally and suppress the defer’d fallback
wt.streamArtifact.mu.Lock()
defer wt.streamArtifact.mu.Unlock()
err := wt.writer.Close(ctx, closeOnError)
finalCloseOnce.Do(func() {}) // no-op to consume the fallback
return err
```

This guarantees cleanup while preserving the intended behavior for normal closes.

shubham19may:
Make retry backoff cancel-aware and add jitter to avoid stampedes

In `utils/utils.go` → `RetryOnBackoff`, the backoff uses `time.Sleep(sleep)` directly. This sleep is not interruptible by context, so if the parent context is canceled during the sleep, the goroutine keeps sleeping until the attempt finishes. Also, without jitter, multiple workers can synchronize and hammer dependencies at the same times.

Suggestions:
1) Make the backoff cancel-aware and add jitter:

```go
// consider signature change to accept ctx
func RetryOnBackoff(ctx context.Context, attempts int, base time.Duration, f func(attempt int) error) error {
    var err error
    for i := 0; i < attempts; i++ {
        if err = f(i); err == nil {
            return nil
        }
        // stop on non-retryable errors
        if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
            return err
        }
        if i == attempts-1 { // no more retries
            break
        }
        // exponential backoff with full jitter
        max := base << i
        jitter := time.Duration(rand.Int63n(int64(max)))
        select {
        case <-ctx.Done():
            return ctx.Err()
        case <-time.After(jitter):
        }
    }
    return err
}
```

- Call sites (e.g., Backfill/CDC/Incremental) should pass their attempt context (`backfillCtx`/`cdcCtx`/`incrementalCtx`).
- With this, the string matching for "context canceled" can be dropped in favor of `errors.Is`.

2) Alternatively, use a library:
- `github.com/cenkalti/backoff/v4` with `backoff.WithContext(ctx, eb)` provides exponential backoff and jitter out-of-the-box.

utils/utils.go
```diff
@@ -424,3 +424,22 @@ func GenerateDestinationDetails(namespace, name string, sourceDatabase *string)
...
 	return dbName, Reformat(name)
 }
+
+// RetryOnBackoff retries the function f up to attempts times with a backoff sleep between attempts.
+func RetryOnBackoff(attempts int, sleep time.Duration, f func(attempt int) error) (err error) {
+	for cur := range attempts {
+		if err = f(cur); err == nil {
+			return nil
+		}
+		if strings.Contains(err.Error(), constants.DestError) || strings.Contains(err.Error(), "context canceled") {
```
shubham19may: this is not even getting used anywhere write? i mean dest error thing? is there some dest error string that is prefixed when we throw the error so it can be catched in this condition?

destination/iceberg/iceberg.go
```diff
@@ -233,6 +233,10 @@ func (i *Iceberg) Close(ctx context.Context) error {
 		}
 	}()
 
+	if closeOnError {
```
shubham19may: also we do need to check ctx.error here? coz if the global context gets closed because some other thread goes in error?

destination/writers.go
```diff
@@ -252,10 +252,10 @@ func (wt *WriterThread) flush(ctx context.Context, buf []types.RawRecord) (err e
 	return nil
 }
 
-func (wt *WriterThread) Close(ctx context.Context) error {
+func (wt *WriterThread) Close(ctx context.Context, closeOnError bool) error {
 	select {
 	case <-ctx.Done():
-		return fmt.Errorf("context closed")
+		return wt.writer.Close(ctx, closeOnError)
```
shubham19may: how does the closeOnError matter here? its anyways main or subcontext is errored.

destination/iceberg/iceberg.go
hash-data: it is for the cases when ctx not cancelled, let us say some logic returning error on certain condition failing. like lsn mismatch etc.

destination/parquet/parquet.go
hash-data: continue actually, I test this previously and it was able to close files. I guess the array was not filled at that time. Updating this

destination/iceberg/iceberg.go
hash-data: This is for a specific thread. Even for other contexts cancel out, the global will not cancel until all thread return. We are separating ctx in each thread now.

destination/writers.go
hash-data: it is same question right? 
image

utils/utils.go
hash-data: checking it

utils/utils.go
```diff
@@ -436,3 +436,27 @@ func SplitAndTrim(s string) []string {
...
+		if err = f(cur); err == nil {
+			return nil
+		}
+
+		// check if error is non retryable
+		for _, nonRetryableError := range constants.NonRetryableErrors {
+			if strings.Contains(err.Error(), nonRetryableError) {
+				break
+			}
+		}
```
hash-data: this is breaking inner loop only

destination/iceberg/iceberg.go
shubham19may: Lets discuss this once. Ideally we can detect if its a retryable error or not from ctx.err. 

why are we handling it separately.

destination/iceberg/iceberg.go
shubham19may: shouldn't we kill the java process here?

destination/writers.go
shubham19may: lets discuss

destination/parquet/parquet.go
```diff
@@ -299,8 +305,13 @@ func (p *Parquet) closePqFiles() error {
 	return nil
 }
 
-func (p *Parquet) Close(_ context.Context) error {
-	return p.closePqFiles()
+func (p *Parquet) Close(ctx context.Context, closeOnError bool) error {
+	select {
+	case <-ctx.Done():
```
shubham19may: reason why we are not checking ctx.done in Iceberg?

drivers/abstract/backfill.go
```diff
@@ -39,16 +39,22 @@ func (a *AbstractDriver) Backfill(ctx context.Context, backfilledStreams chan st
...
 			return fmt.Errorf("failed to create new writer thread: %s", err)
 		}
+
 		logger.Infof("Thread[%s]: created writer for chunk min[%s] and max[%s] of stream %s", threadID, chunk.Min, chunk.Max, stream.ID())
+
 		defer func() {
 			// wait for chunk completion
-			if writerErr := inserter.Close(ctx); writerErr != nil {
+			if writerErr := inserter.Close(backfillCtx, err != nil); writerErr != nil {
 				err = fmt.Errorf("failed to insert chunk min[%s] and max[%s] of stream %s, insert func error: %s, thread error: %s", chunk.Min, chunk.Max, stream.ID(), err, writerErr)
```
shubham19may:
```suggestion
				err = fmt.Errorf("failed to insert chunk min[%s] and max[%s] of stream %s, insert func error: %s, writer close error: %s", chunk.Min, chunk.Max, stream.ID(), err, writerErr)
```

drivers/abstract/cdc.go
shubham19may: uses wrong context

drivers/abstract/cdc.go
```diff
@@ -123,13 +151,37 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
...
+					// close all prev writers
+					for _, inserter := range inserters {
+						_ = inserter.Close(cdcCtx, true)
+					}
+
+					// cancel previous cdc context
+					cdcCtxCancel()
+
+					// create new cdc context
+					cdcCtx, cdcCtxCancel = context.WithCancel(gCtx)
```
shubham19may: need to also do preCdc again, as its going to recreate kafka readers. (Previous readers will already would have crossed some x% of messages using fetchMessage method, so they will only give new messages after that)

drivers/abstract/cdc.go
```diff
@@ -123,13 +151,37 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
...
+
```
shubham19may, line 76: don't you think we need preCdc here as well? we need to reinitialize the readers so they read from last LSN committed as if we don't reinitialise them, it will keep on reading from the point it failed (which will make it miss the messages)

utils/utils.go
```diff
@@ -436,3 +436,27 @@ func SplitAndTrim(s string) []string {
...
+func RetryOnBackoff(attempts int, sleep time.Duration, f func(attempt int) error) (err error) {
+	for cur := range attempts {
+		if err = f(cur); err == nil {
+			return nil
+		}
+
+		// check if error is non retryable
+		for _, nonRetryableError := range constants.NonRetryableErrors {
+			if strings.Contains(err.Error(), nonRetryableError) {
+				return nil
```
shubham19may: shouldn't we return error?

drivers/abstract/backfill.go
```diff
@@ -39,16 +39,22 @@ func (a *AbstractDriver) Backfill(ctx context.Context, backfilledStreams chan st
 	})
 	logger.Infof("Starting backfill for stream[%s] with %d chunks", stream.GetStream().Name, len(chunks))
 	// TODO: create writer instance again on retry
-	chunkProcessor := func(ctx context.Context, chunk types.Chunk) (err error) {
+	chunkProcessor := func(gCtx context.Context, chunk types.Chunk) (err error) {
+		// create backfill context, so that main context not affected if backfill retries
+		backfillCtx, backfillCtxCancel := context.WithCancel(gCtx)
```
shubham19may, line 44: > The defer in outer scope only calls the LAST assigned backfillCtxCancel, but if retry logic fails partway, previous cancel functions might not be stored.

Need to understand this AI comment

drivers/abstract/backfill.go
```diff
@@ -67,10 +73,28 @@ func (a *AbstractDriver) Backfill(ctx context.Context, backfilledStreams chan st
...
+			if attempt > 0 {
+				// close prev writer
+				_ = inserter.Close(backfillCtx, true)
+
+				// cancel previous backfill context
+				backfillCtxCancel()
+
+				// create new backfill context
+				backfillCtx, backfillCtxCancel = context.WithCancel(gCtx)
+				threadID = utils.Ternary(attempt == 1, fmt.Sprintf("%s-retry-attempt", threadID), threadID).(string)
```
shubham19may: reason why we are not updating the thread id on attempt 2, 3?

drivers/abstract/cdc.go
```diff
@@ -181,13 +238,44 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
...
 				err = fmt.Errorf("post cdc error: %s, cdc insert thread error: %s", postCDCErr, err)
 			}
 		}()
-		return RetryOnBackoff(a.driver.MaxRetries(), constants.DefaultRetryTimeout, func() error {
-			return a.driver.StreamChanges(ctx, nil, func(ctx context.Context, change CDCChange) error {
+
+		return utils.RetryOnBackoff(a.driver.MaxRetries(), constants.DefaultRetryTimeout, func(attempt int) error {
+			if attempt > 0 {
+				// close previous cdc session
+				_ = a.driver.PostCDC(cdcCtx, nil, false, "")
```
shubham19may: why? can be dangerous as we have not committed data in iceberg but marking state with this cursor

destination/writers.go
```diff
@@ -270,7 +279,11 @@ func (wt *WriterThread) Close(ctx context.Context) error {
 		wt.streamArtifact.mu.Lock()
 		defer wt.streamArtifact.mu.Unlock()
 
-		return wt.writer.Close(ctx)
+		err := wt.writer.Close(ctx, closeOnError)
+		if err != nil {
+			return fmt.Errorf("%s: %s", constants.DestError, err)
```
shubham19may: should we use %w?

shubham19may:
Follow-up review after requested changes

Thanks for the updates — a few remaining items:

1) Missing cancel of previous attempt contexts in some paths
- drivers/abstract/cdc.go (parallel per-stream path):
  In the retry block (`attempt > 0`), you reassign `cdcCtx, cdcCtxCancel = context.WithCancel(gCtx)` without canceling the previous attempt’s `cdcCtx`. Please call `cdcCtxCancel()` before reassigning.

  Reference area:
  ```
  return utils.RetryOnBackoff(..., func(attempt int) error {
      if attempt > 0 {
          _ = inserter.Close(cdcCtx, true)
          // MISSING: cdcCtxCancel()
          cdcCtx, cdcCtxCancel = context.WithCancel(gCtx)
          ...
      }
  })
  ```

- drivers/abstract/incremental.go:
  Similarly, in the retry block you reassign `incrementalCtx` without canceling the previous attempt. Invoke `incrementalCtxCancel()` before `incrementalCtx, incrementalCtxCancel = context.WithCancel(gCtx)`.

2) RetryOnBackoff returns nil on non-retryable errors (hides failures)
- utils/utils.go → `RetryOnBackoff`:
  You now short-circuit on `constants.NonRetryableErrors`, but you `return nil`. That converts a hard failure into success for the caller. This is dangerous because the outer defers often use `err == nil` to commit/update state.

  Fix:
  - Return the encountered error instead of nil:
  ```go
  for _, nonRetryableError := range constants.NonRetryableErrors {
      if strings.Contains(err.Error(), nonRetryableError) {
          return err // not nil
      }
  }
  ```
  - Note the interaction with places that wrap errors as `"destination error: ..."` (e.g., WriterThread). Since `DestError` is in `NonRetryableErrors`, this would currently short-circuit to success; returning the error will correctly bubble it up.

3) Ensure WriterThread always frees destination resources
- destination/writers.go → `WriterThread.Close`:
  - On `wt.group.Block()` error, the function returns without calling `writer.Close(...)`. Add a defer fallback to always close the writer with `closeOnError=true` so destination resources are freed.
  - On the `ctx.Done()` branch, explicitly force `closeOnError` (or OR with `ctx.Err()!=nil`) so a canceled context cannot accidentally be treated as a normal close.

  Suggested pattern:
  ```go
  var closedOnce sync.Once
  defer func() { closedOnce.Do(func() { _ = wt.writer.Close(ctx, true) }) }()

  wt.group.Add(func(ctx context.Context) error { return wt.flush(ctx, wt.buffer) })
  if err := wt.group.Block(); err != nil {
      return fmt.Errorf("failed to flush data while closing: %w", err)
  }

  wt.streamArtifact.mu.Lock()
  defer wt.streamArtifact.mu.Unlock()
  err := wt.writer.Close(ctx, closeOnError)
  closedOnce.Do(func(){}) // consume fallback
  return err
  ```

  And in the canceled branch:
  ```go
  return wt.writer.Close(ctx, true)
  // or: return wt.writer.Close(ctx, closeOnError || ctx.Err() != nil)
  ```

4) Iceberg.Close: short-circuit on canceled context
- destination/iceberg/iceberg.go → `Close`:
  Add a guard to skip commit when the parent ctx is already canceled, not just when `closeOnError` is true.

  ```go
  if closeOnError || ctx.Err() != nil {
      return nil
  }
  ```

5) (Reminder) Backoff cancel-awareness and jitter
- The backoff still uses `time.Sleep` with no jitter and is not cancel-aware. Please consider the earlier recommendation to make it context-aware and add jitter, or use `cenkalti/backoff` with `backoff.WithContext(ctx, ...)`.

Happy to re-check once these are in.

drivers/abstract/backfill.go
hash-data: consistent check

drivers/abstract/cdc.go
```diff
@@ -123,13 +151,37 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
...
+					for streamID := range inserters {
+						streamIdx, _ := utils.ArrayContains(streams, func(s types.StreamInterface) bool { return s.ID() == streamID })
+						// no relation with prev thread id, should not we also provide prev thread id for reference?
+						threadID := utils.Ternary(attempt == 1, fmt.Sprintf("%s_%s-retry-attempt", streamID, utils.ULID()), fmt.Sprintf("%s_%s", streamID, utils.ULID())).(string)
+						inserters[streamID], err = pool.NewWriter(cdcCtx, streams[streamIdx], destination.WithThreadID(threadID))
+						if err != nil {
+							return fmt.Errorf("failed to create new writer thread: %s", err)
+						}
+					}
+				}
```
hash-data:
```suggestion
	// Global CDC (MySQL/Postgres)
	a.startCDCTask(pool, "global", streams,
		func(ctx context.Context, send CDCMsgFn) error {
			return a.driver.StreamChanges(ctx, nil, send)
		},
		func(ctx context.Context, success bool) error {
			return a.driver.PostCDC(ctx, nil, success, "")
		},
	)
	return nil
}
```

drivers/abstract/cdc.go
```diff
@@ -155,12 +207,16 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
 	}
 
 	// TODO: For a big table cdc (for all tables) will not start until backfill get finished, need to study alternate ways to do cdc sync
-	a.GlobalConnGroup.Add(func(ctx context.Context) (err error) {
+	a.GlobalConnGroup.Add(func(gCtx context.Context) (err error) {
+		// create cdc context, so that main context not affected if cdc retries
```
hash-data:
```suggestion
// runCDCWithRetry handles the common retry logic, writer management, and error handling for CDC sessions.
func (a *AbstractDriver) runCDCWithRetry(
	gCtx context.Context,
	pool *destination.WriterPool,
	threadPrefix string,
	retryStreams []types.StreamInterface, // streams to pass to PreCDC on retry
	streamAction func(ctx context.Context, getWriter func(types.StreamInterface) (*destination.WriterThread, error)) error,
	postCDC func(ctx context.Context, success bool) error,
) error {
	return utils.RetryOnBackoff(a.driver.MaxRetries(), constants.DefaultRetryTimeout, func(attempt int) (err error) {
```

drivers/abstract/cdc.go
```diff
@@ -169,9 +225,10 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
 		if err != nil {
 			return fmt.Errorf("failed to create writer thread: %s", err)
 		}
+
```
hash-data:
```suggestion

		// Cleanup closure
```

drivers/abstract/abstract.go
```diff
@@ -154,44 +149,149 @@ func (a *AbstractDriver) ClearState(streams []types.StreamInterface) (*types.Sta
...
+// handleWriterCleanup is a helper that creates a defer function for common writer cleanup operations
+// It handles writer close (single or multiple), panic recovery, and calls the provided postProcess function
+// The err parameter should be a pointer to the error variable that will be returned from the function
+// The cancel parameter is used to cancel the context when an error occurs, so other threads can detect the failure
+// The writer parameter can be either:
+//   - *destination.WriterThread for a single writer
+//   - map[string]*destination.WriterThread for multiple writers keyed by stream ID
+//
+// The threadID and closeMessage parameters are optional (empty string means not used) and only apply to single writer cases
+func (a *AbstractDriver) handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *error, writer interface{}, threadID string, postProcess func(ctx context.Context) error) func() {
```
ImDoubD-datazip: no need to provide abstract driver

drivers/abstract/cdc.go
```diff
@@ -37,171 +38,136 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
...
 
-	// handle for kafka differently
-	kafkaDriver, ok := a.GetKafkaInterface()
-	if ok {
-		// set limit to global connection group according to readers available
-		a.GlobalConnGroup = utils.NewCGroupWithLimit(ctx, len(kafkaDriver.GetReaderIDs()))
-		utils.ConcurrentInGroup(a.GlobalConnGroup, kafkaDriver.GetReaderIDs(), func(ctx context.Context, readerID string) (err error) {
+	if parallel {
+		// reset the global connection group
+		a.GlobalConnGroup = utils.NewCGroupWithLimit(mainCtx, a.driver.MaxConnections())
```
ImDoubD-datazip, line 79: no need for this if we setting the readers count in the max connection only.

drivers/abstract/abstract.go
```diff
@@ -154,44 +149,149 @@ func (a *AbstractDriver) ClearState(streams []types.StreamInterface) (*types.Sta
...
+					}
+				}
+			}
+		default:
+			closeErr = fmt.Errorf("unsupported writer type")
+		}
+
+		if closeErr != nil {
+			*err = fmt.Errorf("%w: %s, prev error: %v", constants.ErrNonRetryable, closeErr, *err)
+		}
```
hash-data, line 269: remove this retry

drivers/abstract/abstract.go
```diff
@@ -154,44 +149,149 @@ func (a *AbstractDriver) ClearState(streams []types.StreamInterface) (*types.Sta
...
+
+		postErr := postProcess(ctx)
+		if postErr != nil {
+			*err = fmt.Errorf("%w: post process error: %s, prev error: %v", constants.ErrNonRetryable, postErr, *err)
+		}
+
+		if *err != nil && threadID != "" {
+			*err = fmt.Errorf("%w: thread[%s]: %s", constants.ErrNonRetryable, threadID, *err)
+		}
 	}
```
hash-data, line 289: remove from here as well

drivers/kafka/internal/kafka.go
```diff
@@ -129,6 +136,8 @@ func (k *Kafka) ProduceSchema(_ context.Context, streamName string) (*types.Stre
 	schema.AddTypes(KafkaTimestamp, types.Int64) // Message timestamp
 	stream.WithSchema(schema)
 	stream.SourceDefinedPrimaryKey = types.NewSet(Offset, Partition)
+
+	stream.WithSyncMode(types.STRICTCDC)
```
ImDoubD-datazip: this is not needed, do check line 130, it is already there.

drivers/mongodb/internal/mon.go
```diff
@@ -46,6 +47,10 @@ func (m *Mongo) CDCSupported() bool {
 	return m.CDCSupport
 }
 
+func (m *Mongo) ChangeStreamConfig() (bool, bool, bool) {
+	return true, false, true // concurrent change streams supported, stream can start after finishing full load
```
ImDoubD-datazip: why sequential is true?

drivers/postgres/internal/cdc.go
```diff
@@ -75,26 +76,28 @@ func (p *Postgres) PreCDC(ctx context.Context, streams []types.StreamInterface)
 			// failing sync when lsn mismatch found (from state and confirmed flush lsn), as otherwise on backfill, duplication of data will occur
 			// suggesting to proceed with clear destination
 			if parsed != socket.ConfirmedFlushLSN {
-				return fmt.Errorf("lsn mismatch, please proceed with clear destination. lsn saved in state [%s] current lsn [%s]", parsed, socket.ConfirmedFlushLSN)
+				return fmt.Errorf("%w: lsn mismatch, please proceed with clear destination. lsn saved in state [%s] current lsn [%s]", constants.ErrNonRetryable, parsed, socket.ConfirmedFlushLSN)
```
ImDoubD-datazip: please remove %w, we need to follow %s convention.

pkg/waljs/waljs.go
```diff
@@ -65,7 +65,7 @@ func (w *wal2jsonReplicator) StreamChanges(ctx context.Context, db *sqlx.DB, cal
 			return nil
 		default:
 			if !messageReceived && w.socket.initialWaitTime > 0 && time.Since(cdcStartTime) > w.socket.initialWaitTime {
-				return fmt.Errorf("%s, try increasing it or do full load", constants.NoRecordsFoundError)
+				return fmt.Errorf("%w, try increasing it or do full load", constants.ErrNonRetryable)
```
ImDoubD-datazip, line 59: please make this %s

drivers/abstract/abstract.go
```diff
@@ -154,44 +149,149 @@ func (a *AbstractDriver) ClearState(streams []types.StreamInterface) (*types.Sta
...
+			return fmt.Errorf("error occurred while waiting for connections: %s", err)
+		}
+		return nil
 	}
 
-	// wait for all threads to finish
-	if err := a.GlobalCtxGroup.Block(); err != nil {
-		return fmt.Errorf("error occurred while waiting for context groups: %s", err)
+	// TODO: discuss for moving max retry as flag (job level in ui)
+	return utils.RetryOnBackoff(a.driver.MaxRetries(), constants.DefaultRetryTimeout, readFunc)
```
ImDoubD-datazip: all can we comment the list of all the errors on which we need to retry

destination/writers.go
```diff
@@ -255,7 +256,11 @@ func (wt *WriterThread) flush(ctx context.Context, buf []types.RawRecord) (err e
 func (wt *WriterThread) Close(ctx context.Context) error {
 	select {
 	case <-ctx.Done():
-		return fmt.Errorf("context closed")
+		err := wt.writer.Close(ctx)
+		if err != nil {
+			return fmt.Errorf("%w: %s", constants.ErrNonRetryable, err)
```
ImDoubD-datazip: here if the driver side error occurs, cancel takes place in handleWriterCleanup then writer.close is called, and there is an error in writer closing, then the driver side error will be discarded and retry won't take place.

destination/writers.go
```diff
@@ -255,7 +256,11 @@ func (wt *WriterThread) flush(ctx context.Context, buf []types.RawRecord) (err e
...
+		}
```
ImDoubD-datazip, line 261: we need to set some sort of priority on the errors as well to handle this.

pkg/waljs/waljs.go
ImDoubD-datazip: for isError to work, error wrapping must be done using %w.

destination/parquet/parquet.go
shubham19may, line 414: what if the context got cancelled here?

destination/writers.go
shubham19may: lets discuss

drivers/abstract/abstract.go
```diff
@@ -91,23 +87,22 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
...
 	streamMap.Range(func(_, value any) bool {
 		convStream, _ := value.(*types.Stream)
-		if convStream.SupportedSyncModes.Len() == 0 {
-			convStream.WithSyncMode(types.FULLREFRESH, types.INCREMENTAL)
-		}
-		convStream.SyncMode = utils.Ternary(convStream.SyncMode == "", types.FULLREFRESH, convStream.SyncMode).(types.SyncMode)
 
 		// add default columns
 		for column, typ := range DefaultColumns {
+			if column == constants.CdcTimestamp && !a.driver.CDCSupported() {
```
shubham19may: this is not done here previously, why add now?

drivers/abstract/abstract.go
```diff
@@ -91,23 +87,22 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
...
-			convStream.WithSyncMode(types.CDC, types.STRICTCDC)
+		// priority to default sync mode (cdc -> incremental -> strict_cdc)
+		if convStream.SupportedSyncModes.Exists(types.CDC) && a.driver.CDCSupported() {
 			convStream.SyncMode = types.CDC
-		} else {
-			// remove cdc column as it is not supported
-			convStream.Schema.Properties.Delete(constants.CdcTimestamp)
+		} else if convStream.SupportedSyncModes.Exists(types.INCREMENTAL) {
+			convStream.SyncMode = types.INCREMENTAL
+		} else if convStream.SupportedSyncModes.Exists(types.STRICTCDC) {
```
shubham19may, line 103:
```suggestion
		} else if convStream.SupportedSyncModes.Exists(types.STRICTCDC) && a.driver.CDCSupported() {
```

drivers/abstract/abstract.go
```diff
@@ -91,23 +87,22 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
 	var finalStreams []*types.Stream
 	streamMap.Range(func(_, value any) bool {
 		convStream, _ := value.(*types.Stream)
-		if convStream.SupportedSyncModes.Len() == 0 {
```
shubham19may, line 94: is this not required anymore?

drivers/abstract/abstract.go
```diff
@@ -91,23 +87,22 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
...
+			if column == constants.CdcTimestamp && !a.driver.CDCSupported() {
+				continue
+			}
 			convStream.UpsertField(column, typ, true)
 		}
 
-		_, isKafkaDriver := a.GetKafkaInterface()
-		if a.driver.CDCSupported() && !isKafkaDriver {
-			convStream.WithSyncMode(types.CDC, types.STRICTCDC)
+		// priority to default sync mode (cdc -> incremental -> strict_cdc)
```
shubham19may, line 98: shouldn't it be incremental -> cdc -> strict_cdc ?

drivers/abstract/abstract.go
```diff
@@ -154,44 +149,149 @@ func (a *AbstractDriver) ClearState(streams []types.StreamInterface) (*types.Sta
...
+			logger.Infof("Retrying Read operation (attempt %d)", attempt)
+		}
 
-	// run cdc sync
-	if len(cdcStreams) > 0 {
-		if a.driver.CDCSupported() {
-			if err := a.RunChangeStream(ctx, pool, cdcStreams...); err != nil {
-				return fmt.Errorf("failed to run change stream: %s", err)
-			}
+		if a.driver.MaxConnections() > 0 {
```
shubham19may: use ternary operator

drivers/abstract/abstract.go
```diff
@@ -154,44 +149,149 @@ func (a *AbstractDriver) ClearState(streams []types.StreamInterface) (*types.Sta
...
+		// run incremental sync
+		if len(incrementalStreams) > 0 {
+			if err := a.Incremental(ctx, pool, incrementalStreams...); err != nil {
+				return fmt.Errorf("failed to run incremental sync: %s", err)
+			}
+		}
+
+		// handle standard streams (full refresh)
+		for _, stream := range backfillStreams {
+			stream := stream // capture for closure
```
shubham19may: reason?

drivers/abstract/abstract.go
shubham19may: but this will retry all streams?

drivers/abstract/abstract.go
```diff
@@ -154,44 +149,149 @@ func (a *AbstractDriver) ClearState(streams []types.StreamInterface) (*types.Sta
...
+						closeErr = fmt.Errorf("%s; failed closing writer[%s]: %s", closeErr, streamID, threadErr)
+					}
+				}
+			}
+		default:
+			closeErr = fmt.Errorf("unsupported writer type")
+		}
+
+		if closeErr != nil {
+			*err = fmt.Errorf("%w: %s, prev error: %v", constants.ErrNonRetryable, closeErr, *err)
```
shubham19may: s3 rate limits can be retryable error right, which can very well occur in writer close?

drivers/abstract/abstract.go
```diff
@@ -154,44 +149,149 @@ func (a *AbstractDriver) ClearState(streams []types.StreamInterface) (*types.Sta
...
+// The err parameter should be a pointer to the error variable that will be returned from the function
+// The cancel parameter is used to cancel the context when an error occurs, so other threads can detect the failure
+// The writer parameter can be either:
+//   - *destination.WriterThread for a single writer
+//   - map[string]*destination.WriterThread for multiple writers keyed by stream ID
+//
+// The threadID and closeMessage parameters are optional (empty string means not used) and only apply to single writer cases
+func (a *AbstractDriver) handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *error, writer interface{}, threadID string, postProcess func(ctx context.Context) error) func() {
+	return func() {
+		if cancel == nil {
```
shubham19may: any reason why this will be nil?

drivers/abstract/cdc.go
```diff
@@ -4,19 +4,20 @@ import (
...
 
-func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.WriterPool, streams ...types.StreamInterface) error {
+func (a *AbstractDriver) RunChangeStream(mainCtx context.Context, pool *destination.WriterPool, streams ...types.StreamInterface) error {
 	// run pre cdc of drivers
-	if err := a.driver.PreCDC(ctx, streams); err != nil {
+	if err := a.driver.PreCDC(mainCtx, streams); err != nil {
 		return fmt.Errorf("failed in pre cdc run for driver[%s]: %s", a.driver.Type(), err)
 	}
 
+	sequential, parallel, concurrent := a.driver.ChangeStreamConfig()
```
shubham19may: what is difference between parallel and concurrent. both grammar wise means the same

drivers/abstract/interface.go
```diff
@@ -35,13 +35,8 @@ type DriverInterface interface {
...
-	StreamChanges(ctx context.Context, stream types.StreamInterface, processFn CDCMsgFn) error
-	PostCDC(ctx context.Context, stream types.StreamInterface, success bool, readerID string) error // to save state
-}
-
-type KafkaInterface interface {
-	DriverInterface
-	GetReaderIDs() []string
-	PartitionStreamChanges(ctx context.Context, readerID string, processFn CDCMsgFn) error
+	StreamChanges(ctx context.Context, identifier int, processFn CDCMsgFn) error
+	PostCDC(ctx context.Context, identifier int) error // to save state
```
shubham19may, line 41: what is identifier?

drivers/abstract/interface.go
```diff
@@ -35,13 +35,8 @@ type DriverInterface interface {
...
 	PreCDC(ctx context.Context, streams []types.StreamInterface) error // to init state
-	StreamChanges(ctx context.Context, stream types.StreamInterface, processFn CDCMsgFn) error
-	PostCDC(ctx context.Context, stream types.StreamInterface, success bool, readerID string) error // to save state
-}
-
-type KafkaInterface interface {
-	DriverInterface
-	GetReaderIDs() []string
-	PartitionStreamChanges(ctx context.Context, readerID string, processFn CDCMsgFn) error
+	StreamChanges(ctx context.Context, identifier int, processFn CDCMsgFn) error
```
shubham19may, line 40: again what is identifier? variable name doesn't make sense.

drivers/abstract/cdc.go
```diff
@@ -37,171 +38,136 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
...
-					err = fmt.Errorf("failed to insert cdc record of stream %s, insert func error: %s, thread error: %s", stream.ID(), err, threadErr)
+	} else if sequential {
+		// TODO: For a big table cdc (for all tables) will not start until backfill get finished, need to study alternate ways to do cdc sync
+		a.GlobalConnGroup.Add(func(gCtx context.Context) (err error) {
+			// create cdc context, so that main context not affected if cdc retries
+			cdcCtx, cdcCtxCancel := context.WithCancel(gCtx)
+			defer cdcCtxCancel()
+
+			// Set up inserters for each stream
+			inserters := make(map[string]*destination.WriterThread)
```
shubham19may: can be combined with kafka cdc. this and below part is common (we can dump it into a function)

drivers/abstract/incremental.go
```diff
@@ -48,73 +47,56 @@ func (a *AbstractDriver) Incremental(ctx context.Context, pool *destination.Writ
...
+			inserter, err := pool.NewWriter(incrementalCtx, stream, destination.WithThreadID(threadID))
+			if err != nil {
+				return fmt.Errorf("failed to create new writer thread: %s", err)
+			}
+
+			logger.Infof("Thread[%s]: created incremental writer for stream %s", threadID, streams[index].ID())
+
+			defer a.handleWriterCleanup(incrementalCtx, incrementalCtxCancel, &err, inserter, threadID,
+				func(ctx context.Context) error {
+					if ctx.Err() != nil {
```
shubham19may, line 86: why specially this is handled here in incremental but not in cdc?

drivers/mongodb/internal/mon.go
shubham19may: why is sequential true?

drivers/mysql/internal/mysql.go
```diff
@@ -43,6 +44,10 @@ func (m *MySQL) CDCSupported() bool {
 	return m.CDCSupport
 }
 
+func (m *MySQL) ChangeStreamConfig() (bool, bool, bool) {
```
shubham19may: put this in cdc.go in mysql

drivers/postgres/internal/postgres.go
```diff
@@ -52,6 +52,10 @@ func (p *Postgres) CDCSupported() bool {
 	return p.CDCSupport
 }
 
+func (p *Postgres) ChangeStreamConfig() (bool, bool, bool) {
```
shubham19may: should be cdc.go

drivers/abstract/cdc.go
```diff
@@ -37,171 +38,136 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
...
+		return fmt.Errorf("failed to process cdc streams: %s", err)
 	}
 
-	// handle for kafka differently
-	kafkaDriver, ok := a.GetKafkaInterface()
-	if ok {
-		// set limit to global connection group according to readers available
-		a.GlobalConnGroup = utils.NewCGroupWithLimit(ctx, len(kafkaDriver.GetReaderIDs()))
-		utils.ConcurrentInGroup(a.GlobalConnGroup, kafkaDriver.GetReaderIDs(), func(ctx context.Context, readerID string) (err error) {
+	if parallel {
```
shubham19may: why this is not inside of the above function?

code structure looks weird

drivers/kafka/internal/cdc.go
```diff
@@ -50,7 +50,69 @@ func (k *Kafka) PreCDC(ctx context.Context, streams []types.StreamInterface) err
...
+		// Update global state with consumer group ID and affected streams
+		streamIDs := make([]string, 0, len(syncedStreams))
+		for streamID := range syncedStreams {
+			streamIDs = append(streamIDs, streamID)
+		}
+
+		k.state.SetGlobal(map[string]any{"consumer_group_id": k.consumerGroupID}, streamIDs...)
+		logger.Infof("updated global state with consumer_group_id: %s for %d streams", k.consumerGroupID, len(streamIDs))
+
+		k.checkpointMessage.Delete(readerID)
```
shubham19may: StreamChanges stores checkpoints with checkpointMessage.Store(readerID int, ...), but PostCDC deletes with Delete(readerID string). The int-keyed entry is never removed, so per-reader checkpoints linger (memory growth) and may be re-committed on later runs. Fix: checkpointMessage.Delete(readerIdx) (use the int key).

destination/writers.go
```diff
@@ -255,7 +256,11 @@ func (wt *WriterThread) flush(ctx context.Context, buf []types.RawRecord) (err e
...
+		return nil
 	default:
 		defer wt.stats.ThreadCount.Add(-1)
```
shubham19may, line 268: if wt.group.Block() returns an error, the function returns immediately and never calls writer.Close, so destination-side resources stay open. Add a defer fallback to always call writer.Close(ctx) (likely with ErrNonRetryable wrapping) even when the flush fails.

utils/utils.go
```diff
@@ -437,6 +438,28 @@ func SplitAndTrim(s string) []string {
 	return result
 }
 
+// RetryOnBackoff retries the function f up to attempts times with a backoff sleep between attempts.
+func RetryOnBackoff(attempts int, sleep time.Duration, f func(attempt int) error) (err error) {
```
shubham19may: RetryOnBackoff sleeps with time.Sleep (not interruptible) and uses deterministic backoff. If the parent ctx is canceled during sleep, the goroutine waits the full duration. Also still vulnerable to retry stampedes.

destination/writers.go
```diff
@@ -255,7 +256,11 @@ func (wt *WriterThread) flush(ctx context.Context, buf []types.RawRecord) (err e
 func (wt *WriterThread) Close(ctx context.Context) error {
 	select {
 	case <-ctx.Done():
-		return fmt.Errorf("context closed")
+		err := wt.writer.Close(ctx)
```
shubham19may, line 258: > In the ctx.Done() branch, it just calls writer.Close(ctx); if callers mistakenly pass a live ctx, it could behave as a normal close. Consider return wt.writer.Close(ctx) being fine because ctx is canceled, but making it explicit (Close(ctx) is okay now since writers check ctx.Err()).

AI comment

drivers/abstract/cdc.go
hash-data: partition count we only get after pre CDC

drivers/mongodb/internal/mon.go
hash-data: missed as i was trying to merge kafka and mongo

drivers/abstract/abstract.go
hash-data: actually we can't, it is intentional where we should retry. 
Error which can be recovered on retry are retryable.

destination/writers.go
hash-data: priority thing

drivers/abstract/abstract.go
hash-data: yes i removed driver level checks. so I have modified it

drivers/abstract/abstract.go
hash-data: no required as it exists means cdc is there

drivers/abstract/abstract.go
hash-data: yes logic got changed

drivers/abstract/abstract.go
hash-data: it is different, 
if cdc is supported by stream make it default sync mode for streams, if incremental then incremental, if strict CDC(only in kafka) then make it default

drivers/abstract/abstract.go
hash-data: removed it, thought to add param checks

drivers/abstract/cdc.go
hash-data: parallel means all stream can run parallel, after CDC. concurrent means just after backfill CDC can be started for the stream

drivers/abstract/interface.go
hash-data: it is index for stream or reader that is being used and set in precdc as array

drivers/mongodb/internal/mon.go
hash-data: it is a mistake, was trying to merge mongo with kafka

destination/parquet/parquet.go
```diff
@@ -245,65 +247,97 @@ func (p *Parquet) closePqFiles() error {
...
+				removeLocalFile(filePath, "closing parquet files due to retry attempt")
+				continue
+			}
+
+			if p.s3Client != nil {
+				// Construct S3 key path
+				s3KeyPath := basePath
+				if p.config.Prefix != "" {
+					s3KeyPath = filepath.Join(p.config.Prefix, s3KeyPath)
+				}
```
hash-data:
```suggestion
			if p.s3Client != nil {
				// Construct S3 key path
```

drivers/abstract/abstract.go
shubham19may: lets discuss

drivers/abstract/abstract.go
```diff
@@ -158,8 +155,27 @@ func (a *AbstractDriver) Read(ctx context.Context, pool *destination.WriterPool,
 	// run cdc sync
 	if len(cdcStreams) > 0 {
 		if a.driver.CDCSupported() {
-			if err := a.RunChangeStream(ctx, pool, cdcStreams...); err != nil {
-				return fmt.Errorf("failed to run change stream: %s", err)
+			err := utils.RetryOnBackoff(ctx, a.driver.MaxRetries(), constants.DefaultRetryTimeout, func(ctx context.Context) error {
```
shubham19may: once lets discuss scenario where a new stream was added. it first has to do backfill and then cdc should start. 

or what will it look like when we have watermarking 🤔

drivers/abstract/abstract.go
```diff
@@ -158,8 +155,27 @@ func (a *AbstractDriver) Read(ctx context.Context, pool *destination.WriterPool,
...
+				if a.GlobalConnGroup.Ctx().Err() != nil {
```
shubham19may: bad way to handle cdc. this can be risky. ideally each submit to globalconngroup should be self-sufficient to fully-retry itself. (It should have all ways to retry itself)

drivers/oracle/internal/testdata/test_streams.json
```diff
@@ -1 +1 @@
...
...n_name":"col_smallint"},"COL_TIMESTAMP":{"type":["timestamp_micro","null"],"destination_column_name":"col_timestamp"},"COL_TIMESTAMPLTZ":{"type":["timestamp_micro","null"],"destination_column_name":"col_timestampltz"},"COL_TIMESTAMPTZ":{"type":["timestamp_micro","null"],"destination_column_name":"col_timestamptz"},"COL_VARCHAR2":{"type":["string","null"],"destination_column_name":"col_varchar2"},"ID":{"type":["number"],"destination_column_name":"id"},"_olake_id":{"type":["string","null"],"destination_column_name":"_olake_id"},"_olake_timestamp":{"type":["timestamp_micro","null"],"destination_column_name":"_olake_timestamp"},"_op_type":{"type":["string","null"],"destination_column_name":"_op_type"}}},"supported_sync_modes":["full_refresh","incremental"],"source_defined_primary_key":["ID"],"available_cursor_fields":["COL_CLOB","COL_CURSOR","COL_BIGINT","COL_FLOAT","COL_INT","COL_SMALLINT","COL_VARCHAR2","COL_DOUBLE_PRECISION","COL_TIMESTAMP","COL_TIMESTAMPLTZ","COL_INTEGER","ID","COL_CHARACTER","COL_DECIMAL","COL_CHAR","COL_DATE","COL_NCLOB","COL_BLOB","COL_TIMESTAMPTZ"],"sync_mode":"incremental","destination_database":"oracle:myuser","destination_table":"oracle_test_table_olake"}}]}
```
shubham19may: any reason to change this from full_refresh to incremental?

utils/concurrent.go
```diff
@@ -148,15 +155,29 @@ func (g *CxGroup) Block() error {
 	return g.executor.Wait()
 }
 
-func ConcurrentInGroup[T any](group *CxGroup, array []T, execute func(ctx context.Context, one T) error) {
-	for _, one := range array {
+func ConcurrentInGroupWithRetry[T any](group *CxGroup, array []T, retryCount int, execute func(ctx context.Context, index int, one T) error) {
+	for idx, one := range array {
+		select {
+		case <-group.ctx.Done():
+			break
```
shubham19may: ineffective break statement. Did you mean to break out of the outer loop?

drivers/oracle/internal/testdata/test_streams.json
```diff
@@ -1 +1 @@
..."],"destination_column_name":"col_timestamp"},"COL_TIMESTAMPLTZ":{"type":["timestamp_micro","null"],"destination_column_name":"col_timestampltz"},"COL_TIMESTAMPTZ":{"type":["timestamp_micro","null"],"destination_column_name":"col_timestamptz"},"COL_VARCHAR2":{"type":["string","null"],"destination_column_name":"col_varchar2"},"ID":{"type":["number"],"destination_column_name":"id"},"_olake_id":{"type":["string","null"],"destination_column_name":"_olake_id"},"_olake_timestamp":{"type":["timestamp_micro","null"],"destination_column_name":"_olake_timestamp"},"_op_type":{"type":["string","null"],"destination_column_name":"_op_type"}}},"supported_sync_modes":["full_refresh","incremental"],"source_defined_primary_key":["ID"],"available_cursor_fields":["COL_CLOB","COL_CURSOR","COL_BIGINT","COL_FLOAT","COL_INT","COL_SMALLINT","COL_VARCHAR2","COL_DOUBLE_PRECISION","COL_TIMESTAMP","COL_TIMESTAMPLTZ","COL_INTEGER","ID","COL_CHARACTER","COL_DECIMAL","COL_CHAR","COL_DATE","COL_NCLOB","COL_BLOB","COL_TIMESTAMPTZ"],"sync_mode":"full_refresh","destination_database":"oracle:myuser","destination_table":"oracle_test_table_olake","default_stream_properties":{"normalization":true,"append_mode":false}}}]}
```
shubham19may, line 1: reason to change this?

drivers/abstract/abstract.go
```diff
@@ -91,23 +86,24 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
 	var finalStreams []*types.Stream
 	streamMap.Range(func(_, value any) bool {
 		convStream, _ := value.(*types.Stream)
-		if convStream.SupportedSyncModes.Len() == 0 {
-			convStream.WithSyncMode(types.FULLREFRESH, types.INCREMENTAL)
```
shubham19may, line 95: s3, db2 needs to added with these

drivers/abstract/abstract.go
shubham19may: CDCSupported returns true in Kafka, which currently wasn't the case previously as we had check `!isKafkaDriver` before

drivers/abstract/abstract.go
```diff
@@ -197,3 +193,91 @@ func (a *AbstractDriver) Read(ctx context.Context, pool *destination.WriterPool,
...
+
+		if closeErr != nil {
+			*err = utils.Ternary(*err == nil, closeErr, fmt.Errorf("%s: prev error: %w", closeErr, *err)).(error)
+		}
+
+		// check for panics before post-processing
+		if r := recover(); r != nil {
+			*err = utils.Ternary(*err == nil, fmt.Errorf("panic recovered: %v", r), fmt.Errorf("%s: prev error: %w", r, *err)).(error)
+		}
+
```
shubham19may, line 275: what happens when writer thread close returns error, postProcess will still continue?

if yes, that will update the state right?

drivers/s3/internal/sync_test.go
```diff
@@ -434,15 +434,13 @@ func TestCDCNotSupported(t *testing.T) {
 	})
 
 	t.Run("StreamChanges returns error", func(t *testing.T) {
-		stream := types.NewStream("test", "s3", nil)
```
shubham19may, line 437: we don't have any stream here, is that ok?

drivers/mysql/internal/cdc.go
```diff
@@ -11,18 +11,24 @@ import (
 	"github.com/datazip-inc/olake/utils"
 )
 
-func (m *MySQL) prepareBinlogConn(ctx context.Context, globalState MySQLGlobalState, streams []types.StreamInterface) (*binlog.Connection, error) {
-	if !m.CDCSupport {
-		return nil, fmt.Errorf("invalid call; %s not running in CDC mode", m.Type())
+func (m *MySQL) prepareBinlogConn(ctx context.Context) (*binlog.Connection, error) {
+	savedState := m.state.GetGlobal()
```
shubham19may, line 15: Add todo, we might need to think about how to manipulate state for future auto-scaling efforts via externally running pods

drivers/mysql/internal/cdc.go
```diff
@@ -51,34 +61,34 @@ func (m *MySQL) PreCDC(ctx context.Context, streams []types.StreamInterface) err
...
+}
 
-	var mySQLGlobalState MySQLGlobalState
-	if err := utils.Unmarshal(globalState.State, &mySQLGlobalState); err != nil {
-		return fmt.Errorf("failed to unmarshal global state: %s", err)
-	}
-
-	conn, err := m.prepareBinlogConn(ctx, mySQLGlobalState, streams)
+func (m *MySQL) StreamChanges(ctx context.Context, streamIndex int, OnMessage abstract.CDCMsgFn) error {
+	conn, err := m.prepareBinlogConn(ctx)
```
shubham19may, line 69: if the last BinLogConn is still active, shouldn't we clean it first?
or will that get cleaned because of context it was created with got closed after retry?

drivers/postgres/internal/cdc.go
```diff
@@ -29,72 +30,64 @@ func (p *Postgres) prepareWALJSConfig(streams ...types.StreamInterface) (*waljs.
...
+
 func (p *Postgres) PreCDC(ctx context.Context, streams []types.StreamInterface) error {
-	config, err := p.prepareWALJSConfig(streams...)
+	slot, err := waljs.GetSlotPosition(ctx, p.client, p.cdcConfig.ReplicationSlot)
+	if err != nil {
+		return fmt.Errorf("failed to get slot position: %s", err)
+	}
+
+	globalState := p.state.GetGlobal()
+	if globalState == nil || globalState.State == nil {
```
shubham19may, line 44: postgresGlobalState.LSN == "" Check missed?

shubham19may:
thread_id should be recognizable for retries (that is a retry thread of previous thread)

shubham19may:
mssql postcdc cleanup

PR #603, opened 2025-10-25
feat: Kafka as source V0

### LINKED ISSUE #87: [Feat] Kafka Source

ImDoubD-datazip:
# Description
Kafka as source has been added as Olake's latest driver. The messages from the topics of Kafka will be synced directly to destination. Currently modes are supported: streaming cdc. 
- In case of streaming cdc : 

#### TO FIX: 

   - 1st: process the messages 

Fixes #87 

## Type of change

# How Has This Been Tested?

cdc streaming Sync:
image

# How to test Kafka as source

Example source.json:
Using PLAINTEXT protocol
```json
{
    "protocol": {
    },
    "max_threads": 3,
    "auto_offset_reset": "latest"
}
```
Using SASL-SSL protocol
```json
{
    "protocol": {
    },
    "max_threads": 3,
    "auto_offset_reset": "earliest"
}
```

## Documentation

### DIFF
--- a/drivers/abstract/abstract.go
+++ b/drivers/abstract/abstract.go
@@ -58,9 +58,9 @@ func (a *AbstractDriver) Type() string {
 	return a.driver.Type()
 }
 
-func (a *AbstractDriver) IsKafkaDriver() bool {
-	_, ok := a.driver.(KafkaInterface)
-	return ok
+func (a *AbstractDriver) GetKafkaInterface() (KafkaInterface, bool) {
+	kafkaInterface, ok := a.driver.(KafkaInterface)
+	return kafkaInterface, ok
 }
 
 func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error) {
@@ -101,13 +101,15 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error) 
 			convStream.UpsertField(column, typ, true)
 		}
 
-		if a.driver.CDCSupported() && !a.IsKafkaDriver() {
+		_, isKafkaDriver := a.GetKafkaInterface()
+		if a.driver.CDCSupported() && !isKafkaDriver {
 			convStream.WithSyncMode(types.CDC, types.STRICTCDC)
 			convStream.SyncMode = types.CDC
 		} else {
 			// remove cdc column as it is not supported
 			convStream.Schema.Properties.Delete(constants.CdcTimestamp)
 		}
+
 		finalStreams = append(finalStreams, convStream)
 		return true
 	})
--- a/drivers/abstract/cdc.go
+++ b/drivers/abstract/cdc.go
@@ -105,54 +105,47 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
 		return nil
 	}
 
-	// run cdc for kafka
-	if a.IsKafkaDriver() {
-		kafkaDriver, _ := a.driver.(KafkaInterface)
-		// global connection group according to reader tasks
-		a.GlobalConnGroup = utils.NewCGroupWithLimit(ctx, len(kafkaDriver.GetReaderTasks()))
-		utils.ConcurrentInGroup(a.GlobalConnGroup, kafkaDriver.GetReaderTasks(), func(ctx context.Context, readerID string) (err error) {
-			writers := make(map[string]*destination.WriterThread)
+	// handle for kafka differently
+	kafkaDriver, ok := a.GetKafkaInterface()
+	if ok {
+		// set limit to global connection group according to readers available
+		a.GlobalConnGroup = utils.NewCGroupWithLimit(ctx, len(kafkaDriver.GetReaderIDs()))
+		utils.ConcurrentInGroup(a.GlobalConnGroup, kafkaDriver.GetReaderIDs(), func(ctx context.Context, readerID string) (err error) {
+			inserters := make(map[string]*destination.WriterThread)
 			defer func() {
-				for key, wr := range writers {
-					if wr == nil {
-						continue
-					}
-					if threadErr := wr.Close(ctx); threadErr != nil {
-						err = fmt.Errorf("failed closing writer[%s]: %s", key, threadErr)
+				for streamID, inserter := range inserters {
+					if threadErr := inserter.Close(ctx); threadErr != nil {
+						err = fmt.Errorf("failed closing writer[%s]: %s", streamID, threadErr)
 					}
 				}
 
 				if r := recover(); r != nil {
-					err = fmt.Errorf("panic recovered in cdc: %v, prev error: %s", r, err)
+					err = fmt.Errorf("panic recovered in cdc: %s, prev error: %v", r, err)
 				}
 
 				postCDCErr := a.driver.PostCDC(ctx, nil, err == nil, readerID)
 				if postCDCErr != nil {
-					err = fmt.Errorf("post cdc error: %s, cdc insert thread error: %s", postCDCErr, err)
+					err = fmt.Errorf("post cdc error: %s, cdc insert thread error: %v", postCDCErr, err)
 				}
 			}()
 			return RetryOnBackoff(a.driver.MaxRetries(), constants.DefaultRetryTimeout, func() error {
 				return kafkaDriver.PartitionStreamChanges(ctx, readerID, func(ctx context.Context, message CDCChange) error {
-					if message.Stream == nil {
-						return nil
-					}
-					inserter := writers[message.Stream.ID()]
+					inserter := inserters[message.Stream.ID()]
 					if inserter == nil {
-						// dedicated writer for this reader (with stream)
+						// reader can read from multiple streams in kafka which we cant decide (r1 -> s1, s2, r2 -> s2, s3)
+						// so creating writer for stream which read by reader in kafka
 						threadID := fmt.Sprintf("%s_%s", readerID, message.Stream.ID())
-						var err error
 						inserter, err = pool.NewWriter(ctx, message.Stream, destination.WithThreadID(threadID))
 						if err != nil {
 							return fmt.Errorf("failed to create writer for stream %s: %s", message.Stream.ID(), err)
 						}
-						writers[message.Stream.ID()] = inserter
+						inserters[message.Stream.ID()] = inserter
 						logger.Infof("Thread[%s]: created cdc writer for stream %s", threadID, message.Stream.ID())
 					}
-					pkFields := message.Stream.GetStream().SourceDefinedPrimaryKey.Array()
 					return inserter.Push(ctx, types.CreateRawRecord(
-						utils.GetKeysHash(message.Data, pkFields...),
+						utils.GetKeysHash(message.Data, message.Stream.GetStream().SourceDefinedPrimaryKey.Array()...),
 						message.Data,
-						"c",
+						"c", // kafka message is always a create operation
 						&message.Timestamp,
 					))
 				})
@@ -160,6 +153,7 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
 		})
 		return nil
 	}
+
 	// TODO: For a big table cdc (for all tables) will not start until backfill get finished, need to study alternate ways to do cdc sync
 	a.GlobalConnGroup.Add(func(ctx context.Context) (err error) {
 		// Set up inserters for each stream
--- a/drivers/abstract/interface.go
+++ b/drivers/abstract/interface.go
@@ -42,6 +42,6 @@ type DriverInterface interface {
 
 type KafkaInterface interface {
 	DriverInterface
-	GetReaderTasks() []string
+	GetReaderIDs() []string
 	PartitionStreamChanges(ctx context.Context, readerID string, processFn CDCMsgFn) error
 }
--- a/drivers/kafka/internal/cdc.go
+++ b/drivers/kafka/internal/cdc.go
@@ -11,7 +11,6 @@ import (
 	"github.com/datazip-inc/olake/types"
 	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/logger"
-	"github.com/datazip-inc/olake/utils/typeutils"
 	"github.com/segmentio/kafka-go"
 )
 
@@ -20,8 +19,10 @@ func (k *Kafka) PreCDC(ctx context.Context, streams []types.StreamInterface) err
 		return fmt.Errorf("no valid streams found for CDC")
 	}
 
-	// Generate a new consumer group ID if not configured
 	var groupID string
+
+	// NOTE: in kafka we are giving priority of available consumer group id from state over config
+	// get consumer group id from global state
 	if globalState := k.state.GetGlobal(); globalState != nil && globalState.State != nil {
 		if stateMap, ok := globalState.State.(map[string]any); ok {
 			if consumerGroupID, exists := stateMap["consumer_group_id"]; exists {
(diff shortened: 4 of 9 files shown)


drivers/abstract/abstract.go
```diff
@@ -34,11 +34,12 @@ var DefaultColumns = map[string]types.DataType{
 }
 
 func NewAbstractDriver(ctx context.Context, driver DriverInterface) *AbstractDriver {
-	return &AbstractDriver{
+	abstractDriver := &AbstractDriver{
```
hash-data: not required

drivers/abstract/abstract.go
```diff
@@ -86,8 +92,10 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
 	var finalStreams []*types.Stream
 	streamMap.Range(func(_, value any) bool {
 		convStream, _ := value.(*types.Stream)
-		convStream.WithSyncMode(types.FULLREFRESH, types.INCREMENTAL)
-		convStream.SyncMode = types.FULLREFRESH
+		if !a.IsKafkaDriver() {
```
hash-data: move to kafka only it will be set there

drivers/abstract/abstract.go
```diff
@@ -86,8 +92,10 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
...
+			convStream.WithSyncMode(types.FULLREFRESH, types.INCREMENTAL)
```
hash-data, line 95: is full refresh supported in kafka?

drivers/abstract/backfill.go
```diff
@@ -15,6 +15,13 @@ import (
 )
 
 func (a *AbstractDriver) Backfill(ctx context.Context, backfilledStreams chan string, pool *destination.WriterPool, stream types.StreamInterface) error {
+	if a.driver.Type() == string(constants.Kafka) {
```
hash-data: we can mention in chunk iterator function of kafka that backfill not supported

drivers/abstract/interface.go
```diff
@@ -37,5 +37,11 @@ type DriverInterface interface {
 	CDCSupported() bool
 	PreCDC(ctx context.Context, streams []types.StreamInterface) error // to init state
 	StreamChanges(ctx context.Context, stream types.StreamInterface, processFn CDCMsgFn) error
-	PostCDC(ctx context.Context, stream types.StreamInterface, success bool) error // to save state
+	PostCDC(ctx context.Context, stream types.StreamInterface, success bool, readerID string) error // to save state
+}
+
+type KafkaInterface interface {
```
hash-data, line 43: should not we rename it to messageQueueInterface or something similar ?

drivers/abstract/interface.go
```diff
@@ -37,5 +37,11 @@ type DriverInterface interface {
 	CDCSupported() bool
 	PreCDC(ctx context.Context, streams []types.StreamInterface) error // to init state
 	StreamChanges(ctx context.Context, stream types.StreamInterface, processFn CDCMsgFn) error
-	PostCDC(ctx context.Context, stream types.StreamInterface, success bool) error // to save state
+	PostCDC(ctx context.Context, stream types.StreamInterface, success bool, readerID string) error // to save state
```
hash-data, line 40: dont change cdc interface if you are designing new interface

drivers/abstract/cdc.go
```diff
@@ -104,6 +104,60 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
 		// parallel change streams already processed
 		return nil
 	}
+
+	// run cdc for kafka
+	if a.IsKafkaDriver() {
```
hash-data: should not we create a new file for it ? (as it already a new interface)

drivers/abstract/cdc.go
```diff
@@ -104,6 +104,60 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
...
+						// dedicated writer for this reader (with stream)
+						threadID := fmt.Sprintf("%s_%s", readerID, message.Stream.ID())
+						var err error
+						inserter, err = pool.NewWriter(ctx, message.Stream, destination.WithThreadID(threadID))
+						if err != nil {
+							return fmt.Errorf("failed to create writer for stream %s: %s", message.Stream.ID(), err)
+						}
+						writers[message.Stream.ID()] = inserter
+						logger.Infof("Thread[%s]: created cdc writer for stream %s", threadID, message.Stream.ID())
+					}
```
hash-data, line 144: as per discussion we have to create writer per partition or reader right?

drivers/kafka/internal/kafka.go
```diff
@@ -0,0 +1,281 @@
...
+	var topicNames []string
+	for _, topic := range resp.Topics {
+		topicNames = append(topicNames, topic.Name)
+	}
+	return topicNames, nil
+}
+
+func (k *Kafka) ProduceSchema(_ context.Context, streamName string) (*types.Stream, error) {
+	logger.Infof("[KAFKA] producing schema for topic [%s]", streamName)
+	stream := types.NewStream(streamName, "topics", nil)
```
hash-data: make this namespace as constant, as well as key which are being added make them constants in same file

drivers/kafka/internal/config.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type Config struct {
+	BootstrapServers            string         `json:"bootstrap_servers"`
+	ConsumerGroupID             string         `json:"consumer_group_id,omitempty"`
+	AutoOffsetReset             string         `json:"auto_offset_reset,omitempty"`
+	Protocol                    ProtocolConfig `json:"protocol"`
+	MaxThreads                  int            `json:"max_threads"`
+	RetryCount                  int            `json:"backoff_retry_count"`
+	ThreadsEqualTotalPartitions bool           `json:"threads_equal_total_partitions,omitempty"`
```
hash-data, line 16: why not if max threads  <= 0  we should make it equal to partitions?

pkg/kafka/reader.go
```diff
@@ -0,0 +1,224 @@
...
+	return &ReaderManager{
+		config:          config,
+		partitionIndex:  make(map[string]types.PartitionMetaData),
+		readers:         make(map[string]*kafka.Reader),
+		readerClientIDs: make(map[string]string),
+	}
+}
+
+// CreateReaders creates Kafka readers based on the provided streams and configuration
+func (krm *ReaderManager) CreateReaders(ctx context.Context, streams []types.StreamInterface, consumerGroupID string) error {
```
hash-data:
```suggestion
func (r *ReaderManager) CreateReaders(ctx context.Context, streams []types.StreamInterface, consumerGroupID string) error {
```

pkg/kafka/reader.go
```diff
@@ -0,0 +1,224 @@
...
+			Stream:      stream,
+			PartitionID: idx.Partition,
+			EndOffset:   idx.LastOffset,
+			StartOffset: startOffset,
+		}
+
+		// update topic's partition index
+		if krm.partitionIndex == nil {
+			krm.partitionIndex = make(map[string]types.PartitionMetaData)
+		}
```
hash-data: it is already being set to empty map so duplicate here

pkg/kafka/types.go
```diff
@@ -0,0 +1,34 @@
...
+	AdminClient                 *kafka.Client
+}
+
+// ReaderManager manages Kafka readers and their metadata
+type ReaderManager struct {
+	config             ReaderConfig
+	readers            map[string]*kafka.Reader
+	partitionIndex     map[string]types.PartitionMetaData
+	readerClientIDs    map[string]string
+	readerLastMessages sync.Map // map[string]map[partKey]kafka.Message
```
hash-data: can you add use case of each as comment

pkg/kafka/reader.go
```diff
@@ -0,0 +1,224 @@
...
+		return nil
+	}
+
+	krm.readers = make(map[string]*kafka.Reader)
+	krm.readerClientIDs = make(map[string]string)
+
+	// reader tasks according to concurrency policy
+	readersToCreate := utils.Ternary(krm.ShouldMatchPartitionCount(), totalPartitions, utils.Ternary(krm.config.MaxThreads >= totalPartitions, totalPartitions, krm.config.MaxThreads).(int)).(int)
+
+	for i := 0; i < readersToCreate; i++ {
```
hash-data: use range loop

pkg/kafka/reader.go
```diff
@@ -0,0 +1,224 @@
...
+		readerID := fmt.Sprintf("grp_%s", utils.ULID())
```
hash-data:
```suggestion
		readerID := fmt.Sprintf("group_%s", utils.ULID())
```

pkg/kafka/reader.go
```diff
@@ -0,0 +1,224 @@
...
+			GroupID: consumerGroupID,
+			GroupTopics: func() []string {
+				topics := make([]string, 0, len(streams))
+				for _, s := range streams {
+					topics = append(topics, s.Name())
+				}
+				return topics
+			}(),
+			MinBytes:       1,
+			MaxBytes:       10e6,
```
hash-data: let us use default values

drivers/abstract/cdc.go
```diff
@@ -104,6 +104,60 @@ func (a *AbstractDriver) RunChangeStream(ctx context.Context, pool *destination.
 		// parallel change streams already processed
 		return nil
 	}
+
+	// run cdc for kafka
+	if a.IsKafkaDriver() {
+		kafkaDriver, _ := a.driver.(KafkaInterface)
+		utils.ConcurrentInGroup(a.GlobalConnGroup, kafkaDriver.GetReaderTasks(), func(ctx context.Context, readerID string) (err error) {
```
hash-data: so here, 
-> ideally we can get partition in new interface and call utils.Concurrent in group instead of typecasting to another driver interface.

I will check a way if we can skip typecasting to kafka interface

drivers/abstract/cdc.go
hash-data: there are concurrency issues with this implementation

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+
+		// Process message data
+		messageData := func() map[string]any {
+			var result map[string]any
+			if message.Value == nil {
+				logger.Warnf("received nil message value at offset %d for topic %s, partition %d", message.Offset, message.Topic, message.Partition)
+				return nil
+			}
+			if err := json.Unmarshal(message.Value, &result); err != nil {
```
hash-data, line 92: once check err on all places and its formatting, let us make it according to project

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,229 @@
...
+				logger.Errorf("failed to unmarshal message value: %v", err)
+				return nil
```
hash-data, line 94: so we are skipping message if we are unable to process that?

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,229 @@
...
+		// Skip messages before start offset (can happen during retries or partition reassignment)
+		if currentPartitionMeta.StartOffset > 0 && message.Offset < currentPartitionMeta.StartOffset {
+			continue
+		}
+
+		// Process message data
+		messageData := func() map[string]any {
+			var result map[string]any
+			if message.Value == nil {
+				logger.Warnf("received nil message value at offset %d for topic %s, partition %d", message.Offset, message.Topic, message.Partition)
```
hash-data, line 89: for nil values, dont we have to store record with empty value ?

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,229 @@
...
+				return nil
+			}
+			result["partition"] = message.Partition
+			result["offset"] = message.Offset
+			result["key"] = string(message.Key)
+			result["kafka_timestamp"] = message.Time.UnixMilli()
+			return result
+		}()
+		if messageData == nil {
+			// Even for nil data, we need to track the message for offset commit and check if partition is complete
```
hash-data: if message is nil there can be marshalling error or nil value. So, checking for partition copletition is valid in both cases?

hash-data: also storing the message value as last message?

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,229 @@
...
+		lastMessages[currentPartitionKey] = message
+
+		// Check if partition is complete
+		if message.Offset >= currentPartitionMeta.EndOffset-1 {
+			shouldExit, err := k.checkPartitionCompletion(ctx, readerID, currentPartitionKey, completedPartitions, observedPartitions)
+			if err != nil {
+				return err
+			}
+			if shouldExit {
+				return nil
```
hash-data: why we are not validating the partition completion from the end ofset that we received in pre cdc ?

drivers/kafka/internal/cdc.go
```diff
@@ -0,0 +1,229 @@
...
+			}
+		}
+	}
+}
+
+func (k *Kafka) PostCDC(ctx context.Context, stream types.StreamInterface, noErr bool, readerID string) error {
```
hash-data, line 131: do we want to run postcdc for each reader or we want it just like precdc for all reader together?

drivers/kafka/resources/spec.json
```diff
@@ -0,0 +1,126 @@
...
+        "consumer_group_id": {
+            "type": "string",
+            "title": "Consumer Group ID",
+            "description": "Kafka consumer group ID (optional)"
+        },
+        "max_threads": {
+            "type": "integer",
+            "title": "Max Threads",
+            "description": "Maximum number of parallel processing threads",
+            "minimum": 1,
```
hash-data: let us set to 0 which means equal to partitions

drivers/mongodb/internal/testdata/test_streams.json
```diff
@@ -1 +1 @@
...
...stamp"},"id_bigint":{"type":["integer"],"destination_column_name":"id_bigint"},"id_bool":{"type":["boolean"],"destination_column_name":"id_bool"},"id_double":{"type":["number"],"destination_column_name":"id_double"},"id_int":{"type":["integer_small"],"destination_column_name":"id_int"},"id_maxkey":{"type":["unknown"],"destination_column_name":"id_maxkey"},"id_minkey":{"type":["unknown"],"destination_column_name":"id_minkey"},"id_nested":{"type":["object"],"destination_column_name":"id_nested"},"id_nil":{"type":["null"],"destination_column_name":"id_nil"},"id_regex":{"type":["unknown"],"destination_column_name":"id_regex"},"id_timestamp":{"type":["timestamp"],"destination_column_name":"id_timestamp"},"name_varchar":{"type":["string"],"destination_column_name":"name_varchar"}}},"supported_sync_modes":["full_refresh","incremental","cdc","strict_cdc"],"source_defined_primary_key":["_id"],"available_cursor_fields":["id_nested","name_varchar","created_timestamp","id_regex","id_minkey","id_int","id_bigint","id_timestamp","id_maxkey","id_bool","id_nil","_id","id_double"],"sync_mode":"cdc","destination_database":"mongodb:olake_mongodb_test","destination_table":"mongodb_test_table_olake"}}]}
```
hash-data, line 1: why change?

pkg/kafka/balancer.go
```diff
@@ -0,0 +1,51 @@
+package kafka
+
+import (
```
hash-data, line 3: is this balancer is being used?, not reviewed yet will review in next iteration

utils/utils.go
```diff
@@ -250,6 +250,10 @@ func IsJSON(str string) bool {
 
 // GetKeysHash returns md5 hashsum of concatenated map values (sort keys before)
 func GetKeysHash(m map[string]interface{}, keys ...string) string {
+	if len(m) == 0 {
```
hash-data: let us discuss empty message itself should not reach

drivers/abstract/abstract.go
ImDoubD-datazip: no it is not supported. only when to start commit from consumer group side can be told, and that is told using auto offset reset, which is asked by the user.

drivers/abstract/backfill.go
ImDoubD-datazip: I have mentioned there. But i have mentioned here as well for backfill for the kafka based streams to be skipped.

drivers/abstract/interface.go
ImDoubD-datazip: For now this is kafka based only. in other message queses such as kinesis and rabbitMQ, the structure i believe is again different. so there its not exactly partition and offset but something different

drivers/abstract/interface.go
ImDoubD-datazip: only the partition streaming function is there. in case of postcdc, i think it is not necessary to create another function which does the same thing which normal driver postcdc does. only an extra parameter is added.

drivers/abstract/cdc.go
ImDoubD-datazip: we can create. but it works almost same as backfill concurrency , so included it here. if any other new streaming sources if to be added then we can think of something.

drivers/abstract/backfill.go
ImDoubD-datazip: will make it strict cdc

drivers/abstract/interface.go
ImDoubD-datazip: we need to think

pkg/kafka/reader.go
ImDoubD-datazip: i will add comments

drivers/abstract/cdc.go
ImDoubD-datazip: will check on this later

drivers/kafka/internal/cdc.go
ImDoubD-datazip: will have to ask product

drivers/kafka/internal/cdc.go
ImDoubD-datazip: will have to ask product

drivers/kafka/resources/spec.json
ImDoubD-datazip: will ask product

drivers/kafka/internal/config.go
ImDoubD-datazip: this is done in case of bool ThreadsEqualTotalPartitions only for now. if max threads = 0, then default max connection count is set.

pkg/kafka/reader.go
ImDoubD-datazip: ok

pkg/kafka/types.go
ImDoubD-datazip: sure

pkg/kafka/reader.go
ImDoubD-datazip: ok

drivers/kafka/internal/cdc.go
ImDoubD-datazip: ok

drivers/kafka/internal/cdc.go
ImDoubD-datazip: this is for if the last message of the partition had some marshal error or was nil

drivers/kafka/internal/cdc.go
ImDoubD-datazip: we are validating when the specific reader reads all its assigned partitions, then it has to be confirmed from the consumer group that this particular reader id has read all its partitions. after this the loop for the reader will then end.

drivers/kafka/internal/cdc.go
ImDoubD-datazip: yes, because if 2 readers are there and 1 has already read and other is reading from a hot partition, then the waiting reader (the consumer member) might die  because it is there for a while and has not gotten any heartbeat for a while. so readers must be closed concurrently as soon as there work is finished.

pkg/kafka/balancer.go
ImDoubD-datazip: yes this is being used as group balancing algorithm

drivers/kafka/internal/cdc.go
ImDoubD-datazip: clarified from the product, both nil message and marshal error messages have to be skipped.

drivers/kafka/internal/cdc.go
ImDoubD-datazip: no we don;t have to store. clarified with product. we will update on the docs

drivers/mongodb/internal/testdata/test_streams.json
ImDoubD-datazip: it is as per any recent mongodb changes added.

drivers/abstract/abstract.go
ImDoubD-datazip: I tried this, but in supported sync mode still all the sync modes were getting shown, setup happens here finally for the sync modes.

utils/utils.go
ImDoubD-datazip: removed this

drivers/abstract/abstract.go
hash-data: let us discuss on this condition

drivers/abstract/abstract.go
```diff
@@ -86,18 +91,21 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
 	var finalStreams []*types.Stream
 	streamMap.Range(func(_, value any) bool {
 		convStream, _ := value.(*types.Stream)
-		convStream.WithSyncMode(types.FULLREFRESH, types.INCREMENTAL)
-		convStream.SyncMode = types.FULLREFRESH
+		if convStream.SupportedSyncModes.Len() == 0 {
+			convStream.WithSyncMode(types.FULLREFRESH, types.INCREMENTAL)
+			convStream.SyncMode = types.FULLREFRESH
```
hash-data: let us make it simple, adding condition not to override values

hash-data:
Nice Work @ImDoubD-datazip

PR #604, opened 2025-10-28
fix: handle chunk generation for partitioned PostgreSQL tables using max page ID

vikaxsh:
# Description

This PR fixes chunk generation for PostgreSQL tables, ensuring correct CTID range calculation when handling partitioned or large tables.
- Added new query to compute `maxPageID` and `partitionCount` with a 5% safety margin.
- Implemented fallback to `relpages` when `maxPageID` <= 0

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/drivers/postgres/internal/backfill.go
+++ b/drivers/postgres/internal/backfill.go
@@ -233,7 +233,7 @@ func loadPartitionPages(ctx context.Context, db *sql.DB, stream types.StreamInte
 	query := jdbc.PostgresPartitionPages(stream)
 	rows, err := db.QueryContext(ctx, query)
 	if err != nil {
-		return nil, 0, fmt.Errorf("failed to load partition pages: %v", err)
+		return nil, 0, fmt.Errorf("failed to load partition pages: %s", err)
 	}
 	var maxPageID uint32
 	defer rows.Close()
--- a/drivers/postgres/internal/backfill.go
+++ b/drivers/postgres/internal/backfill.go
@@ -121,10 +121,14 @@ func (p *Postgres) splitTableIntoChunks(ctx context.Context, stream types.Stream
 		}
 
 		batchPages := uint32(math.Ceil(float64(constants.EffectiveParquetSize) / float64(blockSize)))
-		batchSize := uint32(math.Ceil(float64(batchPages) / float64(len(partitions))))
+		partionsInRange := PartitionPagesGreaterThan(partitions, 0)
+		batchSize := uint32(math.Ceil(float64(batchPages) / float64(partionsInRange)))
 
 		chunks := types.NewSet[types.Chunk]()
 		for start := uint32(0); start < maxPageID; start += batchSize {
+			lastChunkEnd := start + batchSize
+			partionsInRange := PartitionPagesGreaterThan(partitions, lastChunkEnd)
+			batchSize = uint32(math.Ceil(float64(batchPages) / float64(partionsInRange)))
 			end := start + batchSize
 			if end >= maxPageID {
 				end = ^uint32(0)
@@ -135,8 +139,6 @@ func (p *Postgres) splitTableIntoChunks(ctx context.Context, stream types.Stream
 				Max: fmt.Sprintf("'(%d,0)'", end),
 			})
 
-			partionsInRange := PartitionPagesGreaterThan(partitions, end)
-			batchSize = uint32(math.Ceil(float64(batchPages) / float64(partionsInRange)))
 		}
 
 		return chunks, nil
@@ -262,8 +264,5 @@ func PartitionPagesGreaterThan(partitions []PartitionPage, end uint32) int {
 			count++
 		}
 	}
-	if count == 0 {
-		count = 1 // fallback for non-partitioned tables
-	}
-	return count
+	return utils.Ternary(count == 0, 1, count).(int)
 }
--- a/drivers/postgres/internal/backfill.go
+++ b/drivers/postgres/internal/backfill.go
@@ -73,10 +73,12 @@ func (p *Postgres) GetOrSplitChunks(ctx context.Context, pool *destination.Write
 
 func (p *Postgres) splitTableIntoChunks(ctx context.Context, stream types.StreamInterface) (*types.Set[types.Chunk], error) {
 	generateCTIDRanges := func(stream types.StreamInterface) (*types.Set[types.Chunk], error) {
-		var relpages, maxPageID, blockSize uint32
+		var maxPageID, blockSize uint32
+		var relpages int64
+		var partitionCount int
 		// Fetch actual table page stats
 		storageStatsquery := jdbc.PostgresTableStorageStats(stream)
-		err := p.client.QueryRowContext(ctx, storageStatsquery).Scan(&relpages, &maxPageID)
+		err := p.client.QueryRowContext(ctx, storageStatsquery).Scan(&relpages, &maxPageID, &partitionCount)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get table storage stats: %s", err)
 		}
@@ -87,7 +89,9 @@ func (p *Postgres) splitTableIntoChunks(ctx context.Context, stream types.Stream
 		}
 
 		// Calculate pages per chunk
-		batchSize := uint32(math.Ceil(float64(constants.EffectiveParquetSize) / float64(blockSize)))
+		batchPages := uint32(math.Ceil(float64(constants.EffectiveParquetSize) / float64(blockSize)))
+		batchSize := uint32(math.Ceil(float64(batchPages) / float64(partitionCount)))
+
 		if maxPageID <= 0 {
 			maxPageID = uint32(max(1, int(relpages)))
 		}
--- a/drivers/postgres/internal/backfill.go
+++ b/drivers/postgres/internal/backfill.go
@@ -73,12 +73,7 @@ func (p *Postgres) GetOrSplitChunks(ctx context.Context, pool *destination.Write
 
 func (p *Postgres) splitTableIntoChunks(ctx context.Context, stream types.StreamInterface) (*types.Set[types.Chunk], error) {
 	generateCTIDRanges := func(stream types.StreamInterface) (*types.Set[types.Chunk], error) {
-		var (
-			relpages  uint32
-			maxPageID uint32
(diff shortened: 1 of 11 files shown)


pkg/jdbc/jdbc.go
```diff
@@ -130,9 +130,30 @@ func PostgresBlockSizeQuery() string {
...
+//
+// It returns:
+//   - max_page_id: estimated upper bound of page IDs (using pg_relation_size)
+//   - relpages: estimated page count from pg_class (used as fallback)
+func PostgresTableStorageStats(stream types.StreamInterface) string {
+	return fmt.Sprintf(`
+SELECT
+c.relpages,
+    -- The 1.05 multiplier adds a 5%% safety margin to account for
+    -- page growth or rounding errors in pg_relation_size estimates.
```
vaibhav-datazip: these query comments can be removed

drivers/postgres/internal/backfill.go
```diff
@@ -73,24 +73,28 @@ func (p *Postgres) GetOrSplitChunks(ctx context.Context, pool *destination.Write
...
 			return nil, fmt.Errorf("failed to get block size: %s", err)
 		}
-		batchSize := uint32(math.Ceil(float64(constants.EffectiveParquetSize) / float64(blockSize)))
 
-		relPages = utils.Ternary(relPages == uint32(0), uint32(1), relPages).(uint32)
+		// Calculate pages per chunk
+		batchSize := uint32(math.Ceil(float64(constants.EffectiveParquetSize) / float64(blockSize)))
+		if maxPageID <= 0 {
+			maxPageID = uint32(max(1, int(relpages)))
+		}
```
hash-data: Use ternary, and also test with a normal case (for a table having rel page)

drivers/postgres/internal/backfill.go
```diff
@@ -73,24 +73,28 @@ func (p *Postgres) GetOrSplitChunks(ctx context.Context, pool *destination.Write
 
 func (p *Postgres) splitTableIntoChunks(ctx context.Context, stream types.StreamInterface) (*types.Set[types.Chunk], error) {
 	generateCTIDRanges := func(stream types.StreamInterface) (*types.Set[types.Chunk], error) {
-		var relPages, blockSize uint32
-		relPagesQuery := jdbc.PostgresRelPageCount(stream)
-		err := p.client.QueryRowContext(ctx, relPagesQuery).Scan(&relPages)
+		var relpages, maxPageID, blockSize uint32
+		// Fetch actual table page stats
+		storageStatsquery := jdbc.PostgresTableStorageStats(stream)
+		err := p.client.QueryRowContext(ctx, storageStatsquery).Scan(&relpages, &maxPageID)
```
hash-data: use proper camel case for variable names like storageStatsQuery, relPages etc

pkg/jdbc/jdbc.go
```diff
@@ -130,9 +130,30 @@ func PostgresBlockSizeQuery() string {
...
+    CEIL(1.05 * GREATEST(
+        pg_relation_size(c.oid) / current_setting('block_size')::int,
+        COALESCE((
+            SELECT MAX(pg_relation_size(inhrelid) / current_setting('block_size')::int)
+            FROM pg_inherits
+            WHERE inhparent = c.oid
+        ), 0)
+    )) AS max_page_id
+FROM pg_class c
+JOIN pg_namespace n ON n.oid = c.relnamespace
```
hash-data: for rel pages we were sure for size of chunk, are we sure here as well?

pkg/jdbc/jdbc.go
```diff
@@ -130,9 +130,30 @@ func PostgresBlockSizeQuery() string {
...
+// It returns:
+//   - max_page_id: estimated upper bound of page IDs (using pg_relation_size)
+//   - relpages: estimated page count from pg_class (used as fallback)
+func PostgresTableStorageStats(stream types.StreamInterface) string {
+	return fmt.Sprintf(`
+SELECT
+c.relpages,
+    -- The 1.05 multiplier adds a 5%% safety margin to account for
+    -- page growth or rounding errors in pg_relation_size estimates.
+    CEIL(1.05 * GREATEST(
```
vaibhav-datazip: why is 5% margin needed why not just increase it by 1 as margin ?

drivers/postgres/internal/backfill.go
```diff
@@ -73,24 +73,28 @@ func (p *Postgres) GetOrSplitChunks(ctx context.Context, pool *destination.Write
 
 func (p *Postgres) splitTableIntoChunks(ctx context.Context, stream types.StreamInterface) (*types.Set[types.Chunk], error) {
 	generateCTIDRanges := func(stream types.StreamInterface) (*types.Set[types.Chunk], error) {
-		var relPages, blockSize uint32
-		relPagesQuery := jdbc.PostgresRelPageCount(stream)
-		err := p.client.QueryRowContext(ctx, relPagesQuery).Scan(&relPages)
+		var relpages, maxPageID, blockSize uint32
+		// Fetch actual table page stats
+		storageStatsquery := jdbc.PostgresTableStorageStats(stream)
```
vaibhav-datazip: We are using blockSize in this query , which can also be extracted from this query only instead of firing a separate query for block size.

drivers/postgres/internal/backfill.go
```diff
@@ -73,24 +73,28 @@ func (p *Postgres) GetOrSplitChunks(ctx context.Context, pool *destination.Write
 
 func (p *Postgres) splitTableIntoChunks(ctx context.Context, stream types.StreamInterface) (*types.Set[types.Chunk], error) {
 	generateCTIDRanges := func(stream types.StreamInterface) (*types.Set[types.Chunk], error) {
-		var relPages, blockSize uint32
-		relPagesQuery := jdbc.PostgresRelPageCount(stream)
-		err := p.client.QueryRowContext(ctx, relPagesQuery).Scan(&relPages)
+		var relpages, maxPageID, blockSize uint32
```
vaibhav-datazip: This is not maxPageID it's page count right ? , so we can change the variable name that explains it better

pkg/jdbc/jdbc.go
vikaxsh: these were added to explain reason behind 5% margin

pkg/jdbc/jdbc.go
vikaxsh: it is only for keeping i a safe boundry, similar to as we are using 2^32 for end chunk id

drivers/postgres/internal/backfill.go
vikaxsh: it a very lightweight query won't affect performance

drivers/postgres/internal/backfill.go
vikaxsh: it is an upper bound so keeping it max page id make sense

vaibhav-datazip: its not an id rather number of pages

drivers/postgres/internal/backfill.go
vaibhav-datazip: it will remove the already existing lines of code, why fire query again

pkg/jdbc/jdbc.go
vaibhav-datazip: you can explain on top of the function, why do we need inside the query ?

drivers/postgres/internal/backfill.go
```diff
@@ -73,24 +73,35 @@ func (p *Postgres) GetOrSplitChunks(ctx context.Context, pool *destination.Write
...
-		batchSize := uint32(math.Ceil(float64(constants.EffectiveParquetSize) / float64(blockSize)))
 
-		relPages = utils.Ternary(relPages == uint32(0), uint32(1), relPages).(uint32)
+		// Calculate pages per chunk
+		batchPages := uint32(math.Ceil(float64(constants.EffectiveParquetSize) / float64(blockSize)))
+		batchSize := uint32(math.Ceil(float64(batchPages) / float64(partitionCount)))
+
+		if maxPageID <= 0 {
+			maxPageID = uint32(max(1, int(relPages)))
+		}
```
hash-data: use ternary operator

drivers/postgres/internal/backfill.go
```diff
@@ -178,3 +222,48 @@ func (p *Postgres) nextChunkEnd(ctx context.Context, stream types.StreamInterfac
...
+	Name  string
+	Pages uint32
+}
+
+// loadPartitionPages fetches partition-level relpages using PostgresPartitionPages query.
+func loadPartitionPages(ctx context.Context, db *sql.DB, stream types.StreamInterface) ([]PartitionPage, uint32, error) {
+	query := jdbc.PostgresPartitionPages(stream)
+	rows, err := db.QueryContext(ctx, query)
+	if err != nil {
+		return nil, 0, fmt.Errorf("failed to load partition pages: %v", err)
```
hash-data: use %s in err

drivers/postgres/internal/backfill.go
```diff
@@ -178,3 +222,48 @@ func (p *Postgres) nextChunkEnd(ctx context.Context, stream types.StreamInterfac
...
+	count := 0
+	for _, p := range partitions {
+		if p.Pages > end {
+			count++
+		}
+	}
+	if count == 0 {
+		count = 1 // fallback for non-partitioned tables
+	}
+	return count
```
hash-data: use ternary

pkg/jdbc/jdbc.go
```diff
@@ -130,6 +130,53 @@ func PostgresBlockSizeQuery() string {
 	return `SHOW block_size`
 }
 
+// PostgresPartitionPages returns total relpages for each partition and the parent table.
+// This can be used to dynamically adjust chunk sizes based on partition distribution.
+func PostgresPartitionPages(stream types.StreamInterface) string {
```
hash-data, line 135: check strings formatting

shubham19may (dismissed):

pkg/jdbc/jdbc.go
shubham19may: yes, implemented adoptive chunking as discussed

shubham19may (approved):

PR #613, opened 2025-11-02
fix: reset replication slot before starting performance test

vishalm0509:
# Description

This PR ensures that the PostgreSQL replication slot is reset before running performance tests, which prevents issues caused by using old or stale slots. Since the performance tests do not run frequently, an old slot might accumulate changes or state that can affect the test results or cause errors. By dropping and recreating the slot before each test run, the environment is always fresh, leading to more reliable and predictable performance testing.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- Run performance tests end-to-end
- Ensure in the database that slot is actually reset

# Screenshots or Recordings

## Documentation

- [] Documentation Link: [link to README, olake.io/docs, or olake-docs]

## Related PR's (If Any):

### DIFF
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -75,7 +75,7 @@ jobs:
         mkdir -p ~/.ssh
         echo "${{ secrets.TESTING_SERVER_SSH_KEY }}" | base64 --decode > ~/.ssh/key.pem
         chmod 600 ~/.ssh/key.pem
-        ssh -o StrictHostKeyChecking=no -i ~/.ssh/key.pem testing-server-1@172.31.32.11 "wake-up ${{ matrix.instance_id }}"
+        ssh -o StrictHostKeyChecking=no -o ConnectTimeout=30 -i ~/.ssh/key.pem ${{ secrets.TESTING_SERVER_SSH_USER_HOST }} "/usr/local/bin/wake-up ${{ matrix.instance_id }}"
         echo "Waiting 2 minutes for database to be ready..."
         sleep 120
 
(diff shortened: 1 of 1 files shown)


drivers/postgres/internal/postgres_test_util.go
```diff
@@ -146,6 +146,17 @@ func ExecuteQuery(ctx context.Context, t *testing.T, streams []string, operation
...
+		dropQuery := `SELECT pg_drop_replication_slot('performance_test')
+						WHERE EXISTS (SELECT 1 FROM pg_replication_slots WHERE slot_name = 'performance_test');`
+		_, err := db.ExecContext(ctx, dropQuery)
+		require.NoError(t, err, fmt.Sprintf("failed to execute %s operation", operation), err)
+
+		createQuery := `SELECT pg_create_logical_replication_slot('performance_test', 'pgoutput');`
+		_, err = db.ExecContext(ctx, createQuery)
+		require.NoError(t, err, fmt.Sprintf("failed to execute %s operation", operation), err)
+		return
+
```
vikaxsh: Wouldn't it be better if we could wrap both queries into a single statement?

vishalm0509: I think this is more explicit. Just 2 quick queries. What say

drivers/mysql/internal/mysql_test_util.go
```diff
@@ -141,6 +141,18 @@ func ExecuteQuery(ctx context.Context, t *testing.T, streams []string, operation
...
+		cdcSettings := map[string]string{
+			"binlog_format":       "ROW",
+			"binlog_row_image":    "FULL",
+			"binlog_row_metadata": "FULL",
+		}
+		for variable, value := range cdcSettings {
+			_, err := db.ExecContext(ctx, fmt.Sprintf("SET GLOBAL %s = '%s'", variable, value))
+			require.NoError(t, err, fmt.Sprintf("failed to SET GLOBAL %s = %s", variable, value))
+		}
+		return
```
vikaxsh: can we  Set these  once in mysql.cnf file?

vishalm0509: Removed

PR #615, opened 2025-11-05
feat: Add configurable compression, file size, and row group size for Parquet files

### LINKED ISSUE #572: Ability to configure Compression, Sorting, FileSzie and Row Group size for output Parquet files
### Problem

Size of Parquet file very depends on data and how data stored inside file. Once file is in DataLake its hard to change it and experience of using large and not optimal generated Parquet file will lead to higher cost ( storage, network, CPU) and also show low performance.

### Solution

Parquet spec provides different settings that could be used to make file more efficient like  
- different compression algorithms ( zstd, lz4, etc)
- different  encoding types ( plain/RLE dictionary, delta,e tc) 
- row group size

Its would be hard to implement all Parquet features but some of them that will bring a lot of value would be really important to have

Configure default values with ability to override these settings per table
1. Specify compression codec
2. Specify max Parquet file size
3. Specify row group size
4. Specify column and sort data  based on this column

These settings will give more control and help to make output Parquet files less in size and more efficient to read which will bring better performance and lower infra costs.

vamsi-op:
## Description

This PR implements the feature requested in #572 to provide configuration options for optimizing Parquet file output.

### Changes Implemented

1. **Compression Codec Configuration**
   - Added support for multiple compression algorithms: snappy (default), gzip, zstd, lz4, and uncompressed
   - Configurable via compression field in the writer config
   - Includes validation for supported codecs

2. **Maximum File Size Control**
   - Added max_file_size configuration to control Parquet file size
   - Implements automatic file rotation when size limit is reached
   - Default: 128MB, can be disabled by setting to 0
   - Tracks approximate file size and record count per partition

3. **Row Group Size Configuration**
   - Added row_group_size option for fine-tuning row group size
   - Allows users to balance compression vs memory usage
   - Uses parquet-go default if not specified

4. **Sort Columns (Prepared for future)**
   - Added sort_columns field in config and spec
   - Marked for future implementation with TODO comments
   - Requires buffering and sorting logic to be implemented

### Technical Details

- Updated Config struct with new fields and validation
- Refactored file closing logic into reusable closeAndUploadPartitionFile method
- Enhanced Write method to check file size and rotate files automatically
- Updated spec.json with new configuration options including descriptions and defaults
- Added helper methods: getCompressionCodec(), getMaxFileSize(), getRowGroupSize()

### Benefits

- Smaller file sizes through optimized compression
- Better query performance and reduced storage costs
- Fine-grained control for specific use cases
- Cost savings through reduced storage and network usage

Closes #572


vaibhav-datazip:
@vamsi-op , please rebase your branch to staging, its currently raised on master

vamsi-op:
@vaibhav-datazip Rebased to staging branch

PR #626, opened 2025-11-11
fix: toast values placeholder fix

ImDoubD-datazip:
# Description

During updation in cdc, if updated value is non-toasted value, then a placeholder `_olake_unavailable_value` is put in place for toasted value. This is done as wal does not has logs for toasted columns if replica identity is default, so to remove confusion with 'none', this above placeholder is introduced.

## Type of change

# How Has This Been Tested?

- insertion change of toast column description
Screenshot 2025-11-11 at 3 44 10 PM

- updation changes of non-toasted column (here placeholder for toasted value has occured)
Screenshot 2025-11-11 at 3 45 58 PM

- when null toasted column record inserted (id = 12)
Screenshot 2025-11-11 at 3 47 10 PM

- when toasted column updated to null
image

## Documentation

## Related PR's (If Any):

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -16,6 +16,7 @@ const (
 	MongoPrimaryID         = "_id"
 	OlakeID                = "_olake_id"
 	OlakeTimestamp         = "_olake_timestamp"
+	OlakeUnavailableValue  = "_olake_unavailable_value"
 	OpType                 = "_op_type"
 	CdcTimestamp           = "_cdc_timestamp"
 	DBName                 = "_db"
--- a/pkg/waljs/pgoutput.go
+++ b/pkg/waljs/pgoutput.go
@@ -7,7 +7,9 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


constants/constants.go
```diff
@@ -16,6 +16,7 @@ const (
 	MongoPrimaryID         = "_id"
 	OlakeID                = "_olake_id"
 	OlakeTimestamp         = "_olake_timestamp"
+	OlakeUnavailableValue  = "_olake_unavailable_value"
```
vaibhav-datazip, line 19: shouldn't we show the value of the toast value, which is actually present in the missing row, in the WAL. 

this can be done by querying the missing column

ImDoubD-datazip: it does not comes in WAL. we have to query it or persist the value in state during insertion

shubham19may:
Lets close this as we have not finalised any approach over it yet. This is not solving it for upsert case

PR #627, opened 2025-11-11
fix: mongodb cursor _id not syncing records

vaibhav-datazip:
# Description

In mongodb the cursor value in case of ObjectId was getting saved in the format of ObjectId(number). which was getting parsed as a string which led to id <= string, causing no data to get synced . This has been fixed by saving cursor of object id type as hex in state.

Also there was a timestamp parsing issue because the mongodb format of time stamp was not listed causing the timestamp to get changed to epoch start.

## Type of change

# How Has This Been Tested?

- [x] Synced Mongodb in incremental mode using various datatypes of cursor field.

## Documentation

### DIFF
--- a/types/stream_configured.go
+++ b/types/stream_configured.go
@@ -85,35 +85,51 @@ func (s *ConfiguredStream) Cursor() (string, string) {
 }
 
 func (s *ConfiguredStream) GetFilter() (Filter, error) {
-	filter := s.StreamMetadata.Filter
+	filter := strings.TrimSpace(s.StreamMetadata.Filter)
 	if filter == "" {
 		return Filter{}, nil
 	}
-	// TODO: handle special characters in column name in filter
-	// example: a>b, a>=b, a<b, a<=b, a!=b, a=b, a="b", a=\"b\" and c>d, a="b" or c>d
-	var FilterRegex = regexp.MustCompile(`^(\w+)\s*(>=|<=|!=|>|<|=)\s*(\"[^\"]*\"|\d*\.?\d+|\w+)\s*(?:(and|or)\s*(\w+)\s*(>=|<=|!=|>|<|=)\s*(\"[^\"]*\"|\d*\.?\d+|\w+))?\s*$`)
+	// FilterRegex supports the following filter patterns:
+	// Single condition:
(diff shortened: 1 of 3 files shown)


drivers/abstract/incremental.go
vaibhav-datazip: update the comment as now this function will also handle the Object ID datatype

drivers/mongodb/internal/incremental.go
```diff
@@ -152,19 +152,25 @@ func (m *Mongo) ThresholdFilter(stream types.StreamInterface) (bson.A, error) {
...
 			return nil, fmt.Errorf("failed to convert secondary cursor value: %s", err)
 		}
-		conditions = append(conditions, bson.D{
-			{Key: secondaryCursor, Value: bson.D{{Key: "$lte", Value: formattedSecondaryValue}}},
+		condition := buildMongoCondition(types.Condition{
+			Column:   secondaryCursor,
+			Value:    fmt.Sprintf("%v", formattedSecondaryValue),
+			Operator: "<=",
 		})
+		conditions = append(conditions, condition)
```
ImDoubD-datazip:
```
conditions = append(conditions, buildMongoCondition(types.Condition{
			Column:   primaryCursor,
			Value:    fmt.Sprintf("%v", formattedPrimaryValue),
			Operator: "<=",
		}))
```

drivers/abstract/incremental.go
ImDoubD-datazip: please remove abstract driver method receiver as it is of no use. also please add a comment what does this do.

drivers/abstract/incremental.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/logger"
 	"github.com/datazip-inc/olake/utils/typeutils"
+	"go.mongodb.org/mongo-driver/bson/primitive"
```
shubham19may: IMO abstract should never have knowledge of respective driver types

shubham19may: we can create respective methods for each driver to override its own logic of formatting cursor key.

PR #631, opened 2025-11-12
fix: clear destination global changes

ImDoubD-datazip:
# Description

- If all the global persisted streams are dropped, global state is marked nil.
- Also on lsn mismatch, sync is now set to be failing as previously on backfill this will lead to duplication of data.
- Also on in incremental, on primary and secondary values being nil through out, sync will be failing as previously on backfill duplication of data will take place

## Type of change

# How Has This Been Tested?
This has been tested running clear destination with all the streams to be dropped.

## Documentation

## Related PR's (If Any):

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -46,4 +46,4 @@ const (
 
 var RelationalDrivers = []DriverType{Postgres, MySQL, Oracle}
 
-var NonRetryableErrors = []string{DestError, "context canceled", NoRecordsFoundError, LSNNotUpdatedError}
+var NonRetryableErrors = []string{DestError, "context canceled", NoRecordsFoundError, LSNNotUpdatedError, "lsn mismatch"}


hash-data:
let us remove destination db as well from specs of destination

utils/spec/uischema.go
```diff
@@ -190,16 +190,16 @@ const IcebergUISchema = `{
...
-      { "rest_signing_region": 12, "rest_signing_v_4": 12, "scope": 12 },
+      { "catalog_type": 12, "rest_catalog_url": 12 },
+      { "hive_uri": 12, "jdbc_url": 12 },
+      { "jdbc_username": 12, "jdbc_password": 12 },
+      { "iceberg_s3_path": 12, "hive_clients": 12 },
+      { "s3_use_ssl": 12, "hive_sasl_enabled": 12 },
+      { "s3_path_style": 12, "rest_auth_type": 12 },
+      { "token": 12, "oauth2_uri": 12 },
+      { "credential": 12, "no_identifier_fields": 12 },
+      { "rest_signing_name": 12, "rest_signing_region": 12 },
```
hash-data: why this change

vishalm0509: I have made some changes now. This is changed because we have removed `iceberg_db` from the iceberg spec.

vishalm0509:
@ImDoubD-datazip I have tested it after the spec related changes with the released image

drivers/abstract/incremental.go
```diff
@@ -37,14 +37,15 @@ func (a *AbstractDriver) Incremental(ctx context.Context, pool *destination.Writ
 				return fmt.Errorf("failed to fetch max cursor values: %s", err)
 			}
 
+			// if all primary and secondary (if given) cursor values are nil, then failing sync and suggesting to proceed with clear destination, as otherwise on backfill, duplication will take place
```
hash-data: a.state.resetcursor must not be called

drivers/postgres/internal/cdc.go
```diff
@@ -72,13 +72,10 @@ func (p *Postgres) PreCDC(ctx context.Context, streams []types.StreamInterface)
...
-			// TODO: handle cursor mismatch with user input (Example: user provide if it has to fail or do full load with new resume token)
-			// if confirmed flush lsn is not same as stored in state
+			// failing sync when lsn mismatch found (from state and confirmed flush lsn), as otherwise on backfill, duplication of data will occur
+			// suggesting to proceed with clear destination
 			if parsed != socket.ConfirmedFlushLSN {
-				logger.Warnf("lsn mismatch, backfill will start again. prev lsn [%s] current lsn [%s]", parsed, socket.ConfirmedFlushLSN)
-				if err := fullLoadAck(); err != nil {
-					return fmt.Errorf("failed to ack lsn for full load: %s", err)
-				}
+				return fmt.Errorf("lsn mismatch, please proceed with clear destination. lsn saved in state [%s] current lsn [%s]", parsed, socket.ConfirmedFlushLSN)
```
hash-data, line 78: add this to non retryable error

drivers/abstract/incremental.go
ImDoubD-datazip: removed for now.

PR #634, opened 2025-11-14
fix: lsn not getting confirmed after acknowledgment

hash-data:
# Description
After acknowledging the confirmed flush, the LSN is not getting updated

Added a loop to wait for the confirmed flush lsn to be updated. Loop breaks if the lsn is not updated in 5 minutes.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/pkg/waljs/replicator.go
+++ b/pkg/waljs/replicator.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/datazip-inc/olake/constants"
 	"github.com/datazip-inc/olake/drivers/abstract"
 	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/logger"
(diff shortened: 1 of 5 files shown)


pkg/waljs/replicator.go
```diff
@@ -147,7 +154,34 @@ func AcknowledgeLSN(ctx context.Context, socket *Socket, fakeAck bool) error {
...
+	}
+
+	// wait for slot to be updated
+	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+
+	ticker := time.NewTicker(3 * time.Second)
+	defer ticker.Stop()
+
+	for {
```
ImDoubD-datazip, line 170: Can you just add a comment of why we are doing this, so that if any full proof solution comes for this then we can update this.

hash-data: we not have information for the disrepency once we have we will add a comment, for now there is `// wait for slot to be updated`

ImDoubD-datazip: ok

ImDoubD-datazip (approved):

PR #646, opened 2025-11-20
fix: time parsing of string values

vaibhav-datazip:
# Description

Earlier some of the string values were incorrectly getting converted to epoch start time which resulted in incorrect data in the destination.

- this also adds some of the missing parsing cases in reformat.go
- makes array of values output consistent
- adds version in state file and run certain parts based on that version to maintain backward compatibility 

## Type of change

# How Has This Been Tested?

- [x] Ran sync using postgres driver , with and without state (cdc mode)
- [x] Ran postgres sync with various datatype supported
- [x] Ran mysql sync with invalid date time to check its getting converted to epoch start. incase if its in string format and unable to parse then return as it is. (ran for cdc)
- [x] for older version of state file backward compatibility is being maintained. 

## Cases where the sync will fail
* In mongodb the field is a string but matches the timestamp format, during discover this will be treated as timestamp. But as this field was originally a string and later on some string comes in a format which dosent match timestamp format  the parsing will fail which will result in sync failure.

# Screenshots or Recordings
- How  array of values is made consistent 
  - during full refresh the data is :
     Screenshot 2025-11-21 at 1 09 51 PM

  - after cdc sync earlier it was getting synced as 
     Screenshot 2025-11-21 at 1 07 56 PM

  - now corrected , synced similar to the format of full refresh  
    Screenshot 2025-11-21 at 1 12 21 PM

## Documentation

## Related PR's (If Any):

### DIFF
--- a/protocol/discover.go
+++ b/protocol/discover.go
@@ -90,11 +90,11 @@ func compareStreams() error {
 	if err := logger.FileLoggerWithPath(diffCatalog, viper.GetString(constants.DifferencePath)); err != nil {
 		return fmt.Errorf("failed to write difference streams: %s", err)
 	}
+	logger.Infof("Successfully wrote stream differences")
 	message := types.Message{
 		Type:    types.CatalogMessage,
 		Catalog: diffCatalog,
 	}
 	logger.Info(message)
-	logger.Infof("Successfully wrote stream differences")
 	return nil
 }
--- a/protocol/sync.go
+++ b/protocol/sync.go
@@ -72,9 +72,7 @@ var syncCmd = &cobra.Command{
 		} else {
 			state.Version = constants.LatestStateVersion
 		}
-		// Loaded state version is used during the sync process to maintain backward compatibility
-		// if the state file dosen't exists it's created with latest version
-		// if state file dosen't have version it's set as 0
+		// set state version to global variable to use throughout olake cli instance
 		constants.LoadedStateVersion = state.Version
 
 		state.RWMutex = &sync.RWMutex{}
--- a/utils/typeutils/reformat.go
+++ b/utils/typeutils/reformat.go
@@ -233,8 +233,6 @@ func ReformatDate(v interface{}, isTimestampInDB bool) (time.Time, error) {
 // parseStringTimestamp expects value and isTimestampInDB boolean
 // if this is a timestamp and unable to parse into correct format it will return epoch start time
 // if its a string and unable to parse into correct time format it will be returned as string
-// TODO: for MongoDB currently the string in timestamp format is being passed in parseStringTimestamp
-// this can lead to sync failure if further-on in the database the string is of invalid timestamp format
 func parseStringTimestamp(value string, isTimestampInDB bool) (time.Time, error) {
 	// Check if the string starts with a date pattern (YYYY-MM-DD)
 	startsWithDatePattern := func(value string) bool {
--- a/utils/typeutils/datatype.go
+++ b/utils/typeutils/datatype.go
@@ -40,6 +40,7 @@ func TypeFromValue(v interface{}) types.DataType {
 	case reflect.Float64:
 		return types.Float64
 	case reflect.String:
+		// NOTE: If the string is in correct datetime format, it will be detected as timestamp and returned as timestamp datatype
 		t, err := ReformatDate(v, false)
 		if err == nil {
 			return detectTimestampPrecision(t)


protocol/clear.go
```diff
@@ -30,14 +32,23 @@ var clearCmd = &cobra.Command{
...
-			Type: types.StreamType,
+			Type:    types.StreamType,
+			Version: 0,
 		}
 		if statePath != "" {
 			if err := utils.UnmarshalFile(statePath, state, false); err != nil {
-				return err
+				if strings.Contains(err.Error(), "does not exist") {
+					state.Version = constants.StateVersion
+				} else {
```
hash-data: so if state does not have version then you are setting 1, how this is helping us?

pkg/waljs/pgoutput.go
```diff
@@ -161,7 +161,7 @@ func (p *pgoutputReplicator) tupleValuesToMap(rel *pglogrepl.RelationMessage, tu
 
 		// Convert according to OID to string
 		typeName := oidToString(colType)
-		val, err := p.socket.changeFilter.converter(string(col.Data), typeName)
+		val, err := p.socket.changeFilter.converter(col.Data, typeName)
```
hash-data, line 164: any consequences like bytes data ?

protocol/clear.go
```diff
@@ -30,14 +32,23 @@ var clearCmd = &cobra.Command{
...
+					return err
+				}
 			}
+		} else {
+			state.Version = constants.StateVersion
```
hash-data: let us think of omitempty kind of flag in json where you dont need to add such check

protocol/sync.go
```diff
@@ -63,15 +63,23 @@ var syncCmd = &cobra.Command{
...
 		if statePath != "" {
 			if err := utils.UnmarshalFile(statePath, state, false); err != nil {
-				return err
+				if strings.Contains(err.Error(), "does not exist") {
+					state.Version = constants.StateVersion
+				} else {
+					return err
+				}
 			}
+		} else {
```
hash-data, line 72: same here

hash-data: we dont need to have state version to be set here think in that way. it will make state complex to handle

types/state.go
```diff
@@ -65,15 +65,29 @@ func (s *State) initStreamState(stream *ConfiguredStream) *StreamState {
...
 	ErrStateCursorMissing = errors.New("cursor field missing from state")
+	// stateVersion stores the current state file version for backward compatibility checks
+	// Defaults to 0 for old state files without version
+	stateVersion int32
 )
 
+// SetStateVersion sets the global state version (called when state is loaded or created)
+func SetStateVersion(version int32) {
+	atomic.StoreInt32(&stateVersion, int32(version))
+}
```
hash-data: why is it required to be atomic?

types/state.go
```diff
@@ -65,15 +65,29 @@ func (s *State) initStreamState(stream *ConfiguredStream) *StreamState {
...
+
+// GetStateVersion returns the current state version (defaults to 0 if not set)
+func GetStateVersion() int32 {
+	return int32(atomic.LoadInt32(&stateVersion))
+}
```
hash-data: why it is returning from stateVersion variable, why not s.Version kind of ?

utils/typeutils/reformat.go
```diff
@@ -265,6 +271,12 @@ func parseStringTimestamp(value string) (time.Time, error) {
 		}
 	}
 
+	// time unable to be parsed string will be returned as it state version >= 1 (backward compatibility)
+	if !isTimestamp && types.GetStateVersion() >= constants.StateVersion {
```
hash-data: we might need to have a state.txt file where we are going to track state version and what changed there. 

Also here constant variable not making sense.

utils/typeutils/reformat.go
```diff
@@ -308,6 +320,13 @@ func ReformatInt64(v any) (int64, error) {
 			return int64(0), fmt.Errorf("failed to change string %v to int64: %v", v, err)
 		}
 		return intValue, nil
+	case []uint8:
+		strVal := string(v)
+		intValue, err := strconv.ParseInt(strVal, 10, 64)
+		if err != nil {
+			return int64(0), fmt.Errorf("failed to change []byte %v to int64: %v", v, err)
+		}
+		return intValue, nil
```
hash-data, line 331: all these kind of changes also not supposed to come in older syncs, like above array and boolean function are being updated.

types/state.go
```diff
@@ -65,15 +65,29 @@ func (s *State) initStreamState(stream *ConfiguredStream) *StreamState {
...
+	// Defaults to 0 for old state files without version
+	stateVersion int32
 )
 
+// SetStateVersion sets the global state version (called when state is loaded or created)
+func SetStateVersion(version int32) {
+	atomic.StoreInt32(&stateVersion, int32(version))
+}
+
+// GetStateVersion returns the current state version (defaults to 0 if not set)
```
hash-data: we can have ctx where we have already set or we can also set in environment about state version. Let us discuss what is better approach.

pkg/waljs/pgoutput.go
vaibhav-datazip: that is handled as we have added bytes conversion in every reformat function now.

protocol/clear.go
vaibhav-datazip: This is not checking version in state file, this is checking the state file exists or not . 

if not , it will use state version as the current saved in constants.go

constants/constants.go
```diff
@@ -32,6 +32,7 @@ const (
 	DestinationDatabasePrefix = "DESTINATION_DATABASE_PREFIX"
 	// EffectiveParquetSize is the effective size in bytes considering 256mb targeted parquet size, compression ratio as 8
 	EffectiveParquetSize = int64(256) * 1024 * 1024 * int64(8)
+	StateVersion         = 1 // version of state file to be used for backward compatibility
```
vikaxsh: the naming is confusing here name it to latestStateVersion

constants/constants.go
```diff
@@ -47,3 +48,6 @@ const (
 var RelationalDrivers = []DriverType{Postgres, MySQL, Oracle}
 
 var NonRetryableErrors = []string{DestError, "context canceled", NoRecordsFoundError, LSNNotUpdatedError, "lsn mismatch"}
+
+// CurrentStateVersion stores the version of the state file being used
+var CurrentStateVersion = 1
```
vikaxsh:
```suggestion
var LoadedStateVersion = 1
```

protocol/sync.go
```diff
@@ -61,15 +61,20 @@ var syncCmd = &cobra.Command{
...
 		state = &types.State{
-			Type: types.StreamType,
+			Type:    types.StreamType,
+			Version: constants.StateVersion, // Default to current version
 		}
+
+		// Load existing state if available, otherwise use empty state
 		if statePath != "" {
 			if err := utils.UnmarshalFile(statePath, state, false); err != nil {
+				logger.Debugf("State file not available, starting with empty state: %v", err)
```
vikaxsh: the log is saying "starting with" and returning error, Is this the intended behavior?

protocol/clear.go
```diff
@@ -30,14 +31,21 @@ var clearCmd = &cobra.Command{
...
+
+		// Load existing state if available, otherwise use empty state
 		if statePath != "" {
 			if err := utils.UnmarshalFile(statePath, state, false); err != nil {
+				logger.Debugf("State file not available, starting with empty state: %v", err)
 				return err
 			}
 		}
+		constants.CurrentStateVersion = constants.StateVersion
+		state.Version = constants.CurrentStateVersion
```
vikaxsh: why are we using latest state version for clear destination and not for sync?

utils/typeutils/reformat.go
```diff
@@ -145,19 +151,19 @@ func ReformatBool(v interface{}) (bool, error) {
 }
 
 // reformat date
-func ReformatDate(v interface{}) (time.Time, error) {
+func ReformatDate(v interface{}, isTimestamp bool) (time.Time, error) {
```
vikaxsh: please add an comment explaining about isTimestamp

utils/typeutils/reformat.go
```diff
@@ -266,6 +272,12 @@ func parseStringTimestamp(value string) (time.Time, error) {
 		}
 	}
 
+	// time unable to be parsed string will be returned as it state version >= 1 (backward compatibility)
+	if !isTimestamp && constants.CurrentStateVersion >= constants.StateVersion {
+		// return the string with no error in case passed value was a stirng as
```
vikaxsh: comment is saying returning with no error, but code return an error

constants/constants.go
vikaxsh: I have concerns about the state versioning approach: Each new version requires writing migration logic (v0→v1, v1→v2, etc.)
We need to maintain backward compatibility across all versions
It's easy to forget what behavior changed in which version

let's discuss

utils/typeutils/reformat.go
```diff
@@ -266,6 +275,11 @@ func parseStringTimestamp(value string) (time.Time, error) {
 		}
 	}
 
+	// time unable to be parsed string will be returned as it state version >= 1 (backward compatibility)
+	if !isTimestamp && constants.LoadedStateVersion >= constants.LatestStateVersion {
```
vikaxsh: in which case loadedStateVersion could be greater then latest?

vikaxsh: shouldn't we check this for loaded version == 0 , after that it fixed?

protocol/sync.go
```diff
@@ -61,15 +61,19 @@ var syncCmd = &cobra.Command{
 
 		syncID = utils.ComputeConfigHash(configPath, destinationConfigPath)
 
-		// default state
+		// Initialize state
 		state = &types.State{
-			Type: types.StreamType,
+			Type:    types.StreamType,
+			Version: 0, // Default to 0 if state file doesn't have version
```
vikaxsh: do we need to initialize version as 0?

vaibhav-datazip: its a good practice as it tells clearly in the code itself that version will be defaulted to 0.

utils/typeutils/reformat.go
vaibhav-datazip: there is a flaw in this logic I wrote as well as what you told.
 if you try to understand, what this does is it checks if loaded state is >= latest state version it will run this thing, but imagine in future we increase latest state version as well. in that case this condition wont even run for state version 1, we want this to run when state version !=0 right ?

I will correct it, this will also solve your above comment.

utils/typeutils/reformat.go
```diff
@@ -224,7 +230,10 @@ func ReformatDate(v interface{}) (time.Time, error) {
 	return parsed, nil
 }
 
-func parseStringTimestamp(value string) (time.Time, error) {
+// parseStringTimestamp expects value and isTimestamp boolean
+// if this is a timestamp and unable to parse into correct format it will return epoch start time
+// if its a string and unable to parse into correct time format it will be returned as string
+func parseStringTimestamp(value string, isTimestamp bool) (time.Time, error) {
```
hash-data:
```suggestion
func parseStringTimestamp(value string, isTimestampInDB bool) (time.Time, error) {
```

hash-data: or something that says it is from db

drivers/mongodb/internal/backfill.go
```diff
@@ -415,7 +415,7 @@ func buildMongoCondition(cond types.Condition) bson.D {
 		if strings.ToLower(val) == "true" || strings.ToLower(val) == "false" {
 			return strings.ToLower(val) == "true"
 		}
-		if timeVal, err := typeutils.ReformatDate(val); err == nil {
+		if timeVal, err := typeutils.ReformatDate(val, false); err == nil {
```
hash-data, line 418: we must fail in this case if it is not reformatable, (kind of invalid string in state, which will return 1970 value)

protocol/clear.go
```diff
@@ -30,14 +31,18 @@ var clearCmd = &cobra.Command{
...
+			Type:    types.StreamType,
+			Version: constants.LatestStateVersion, // Default to 0 if state file doesn't have version
 		}
 		if statePath != "" {
 			if err := utils.UnmarshalFile(statePath, state, false); err != nil {
 				return err
 			}
 		}
+		state.Version = constants.LatestStateVersion
+		constants.LoadedStateVersion = state.Version
```
hash-data: there is a bug on this line, 
-> versioning is not stream level property 
-> clear destination can run for single stream as well

protocol/clear.go
```diff
@@ -30,14 +31,18 @@ var clearCmd = &cobra.Command{
...
+		// Initialize state - clear command needs existing state for stream classification
 		state = &types.State{
-			Type: types.StreamType,
+			Type:    types.StreamType,
+			Version: constants.LatestStateVersion, // Default to 0 if state file doesn't have version
 		}
 		if statePath != "" {
 			if err := utils.UnmarshalFile(statePath, state, false); err != nil {
 				return err
 			}
```
hash-data: this is getting handled at two places can we abstract it out

protocol/sync.go
hash-data: in clear destination it is 1 here 0 (inconsistent)

utils/typeutils/reformat.go
```diff
@@ -137,27 +141,29 @@ func ReformatBool(v interface{}) (bool, error) {
 		default:
 			return false, fmt.Errorf("found to be boolean, but value is not boolean : %v", v)
 		}
+	case []uint8:
+		return string(booleanValue) == "t", nil
```
hash-data, line 145: can it be "true" as well?

utils/typeutils/reformat.go
```diff
@@ -109,7 +110,10 @@ func ReformatValue(dataType types.DataType, v any) (any, error) {
 		if value, isArray := v.([]any); isArray {
 			return value, nil
 		}
-
+		// Handle byte arrays, convert to string
```
hash-data, line 113: not added state version check, is that fine?

utils/typeutils/time.go
```diff
@@ -13,7 +13,7 @@ type Time struct {
 func (ct *Time) UnmarshalJSON(b []byte) error {
 	// Remove the quotes around the date string
 	str := strings.Trim(string(b), "\"")
-	time, err := parseStringTimestamp(str)
+	time, err := parseStringTimestamp(str, false)
```
hash-data: it is time unmarshalling only so why false?

utils/typeutils/reformat.go
vaibhav-datazip: the case which I encountered had output as t and f, will test it with other drivers as well after all code changes are done after review.

protocol/clear.go
vaibhav-datazip: should we make versioning a stream level property then ?

utils/typeutils/reformat.go
vaibhav-datazip: this is going to return string only, so the sync wont break

utils/typeutils/time.go
vaibhav-datazip: this is being extracted from string only, so format input might not be generic (commonly used by db). that's why added false here.

protocol/clear.go
vaibhav-datazip: okay will remove the changes from clear.go

utils/typeutils/reformat.go
vaibhav-datazip: the bytes array is only getting returned in case of postgress from all the available drivers as of now. and it only returns t and f

protocol/clear.go
vaibhav-datazip: removed this from clear.go

protocol/sync.go
```diff
@@ -61,15 +61,19 @@ var syncCmd = &cobra.Command{
...
 		}
 		if statePath != "" {
 			if err := utils.UnmarshalFile(statePath, state, false); err != nil {
 				return err
 			}
+		} else {
+			state.Version = constants.LatestStateVersion
```
hash-data, line 73: can we set the latest version before unmarshal

vaibhav-datazip: no setting this some value will set state.version as the latest version , after that during unmarshal , if version isn't available in file it keep the earlier set version and hence the original file version was 0 but after this sync with state it will change to latest.

did a test for this and got this result

protocol/sync.go
```diff
@@ -61,15 +61,18 @@ var syncCmd = &cobra.Command{
...
 			Type: types.StreamType,
 		}
 		if statePath != "" {
 			if err := utils.UnmarshalFile(statePath, state, false); err != nil {
 				return err
 			}
+		} else {
+			state.Version = constants.LatestStateVersion
 		}
+		constants.LoadedStateVersion = state.Version
```
hash-data, line 76: add comment on top of it what we are doing

constants/state_version.go
```diff
@@ -0,0 +1,26 @@
...
+//
+// Version History:
+//   - Version 0: Legacy format (backward compatibility)
+//     * More lenient date/timestamp parsing behavior
+//     * When a string cannot be parsed as a timestamp, it returns epoch time (1970-01-01)
+//     * Used for state files created before version 1 was introduced
+//
+//   - Version 1: Current format (introduced stricter validation)
+//     * Stricter date/timestamp parsing validation
+//     * When a string cannot be parsed as a timestamp (and isTimestampInDB is false), it will be returned as string
```
hash-data: not required to mention about isTimestampInDB

hash-data: can you mention the issue that we solved using this, like string fallback on 1970

protocol/sync.go
hash-data: not added yet

protocol/discover.go
```diff
@@ -85,6 +85,13 @@ func compareStreams() error {
 	}
 
 	diffCatalog := types.GetStreamsDelta(&oldStreams, &newStreams)
+	// log the difference catalog to stdout
+	message := types.Message{
+		Type:    types.CatalogMessage,
+		Catalog: diffCatalog,
+	}
+	logger.Info(message)
+
```
hash-data: move after file log

hash-data:
add todo as well on mongodb issue

protocol/discover.go
```diff
@@ -86,15 +86,15 @@ func compareStreams() error {
 
 	diffCatalog := types.GetStreamsDelta(&oldStreams, &newStreams)
 	// log the difference catalog to stdout
+
+	if err := logger.FileLoggerWithPath(diffCatalog, viper.GetString(constants.DifferencePath)); err != nil {
+		return fmt.Errorf("failed to write difference streams: %s", err)
+	}
```
hash-data, line 92: move line 98 as well

protocol/sync.go
```diff
@@ -72,6 +72,9 @@ var syncCmd = &cobra.Command{
 		} else {
 			state.Version = constants.LatestStateVersion
 		}
+		// Loaded state version is used during the sync process to maintain backward compatibility
+		// if the state file dosen't exists it's created with latest version
+		// if state file dosen't have version it's set as 0
```
hash-data: simple comment:
set state version to global variable to use throughout olake cli instance

utils/typeutils/reformat.go
```diff
@@ -233,6 +233,8 @@ func ReformatDate(v interface{}, isTimestampInDB bool) (time.Time, error) {
 // parseStringTimestamp expects value and isTimestampInDB boolean
 // if this is a timestamp and unable to parse into correct format it will return epoch start time
 // if its a string and unable to parse into correct time format it will be returned as string
+// TODO: for MongoDB currently the string in timestamp format is being passed in parseStringTimestamp
+// this can lead to sync failure if further-on in the database the string is of invalid timestamp format
```
hash-data: actually we can remove this for now

vaibhav-datazip: but we should create the issue ? so that someone knows that we are aware about it

hash-data (approved):

PR #655, opened 2025-11-28
chore: upgrading the initial wait time in postgres

hash-data:
# Description
Increasing the initial wait time for first message in postgres. 
In previous implementation :
-  there was upper bound limit of 2400 seconds, this pr removed that limit. 
-  the default timeout 1200 seconds also increased to 10800 seconds (3 hours).

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/drivers/postgres/resources/spec.json
+++ b/drivers/postgres/resources/spec.json
@@ -93,9 +93,8 @@
                         "initial_wait_time": {
                             "type": "integer",
                             "title": "Initial Wait Time",
-                            "description": "Idle timeout for wal catchup to Saved LSN, Range: (120,2400]",
-                            "minimum": 120,
-                            "maximum": 2400
+                            "description": "Idle timeout for wal catchup to Saved LSN",
+                            "minimum": 120
                         },
                         "replication_slot": {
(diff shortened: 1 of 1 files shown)


utils/typeutils/reformat.go
```diff
@@ -265,7 +265,6 @@ func parseStringTimestamp(value string) (time.Time, error) {
 		}
 	}
 
-	logger.Warnf("Invalid datetime detected, failed to parse: %s. Converting to epoch start time (1970-01-01 00:00:00 UTC)", value)
```
shubham19may, line 268: can we keep it, so if there is any bad date format, we can detect? or this is being changed by Vaibhav PR?

shubham19may (dismissed):

utils/typeutils/reformat.go
hash-data: I am removing it as we know issue now, about string. It is just adding logs which is going to be problem while doing full load.

PR #657, opened 2025-11-28
feat: OLake full refresh sync performance improvement

vaibhav-datazip:
# Description

Olake's Performance improvement PR.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/destination/iceberg/iceberg.go
+++ b/destination/iceberg/iceberg.go
@@ -352,6 +352,7 @@ func (i *Iceberg) FlattenAndCleanData(ctx context.Context, records []types.RawRe
 
 			// if schema difference is not detected, detect schema difference
 			if !diffThreadSchema.Load() {
+				// when detectChange is true, the function does not modify schema parameter
 				if changeDetected, err := detectOrUpdateSchema(records[idx], true, i.schema, i.schema); err != nil {
 					return fmt.Errorf("failed to detect schema: %s", err)
 				} else if changeDetected {
--- a/pkg/jdbc/reader.go
+++ b/pkg/jdbc/reader.go
@@ -87,6 +87,7 @@ func normalizeDataTypeAndConvert(rawData any, colType *sql.ColumnType, converter
 	return conv, nil
 }
 
+// TODO: Use MapScanConcurrent instead of MapScan for incremental as well
 func MapScan(rows *sql.Rows, dest map[string]any, converter func(value interface{}, columnType string) (interface{}, error)) error {
 	columns, colTypes, err := getColumnMetadata(rows)
 	if err != nil {


destination/iceberg/iceberg.go
```diff
@@ -38,6 +39,23 @@ type PartitionInfo struct {
...
+	"string":      {"string": true},
+	"timestamptz": {"timestamptz": true},
+}
+
+// promotionTransitions tracks when type promotion is required
+var promotionTransitions = map[string]map[string]bool{
+	"int":   {"long": true},
+	"float": {"double": true},
+}
+
```
vikaxsh, line 55: how is is beneficial? both the old implementation and the new lookup-table implementation are technically O(1).

destination/iceberg/iceberg.go
```diff
@@ -336,149 +354,171 @@ func (i *Iceberg) Type() string {
...
+	now := time.Now().UTC()
+	diffThreadSchema := atomic.Bool{}
+
+	var mu sync.Mutex
+	newColumns := make(map[string]string) // columns not in threadSchema
+	promotions := make(map[string]string) // columns requiring promotion
+
+	err := utils.Concurrent(ctx, records, runtime.GOMAXPROCS(0)*16, func(_ context.Context, record types.RawRecord, idx int) error {
+		records[idx].Data[constants.OlakeID] = record.OlakeID
+		records[idx].Data[constants.OlakeTimestamp] = now
```
vikaxsh: why are we using same timestamp for all the records?

destination/iceberg/iceberg.go
```diff
@@ -336,149 +354,171 @@ func (i *Iceberg) Type() string {
...
-					valid := validIcebergType(finalSchema[key], detectedIcebergType)
-					if !valid {
-						return false, fmt.Errorf(
-							"failed to validate schema for field[%s] (detected two different types in batch), expected type: %s, detected type: %s",
-							key, finalSchema[key], detectedIcebergType,
-						)
-					}
+		flattener := typeutils.NewFlattener()
+		flattenedRecord, err := flattener.Flatten(record.Data)
+		typeutils.ReleaseFlattener(flattener)
```
vikaxsh: don't we need to call this for parquet writer as well?

pkg/jdbc/reader.go
```diff
@@ -97,3 +98,88 @@ func MapScan(rows *sql.Rows, dest map[string]any, converter func(value interface
...
+	ctx := setter.ctx
+	valuesCh := make(chan []any)
+	doneCh := make(chan error, 1)
+
+	var (
+		columns  []string
+		colTypes []*sql.ColumnType
+	)
+
+	go func() {
```
vikaxsh: could we use ConcurrentC or should we instead define a dedicated concurrency utility in utils/concurrent.go?

pkg/jdbc/reader.go
```diff
@@ -97,3 +98,88 @@ func MapScan(rows *sql.Rows, dest map[string]any, converter func(value interface
...
+				precision, scale, hasPrecisionScale := colTypes[i].DecimalSize()
+				if datatype == "NUMBER" && hasPrecisionScale && scale == 0 {
+					datatype = utils.Ternary(precision > 9, "int64", "int32").(string)
+				}
+				conv, err := converter(rawData, datatype)
+				if err != nil && err != typeutils.ErrNullValue {
+					doneCh <- fmt.Errorf("failed to convert value for column %s: %w", col, err)
+					return
+				}
+				record[col] = conv
```
vikaxsh, line 183: Could we extract the shared common code between mapScan and here

destination/iceberg/iceberg.go
vaibhav-datazip: earlier we were using `promotionTransitions` function to do this , which first converted to string and then matched with switch case which was slow and also took extra memory every time , current implementation is better than that

destination/iceberg/iceberg.go
vaibhav-datazip: this is olake timestamp of when the record is getting processed. every time fetching the time stamp will take extra time, this is efficient and wont have much affect.

destination/iceberg/iceberg.go
vaibhav-datazip: old way will work as well , but I will do the change there as this is better .

pkg/jdbc/reader.go
vaibhav-datazip: concurrentc wont be sequential , that's needed here , so we should stick to this only. also its only used here so I don't think we need to define this pattern in concurrent.go

pkg/jdbc/reader.go
vaibhav-datazip: done

destination/iceberg/iceberg.go
vikaxsh: how it is beneficial in terms of memory? you are keeping a pre computed map in memory

vaibhav-datazip: yes, creating a map single time and using it whenever needed . isn't that better than creating a new string every time to check switch case ?

destination/iceberg/iceberg.go
```diff
@@ -231,149 +249,171 @@ func (i *Iceberg) Type() string {
...
-					diffThreadSchema.Store(true)
+
+			detectedIcebergType := detectedType.ToIceberg()
+
+			if existingType, existInSchema := i.schema[key]; existInSchema {
+				// Check if promotion is required using lookup table
+				if isPromotionRequired(existingType, detectedIcebergType) {
+					localPromotions[key] = detectedIcebergType
+					hasChanges = true
+				} else if !isValidTransition(existingType, detectedIcebergType) {
```
vikaxsh: why are you calling isValidTransition again here?

destination/iceberg/iceberg.go
```diff
@@ -231,149 +249,171 @@ func (i *Iceberg) Type() string {
...
+			diffThreadSchema.Store(true)
+			mu.Lock()
+			for k, v := range localNewCols {
+				if existing, ok := newColumns[k]; ok {
+					newColumns[k] = getCommonAncestorType(existing, v)
+				} else {
+					newColumns[k] = v
 				}
 			}
+			for k, v := range localPromotions {
```
vikaxsh: here last writer would win
for example
record 1: col = int
record 2: col = double

int ->  long is allowed

int ->  double is allowed

long -> double is allowed

destination/iceberg/iceberg.go
```diff
@@ -36,6 +37,23 @@ type Iceberg struct {
 	// It defines when to complete the Java writer and when schema evolution is required.
 }
 
+// Pre-computed lookup tables for type validation
+var validTypeTransitions = map[string]map[string]bool{
+	"int":         {"int": true, "long": true},
+	"long":        {"long": true, "int": true},
```
vikaxsh, line 42: are we allowing long to int transition?

utils/typeutils/datatype.go
```diff
@@ -10,13 +10,90 @@ import (
...
+			return types.Null
+		}
+		return detectTimestampPrecision(*val)
+	}
+
+	return typeFromValueReflect(v)
+}
+
+// typeFromValueReflect handles types that require reflection
+func typeFromValueReflect(v interface{}) types.DataType {
```
vikaxsh, line 96: It’d be good to confirm how much performance gain we actually get from this, if it’s not significant, we can revert to the simpler reflect-based version. otherwise every time we have to make changes in both the function for any datatype changes

utils/typeutils/flatten.go
```diff
@@ -19,55 +18,83 @@ type FlattenerImpl struct {
...
-				destination[key] = tm
-			} else {
-				destination[key] = fmt.Sprint(value)
-			}
+		// Fallback for other slice/map types
+		b, err := json.Marshal(v)
+		if err != nil {
+			return err
 		}
+		destination[reformattedKey] = string(b)
```
vikaxsh: can we merge this code

pkg/jdbc/reader.go
```diff
@@ -97,3 +117,77 @@ func MapScan(rows *sql.Rows, dest map[string]any, converter func(value interface
...
+				rawData := vals[i]
+				conv, err := normalizeDataTypeAndConvert(rawData, colTypes[i], converter)
+				if err != nil {
+					doneCh <- fmt.Errorf("failed to convert value for column %s: %w", col, err)
+					return
+				}
+				record[col] = conv
+			}
+			if err := OnMessage(ctx, record); err != nil {
+				doneCh <- err
```
vikaxsh: can we send error to channel in a defer function , from this we don't have to write logic of sending values into channel multiple times

utils/typeutils/flatten.go
```diff
@@ -19,55 +18,83 @@ type FlattenerImpl struct {
...
+
+func ReleaseFlattener(f Flattener) {
+	if impl, ok := f.(*FlattenerImpl); ok {
+		flattenerPool.Put(impl)
+	}
+}
+
+func getReformattedKey(key string) string {
+	if cached, ok := keyCache.Load(key); ok {
+		return cached.(string)
```
vaibhav-datazip, line 32: although this might not generally create problem, but what if the key exceeds a huge number, in that case memory overflow might occur. Wouldn't it be better to use a LRU cache here ?

destination/iceberg/iceberg.go
vaibhav-datazip: this is the only place we are calling it right now?, if its a not a valid transition we will throw error.

destination/iceberg/iceberg.go
vaibhav-datazip: no but promotions will only note actual promotion right ?

i.e int->long and float -> double 

so promotion will look like

promote column id to long 
or promote column marks to double 

that shouldn't create a problem

utils/typeutils/datatype.go
vaibhav-datazip: if the parsing fails in normal type from value we are falling back to reflect only, that shouldn't create a problem right ?

destination/iceberg/iceberg.go
vaibhav-datazip: it was there in the code earlier as well, so kept it. can be discussed

```
switch fmt.Sprintf("%s->%s", oldType, newType) {
	case "int->long", "float->double", "long->int", "double->float":
		return true
```

utils/typeutils/datatype.go
vikaxsh: why to keep fallback to reflect if there is no significant performance improvement

destination/iceberg/iceberg.go
vikaxsh: for example a column contains int and long values what would be evolved datatpye? long but here it could overwrite the last detected datatype, could be inconsistency

vaibhav-datazip: no it will be saved as long only right ? because int will get promoted to long and array will contain only long

hash-data:
couple of questions
-> how much improvement we got in each step : like via reflect change, using flatten with sync pool and map scanner?
-> how much memory usage we have compare to prev one?

vaibhav-datazip:
> couple of questions -> how much improvement we got in each step : like via reflect change, using flatten with sync pool and map scanner? -> how much memory usage we have compare to prev one?

| condition | speed after 400 sec in lakhs rps | memory usage in mb |
|---------| -----| ------ |
| current code | 4.13 | 5379 |
| without reflect change  | 4.11 | 5326 |
| without flatten.go changes | 3.1 | 5112 |
| without the flatten with sync pool | 4.12 | 5278 |

destination/parquet/parquet.go
```diff
@@ -330,7 +330,9 @@ func (p *Parquet) FlattenAndCleanData(ctx context.Context, records []types.RawRe
 			records[idx].Data[constants.CdcTimestamp] = *record.CdcTimestamp
 		}
 
-		flattenedRecord, err := typeutils.NewFlattener().Flatten(record.Data)
+		flattener := typeutils.NewFlattener()
+		flattenedRecord, err := flattener.Flatten(record.Data)
+		typeutils.ReleaseFlattener(flattener)
```
hash-data: for now let us remove sync pool thing from flattener, as the number of object creation is = total record in chunk/ records in 1 batch. 
We can think of flattener object creation when we setup the writer.

destination/iceberg/iceberg.go
```diff
@@ -231,149 +249,171 @@ func (i *Iceberg) Type() string {
...
-			records[idx].Data[constants.OlakeID] = record.OlakeID
-			records[idx].Data[constants.OlakeTimestamp] = time.Now().UTC()
-			records[idx].Data[constants.OpType] = record.OperationType
-			if record.CdcTimestamp != nil {
-				records[idx].Data[constants.CdcTimestamp] = *record.CdcTimestamp
-			}
+		localNewCols := make(map[string]string)
+		localPromotions := make(map[string]string)
+		hasChanges := false
+
```
hash-data: let us keep the older logic, not seems it is going to give much improvement

drivers/mysql/internal/backfill.go
```diff
@@ -54,15 +54,7 @@ func (m *MySQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
...
-		// Capture and process rows
-		return setter.Capture(func(rows *sql.Rows) error {
-			record := make(types.Record)
-			err := jdbc.MapScan(rows, record, m.dataTypeConverter)
-			if err != nil {
-				return fmt.Errorf("failed to scan record data as map: %s", err)
-			}
-			return OnMessage(ctx, record)
-		})
+		return jdbc.MapScanConcurrent(setter, m.dataTypeConverter, OnMessage)
```
hash-data, line 57:
```suggestion
		return jdbc.MapScan(setter, m.dataTypeConverter, OnMessage)
```

utils/typeutils/datatype.go
```diff
@@ -10,13 +10,90 @@ import (
...
+		if val == nil {
+			return types.Null
+		}
+		return types.Float32
+	case *float64:
+		if val == nil {
+			return types.Null
+		}
+		return types.Float64
+	case *string:
```
hash-data, line 76: add timestampl thing in it as well

utils/typeutils/datatype.go
```diff
@@ -10,13 +10,90 @@ import (
...
+	return typeFromValueReflect(v)
+}
+
+// typeFromValueReflect handles types that require reflection
+func typeFromValueReflect(v interface{}) types.DataType {
 	valType := reflect.TypeOf(v)
+	if valType == nil {
+		return types.Null
+	}
+	// Handle pointers
```
hash-data, line 101: is it required?

utils/typeutils/datatype.go
```diff
@@ -10,13 +10,90 @@ import (
...
+		return types.Array
+	case map[string]any:
+		return types.Object
+	case *bool:
+		if val == nil {
+			return types.Null
+		}
+		return types.Bool
+	case *int:
+		if val == nil {
```
hash-data, line 52: is this nil check required?

utils/typeutils/flatten.go
```diff
@@ -19,55 +18,69 @@ type FlattenerImpl struct {
...
 		}
-		destination[key] = string(b)
-	case reflect.Bool, reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
-		reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64,
-		reflect.Float32, reflect.Float64, reflect.String:
-		destination[key] = value
+		return nil
+	}
+
+	reformattedKey := getReformattedKey(key)
```
hash-data, line 59: call it at line 62 !!

utils/utils.go
```diff
@@ -376,26 +376,19 @@ func NormalizedEqual(strune1, strune2 string) bool {
...
+
+	for _, r := range key {
+		if (r >= 'a' && r <= 'z') || (r >= '0' && r <= '9') {
+			result = append(result, r)
+		} else if r >= 'A' && r <= 'Z' {
+			result = append(result, r+32)
 		} else {
-			result.WriteRune('_')
+			result = append(result, '_')
 		}
```
hash-data: let us do profiling of this as well

utils/typeutils/flatten_test.go
```diff
@@ -475,7 +475,7 @@ func TestFlattenInternal(t *testing.T) {
 				Age  int
 			}{"John", 30},
 			expected: types.Record{
-				"struct": "{John 30}",
+				"struct": `{"Name":"John","Age":30}`,
```
hash-data, line 478: any reason for this change?

pkg/jdbc/reader.go
hash-data: yes we should use similar kind of function and make this function easy to read.

destination/parquet/parquet.go
vaibhav-datazip: but we are calling new flattener for each record in FlattenAndCleanData function instead of creating each time isn't using the free object from pool better, this also helps in avoiding gc pressure.

destination/iceberg/iceberg.go
vaibhav-datazip: replaced this with old schema evolution logic in commit [839ee0e](https://github.com/datazip-inc/olake/pull/657/commits/839ee0e6b8a42deec90e3f4a117cb03e209375ef) . here are the stats 

I have also added one more commit [0465f77](https://github.com/datazip-inc/olake/pull/657/commits/0465f774f7324f95b92b6f02d81078bcb140b728) which keeps the new changes but are much simplified 


| condition | speed after 400 sec in lakhs rps | memory usage in mb |
|---------| -----| ------ |
| with old schema evolution logic | 3.45 | 5288 |
| with new change  | 4.07 | 5526 |
| with changes simplified | 4.1 | 5385 |

drivers/mysql/internal/backfill.go
vaibhav-datazip: MapScan is still used in incremental syncs

utils/typeutils/flatten_test.go
vaibhav-datazip: now the structure is getting synced like this because of change in reformat related function

hash-data: this is wrong

destination/parquet/parquet.go
hash-data: let us do profiling for it as well

vaibhav-datazip: removed the optimized flattener pool and switched back to simple flattener in this commit [4588be6](https://github.com/datazip-inc/olake/pull/657/commits/4588be64e8feb8974aff975be16a69dc87202af3) , below are the profiling results

| condition | speed after 900 sec in lakhs rps | memory usage in mb |
|---------| -----| ------ |
| current code (with flattener pool)| 4.21 | 5489 |
| without flattener pool (original code without optimizations)  | 4.16 | 5333 |

utils/utils.go
vaibhav-datazip: haven't pushed commit for this change,as it only need to revert utils.go in its original state which was present before this PR

| condition | speed after 400 sec in lakhs rps | memory usage in mb |
|---------| -----| ------ |
| optimized changes (no write rune or write byte used) | 4.13 | 5346 |
| original code (uses write byte and write rune)  | 4.11 | 5415 |

utils/typeutils/datatype.go
vaibhav-datazip: yeah just in case reflect is able to parse some value which was not able to be done in TypeFromValue

utils/typeutils/datatype.go
vaibhav-datazip: pointer values might be nil, have removed repetition

utils/typeutils/flatten.go
vaibhav-datazip: but in that case if value ==nil and f.omitNilValues returns true , we are calling that function for no reason

utils/typeutils/datatype.go
vaibhav-datazip: although the code is looking repetitive still abstracting out is causing one problem , i.e. its giving typed nil thing

Assigning a nil pointer (or other nil-capable type) to an interface creates an interface with a non-nil dynamic type but a nil value. This can lead to unexpected results when comparing the interface to nil.

if we abstract out this into function there will again be need to call reflect for pointer types:

```
typeFromPtr := func(ptr interface{}, getType func() types.DataType) types.DataType {
		if ptr == nil {
			return types.Null
		}
		// Check for typed nil pointers using reflection
		val := reflect.ValueOf(ptr)
		if val.Kind() == reflect.Pointer && val.IsNil() {
			return types.Null
		}
		return getType()
	}
```

utils/typeutils/flatten_test.go
vaibhav-datazip: earlier, in `flatten.go` in internal flatten function we were doing 

```
if tm, ok := value.(time.Time); ok {
				destination[key] = tm
			} else {
				destination[key] = fmt.Sprint(value)
			}
```


now its 

```
		// Fallback for other types
		b, err := json.Marshal(v)
		if err != nil {
			return err
		}
		destination[reformattedKey] = string(b)
```

vaibhav-datazip:
pushed changes after removal of flattener pool and reformat function changes

| condition | speed after 400 sec in lakhs rps | memory usage in mb |
|---------| -----| ------ |
| removed pool and flattener | 4.10 | 5388 |

vaibhav-datazip:
| condition | speed after 400 sec in lakhs rps | memory usage in mb |
|---------| -----| ------ |
|  with changes till now | 4.11 | 5356 |
| with common flattener for all records (tested for iceberg) |3.7 |5576 |

```
flattener := typeutils.NewFlattener()

	err := utils.Concurrent(ctx, records, runtime.GOMAXPROCS(0)*16, func(_ context.Context, record types.RawRecord, idx int) error {
		records[idx].Data[constants.OlakeID] = record.OlakeID
		records[idx].Data[constants.OlakeTimestamp] = now
		records[idx].Data[constants.OpType] = record.OperationType
		if record.CdcTimestamp != nil {
			records[idx].Data[constants.CdcTimestamp] = *record.CdcTimestamp
		}

		flattenedRecord, err := flattener.Flatten(record.Data)
```

destination/iceberg/iceberg.go
```diff
@@ -335,7 +352,7 @@ func (i *Iceberg) FlattenAndCleanData(ctx context.Context, records []types.RawRe
 
 			// if schema difference is not detected, detect schema difference
 			if !diffThreadSchema.Load() {
-				if changeDetected, err := detectOrUpdateSchema(records[idx], true, i.schema, copySchema(i.schema)); err != nil {
+				if changeDetected, err := detectOrUpdateSchema(records[idx], true, i.schema, i.schema); err != nil {
```
hash-data, line 356: can you add a comment on why we are not copying here so that no one get confused or also we can send empty map instead of i.stream

drivers/mysql/internal/backfill.go
hash-data: add a todo for it

vaibhav-datazip:
- Tested by syncing fhv_trips and trips2 from oracle to iceberg , prefetch rows set as 10k, 40 threads.

```
{
  "memory": {
    "min": 112,
    "max": 6324,
    "mean": 4605.117840684657,
    "count": 1519
  },
  "speed": {
    "min": 0,
    "max": 529151.13,
    "mean": 498147.951869651,
    "count": 1519
  },
  "system_memory_used_gb": {
    "min": 19.261314392089844,
    "max": 93.16106796264648,
    "mean": 73.18980524993555,
    "count": 1519
  },
  "cpu_usage_percent": {
    "min": 0.24679016596459222,
    "max": 82.86927048696563,
    "mean": 74.82028569678994,
    "count": 1519
  }
}
```

stats.json

```
{
    "Estimated Remaining Time": "0.00 s",
    "Memory": "123 mb",
    "Seconds Elapsed": "7616.00",
    "Speed": "526337.56 rps",
    "Synced Records": 4008587913,
    "Writer Threads": 0
}
```

hash-data (approved):

PR #677, opened 2025-12-11
fix: MariaDB CDC failing due to query error

### LINKED ISSUE #675: MariaDB Source Support
### Problem

When trying to create a CDC task with a MariaDB RDS instance the version returned from Maria does not correlate to the appropriate MySQL version and the check within pkg/jdbc/jdbc.go returns the wrong command to determine whether CDC can be enabled on the task.

`query := utils.Ternary(majorVersion > 8 || (majorVersion == 8 && minorVersion >= 4), jdbc.MySQLMasterStatusQueryNew(), jdbc.MySQLMasterStatusQuery()).(string)`

In our example we are trying to connect to MariaDB 10.11.13 which returns the func `MySQLMasterStatusQueryNew` and not the supported command in func `MySQLMasterStatusQuery`. 

The error surfaced associated with this is `Query 1 ERROR at Line 1: : You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'LOG STATUS' at line 1`.

### Solution

Due to the compatibility between MariaDB and MySQL especially in the CDC space I wouldn't recommend cloning the driver and changing the version logic, I think the overhead there would be more than it's worth. 

The suggestion I had was to evaluate the Engine string within the pkg/jdbc/jdbc.go `MySQLVersion` function and resolve to the associated MariaDB version to MySQL for example in our instance:

MariaDB: `SELECT @@version` = 10.11.13-MariaDB-log

Within the `MySQLVersion` check for the string instance of `MariaDB` and if so present resolve to a MySQL version string. Ideally `10.11.13-MariaDB-log` would resolve to the `MySQLMasterStatusQuery` function which I think means means majorVersion < 8.4 (Not a dev or a go coder so trying to work through it :)).

### Impact

Dev roll out is ready, our one MySQL server is chipping along all the other servers are MariaDB, if it's possible to hook in a comparison change with low effort that would completely unblock us and allow us to keep going on testing the roll out in the development environment.

Thanks for the consideration :)

vaibhav-datazip:
# Description

This PR adds support for MariaDB as a CDC source by fixing the binlog position query selection and result scanning logic.

**Problem:**
When connecting to a MariaDB instance (e.g., MariaDB 10.11.13), the code was incorrectly:
1. Treating MariaDB version numbers as MySQL versions (MariaDB 10.11 was parsed as MySQL 10.11)
2. Using the wrong query: `SHOW BINARY LOG STATUS` (MySQL 8.4+ syntax) instead of `SHOW MASTER STATUS` (which MariaDB requires)
3. Attempting to scan 5 columns when MariaDB only returns 4 columns from `SHOW MASTER STATUS`

This caused two errors:
- SQL syntax error: `You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'LOG STATUS'`
- Scanning error when trying to read 5 columns from a 4-column result set

**Solution:**
1. Enhanced `MySQLVersion()` function to detect and return the database flavor (MySQL vs MariaDB) by checking if the version string contains "MARIADB"
2. Updated query selection logic to ensure MariaDB always uses `SHOW MASTER STATUS` (regardless of version number)
3. Added flavor-based scanning logic to handle different column counts:
   - MySQL: 5 columns (File, Position, Binlog_Do_DB, Binlog_Ignore_DB, Executed_Gtid_Set)
   - MariaDB: 4 columns (File, Position, Binlog_Do_DB, Binlog_Ignore_DB)

**Changes:**
- `pkg/jdbc/jdbc.go`: Modified `MySQLVersion()` to return database flavor along with version numbers
- `pkg/binlog/binlog.go`: Updated `GetCurrentBinlogPosition()` to use flavor-aware query selection and column scanning

Fixes #675

## Type of change

# How Has This Been Tested?

- [x] Tested with MariaDB 10.11.13 (Docker)
  - Verified `SHOW MASTER STATUS` query executes successfully
  - Verified binlog position is correctly retrieved and scanned
  - Confirmed CDC task creation works without errors
- [x] Tested with MySQL (backward compatibility)
  - Verified existing MySQL functionality remains unchanged
  - Confirmed query selection logic works for MySQL 8.3 and below

## Documentation


pkg/jdbc/jdbc.go
```diff
@@ -402,28 +402,33 @@ func MySQLTableColumnsQuery() string {
 
 // MySQLVersion returns the version of the MySQL server
 // It returns the major and minor version of the MySQL server
```
vikaxsh:
```suggestion
// It returns the major and minor version of the MySQL server ans also database flavor ("MySQL" or "MariaDB")
```

vikaxsh:
please add docs link as well in pr description

vaibhav-datazip:
> please add docs link as well in pr description

will add once @nayanj98  raises a pr on olake documentation repo

PR #685, opened 2025-12-16
feat: dynamic rps benchmarking

### LINKED ISSUE #654: improvement: improve performance test by comparing RPS against average of last 5 runs instead of static benchmarks
## Description
Our current performance test uses hardcoded benchmark RPS values:
```go
// utils/testutils/test_utils.go
var benchmarks = map[constants.DriverType]benchmarkStats{
    constants.MySQL:    {Backfill: 15906.40, CDC: 15648.93},
    constants.Postgres: {Backfill: 12000, CDC: 1500},
    constants.Oracle:   {Backfill: 3500, CDC: 0},
    constants.MongoDB:  {Backfill: 0, CDC: 0},
}
```

During test execution, we check whether the current run’s RPS (from stats.json) is at least 90% of the static benchmark.
This approach is rigid and doesn’t adapt as performance evolves or hardware changes.

---

## Goal
Replace the static benchmark comparison with a dynamic benchmark calculated using the average RPS of the last 5 successful runs.

---
## Idea

1. Store the history of the last 5 RPS values for each driver + test type (Backfill & CDC).
2. On each new performance test run:
    - Load the last 5 historical RPS values for the given driver/test-type.  
    - Compute their average.
    - Compare the current RPS against 90% of that average.
3. Ensure that the test:
    - Falls back gracefully for the first few runs (e.g., <5 history entries).
    - Updates the history by adding the current run and removing the oldest entry.

---

## Checkbox

- [ ] Past 5 runs are persisted and readable during subsequent runs.
- [ ] Performance test compares current RPS with 90% of the average of those 5 runs.
- [ ] Logic supports both Backfill and CDC independently.
- [ ] No static hardcoded benchmark values are used anymore.

manish-singh-bisht:
# Description

Fixes #654 

## Type of change

# How Has This Been Tested?

1) tested locally for the functionality, added lot of rows in the db to get the rps as in images below.

2) also have tested this in the github actions in my own fork, https://github.com/manish-singh-bisht/olake/actions/runs/20350650406/job/58474101019 , only for postgres

failed rps

rps data

rps success

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/utils/testutils/test_utils.go
+++ b/utils/testutils/test_utils.go
@@ -80,7 +80,7 @@ type history struct {
 	UpdatedAt time.Time `json:"updated_at"`
 }
 
-// benchmarkStore stores the benchmark RPS values for backfill and CDC modes.
+// benchmarkStore stores the benchmark RPS history for backfill and CDC modes.
 type benchmarkStore struct {
 	Backfill history `json:"backfill"`
 	CDC      history `json:"cdc"`


manish-singh-bisht:
@vishalm0509 ready for review.

vishalm0509:
**Self-note / future idea to explore:**

>Consider changing the artifact download logic so that we don’t only read from the “last successful” workflow run. Instead, always download the most recent run’s artifact for each driver (regardless of success/failure), and then only update/write the benchmarks.json when that driver’s performance test actually passes.
 **Implications:**
  • If MySQL fails but Postgres succeeds in a mixed run, Postgres would update its baseline while MySQL would keep its previous one.
  • This gives us per-driver history continuity even when the workflow overall is partially failing.
  • Worth thinking through edge cases and implementation complexity before proceeding.
  
 Just noting it here, let's proceed after internal discussions

utils/utils.go
```diff
@@ -75,6 +75,17 @@ func Ternary(cond bool, a, b any) any {
...
+	if len(values) == 0 {
+		return T(0)
+	}
+	sum := T(0)
+	for _, v := range values {
+		sum += v
+	}
+	return sum / T(len(values))
+}
+
```
vishalm0509: - This function uses integer division when T is an integer type, which drops the fractional part. That means it will truncate averages (average of 1 and 2 becomes 1 instead of 1.5), which may not be what we want for an average.

- Adding all the numbers using the same type T can overflow when the values are large or there are many of them. It’s safer to add them up in a bigger type first before dividing. Let's say Float64. Think about it

utils/testutils/benchmarks.go
```diff
@@ -0,0 +1,103 @@
...
+		store.Backfill.RPS,
+		store.CDC.RPS,
+	).([]float64)
+
+	rpsValues = append(rpsValues, rps)
+
+	// Truncate history to maintain a rolling window of the last maxRPSHistorySize values.
+	if len(rpsValues) > maxRPSHistorySize {
+		rpsValues = rpsValues[len(rpsValues)-maxRPSHistorySize:]
+	}
```
vishalm0509: It'll always be remove the first element right. What do you think 

```go
rpsValues = rpsValues[1:]
```

utils/testutils/benchmarks.go
```diff
@@ -0,0 +1,103 @@
...
+	UpdatedAt time.Time `json:"updated_at"`
+}
+
+type BenchmarkStore struct {
+	Backfill history `json:"backfill"`
+	CDC      history `json:"cdc"`
+	FilePath string  `json:"-"`
+}
+
+func LoadBenchmarks(path string) (*BenchmarkStore, error) {
```
vishalm0509: Why capitalized `LoadBenchmarks`, it's used within the same package right

utils/testutils/benchmarks.go
```diff
@@ -0,0 +1,103 @@
...
+		},
+		FilePath: path,
+	}
+	if err := store.load(); err != nil {
+		return nil, err
+	}
+	return store, nil
+}
+
+func (store *BenchmarkStore) load() error {
```
vishalm0509: Go best practice

```go
func (bs *BenchmarkStore) load() error {
```

or 

```go
func (s *BenchmarkStore) load() error {
```

utils/testutils/test_utils.go
```diff
@@ -1058,26 +1047,45 @@ func (cfg *PerformanceTest) TestPerformance(t *testing.T) {
 	ctx := context.Background()
 
 	// checks if the current rps (from stats.json) is at least 90% of the benchmark rps
-	isRPSAboveBenchmark := func(config TestConfig, isBackfill bool) (bool, error) {
+	isRPSAboveBenchmark := func(config TestConfig, isBackfill bool) (bool, float64, error) {
```
vishalm0509: Since this function not just returns `bool` but also `float64`, can you think of a better name for it

utils/utils.go
manish-singh-bisht: fixed,thanks

utils/testutils/benchmarks.go
manish-singh-bisht: replaced, i was just overly defensive for cases where length is greater than max length+1 (which will never happen)

utils/testutils/benchmarks.go
manish-singh-bisht: fixed,

utils/testutils/benchmarks.go
manish-singh-bisht: changed to `s`, thanks

utils/testutils/test_utils.go
manish-singh-bisht: fixed, now `checkRPSBenchmark`

utils/utils.go
```diff
@@ -75,15 +75,15 @@ func Ternary(cond bool, a, b any) any {
...
 }
 
-func Average[T int | int8 | int16 | int32 | int64 | float32 | float64](values []T) T {
+func Average[T int | int8 | int16 | int32 | int64 | float32 | float64](values []T) float64 {
 	if len(values) == 0 {
-		return T(0)
+		return 0.0
 	}
-	sum := T(0)
+	sum := float64(0)
```
vishalm0509:
```go
var sum float64
```

utils/testutils/benchmarks.go
```diff
@@ -0,0 +1,103 @@
...
+const (
+	maxRPSHistorySize = 5
+)
+
+type history struct {
+	RPS       []float64 `json:"rps"`
+	UpdatedAt time.Time `json:"updated_at"`
+}
+
+type BenchmarkStore struct {
```
vishalm0509: Even this should not be capitalized as it is not exported and used only within package.

utils/utils.go
manish-singh-bisht: fixed

utils/testutils/benchmarks.go
manish-singh-bisht: fixed

utils/testutils/test_utils.go
```diff
@@ -1058,26 +1047,45 @@ func (cfg *PerformanceTest) TestPerformance(t *testing.T) {
...
+		averageRPS, observations := benchmarks.stats(isBackfill)
+
+		// No benchmarks exist yet for this driver/mode
+		// Skip validation to allow initial benchmarking.
+		if observations == 0 {
+			t.Logf("No benchmarks exist yet for %s %s mode, skipping validation", config.Driver, utils.Ternary(isBackfill, "backfill", "cdc").(string))
+			return true, rps, nil
+		}
+
+		t.Logf("CurrentRPS: %.2f, averageRPS: %.2f, observations: %d", rps, averageRPS, observations)
```
vishalm0509:
```go
t.Logf("currentRPS: %.2f, averageRPS: %.2f, observations: %d", rps, averageRPS, observations)
```

utils/testutils/test_utils.go
```diff
@@ -1058,26 +1047,45 @@ func (cfg *PerformanceTest) TestPerformance(t *testing.T) {
 	ctx := context.Background()
 
 	// checks if the current rps (from stats.json) is at least 90% of the benchmark rps
-	isRPSAboveBenchmark := func(config TestConfig, isBackfill bool) (bool, error) {
+	checkRPSBenchmark := func(config TestConfig, isBackfill bool) (bool, float64, error) {
```
vishalm0509: I think `checkBenchmarkRPS` is more understandable. What do you think, any better name ?

utils/testutils/test_utils.go
manish-singh-bisht: fixed

utils/testutils/test_utils.go
manish-singh-bisht: fixed

vishalm0509:
@manish-singh-bisht can we move all the methods from `benchmark.go` to `test_utils.go` itself ?

manish-singh-bisht:
> @manish-singh-bisht can we move all the methods from `benchmark.go` to `test_utils.go` itself ?

shifted

vishalm0509:
@manish-singh-bisht just wanted to confirm if you have verified the working of `github-artifacts` for more than 5 runs.

manish-singh-bisht:
> @manish-singh-bisht just wanted to confirm if you have verified the working of `github-artifacts` for more than 5 runs.

 https://github.com/manish-singh-bisht/olake/actions/runs/20350650406/job/58474101019 , only for postgres

edit: i only checked whether downloading of artifacts from previous run is working for subsequent runs, the working for average 5 runs did that locally

vishalm0509:
> > @manish-singh-bisht just wanted to confirm if you have verified the working of `github-artifacts` for more than 5 runs.
> 
> yes i did in my fork, https://github.com/manish-singh-bisht/olake/actions/runs/20350650406/job/58474101019 , only for postgres
> 
> edit: i only checked whether downloading of artifacts from previous run is working for subsequent runs, the working for average 5 runs did that locally

I checked locally everything's working fine. Good work 🙌. 

Just wanted to confirm the same is the case in the CI as well, at least reading the correct updated artifact for subsequent run.

manish-singh-bisht:
latest
the data of last uploaded artifact, CI

manish-singh-bisht:
yes it reads correct data from last runs

vishalm0509:
@manish-singh-bisht it will be helpful for others if you add some comments for the new types and methods you have created around it on what it does and that they are used by the performance test particularly. 

Just a simple, brief comment for each method and the types would work. Then let's merge it

manish-singh-bisht:
> @manish-singh-bisht it will be helpful for others if you add some comments for the new types and methods you have created around it on what it does and that they are used by the performance test particularly.
> 
> Just a simple, brief comment for each method and the types would work. Then let's merge it

done

vishalm0509 (approved):
LGTM

PR #694, opened 2025-12-22
feat: DB2 LUW (Linux, Unix, Windows) as source connector

ImDoubD-datazip:
# Description

DB2 LUW as source connector is added. As of now it supports 2 sync modes:
- Full Refresh
- Incremental

Chunking is either done via primary keys (if primary keys are present). Else RID based chunking is done (for tables with no primary keys).

## Pre-Requisite
To run DB2 LUW, one needs IBM data server ODBC and CLI driver to be installed in the machine.

Steps to run DB2 LUW in your machine:
- Download the IBM data server ODBC and CLI driver as per OS. 
   - [Linux (x86-64)](https://public.dhe.ibm.com/ibmdl/export/pub/software/data/db2/drivers/odbc_cli/linuxx64_odbc_cli.tar.gz)
   - [MacOS (arm64)](https://public.dhe.ibm.com/ibmdl/export/pub/software/data/db2/drivers/odbc_cli/macarm64_odbc_cli.tar.gz)
   - [MacOS (64-bit intel/apple silicon)](https://public.dhe.ibm.com/ibmdl/export/pub/software/data/db2/drivers/odbc_cli/macos64_odbc_cli.tar.gz)
   - [Windows (64 bit)](https://public.dhe.ibm.com/ibmdl/export/pub/software/data/db2/drivers/odbc_cli/ntx64_odbc_cli.zip)

- Now extract the downloadable content from the above URLs, a folder named `clidriver` will be created in your directory.

- Now set the environment variables in pc terminal, necessary to run DB2 LUW.
```
# IBM DB2 CLI environment
export IBM_DB_HOME=/pathto/clidriver
export PATH=$IBM_DB_HOME/bin:$PATH
export CGO_CFLAGS="-I$IBM_DB_HOME/include"
export CGO_LDFLAGS="-L$IBM_DB_HOME/lib -Wl,-rpath,$IBM_DB_HOME/lib"
export DYLD_LIBRARY_PATH=$IBM_DB_HOME/lib
```

- Now try to run the discover command and sync.

## Type of change

# How Has This Been Tested?

This has been tested on DB2 LUW VM Instance. Full refresh and incremental were tested in this.

## Documentation

## Related PR's (If Any):

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -31,9 +31,9 @@ const (
 	// DestinationDatabasePrefix is used as prefix for destination database name
 	DestinationDatabasePrefix = "DESTINATION_DATABASE_PREFIX"
 	// EffectiveParquetSize is the effective size in bytes considering 256mb targeted parquet size, compression ratio as 8
-	EffectiveParquetSize   = int64(256) * 1024 * 1024 * int64(8)
-	DB2TimestampFormat     = "2006-01-02 15:04:05.000000"
-	DefaultTimestampFormat = "2006-01-02T15:04:05.000000000Z"
+	EffectiveParquetSize        = int64(256) * 1024 * 1024 * int64(8)
+	DB2StateTimestampFormat     = "2006-01-02 15:04:05.000000"
+	DefaultStateTimestampFormat = "2006-01-02T15:04:05.000000000Z"
 )
 
 type DriverType string
--- a/drivers/abstract/incremental.go
+++ b/drivers/abstract/incremental.go
@@ -164,11 +164,11 @@ func (a *AbstractDriver) getMaxIncrementCursorFromData(primaryCursor, secondaryC
 func (a *AbstractDriver) FormatCursorValue(cursorValue any) any {
 	switch v := cursorValue.(type) {
 	case time.Time:
-		// DB2 TIMESTAMP does NOT store timezone information. Applying v.UTC().Format() changes the actual time value for db2.
+		// db2 timestamp does NOT store timezone information. Applying v.UTC() changes the actual time value for db2.
 		if a.driver.Type() == string(constants.DB2) {
-			return v.Format(constants.DB2TimestampFormat)
+			return v.Format(constants.DB2StateTimestampFormat)
 		}
-		return v.UTC().Format(constants.DefaultTimestampFormat)
+		return v.UTC().Format(constants.DefaultStateTimestampFormat)
 	case primitive.ObjectID:
 		return v.Hex()
 	default:
--- a/drivers/db2/internal/db2.go
+++ b/drivers/db2/internal/db2.go
@@ -178,6 +178,15 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
 	if columnType == "TIME" {
 		return typeutils.ReformatTimeValue(value)
 	}
+
 	olakeType := typeutils.ExtractAndMapColumnType(columnType, db2TypeToDataTypes)
+
+	// in db2, string based types come in byte format
+	if olakeType == types.String {
+		if v, ok := value.([]byte); ok {
+			return strings.TrimSpace(string(v)), nil
+		}
+	}
+
 	return typeutils.ReformatValue(olakeType, value)
 }
--- a/utils/typeutils/reformat.go
+++ b/utils/typeutils/reformat.go
@@ -101,8 +101,7 @@ func ReformatValue(dataType types.DataType, v any) (any, error) {
 		case bool:
 			return fmt.Sprintf("%t", v), nil
 		case []byte: // byte slice
-			// in db2, char and varchar comes in byte format
-			return strings.TrimSpace(string(v)), nil
+			return string(v), nil
 		default:
 			return fmt.Sprintf("%v", v), nil
 		}


pkg/jdbc/jdbc.go
```diff
@@ -770,3 +770,205 @@ func ThresholdFilter(ctx context.Context, opts DriverOptions) (string, []any, er
...
+// DB2 Specific Queries
+
+// DB2DiscoveryQuery returns the query to discover tables in a DB2 database with filter for 'T' (Table) and 'V' (View).
+func DB2DiscoveryQuery() string {
+	return `
+		SELECT
+			TRIM(TABSCHEMA) AS table_schema,
+			TRIM(TABNAME) AS table_name
+		FROM SYSCAT.TABLES
+		WHERE TYPE IN ('T', 'V')
```
vishalm0509: Should we select `View` also ?

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,215 @@
...
+		return nil, fmt.Errorf("failed to get approx row count: %s", err)
+	}
+	if approxRowCount == -1 {
+		var hasRows bool
+		existsQuery := jdbc.DB2TableExistQuery(stream)
+		err := d.client.QueryRowContext(ctx, existsQuery).Scan(&hasRows)
+
+		if err != nil {
+			return nil, fmt.Errorf("failed to check if table has rows: %s", err)
+		}
```
vishalm0509, line 81:
```suggestion
		err := d.client.QueryRowContext(ctx, existsQuery).Scan(&hasRows)
		if err != nil {
			return nil, fmt.Errorf("failed to check if table has rows: %s", err)
		}
```

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,215 @@
...
+			chunks.Insert(types.Chunk{
+				Min: utils.ConvertToString(currentVal),
+				Max: nil,
+			})
+		}
+
+		return chunks, nil
+	}
+	// split chunks via physical identifier RID()
+	splitViaRID := func(ctx context.Context, stream types.StreamInterface) (*types.Set[types.Chunk], error) {
```
vishalm0509, line 174: is it safe to use RID for chunking ?

ImDoubD-datazip: for non-primary key, I think we should use it because in table with no primary keys there is less chance of any column to be indexed. so for those kind of tables, it is better to use RID.

pkg/jdbc/jdbc.go
ImDoubD-datazip: will ask product and do changes accordingly.

vishalm0509:
**Column type: `DBCLOB`**

- In the databae
Screenshot 2025-12-30 at 3 58 59 PM

- In destination (Glue)
Screenshot 2025-12-30 at 3 59 14 PM

vishalm0509:
**Column type: `varbinary`**

- `"varbinary": types.String`

- Also column `col_varbinary`

--- 

- In the database
    Screenshot 2025-12-30 at 4 12 10 PM

- In the destination
    Screenshot 2025-12-30 at 4 11 25 PM

vishalm0509:
- **Column name: `col_long_vargraphic`**
- Column type: `LONG VARGRAPHIC`
- Not mapped in OLake

---
- In the database
    Screenshot 2025-12-30 at 4 15 10 PM

- In the destination
Screenshot 2025-12-30 at 4 15 22 PM

vishalm0509:
**`col_timestamp`**

- Database: `2024-01-01-10.15.30.123456`
- Destination: `2024-01-01 10:15:30.123000 UTC`

vishalm0509:
**`col_time`** (also col_date)

- Datatype: `TIME`
- Databae: `10:15:30`
- Destination: `0001-01-01 10:15:30 +0000 UTC`

> "Same case with `MSSQL (SQL Server)` also"

vishalm0509:
**Column: `CHAR_ONE `**

- Datatype: `CHARACTER`
- Database: `CHAR_ONE `
- Destination: `"Q0hBUl9PTkUgIA=="`

vishalm0509:
- Table details
      Screenshot 2025-12-30 at 4 51 08 PM

- Database: `99999999999999.9999`
- Destination: `1.0E14`

--- 

I believe this is expected

```go
	// Can handle up to 15 significant digits accurately (e.g., DECIMAL(15,2) or DECIMAL(15,7))
	// Values with 16 digits may have minor rounding. Beyond 16 (from 17) digits will have precision loss.
	"numeric": types.Float64,
	"decimal": types.Float64,
```

But just letting you know, check once

vishalm0509:
**Incremental test**

Table: `DB2_ALL_DATATYPES` 

```json
"sync_mode": "incremental",
"cursor_field": "COL_TIMESTAMP:COL_TIME",
```

Screenshot 2025-12-30 at 4 56 59 PM

--- 

I checked for `col_bigInt`, it's working fine. We need to check for `timestamp` based cols similar to Oracle

Dockerfile
```diff
@@ -1,41 +1,81 @@
...
+FROM golang:1.24-bookworm AS builder
 
 WORKDIR /home/app
 COPY . .
 
 ARG DRIVER_NAME=olake
+
+# DB2 conditional setup
+RUN if [ "$DRIVER_NAME" = "db2" ]; then \
+  go run github.com/ibmdb/go_ibm_db/installer@latest; \
```
schitizsharma: We use `@latest` here but what if one day a new driver version is released which is not backward compatible? 
We should always keep static driver version I believe.

Dockerfile
```diff
@@ -1,41 +1,81 @@
...
+else \
+  mkdir -p /go/pkg/mod/github.com/ibmdb/clidriver; \
```
schitizsharma, line 14: You should maybe add a comment that this `mkdir` is an edge case handling for the `db2-stage` later in this Dockerfile.

schitizsharma: Because otherwise on first read it looks like "if driver is not db2 then create a directory specific to db2". Kinda confusing why we should be doing this.

Dockerfile
```diff
@@ -1,41 +1,81 @@
...
+# Copy DB2 CLI Driver
+COPY --from=builder /go/pkg/mod/github.com/ibmdb/clidriver /opt/clidriver
+
+# Set DB2 CLI environment variables
+ENV IBM_DB_HOME=/opt/clidriver
+ENV PATH=$IBM_DB_HOME/bin:$PATH
+ENV CGO_CFLAGS="-I$IBM_DB_HOME/include"
+ENV CGO_LDFLAGS="-L$IBM_DB_HOME/lib -Wl,-rpath,$IBM_DB_HOME/lib"
+ENV LD_LIBRARY_PATH=$IBM_DB_HOME/lib
+
```
schitizsharma, line 80: We set it already above in the builder image so we need to add the ENV variables back here. Right?

release-tool.sh
```diff
@@ -59,6 +59,7 @@ function release() {
     local version=$1
     local platform=$2
     local branch=${3:-master}
+    local target_stage=${4:-driver-stage}
```
schitizsharma, line 62: Maybe add a comment that in case of `db2` we only build till `db2-stage` and otherwise go till `driver-stage`

Dockerfile
ImDoubD-datazip: yes

Dockerfile
ImDoubD-datazip: yes

drivers/abstract/incremental.go
```diff
@@ -157,3 +159,18 @@ func (a *AbstractDriver) getMaxIncrementCursorFromData(primaryCursor, secondaryC
 	}
 	return primaryCursorValue, secondaryCursorValue
 }
+
+// FormatCursorValue is used to make time format and object id format consistent to be saved in state
+func (a *AbstractDriver) FormatCursorValue(cursorValue any) any {
+	switch v := cursorValue.(type) {
+	case time.Time:
+		if a.driver.Type() == string(constants.DB2) {
+			return v.Format("2006-01-02 15:04:05.000000")
```
vaibhav-datazip: isn't there a timestamp aware format for db2 ?

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	// if PK present then PK based chunking else RID based chunking
+	pkColumns := stream.GetStream().SourceDefinedPrimaryKey.Array()
+
+	var stmt string
+	if len(pkColumns) > 0 {
+		stmt = jdbc.DB2PKChunkScanQuery(stream, pkColumns, chunk, filter)
+	} else if stream.Self().StreamMetadata.ChunkColumn != "" {
+		stmt = jdbc.DB2PKChunkScanQuery(stream, []string{stream.Self().StreamMetadata.ChunkColumn}, chunk, filter)
```
vaibhav-datazip: shouldn't chunk column be given priority over primary key , as its user defined ?

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,217 @@
...
+
+			args := make([]interface{}, 0)
+			for columnIndex := 0; columnIndex < len(pkColumns); columnIndex++ {
+				// For each column combination in the WHERE clause, we need to add the necessary parts
+				for partIndex := 0; partIndex <= columnIndex && partIndex < len(columns); partIndex++ {
+					args = append(args, columns[partIndex])
+				}
+			}
+			var nextValRaw interface{}
+			err := d.client.QueryRowContext(ctx, query, args...).Scan(&nextValRaw)
```
vaibhav-datazip, line 150: if this is similar to what we have in mysql , can we abstract this out and then use the same at both places

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,217 @@
...
+		// pages size and number of pages
+		var pageSize, nPages int64
+		pageStatsQuery := jdbc.DB2PageStatsQuery(stream)
+		if err := d.client.QueryRowContext(ctx, pageStatsQuery).Scan(&pageSize, &nPages); err != nil {
+			return nil, fmt.Errorf("failed to get the page size and number of pages: %s", err)
+		}
+
+		// pages to be in a chunk
+		pagesPerChunk := int64(math.Ceil(float64(int64(constants.EffectiveParquetSize)) / float64(pageSize)))
+		nPages = utils.Ternary(nPages <= 0, int64(1), nPages).(int64)
```
vaibhav-datazip, line 189: can pages be less than 0 ?

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,217 @@
...
+
+		// number of chunks
+		targetChunkCount := int64(math.Ceil(float64(nPages) / float64(pagesPerChunk)))
```
vaibhav-datazip: this can be renamed numberOfChunks, and rid one can be named ridPerChunk

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,217 @@
...
+
+		totalRidRange := maxRID - minRID
+		// distance between start and end RID of a chunk
+		ridInterval := int64(math.Ceil(float64(totalRidRange) / float64(targetChunkCount)))
+		chunks := types.NewSet[types.Chunk]()
+		for start := minRID; start <= maxRID; start += ridInterval {
+			end := start + ridInterval
+			if end > maxRID+1 {
+				end = maxRID + 1
+			}
```
vaibhav-datazip: you can use ternary function here

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,217 @@
...
+		return splitViaPrimaryKey(ctx, stream)
+	}
+	return splitViaRID(ctx, stream)
+}
+
+func (d *DB2) getTableExtremes(ctx context.Context, stream types.StreamInterface, pkColumns []string) (min, max any, err error) {
+	query := jdbc.DB2MinMaxPKQuery(stream, pkColumns)
+	err = d.client.QueryRowContext(ctx, query).Scan(&min, &max)
+	return min, max, err
+}
```
vaibhav-datazip, line 218: is this function being used at multiple places, if not then why haven't we called this query directly where this is being used ?

drivers/db2/internal/cdc.go
```diff
@@ -0,0 +1,16 @@
+package driver
+
+import (
+	"context"
+
+	"github.com/datazip-inc/olake/drivers/abstract"
+	"github.com/datazip-inc/olake/types"
+)
+
```
vaibhav-datazip, line 9: you can add a comment here of cdc not supported yet similar to oracle

drivers/db2/internal/config.go
```diff
@@ -0,0 +1,80 @@
...
+	SSLConfiguration *utils.SSLConfig  `json:"ssl"`
+	MaxThreads       int               `json:"max_threads"`
+	RetryCount       int               `json:"retry_count"`
+	SSHConfig        *utils.SSHConfig  `json:"ssh_config"`
+}
+
+func (c *Config) BuildDSN() string {
+	// Base DSN
+	dsn := fmt.Sprintf(
+		"HOSTNAME=%s;PORT=%d;DATABASE=%s;UID=%s;PWD=%s;",
```
vaibhav-datazip, line 26: have we tested it for hostname and password with special characters ?

drivers/db2/internal/db2.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	return stream, nil
+}
+
+func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface{}, error) {
+	if value == nil {
+		return nil, typeutils.ErrNullValue
+	}
+
+	if columnType == "TIME" {
```
vaibhav-datazip, line 178: can column type contain other values of timestamp except of TIME , like TIMESTAMP-TZ TIMEDATE etc. in that case shouldn't we use string contains TIME

utils/typeutils/reformat.go
```diff
@@ -96,8 +98,9 @@ func ReformatValue(dataType types.DataType, v any) (any, error) {
 			return v, nil
 		case bool:
 			return fmt.Sprintf("%t", v), nil
+		// in db2, char and varchar comes in byte format
 		case []byte: // byte slice
-			return string(v), nil
+			return strings.TrimSpace(string(v)), nil
```
vaibhav-datazip: shouldn't we keep the original string ? trimspace will remove the leading or trailing spaces .

utils/typeutils/reformat.go
```diff
@@ -221,6 +224,11 @@ func ReformatDate(v interface{}) (time.Time, error) {
 		parsed = parsed.AddDate(-(parsed.Year() - 9999), 0, 0)
 	}
 
+	// for DB2, normallizing time cursor when year is 1 -> 1970-01-01
+	if parsed.Year() == 1 {
+		parsed = time.Date(1970, 1, 1, parsed.Hour(), parsed.Minute(), parsed.Second(), parsed.Nanosecond(), time.UTC)
+	}
```
vaibhav-datazip: wont this affect condition when other db , for eg. mongodb has year value as 1 ?

pkg/jdbc/jdbc.go
vaibhav-datazip: As this is being called for db2 case as well you can rename this function appropriately

vaibhav-datazip: we might need to add db2 here as well, because value formatting conditions have been defined for db2 inside `OracleIncrementalValueFormatter` function

pkg/jdbc/jdbc.go
```diff
@@ -770,3 +791,205 @@ func ThresholdFilter(ctx context.Context, opts DriverOptions) (string, []any, er
...
+			c.NULLS     AS is_nullable,
+			k.COLNAME   AS pk_column
+		FROM SYSCAT.COLUMNS c
+			LEFT JOIN SYSCAT.KEYCOLUSE k ON c.TABSCHEMA = k.TABSCHEMA AND c.TABNAME = k.TABNAME AND c.COLNAME = k.COLNAME
+		WHERE c.TABSCHEMA = ? AND c.TABNAME = ?
+		ORDER BY c.COLNO
+	`
+}
+
+// DB2RowCountQuery uses generic CARD from SYSCAT.TABLES (CARD is approx).
```
vaibhav-datazip: function name is different from comment similarly for `DB2MinMaxPKQuery`

pkg/jdbc/jdbc.go
```diff
@@ -770,3 +791,205 @@ func ThresholdFilter(ctx context.Context, opts DriverOptions) (string, []any, er
...
+		chunkCond = fmt.Sprintf("%s >= %v", ridFunc, chunk.Min)
+	case chunk.Max != nil:
+		chunkCond = fmt.Sprintf("%s < %v", ridFunc, chunk.Max)
+	}
+
+	// user-based filter
+	if filter != "" {
+		return fmt.Sprintf("SELECT * FROM %s WHERE (%s) AND (%s)", quotedTable, chunkCond, filter)
+	}
+	return fmt.Sprintf("SELECT * FROM %s WHERE %s", quotedTable, chunkCond)
```
vaibhav-datazip, line 854: ternary function can be used

utils/typeutils/reformat.go
ImDoubD-datazip: this is when byte array is returned. in case of db2, sometime it saves the string with some extra spaces, and when sync runs, the db2 driver return string in byte format with those leading spaces. we don't need those.

drivers/db2/internal/db2.go
ImDoubD-datazip: in db2 there are only 2 types, TIME and TIMESTAMP. There is nothing like timestamp_tz, timedate. It returns only time. for time, we need the only time format.

drivers/db2/internal/backfill.go
ImDoubD-datazip: same format followed as postgres.

drivers/db2/internal/db2.go
vaibhav-datazip: so why won't we need `TIMESTAMP` check here ?

utils/typeutils/reformat.go
vaibhav-datazip: no my concern is what if someone has intentionally added such spaces in string , that will result in incorrect data. 

eg. if someone added a data "   check space" or "check space  "  currently after passing through this it will give "check space".

drivers/db2/internal/backfill.go
ImDoubD-datazip: yes, in case when stats command is not run, this might return -1

drivers/db2/internal/backfill.go
ImDoubD-datazip: no one uses chunk column, sync is mostly first strategy based as seen till date from user side

drivers/abstract/incremental.go
ImDoubD-datazip: This is the format we require for DB2 timestamp to be saved in state.

drivers/db2/internal/config.go
ImDoubD-datazip: not yet

utils/typeutils/reformat.go
ImDoubD-datazip: this is not needed, added during implementation start for db2. removing this.

pkg/jdbc/jdbc.go
ImDoubD-datazip: if we use ternary, it won't look clean. I have followed what other drivers follow

drivers/db2/internal/backfill.go
ImDoubD-datazip: added a TODO as of now. otherwise will have to test the ongoing mysql sync as well. also the chunking strategy might change in the near future so then and there it can be done.

drivers/db2/internal/config.go
vaibhav-datazip: please do once

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,215 @@
...
+	if approxRowCount == -1 {
+		var hasRows bool
+		existsQuery := jdbc.DB2TableExistQuery(stream)
+		err := d.client.QueryRowContext(ctx, existsQuery).Scan(&hasRows)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check if table has rows: %s", err)
+		}
+
+		if hasRows {
+			return nil, fmt.Errorf("stats not populated for table[%s]. Please run command:\tRUNSTATS ON TABLE %s.%s WITH DISTRIBUTION AND DETAILED INDEXES ALL;\t to update table statistics", stream.ID(), stream.Namespace(), stream.Name())
```
vaibhav-datazip: instead of writing Please run command, you can mention Please run CLP command:

build.sh
```diff
@@ -8,6 +8,144 @@ function fail() {
...
+            if [[ "$arch_type" == "arm64" ]]; then
+                download_url="https://public.dhe.ibm.com/ibmdl/export/pub/software/data/db2/drivers/odbc_cli/macarm64_odbc_cli.tar.gz"
+                file_name="macarm64_odbc_cli.tar.gz"
+            else
+                download_url="https://public.dhe.ibm.com/ibmdl/export/pub/software/data/db2/drivers/odbc_cli/macos64_odbc_cli.tar.gz"
+                file_name="macos64_odbc_cli.tar.gz"
+            fi
+            ;;
+        "Linux")
+            download_url="https://public.dhe.ibm.com/ibmdl/export/pub/software/data/db2/drivers/odbc_cli/linuxx64_odbc_cli.tar.gz"
```
schitizsharma, line 32: So for linux arm64 don't exist. Is that why we skip the case?

build.sh
```diff
@@ -8,6 +8,144 @@ function fail() {
...
+        if go run github.com/ibmdb/go_ibm_db/installer@v0.5.4 2>/dev/null; then
+            echo "go_ibm_db installer succeeded"
+        else
+            echo "go_ibm_db installer failed, falling back to direct download..."
+            
+            # Clean up any partial downloads from the failed go installer
+            rm -rf "$install_dir/clidriver" 2>/dev/null
+            rm -f "$install_dir"/*.tar.gz 2>/dev/null
+            rm -f "$install_dir"/*.zip 2>/dev/null
+
```
schitizsharma: Let's just do curl. Keep it simple silly!

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,215 @@
...
+		logger.Warnf("Table %s is empty, skipping chunking", stream.ID())
+		return types.NewSet[types.Chunk](), nil
+	}
+	pool.AddRecordsToSyncStats(approxRowCount)
+	return d.splitTableIntoChunks(ctx, stream)
+}
+
+func (d *DB2) splitTableIntoChunks(ctx context.Context, stream types.StreamInterface) (*types.Set[types.Chunk], error) {
+	// split chunks via primary key
+	splitViaPrimaryKey := func(ctx context.Context, stream types.StreamInterface) (*types.Set[types.Chunk], error) {
```
vaibhav-datazip, line 96: I have tried syncing following table with 3 records

```
CREATE TABLE ALL_DB2_TYPES (
    COL_SMALLINT       SMALLINT,
    COL_INTEGER        INTEGER,
    COL_BIGINT         BIGINT,
    COL_DECIMAL        DECIMAL(10,2),
    COL_NUMERIC        NUMERIC(8,4),
    COL_REAL           REAL,
    COL_DOUBLE         DOUBLE,
    COL_DECFLOAT16     DECFLOAT(16),
    COL_DECFLOAT34     DECFLOAT(34),
    COL_CHAR10         CHAR(10),
    COL_VARCHAR50      VARCHAR(50),
    COL_VARGRAPHIC50   VARGRAPHIC(50),
    COL_GRAPHIC10      GRAPHIC(10),
    COL_LONGVARCHAR    LONG VARCHAR,
    COL_LONGVARGRAPHIC LONG VARGRAPHIC,
    COL_CHAR_BIT       CHAR(10) FOR BIT DATA,
    COL_VARCHAR_BIT    VARCHAR(20) FOR BIT DATA,
    COL_VARBINARY      VARBINARY(50),
    COL_DATE           DATE,
    COL_TIME           TIME,
    COL_TIMESTAMP      TIMESTAMP,
    COL_BOOLEAN        BOOLEAN,
    COL_CLOB           CLOB(1M),
    COL_DBCLOB         DBCLOB(500K),
    COL_BLOB           BLOB(500K),
    COL_XML            XML
);
```

I got the following error which syncing 

```
2026-01-09T09:06:38Z DEBUG Starting backfill for DB2INST1.ALL_DB2_TYPES with chunk {6 7} using query: SELECT * FROM "DB2INST1"."ALL_DB2_TYPES" WHERE RID("DB2INST1"."ALL_DB2_TYPES") >= 6 AND RID("DB2INST1"."ALL_DB2_TYPES") < 7
2026-01-09T09:06:38Z INFO Sync completed, wait 5 seconds cleanup in progress...
2026-01-09T09:06:43Z FATAL error occurred while reading records: error occurred while waiting for connections: thread[DB2INST1.ALL_DB2_TYPES_01KEH03WW4FHZKVDDVHQFEKQ43]: failed to insert chunk min[%!s(int64=4)] and max[%!s(int64=6)] of stream DB2INST1.ALL_DB2_TYPES, insert func error: %!s(<nil>), thread error: failed to flush data while closing: failed to write records: failed to send batch: rpc error: code = Internal desc = grpc: error while marshaling: string field contains invalid UTF-8
```

build.sh
ImDoubD-datazip: yes

build.sh
ImDoubD-datazip: ok

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,215 @@
...
+
+		logger.Warnf("Table %s is empty, skipping chunking", stream.ID())
+		return types.NewSet[types.Chunk](), nil
+	}
+	pool.AddRecordsToSyncStats(approxRowCount)
+	return d.splitTableIntoChunks(ctx, stream)
+}
+
+func (d *DB2) splitTableIntoChunks(ctx context.Context, stream types.StreamInterface) (*types.Set[types.Chunk], error) {
+	// split chunks via primary key
```
vaibhav-datazip, line 95: with 0 records I tried syncing the following table 

```
CREATE TABLE ALL_DB2_TYPES (
    COL_SMALLINT       SMALLINT,
    COL_INTEGER        INTEGER,
    COL_BIGINT         BIGINT,
    COL_DECIMAL        DECIMAL(10,2),
    COL_NUMERIC        NUMERIC(8,4),
    COL_REAL           REAL,
    COL_DOUBLE         DOUBLE,
    COL_DECFLOAT16     DECFLOAT(16),
    COL_DECFLOAT34     DECFLOAT(34),
    COL_CHAR10         CHAR(10),
    COL_VARCHAR50      VARCHAR(50),
    COL_VARGRAPHIC50   VARGRAPHIC(50),
    COL_GRAPHIC10      GRAPHIC(10),
    COL_LONGVARCHAR    LONG VARCHAR,
    COL_LONGVARGRAPHIC LONG VARGRAPHIC,
    COL_CHAR_BIT       CHAR(10) FOR BIT DATA,
    COL_VARCHAR_BIT    VARCHAR(20) FOR BIT DATA,
    COL_VARBINARY      VARBINARY(50),
    COL_DATE           DATE,
    COL_TIME           TIME,
    COL_TIMESTAMP      TIMESTAMP,
    COL_BOOLEAN        BOOLEAN,
    COL_CLOB           CLOB(1M),
    COL_DBCLOB         DBCLOB(500K),
    COL_BLOB           BLOB(500K),
    COL_XML            XML
);
```

got this error

```
2026-01-09T08:56:32Z INFO Sync completed, wait 5 seconds cleanup in progress...
2026-01-09T08:56:37Z FATAL error occurred while reading records: error occurred while waiting for context groups: failed to get or split chunks: failed to get the min and max rid: sql: Scan error on column index 0, name "1": converting NULL to int64 is unsupported
```

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,215 @@
...
+	reader := jdbc.NewReader(ctx, stmt, func(ctx context.Context, query string, queryArgs ...any) (*sql.Rows, error) {
+		return d.client.QueryContext(ctx, query, args...)
+	})
+
+	return reader.Capture(func(rows *sql.Rows) error {
+		record := make(types.Record)
+		if err := jdbc.MapScan(rows, record, d.dataTypeConverter); err != nil {
+			return fmt.Errorf("failed to scan record data as map: %s", err)
+		}
+		return OnMessage(ctx, record)
```
vaibhav-datazip: what isolation mode are we using here ?

drivers/db2/internal/datatype_conversion.go
```diff
@@ -0,0 +1,47 @@
...
+	"row":             types.String,
+
+	// binary
+	"blob":      types.String,
+	"binary":    types.String,
+	"varbinary": types.String,
+	"dbclob":    types.String,
+
+	// date / time
+	"time":      types.String,
```
vaibhav-datazip, line 44: the time I am seeing in dbeaver is different from iceberg

Image

in iceberg

Image

drivers/db2/internal/datatype_conversion.go
```diff
@@ -0,0 +1,47 @@
...
+	// numeric / decimal
+	"real":     types.Float32,
+	"float":    types.Float64,
+	"numeric":  types.Float64,
+	"double":   types.Float64,
+	"decimal":  types.Float64,
+	"decfloat": types.Float64,
+
+	// boolean
+	"boolean": types.Bool,
```
vaibhav-datazip, line 20: there is not boolean type in in db2, found this while testing

vaibhav-datazip (commented):
Tested for both full-refresh and incremental mode on OLake-CLI
- using only primary cursor
- using fallback cursor aswell
- using float , int, string , timestamp as cursor values
- filter using string, timestamp , int .

drivers/db2/internal/backfill.go
ImDoubD-datazip: data in database is invalid utf-8 vlues

drivers/db2/internal/backfill.go
ImDoubD-datazip: read committed or cursor stability as it is called by DB2

drivers/db2/internal/datatype_conversion.go
ImDoubD-datazip: in iceberg it is coming in utc format,  10 -> 4:30 (- 5:30)

drivers/db2/internal/datatype_conversion.go
ImDoubD-datazip: Screenshot 2026-01-09 at 4 39 45 PM

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,215 @@
...
+	if approxRowCount == -1 || approxRowCount == 0 {
+		var hasRows bool
+		existsQuery := jdbc.DB2TableExistQuery(stream)
+		err := d.client.QueryRowContext(ctx, existsQuery).Scan(&hasRows)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check if table has rows: %s", err)
+		}
+
+		if hasRows {
+			return nil, fmt.Errorf("stats not populated for table[%s]. Please run CLP command:\tRUNSTATS ON TABLE %s.%s WITH DISTRIBUTION AND DETAILED INDEXES ALL;\t to update table statistics", stream.ID(), stream.Namespace(), stream.Name())
```
vaibhav-datazip: for LOB (CLOB, DBCLOB, BLOB) and XML columns, don't support distribution statistics. Running RUNSTATS will give error while trying to run the given command in database

```
SQL2310N  The utility could not generate statistics.  Error "-668" was returned.
``` 

please mention about this in the doc as well

drivers/db2/internal/datatype_conversion.go
```diff
@@ -0,0 +1,47 @@
...
+	// integers
+	"smallint": types.Int32,
+	"integer":  types.Int32,
+	"bigint":   types.Int64,
+
+	// numeric / decimal
+	"real":     types.Float32,
+	"float":    types.Float64,
+	"numeric":  types.Float64,
+	"double":   types.Float64,
```
vaibhav-datazip, line 15: testing with DECFLOAT34 datatype, in db it was 

Image



in iceberg its 

Image

is this due to spark that we are getting in scientific notation, is there a way we can get as its in db

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,215 @@
...
+	} else {
+		stmt = jdbc.DB2RidChunkScanQuery(stream, chunk, filter)
+	}
+
+	logger.Debugf("Starting backfill for %s with chunk %v using query: %s", stream.ID(), chunk, stmt)
+
+	reader := jdbc.NewReader(ctx, stmt, func(ctx context.Context, query string, queryArgs ...any) (*sql.Rows, error) {
+		return d.client.QueryContext(ctx, query, args...)
+	})
+
```
vaibhav-datazip, line 64: in some cases blank values are taking NULL in others its blank

Image

Image

drivers/db2/internal/backfill.go
vaibhav-datazip: this will omit xml and lob types, which can cause incorrect avg row size , causing inconsistent file size

drivers/db2/internal/backfill.go
vaibhav-datazip: can we add a comment for this also add it in the docs

drivers/db2/internal/backfill.go
ImDoubD-datazip: changes done

drivers/db2/internal/backfill.go
vaibhav-datazip: still getting this error

ImDoubD-datazip: fix is done.

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,223 @@
...
+	if approxRowCount == -1 || approxRowCount == 0 {
+		var hasRows bool
+		existsQuery := jdbc.DB2TableExistQuery(stream)
+		err := d.client.QueryRowContext(ctx, existsQuery).Scan(&hasRows)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check if table has rows: %s", err)
+		}
+
+		if hasRows {
+			return nil, fmt.Errorf("stats not populated for table[%s]. Please run CLP command:\tRUNSTATS ON TABLE %s.%s WITH DETAILED INDEXES ALL;\t to update table statistics", stream.ID(), stream.Namespace(), stream.Name())
```
vaibhav-datazip: getting this error

SQL Error [42601]: An unexpected token "DETAILED" was found following "WITH".  Expected tokens may include:  "DISTRIBUTION".. SQLCODE=-104, SQLSTATE=42601, DRIVER=4.33.31

Image

drivers/postgres/internal/postgres_test_util.go
```diff
@@ -203,7 +203,7 @@ func insertTestData(t *testing.T, ctx context.Context, db *sqlx.DB, tableName st
 			col_uuid, col_varbit, col_xml,
 			col_point, col_polygon, col_circle
 		) VALUES (
-			%d, 123456789012345, DEFAULT, TRUE, 'c', 'char_val',
+			%d, 123456789012345, DEFAULT, TRUE, 'c', 'charac_val',
```
vaibhav-datazip, line 206: why were the changes in postgres test file made ?

ImDoubD-datazip: because spacing is left due to the datatype length we give, so had to change the data to match the length.

drivers/db2/internal/backfill.go
ImDoubD-datazip: changes done.

vaibhav-datazip: which will be better query for our use case ?


the previous one which was
```
RUNSTATS ON TABLE %s.%s WITH DISTRIBUTION AND DETAILED INDEXES ALL
```

or the

```
RUNSTATS ON TABLE %s.%s AND INDEXES ALL;
```

vaibhav-datazip (commented):
I am still getting the utf-8 error for this data in the table , but I don't see any utf8 values ?


| COL_SMALLINT | COL_INTEGER | COL_BIGINT | COL_DECIMAL | COL_NUMERIC | COL_REAL | COL_DOUBLE | COL_DECFLOAT16 | COL_DECFLOAT34 | COL_CHAR10 | COL_VARCHAR50 | COL_VARGRAPHIC50 | COL_GRAPHIC10 | COL_LONGVARCHAR | COL_LONGVARGRAPHIC | COL_DATE | COL_TIME | COL_TIMESTAMP | COL_BOOLEAN | COL_CLOB | COL_DBCLOB | COL_BLOB |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 100 | 20000 | 30000000 | 123.45 | 67.8912 | 1.5 | 3.14159 | 110 | 2200 | Row1Data | Full row standard data | GraphicData | GraphicFix | Long varchar text content | Long graphic content | 2026-02-15 | 10:00:00 | 2026-02-15-10.00.00.123456 | true | Standard CLOB text block | Standard DBCLOB text | [BLOB] |
| | 50 | | | 5.5555 | | 1.0 | | | | Sparse Row Data | | | | | 2026-03-01 | | 2026-01-13-19.39.12.588394 | false | | | |
| 0 | -99999 | 0 | 0.00 | 0.0000 | -1.5 | 0.0 | -100000 | 0 | | | | | | | 2026-01-01 | 00:00:00 | 2026-01-01-00.00.00.000000 | | | | [BLOB] |
| 555 | | 888 | | | | | | | | | | | | | | | | true | | | [BLOB] |
| | | | | | | | | | TextOnly | Checking text fields | DoubleByte | Graphics | Long text area content goes here | Long graphic content here | | | | | A very large clob entry that contains simple ascii text | DBCLOB content | |

this is the error

```
2026-01-13T20:15:31Z INFO Thread[DB2INST1.ALL_DB2_TYPES_01KEWFZKBYB5JYVFJW0J1MT433:50054] 01:45:31.687 [main] INFO  io.debezium.server.iceberg.OlakeRpcServer - Server started on port 50054 with max message size: 1024MB
2026-01-13T20:15:31Z INFO Thread[DB2INST1.ALL_DB2_TYPES_01KEWFZKBYB5JYVFJW0J1MT433]: Connected to new iceberg writer on port 50054
2026-01-13T20:15:31Z INFO Thread[DB2INST1.ALL_DB2_TYPES_01KEWFZKBYB5JYVFJW0J1MT433]: created writer for chunk min[%!s(int64=8)] and max[%!s(int64=9)] of stream DB2INST1.ALL_DB2_TYPES
2026-01-13T20:15:31Z INFO Sync completed, wait 5 seconds cleanup in progress...
2026-01-13T20:15:36Z FATAL error occurred while reading records: error occurred while waiting for connections: thread[DB2INST1.ALL_DB2_TYPES_01KEWFZD09QC8401505JMB5W4C]: failed to insert chunk min[%!s(int64=4)] and max[%!s(int64=8)] of stream DB2INST1.ALL_DB2_TYPES, insert func error: %!s(<nil>), thread error: failed to flush data while closing: failed to write records: failed to send batch: rpc error: code = Internal desc = grpc: error while marshaling: string field contains invalid UTF-8
admin@ip-192-168-0-104 volake % 
```

drivers/db2/internal/backfill.go
vaibhav-datazip: this error occurs when some column is deleted or added and can be solved by first running this command

```
CALL SYSPROC.ADMIN_CMD('REORG TABLE DB2INST1.ALL_DB2_TYPES');
```

you can read more about it

drivers/db2/internal/backfill.go
ImDoubD-datazip: We only require the 2nd command

ImDoubD-datazip:
non utf-8 data parsing is not supported as of now.

drivers/db2/internal/backfill.go
vaibhav-datazip: why is it better than the first one ?

ImDoubD-datazip: we don't require statistical distribution for our sync. normal indexes during chunk column strategy is required, and syscat table properties like avrrowsize, npages, pagesize is required which we get from runstats.

utils/typeutils/reformat.go
vishalm0509: We are keeping `trimspace`

drivers/db2/internal/config.go
ImDoubD-datazip: can be checked once after feature is launched, not of much priority.

utils/typeutils/reformat.go
vaibhav-datazip: had discussed this with @hash-data , trimspace won't have much affect.

Dockerfile
```diff
@@ -1,41 +1,82 @@
 # Build Stage
-FROM golang:1.24-alpine AS base
+FROM golang:1.24-bookworm AS builder
```
hash-data, line 2: why go update?

Dockerfile
```diff
@@ -1,41 +1,82 @@
...
+# Copy DB2 CLI Driver
+COPY --from=builder /go/pkg/mod/github.com/ibmdb/clidriver /opt/clidriver
+
+# Set DB2 CLI environment variables
+ENV IBM_DB_HOME=/opt/clidriver
+ENV PATH=$IBM_DB_HOME/bin:$PATH
+ENV CGO_CFLAGS="-I$IBM_DB_HOME/include"
+ENV CGO_LDFLAGS="-L$IBM_DB_HOME/lib -Wl,-rpath,$IBM_DB_HOME/lib"
+ENV LD_LIBRARY_PATH=$IBM_DB_HOME/lib
+
```
hash-data, line 80: why setting in all drivers?

Dockerfile
```diff
@@ -1,41 +1,82 @@
...
+FROM runtime-base-stage AS db2-stage
+
+# Copy DB2 CLI Driver
+COPY --from=builder /go/pkg/mod/github.com/ibmdb/clidriver /opt/clidriver
+
+# Set DB2 CLI environment variables
+ENV IBM_DB_HOME=/opt/clidriver
+ENV PATH=$IBM_DB_HOME/bin:$PATH
+ENV CGO_CFLAGS="-I$IBM_DB_HOME/include"
+ENV CGO_LDFLAGS="-L$IBM_DB_HOME/lib -Wl,-rpath,$IBM_DB_HOME/lib"
```
hash-data, line 78: we must do these in db2 only

drivers/abstract/incremental.go
hash-data: why not v.utc and then format?
also is this handling required? we need to remove driver level handling from abstraction. need to add a todo or think in that direction. It will mess framework going forward

utils/typeutils/reformat.go
```diff
@@ -579,13 +586,16 @@ func ReformatGeoType(v any) (any, error) {
...
 
-// FormatCursorValue is used to make time format and object id format consistent to be saved in state
-func FormatCursorValue(cursorValue any) any {
-	if _, ok := cursorValue.(time.Time); ok {
-		return cursorValue.(time.Time).UTC().Format("2006-01-02T15:04:05.000000000Z")
-	}
-	if oid, ok := cursorValue.(primitive.ObjectID); ok {
-		return oid.Hex()
+// ReformatTimeValue is used to make time format consistent as per db
+func ReformatTimeValue(value any) (any, error) {
```
hash-data: returning string only then why any ?

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,223 @@
...
+	var stmt string
+	if stream.Self().StreamMetadata.ChunkColumn != "" {
+		stmt = jdbc.DB2PKChunkScanQuery(stream, []string{stream.Self().StreamMetadata.ChunkColumn}, chunk, filter)
+	} else if len(pkColumns) > 0 {
+		stmt = jdbc.DB2PKChunkScanQuery(stream, pkColumns, chunk, filter)
+	} else {
+		stmt = jdbc.DB2RidChunkScanQuery(stream, chunk, filter)
+	}
+
+	// begin transaction for chunk iteration (by default isolation mode in db2 driver is cursor stability)
```
hash-data: what is default isolation mode? and why we are not updating to repeatable read?

drivers/db2/internal/backfill.go
```diff
@@ -0,0 +1,223 @@
...
+func (d *DB2) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPool, stream types.StreamInterface) (*types.Set[types.Chunk], error) {
+	// Approx Row Count
+	var approxRowCount int64
+	rowCountQuery := jdbc.DB2ApproxRowCountQuery(stream)
+	if err := d.client.QueryRowContext(ctx, rowCountQuery).Scan(&approxRowCount); err != nil {
+		return nil, fmt.Errorf("failed to get approx row count: %s", err)
+	}
+	if approxRowCount == -1 || approxRowCount == 0 {
+		var hasRows bool
+		existsQuery := jdbc.DB2TableExistQuery(stream)
```
hash-data: this is stats for table right not the table existance? please rename functions

drivers/db2/internal/datatype_conversion.go
```diff
@@ -0,0 +1,47 @@
...
+
+var db2TypeToDataTypes = map[string]types.DataType{
+	// integers
+	"smallint": types.Int32,
+	"integer":  types.Int32,
+	"bigint":   types.Int64,
+
+	// numeric / decimal
+	"real":     types.Float32,
+	"float":    types.Float64,
```
hash-data, line 13: float is float64? or 32

drivers/db2/internal/db2.go
```diff
@@ -0,0 +1,183 @@
...
+		return fmt.Errorf("failed to ping db2: %w", err)
+	}
+
+	d.client = client
+	d.config.RetryCount = utils.Ternary(d.config.RetryCount <= 0, 1, d.config.RetryCount+1).(int)
+	return nil
+}
+
+func (d *DB2) StateType() types.StateType {
+	return types.GlobalType
```
hash-data: not global, right now we are not using this variable but later on we might have to add. i explained a case

drivers/db2/internal/db2.go
```diff
@@ -0,0 +1,183 @@
...
+		return nil, fmt.Errorf("failed to list tables: %w", err)
+	}
+	defer rows.Close()
+
+	var streamNames []string
+	for rows.Next() {
+		var schema, name string
+		if err := rows.Scan(&schema, &name); err != nil {
+			return nil, fmt.Errorf("failed to scan table row: %w", err)
+		}
```
hash-data, line 107: are we skipping streams that are system level?

drivers/db2/README.md
```diff
@@ -0,0 +1,219 @@
+# DB2 LUW (Linux, Unix, Windows) Driver
+The DB2 LUW Driver enables data synchronization from DB2 LUW to your desired destination. It supports **Full Refresh** and **Incremental** mode.
```
hash-data: can we add about architecture it supports?

utils/typeutils/reformat.go
```diff
@@ -98,8 +100,13 @@ func ReformatValue(dataType types.DataType, v any) (any, error) {
 			return v, nil
 		case bool:
 			return fmt.Sprintf("%t", v), nil
+		// in db2, char and varchar comes in byte format
```
hash-data: this comment need to be in below case

utils/typeutils/reformat.go
```diff
@@ -98,8 +100,13 @@ func ReformatValue(dataType types.DataType, v any) (any, error) {
...
 		case []byte: // byte slice
-			return string(v), nil
+			val := strings.TrimSpace(string(v))
+			if val == "" {
+				return nil, ErrNullValue
+			}
+			return val, nil
```
hash-data: empty string is supported currently why we want to fail here? the other driver also use this.

release-tool.sh
```diff
@@ -129,7 +131,16 @@ setup_buildx
 
 # Release the driver
 platform="linux/amd64,linux/arm64"
-echo "✅ Releasing driver $DRIVER for version $VERSION on branch $CURRENT_BRANCH to platforms: $platform"
+# if driver is db2, target_stage -> db2-stage else target_stage -> driver-stage
+target_stage="driver-stage"
```
hash-data, line 135: can you explain this a bit

utils/typeutils/reformat.go
ImDoubD-datazip: fixing this. if empty then empty data will come, if null then none will come in destination.

release-tool.sh
ImDoubD-datazip: driver-stage is for the normal drivers where env required for db2 is not set, and image is build for all linux arm64/amd64

drivers/db2/README.md
ImDoubD-datazip: added

drivers/db2/internal/db2.go
ImDoubD-datazip: no system level tables were generated from initial connection to db2.

drivers/db2/internal/db2.go
ImDoubD-datazip: changed it to types.Stream

drivers/db2/internal/datatype_conversion.go
ImDoubD-datazip: float is float64 in db2

drivers/db2/internal/backfill.go
ImDoubD-datazip: added further comment in this

constants/constants.go
```diff
@@ -31,7 +31,9 @@ const (
 	// DestinationDatabasePrefix is used as prefix for destination database name
 	DestinationDatabasePrefix = "DESTINATION_DATABASE_PREFIX"
 	// EffectiveParquetSize is the effective size in bytes considering 256mb targeted parquet size, compression ratio as 8
-	EffectiveParquetSize = int64(256) * 1024 * 1024 * int64(8)
+	EffectiveParquetSize   = int64(256) * 1024 * 1024 * int64(8)
+	DB2TimestampFormat     = "2006-01-02 15:04:05.000000"
+	DefaultTimestampFormat = "2006-01-02T15:04:05.000000000Z"
```
hash-data: state timestamp format

utils/typeutils/reformat.go
```diff
@@ -99,7 +101,8 @@ func ReformatValue(dataType types.DataType, v any) (any, error) {
 		case bool:
 			return fmt.Sprintf("%t", v), nil
 		case []byte: // byte slice
-			return string(v), nil
+			// in db2, char and varchar comes in byte format
+			return strings.TrimSpace(string(v)), nil
```
hash-data: remove this change

utils/typeutils/reformat.go
```diff
@@ -30,13 +30,15 @@ var DateTimeFormats = []string{
...
 	"2006-01-02 15:04:05 -0700 MST",
+	"2006-01-02-15.04.05.000000",
 	"2006-01-02T15:04:05",
 	"2006-01-02T15:04:05.000000",
 	"2006-01-02T15:04:05.999999999Z07:00",
 	"2006-01-02T15:04:05+0000",
 	"2020-08-17T05:50:22.895Z",
 	"2006-01-02 15:04:05.999999-07",
 	"2006-01-02 15:04:05.999999+00",
+	"2006-01-02T15:04:05.000000000Z",
```
hash-data, line 41: these two format added intentionally?

ImDoubD-datazip: one is the format of db2: "2006-01-02-15.04.05.000000", other I have added as it was in state for default drivers.

PR #695, opened 2025-12-23
fix(mysql): prevent precision loss in DECIMAL and NUMERIC types

### LINKED ISSUE #693: Incorrect internal datatype for MySQL DECIMAL
## Description

I have column in MySQL with datatype DECIMAL(11,2) which OLake internally represents as "number_small".
This is float32 and according to spec float 32 could represents 6-7 Significant digits only.
Sources
https://en.wikipedia.org/wiki/Single-precision_floating-point_format
 https://learn.microsoft.com/en-us/cpp/c-language/type-float?view=msvc-170

This causes problem with passing large decimal values to Parquet file ( loosing precision).
This is critical to work financial data with precision 8 numbers and more.

## Steps to reproduce

In Mysql I have these values
157640.87
161864.63

OLake converts it to 
157640.875
161864.625

**Olake version**:   v0.3.5

## Proposed solution
Consider ALL decimal columns as "number" which is double(float64)  which supports 15-16 Significant digits

badalprasadsingh:
# Description

Values like `NUMERIC(11,2)` and `DECIMAL(11,2)` in MySQL were loosing their precision value in iceberg writers (both legacy and arrow) and parquet writer.

For e.g., `NUMERIC(11,2)` value `157640.87` in MySQL gets converted to `157640.875` when queried through Spark.

**Solution**
Using `types.Float64` for both "decimal" and "numeric" case, as it provides 15-17 significant digits.

Fixes #693 

## Type of change

# How Has This Been Tested?

- [x] Legacy Writer of Iceberg (via Spark and Athena)
- [x] Arrow Iceberg of Iceberg (via Spark and Athena)
- [x] Local Parquet Writer (via `parquet-tools` CLI)

## Documentation


drivers/mysql/internal/datatype_conversion.go
```diff
@@ -20,10 +20,13 @@ var mysqlTypeToDataTypes = map[string]types.DataType{
...
 	"real":    types.Float32,
-	"decimal": types.Float32,
-	"numeric": types.Float32,
 	"double":  types.Float64,
 
+	// Decimal and Numeric: Can handle up to 15 significant digits accurately (e.g., DECIMAL(15,2) or DECIMAL(15,7))
+	// Values with 16 digits may have minor rounding. Beyond 16 (from 17) digits will have precision loss.
+	"decimal": types.Float64,
+	"numeric": types.Float64,
+
```
vaibhav-datazip: why are we converting all the values of decimal and numeric to float64, shouldn't we check the scale and precision and then convert it to either float 32 or 64 like its being done in oracle.

drivers/mysql/internal/datatype_conversion.go
```diff
@@ -20,10 +20,13 @@ var mysqlTypeToDataTypes = map[string]types.DataType{
 	// Floating point types
 	"float":   types.Float32,
 	"real":    types.Float32,
-	"decimal": types.Float32,
-	"numeric": types.Float32,
 	"double":  types.Float64,
 
+	// Decimal and Numeric: Can handle up to 15 significant digits accurately (e.g., DECIMAL(15,2) or DECIMAL(15,7))
+	// Values with 16 digits may have minor rounding. Beyond 16 (from 17) digits will have precision loss.
+	"decimal": types.Float64,
```
shubham19may: is this backward compatible? 

what will happen to existing sync?

drivers/mysql/internal/datatype_conversion.go
badalprasadsingh: Yes, we can do it. Instead of always keeping it `float64` for all decimal and numeric cases, we can have a check here:
```go
if precision < 7 {
    return types.Float32
} else {
    return types.Float64
}
```

But need to confirm this once, cc: @hash-data

drivers/mysql/internal/datatype_conversion.go
badalprasadsingh: Yes this is backward compatible.
`Schema Evolution` case will be triggered for existing syncs.

Tested it via running the cdc sync swith state (without this change), added values into the source table, then again ran the cdc sync with state (with this change)

drivers/mysql/internal/datatype_conversion.go
hash-data: we can add a check as putting float64 for small values is not good, as it will come up with storage cost.

badalprasadsingh: Currently in Oracle, as well we use `float64` for all Decimal Types. 

For any driver, to decide on runtime whether to use `float32` or `float64` the DataTypeConvertor function needs to have access to precision and scale values, otherwise it looks up from the hardcoded map. Currently, the dataTypeConverter function does not have access to precision or scale values.

In Oracle, the hardcoded map uses float64 for all Number Types (Decimals included). 
Same doing for MySQL.

drivers/mysql/internal/datatype_conversion.go
```diff
@@ -20,10 +20,13 @@ var mysqlTypeToDataTypes = map[string]types.DataType{
 	// Floating point types
 	"float":   types.Float32,
 	"real":    types.Float32,
-	"decimal": types.Float32,
-	"numeric": types.Float32,
 	"double":  types.Float64,
 
+	// Can handle up to 15 significant digits accurately (e.g., DECIMAL(15,2) or DECIMAL(15,7))
+	// Values with 16 digits may have minor rounding. Beyond 16 (from 17) digits will have precision loss.
+	"numeric": types.Float64,
```
vaibhav-datazip, line 27: I've tried syncing the data using master branch first and then switching to feature branch to test backward  compatibility . The data present in database is this (checked using sql workbench as well as dbeaver)

Image


whereas the synced row of id 2 is 

Image

the f32 column's data after decimal is a little different

badalprasadsingh: discussed this in-person, this issue doesn't exists

image

vaibhav-datazip: What I conclude is that, The sync will give wrong data if the user defined float32 and tried adding value like 12345.6789, because of this the data which is getting stored in database will show rounded off data i.e. 12345.7 as we run select * during backfill but if we are running cdc it fetches data from the logs and as the logs contain the original value inserted by the user which is 12345.6789.

Hence, backfill will show correct data as its in db whereas cdc will show the data which was being inserted by the user.

vaibhav-datazip: had discussed this with @hash-data , this should be consistent with what we see using select * query, like typecasting it to the same format, also you can document your findings in a doc related to this.

badalprasadsingh: Yes, added a TODO for this.

Working on the doc, will send it to the team soon. (Won't sharing the doc on this thread).

drivers/mysql/internal/mysql_test_util.go
```diff
@@ -217,11 +217,11 @@ var ExpectedMySQLData = map[string]interface{}{
 	"id_smallint_unsigned":   int32(102),
 	"id_tinyint":             int32(50),
 	"id_tinyint_unsigned":    int32(51),
-	"price_decimal":          float32(123.45),
+	"price_decimal":          float64(123.45),
 	"price_double":           float64(123.456),
 	"price_double_precision": float64(123.456),
 	"price_float":            float32(123.45),
-	"price_numeric":          float32(123.45),
+	"price_numeric":          float64(123.45),
```
vaibhav-datazip, line 226: shouldn't we add some large values in the test case so that it catches any precision loss next time ?

drivers/mysql/internal/mysql_test_util.go
```diff
@@ -217,11 +218,12 @@ var ExpectedMySQLData = map[string]interface{}{
 	"id_smallint_unsigned":   int32(102),
 	"id_tinyint":             int32(50),
 	"id_tinyint_unsigned":    int32(51),
-	"price_decimal":          float32(123.45),
+	"price_decimal":          float64(123.45),
+	"amount_decimal_9_2":     float64(5330197.27),
```
vikaxsh, line 222: can we add Precision edge cases like DECIMAL(15,7), DECIMAL(16,2) or add an todo for these as well?

drivers/mysql/internal/mysql_test_util.go
badalprasadsingh: done.

pkg/binlog/filter.go
```diff
@@ -94,6 +94,8 @@ func convertRowToMap(row []interface{}, columns []string, columnTypes []string,
 		return nil, fmt.Errorf("column count mismatch: expected %d, got %d", len(columns), len(row))
 	}
 
+	// TODO: float values from binlog are not always same as the output of select * from db,
```
vikaxsh, line 97: why aren't we solving this issue in current pr? if precision loss? in cdc remains, data would be incorrect

drivers/mysql/internal/mysql_test_util.go
badalprasadsingh: done.

pkg/binlog/filter.go
badalprasadsingh: Had a detailed discussion on this, need to do a proper documentation on this, as this will affect each and every data type. Case: where the data inserted doesn't match the data type schema of the column in the table.

thus, not handling this in this PR.

vikaxsh: did we test DECIMAL(9,2) column in cdc mode, was there any issue?

badalprasadsingh: yes I tested, there was no issue in this PR
even the integration test handles that case

vaibhav-datazip (approved):

PR #697, opened 2025-12-23
fix(mysql): update mysql cdc timestamp to microsecond precision

badalprasadsingh:
# Description

`_cdc_timestamp` value for MySQL is now up to millisecond precision.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Legacy Writer: millisecond precision (queried via Spark)
- [x] Arrow Writer: microsecond precision (queried via Spark) - already known

## Documentation

### DIFF
--- a/pkg/binlog/binlog.go
+++ b/pkg/binlog/binlog.go
@@ -116,6 +116,8 @@ func (c *Connection) StreamMessages(ctx context.Context, client *sqlx.DB, callba
 					c.changeFilter.lastGTIDEvent = time.UnixMicro(int64(e.OriginalCommitTimestamp)) // #nosec G115 - timestamp value is always within int64 range
 				}
 
+				// TODO: Investigate MariaDB GTID event structure for microsecond timestamp support.
+
 			case *replication.RowsEvent:
 				messageReceived = true
 				if err := c.changeFilter.FilterRowsEvent(ctx, e, ev, callback); err != nil {


pkg/binlog/filter.go
```diff
@@ -77,7 +77,7 @@ func (f ChangeFilter) FilterRowsEvent(ctx context.Context, e *replication.RowsEv
 		}
 		change := abstract.CDCChange{
 			Stream:    stream,
-			Timestamp: time.Unix(int64(ev.Header.Timestamp), 0),
+			Timestamp: time.Now(),
```
vaibhav-datazip: the earlier code was giving the actual time at which change occurred in database , whereas this change will give the the processing time as `_cdc_timestamp`, right ?

documentation change will also be required if this is the expected behavior

[here](https://olake.io/docs/understanding/terminologies/olake/#2-timestamps) time stamp documentation 

Image

pkg/binlog/filter.go
```diff
@@ -77,7 +77,7 @@ func (f ChangeFilter) FilterRowsEvent(ctx context.Context, e *replication.RowsEv
...
 			Kind:      operationType,
 			Data:      record,
```
vaibhav-datazip, line 89: also I have tried syncing data in cdc mode , the timestamp is till millisecond precision but when queried through spark the last 3 digits are 0.

Image

pkg/binlog/filter.go
badalprasadsingh: Changing this to GTID based implementation

pkg/binlog/filter.go
badalprasadsingh: this is **second** level precision: `2025-12-24 10:32:22`
this is **milliecond** level precision: `2025-12-24 10:35:53.403000`
and, this is **microsecond** level precision: `2025-12-24 10:41:59.397195`

the last three digits has to be zero as millisecond is 10^3, last three digits will get filled in case of microsecond level precision

vaibhav-datazip: are we not supporting microsecond level precision as of now ?

badalprasadsingh: With the new GTID based implementation, we have microsecond level precision for `_cdc_timestamp` column

pkg/binlog/filter.go
```diff
@@ -14,8 +14,9 @@ import (
 
 // ChangeFilter filters binlog events based on the specified streams.
 type ChangeFilter struct {
-	streams   map[string]types.StreamInterface // Keyed by "schema.table"
-	converter func(value interface{}, columnType string) (interface{}, error)
+	streams       map[string]types.StreamInterface // Keyed by "schema.table"
+	converter     func(value interface{}, columnType string) (interface{}, error)
+	lastGTIDEvent *replication.GTIDEvent
```
vikaxsh: what if we change`lastGTIDEvent` from `*replication.GTIDEvent` to `time.Time` to store only the commit timestamp instead of the entire event object? since we requires only timestamp

pkg/binlog/binlog.go
```diff
@@ -111,6 +111,9 @@ func (c *Connection) StreamMessages(ctx context.Context, client *sqlx.DB, callba
 				c.CurrentPos.Pos = uint32(e.Position)
 				logger.Infof("Binlog rotated to %s:%d", c.CurrentPos.Name, c.CurrentPos.Pos)
 
+			case *replication.GTIDEvent:
+				c.changeFilter.TrackGTIDEvent(e)
```
vikaxsh: can we remove intermediate methods (`TrackGTIDEvent` and `getEventTimestamp`) and directly convert and assign the timestamp in the event handling loop?

pkg/binlog/filter.go
badalprasadsingh: done.

pkg/binlog/binlog.go
badalprasadsingh: done.

POV: we still have to make sure seconds remains as the fallback if for any MySQL version less than 8.0.1+ where GTID is not there, it falls back to seconds.

pkg/binlog/binlog.go
```diff
@@ -111,6 +111,11 @@ func (c *Connection) StreamMessages(ctx context.Context, client *sqlx.DB, callba
 				c.CurrentPos.Pos = uint32(e.Position)
 				logger.Infof("Binlog rotated to %s:%d", c.CurrentPos.Name, c.CurrentPos.Pos)
 
+			case *replication.GTIDEvent:
+				if e.OriginalCommitTimestamp > 0 {
+					c.changeFilter.lastGTIDEvent = time.UnixMicro(int64(e.OriginalCommitTimestamp)) // #nosec G115 - timestamp value is always within int64 range
```
vaibhav-datazip, line 116: as we are using gtid now, we'll have to update the documentation also, as this needs to be enabled right ?

a validation can also be added for GTID if its enabled or supported or not and in that case fallback.

pkg/binlog/filter.go
```diff
@@ -75,9 +77,14 @@ func (f ChangeFilter) FilterRowsEvent(ctx context.Context, e *replication.RowsEv
 		if record == nil {
 			continue
 		}
+
+		// Use microsecond-precision timestamp from GTID event (MySQL 8.0.1+) if available,
```
vaibhav-datazip, line 81: GTID was available from version 5.6 in mysql, aren't we using it for any mysql version 5.6 + ?

also what about mariadb ?

pkg/binlog/binlog.go
```diff
@@ -111,6 +111,11 @@ func (c *Connection) StreamMessages(ctx context.Context, client *sqlx.DB, callba
 				c.CurrentPos.Pos = uint32(e.Position)
 				logger.Infof("Binlog rotated to %s:%d", c.CurrentPos.Name, c.CurrentPos.Pos)
```
vaibhav-datazip, line 112: shouldn't we update the gtid event timestamp here as well ?

pkg/binlog/binlog.go
badalprasadsingh: yes, document needs to be updated, cc: @nayanj98 

The current code validation handles this case when the GTID will be off, as it will fallback to seconds. We don't need to have an extra check for that.

pkg/binlog/filter.go
badalprasadsingh: GTID is available from 5.6+, but originalCommitTimestamp is available from 8.01 onwards check [MySQL Link](https://dev.mysql.com/blog-archive/replication-features-in-mysql-8-0-1/)

For MariaDB let's handle it in a different PR or create a "good first issue" for it, let's deliver this first to the client.

pkg/binlog/binlog.go
badalprasadsingh: A RotateEvent occurs when MySQL changes to a new binlog file (due to size limits, or FLUSH LOGS). It's purely a file management event.

POV: If we reset the lastGTIDEvent on rotate, the upcoming events would incorrectly use second precision (the fallback) even though they're part of a transaction with a microsecond GTID timestamp.

pkg/binlog/filter.go
vaibhav-datazip: add a todo for it then

badalprasadsingh: done.

PR #698, opened 2025-12-23
feat: Implement 2-Phase Commit for Iceberg Sync

vaibhav-datazip:
# Description

This PR implements 2-phase commit mechanism for the Iceberg destination to ensure data consistency and prevent duplication during backfills and incremental/cdc syncs. It introduces a reliable way to check the commit status of individual threads (chunks) by querying the destination state.

Implemented for:
- Incremental all drivers 
- Full Refresh all drivers
- CDC : Mongo, Postgres, Mysql, MSSQL

## Type of change

# How Has This Been Tested?

- [x] **Manual Incremental Sync**: Verified that running an incremental sync correctly identifies previously committed thread IDs and skips them without duplication. 
  - stopped the sync after the chunk was committed in iceberg and the state file was not yet modified to chunk.
  - after this the sync is started again with state where the chunk is present but now the chunk is skipped as its already present in the destination table.
- [x] **Thread Status Check**: Validated that `CHECK_THREAD_STATUS` returns "true" only when the Iceberg table properties result explicitly confirms the commit for the given thread ID.
- [x] For cdc of mysql test with multiple streams , and all gets committed to iceberg but stop before state changes, then sync with state again and check data is not duplicated
- [x] Some streams are cdc and some are in other mode
- [x] processing array is empty but next cdc pos variable is still present.
- [x] Update 2-3 streams , and check if anyone is committed but another is not , then add more updates, and then sync , only updates for the first run should get synced (postgres and mysql)

## Documentation

- [ ] Documentation Link:
- [x] N/A

### DIFF
--- a/drivers/mssql/internal/cdc.go
+++ b/drivers/mssql/internal/cdc.go
@@ -27,36 +27,6 @@ type captureInstance struct {
 	startLSN     string
 }
 
-// prepareCaptureInstances discovers capture instances for selected streams
-func (m *MSSQL) prepareCaptureInstances(ctx context.Context, stream types.StreamInterface) ([]captureInstance, error) {
-	// Discover capture instances for selected streams
-	rows, err := m.client.QueryContext(ctx, jdbc.MSSQLCDCDiscoverQuery(stream.ID()))
-	if err != nil {
-		return nil, fmt.Errorf("failed to query MSSQL CDC tables: %s", err)
-	}
-	defer rows.Close()
-
-	var captureInstances []captureInstance
-	for rows.Next() {
-		var (
-			capture  captureInstance
-			startLSN []byte // to track the start LSN of the capture instance
-		)
-
-		if err := rows.Scan(&capture.schema, &capture.table, &capture.instanceName, &startLSN); err != nil {
-			return nil, fmt.Errorf("failed to scan MSSQL CDC table: %s", err)
-		}
-		capture.startLSN = hex.EncodeToString(startLSN)
-		captureInstances = append(captureInstances, capture)
-	}
-
-	if err := rows.Err(); err != nil {
-		return nil, err
-	}
-
-	return captureInstances, nil
-}
-
 func (m *MSSQL) ChangeStreamConfig() (bool, bool, bool) {
 	return false, false, true // concurrent change streams supported, stream can start after finishing full load
 }
@@ -73,22 +43,34 @@ func (m *MSSQL) PreCDC(ctx context.Context, streams []types.StreamInterface) err
 		return fmt.Errorf("failed to get MSSQL current max LSN: %w", err)
 	}
 
-	// check if CDC is enabled for each stream
-	for _, stream := range streams {
-		enabled, err := m.validateCDCStream(ctx, stream.Namespace(), stream.Name())
-		if err != nil {
-			return fmt.Errorf("failed to validate CDC for stream %s.%s: %s", stream.Namespace(), stream.Name(), err)
-		}
+	streamIDs := make([]string, len(streams))
+	for i, s := range streams {
+		streamIDs[i] = s.ID()
+	}
 
-		if !enabled {
+	// Bulk discovery of capture instances for validation and management
+	captureInstancesMap, err := m.prepareCaptureInstancesBulk(ctx, streamIDs)
+	if err != nil {
+		return fmt.Errorf("failed to discover capture instances: %w", err)
+	}
+
+	for _, stream := range streams {
+		if _, found := captureInstancesMap[stream.ID()]; !found {
 			return fmt.Errorf("CDC is not enabled for stream %s.%s", stream.Namespace(), stream.Name())
 		}
 
+		lsnVal := m.state.GetCursor(stream.Self(), cdcCursorKey)
 		// Initialize LSN for each stream if not present
-		if lsnVal := m.state.GetCursor(stream.Self(), cdcCursorKey); lsnVal == nil {
+		if lsnVal == nil {
 			m.state.SetCursor(stream.Self(), cdcCursorKey, currentLSN)
 		}
 	}
+
+	// Manage capture instance creation and deletion
+	if err := m.manageCaptureInstances(ctx, streamIDs, streams, captureInstancesMap); err != nil {
+		return fmt.Errorf("failed to manage CDC capture instances: %w", err)
+	}
+
 	m.streams = streams
 	return nil
 }
@@ -130,8 +112,6 @@ func (m *MSSQL) StreamChanges(ctx context.Context, streamIndex int, metadataStat
 		return nil, fmt.Errorf("failed to prepare capture instance for stream %s.%s: %s", stream.Namespace(), stream.Name(), err)
 	}
 
-	// TODO: research how to handle schema evolution
-
 	// When multiple capture instances exist for the same table (due to schema
 	// evolution), pick the newest instance whose startLSN is <= fromLSN.
 	// This guarantees we continue from an instance that was valid at our last
@@ -143,33 +123,20 @@ func (m *MSSQL) StreamChanges(ctx context.Context, streamIndex int, metadataStat
 	//
 	// Note: we expect column-level data loss (e.g., new columns missing)
 	// in the LSN range between the DDL and when the new capture instance becomes active.
-	var selectedCapture *captureInstance
-	for captureIdx := len(captureInstances) - 1; captureIdx >= 0; captureIdx-- {
-		// Skip if this capture started after fromLSN
-		if captureInstances[captureIdx].startLSN > lsnInState {
-			continue
-		}
-
-		// Select the capture instance
-		selectedCapture = &captureInstances[captureIdx]
-
-		// If a newer capture instance exists, restrict the targetLSN to the newer instance's startLSN
-		nextCaptureIdx := captureIdx + 1
-		if nextCaptureIdx < len(captureInstances) && targetLSN > captureInstances[nextCaptureIdx].startLSN {
-			newerCapture := captureInstances[nextCaptureIdx]
-			logger.Warnf("Newer capture instance [%s] detected for stream %s at LSN %s, but not using it in this sync. Clamping targetLSN to %s. It will be picked up in the next CDC sync", newerCapture.instanceName, stream.ID(), newerCapture.startLSN, newerCapture.startLSN)
-			targetLSN = newerCapture.startLSN
-		}
-
-		break
-	}
-
+	captureIdx, selectedCapture := newestValidInstance(captureInstances, lsnInState)
 	if selectedCapture == nil {
 		return nil, fmt.Errorf(
 			"LSN %s is earlier than the start LSN of available capture instances for stream %s. Please perform full-refresh",
 			lsnInState,
 			stream.ID(),
 		)
+	}
+	// If a newer capture instance exists, restrict the targetLSN to the newer instance's startLSN
+	nextCaptureIdx := captureIdx + 1
+	if nextCaptureIdx < len(captureInstances) && targetLSN > captureInstances[nextCaptureIdx].startLSN {
+		newerCapture := captureInstances[nextCaptureIdx]
+		logger.Warnf("Newer capture instance [%s] detected for stream %s at LSN %s, but not using it in this sync. Clamping targetLSN to %s. It will be picked up in the next CDC sync", newerCapture.instanceName, stream.ID(), newerCapture.startLSN, newerCapture.startLSN)
+		targetLSN = newerCapture.startLSN
 	}
 
 	logger.Infof(
@@ -205,6 +172,143 @@ func (m *MSSQL) PostCDC(ctx context.Context, streamIndex int) error {
 		}
 	}
 	return nil
+}
+
+// prepareCaptureInstancesBulk retrieves all currently defined CDC capture instances for the specified streams.
+func (m *MSSQL) prepareCaptureInstancesBulk(ctx context.Context, streamIDs []string) (map[string][]captureInstance, error) {
+	rows, err := m.client.QueryContext(ctx, jdbc.MSSQLCDCDiscoverQuery(streamIDs))
+	if err != nil {
+		return nil, fmt.Errorf("failed to query MSSQL CDC tables: %s", err)
+	}
+	defer rows.Close()
+
+	captureInstances := make(map[string][]captureInstance, len(streamIDs))
+	for rows.Next() {
+		var (
+			capture  captureInstance
+			startLSN []byte
+		)
+
+		if err := rows.Scan(&capture.schema, &capture.table, &capture.instanceName, &startLSN); err != nil {
+			return nil, fmt.Errorf("failed to scan MSSQL CDC table: %s", err)
+		}
+		capture.startLSN = hex.EncodeToString(startLSN)
+		streamID := fmt.Sprintf("%s.%s", capture.schema, capture.table)
+		captureInstances[streamID] = append(captureInstances[streamID], capture)
+	}
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("error iterating over bulk capture instances: %w", err)
+	}
+
+	return captureInstances, nil
+}
+
+// prepareCaptureInstances discovers capture instances for a stream
+func (m *MSSQL) prepareCaptureInstances(ctx context.Context, stream types.StreamInterface) ([]captureInstance, error) {
+	res, err := m.prepareCaptureInstancesBulk(ctx, []string{stream.ID()})
+	if err != nil {
+		return nil, err
+	}
+	return res[stream.ID()], nil
+}
+
+// manageCaptureInstances manages the lifecycle of CDC capture instances for multiple streams.
+func (m *MSSQL) manageCaptureInstances(ctx context.Context, streamIDs []string, streams []types.StreamInterface, captureInstancesMap map[string][]captureInstance) error {
+	// If manage capture instances is not enabled, do nothing
+	if !m.config.ManageCaptureInstances {
+		return nil
+	}
+
+	// Fetch DDL history for all streams in bulk
+	ddlHistoryQuery := jdbc.MSSQLCDCGetDDLHistoryBulkQuery(streamIDs)
+	rows, err := m.client.QueryContext(ctx, ddlHistoryQuery)
+	if err != nil {
+		return fmt.Errorf("failed to query bulk DDL history: %w", err)
+	}
+	defer rows.Close()
+
+	latestDDLMap := make(map[string]string, len(streamIDs))
+	for rows.Next() {
+		var (
+			schema, table, command string
+			ddlLSN                 []byte
+			ddlTime                time.Time
+			requiredColumnUpdate   bool
+		)
+		if err := rows.Scan(&schema, &table, &requiredColumnUpdate, &command, &ddlLSN, &ddlTime); err != nil {
+			return fmt.Errorf("failed to scan DDL history: %w", err)
+		}
+		streamID := fmt.Sprintf("%s.%s", schema, table)
+		// Since results are ORDER BY ASC, the last one for a table is the latest.
+		latestDDLMap[streamID] = hex.EncodeToString(ddlLSN)
+	}
+	if err := rows.Err(); err != nil {
+		return fmt.Errorf("error iterating over bulk DDL history: %w", err)
+	}
+
+	for _, stream := range streams {
+		streamID := stream.ID()
+		instances := captureInstancesMap[streamID]
+		// if cdc is not enabled it will fail in preCDC
+		// So we are sure that lsn exists in state for this stream
+		currentCursorLSN := m.state.GetCursor(stream.Self(), cdcCursorKey).(string)
+
+		// Find the newest valid instance
+		activeIdx, selected := newestValidInstance(instances, currentCursorLSN)
+		if selected == nil {
+			return fmt.Errorf("LSN %s for stream %s is older than any available scan instances; please perform a full refresh", currentCursorLSN, streamID)
+		}
+
+		// Delete fully consumed older instances and track survivors
+		deletedCount := 0
+		for idx, capture := range instances {
+			if idx != activeIdx && (currentCursorLSN == "" || capture.startLSN <= currentCursorLSN) {
+				query := jdbc.MSSQLCDCDisableCaptureInstanceQuery()
+				_, err := m.client.ExecContext(ctx, query, capture.schema, capture.table, capture.instanceName)
+				if err != nil {
+					return fmt.Errorf("failed to delete obsolete capture instance %s for %s: %w", capture.instanceName, streamID, err)
+				}
+				logger.Infof("Deleted fully consumed CDC capture instance [%s] for stream %s", capture.instanceName, streamID)
+				deletedCount++
+			}
+		}
+
+		// Check if we can create a new one (SQL Server limit is max 2)
+		// We use the count from our discovery minus what we just deleted.
+		if len(instances)-deletedCount == 1 {
+			latestCapture := instances[activeIdx]
+
+			// Check if any DDL event for this table is newer than the latest capture start_lsn
+			if ddlLSN, ok := latestDDLMap[streamID]; ok && ddlLSN > latestCapture.startLSN {
+				// Create a new instance with a safe length (MSSQL limit is 100 chars)
+				streamPart := streamID
+				if len(streamPart) > 75 {
+					streamPart = streamPart[:75]
+				}
+				newInstanceName := fmt.Sprintf("olake_%s_%d", streamPart, time.Now().Unix())
+
+				createCaptureInstanceQuery := jdbc.MSSQLCDCCreateCaptureInstanceQuery()
+				_, err := m.client.ExecContext(ctx, createCaptureInstanceQuery, latestCapture.schema, latestCapture.table, newInstanceName)
+				if err != nil {
+					return fmt.Errorf("failed to create new capture instance for schema drift on %s: %w", streamID, err)
+				}
+				logger.Infof("Detected schema drift and created new CDC capture instance [%s] for stream %s", newInstanceName, streamID)
+			}
+		}
+	}
+
+	return nil
+}
+
+// newestValidInstance selects the newest capture instance whose startLSN is <= currentLSN.
+func newestValidInstance(instances []captureInstance, currentLSN string) (int, *captureInstance) {
+	for i := len(instances) - 1; i >= 0; i-- {
+		// if currentLSN is empty, it means we are starting fresh, so we pick the latest instance
+		if currentLSN == "" || instances[i].startLSN <= currentLSN {
+			return i, &instances[i]
+		}
+	}
+	return -1, nil
 }
 
 // fetchTableChangesInLSNRange fetches and emits CDC changes for a single table/capture-instance within an LSN range.
--- a/drivers/mssql/internal/backfill.go
+++ b/drivers/mssql/internal/backfill.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/datazip-inc/olake/constants"
 	"github.com/datazip-inc/olake/destination"
@@ -124,9 +125,8 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 			if len(pkCols) == 0 {
 				return nil
 			}
-
 			// Get the minimum and maximum values for the primary key columns
-			minVal, maxVal, err := m.getTableExtremesMSSQL(ctx, stream, pkCols, tx)
+			minVal, maxVal, err := m.getTableExtremes(ctx, stream, pkCols, tx)
 			if err != nil {
 				return fmt.Errorf("failed to get table extremes: %s", err)
 			}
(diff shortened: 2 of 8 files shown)


destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeRowsIngester.java
```diff
@@ -110,6 +110,20 @@ public void sendRecords(IcebergPayload request, StreamObserver<RecordIngest.Reco
...
+                    
+                    RecordIngest.RecordIngestResponse statusResponse = RecordIngest.RecordIngestResponse.newBuilder()
+                        .setSuccess(committed)
+                        .setResult(String.valueOf(committed))
+                        .build();
+                    responseObserver.onNext(statusResponse);
+                    responseObserver.onCompleted();
+                    LOGGER.info("{} Processed check commit status for thread: {}, committed: {}", requestId, threadId, committed);
+                    break;
+
```
vaibhav-datazip: Instead of checking per thread, get the thread status of whole table. This will reduce api calls in turn increasing speed.

But this might fill the map created for checking thread status, when the logs get too big in iceberg properties table. This will increase memory utilization.

drivers/abstract/backfill.go
```diff
@@ -37,51 +37,69 @@ func (a *AbstractDriver) Backfill(ctx context.Context, backfilledStreams chan st
...
 					backfilledStreams <- stream.ID()
 				}
-			} else {
-				err = fmt.Errorf("thread[%s]: %s", threadID, err)
+				logger.Infof("finished chunk min[%v] and max[%v] of stream %s", chunk.Min, chunk.Max, stream.ID())
+				return nil
+			}, nil)()
+
+		// Check status if chunk was in preparing state
+		if chunk.Status == "preparing" {
```
vikaxsh: why do we need to add status in each chunk? why can't we just check for every chunk or we can fetch commit status for all threadIDs in once

drivers/abstract/backfill.go
```diff
@@ -37,51 +37,69 @@ func (a *AbstractDriver) Backfill(ctx context.Context, backfilledStreams chan st
...
+			committed, err := inserter.IsThreadCommitted(backfillCtx, threadID)
```
vikaxsh: once check should we add any backoff retry here?

pkg/waljs/types.go
```diff
@@ -24,7 +24,9 @@ type Config struct {
 }
 
 type WALState struct {
-	LSN string `json:"lsn"`
+	LSN        string   `json:"lsn"`
+	NextCDCPos string   `json:"next_cdc_pos,omitempty"` // For 2PC recovery - position before writers commit
+	Processing []string `json:"processing,omitempty"`   // Stream IDs currently being processed in CDC
```
vikaxsh: what if we can remove this and call for each stream with threadID ans check commit status or we can fetch status in batch and check locally stream wise

pkg/waljs/types.go
```diff
@@ -24,7 +24,9 @@ type Config struct {
 }
 
 type WALState struct {
-	LSN string `json:"lsn"`
+	LSN        string   `json:"lsn"`
+	NextCDCPos string   `json:"next_cdc_pos,omitempty"` // For 2PC recovery - position before writers commit
```
vikaxsh: can we store this as well in iceberg table metadata properties and max cursor becomes a property of the committed batch itself?
CDC progress is whatever the destination says is committed.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -120,102 +118,67 @@ public void commitThread(String threadId, Table table) {
...
+          dataFiles.forEach(appendFiles::appendFile);
+          appendFiles.commit();
         }
-  
-        txn.commitTransaction();
-  
-        LOGGER.info("Txn commit success: {} data files + {} equality delete files for thread: {}",
+
+        // 3. Final Commit to Catalog (Creates ONE metadata file)
+        transaction.commitTransaction();
```
vikaxsh: Open question:
Is it possible to support a transaction-level commit across all streams, where either all streams in a CDC batch are committed together or none are?

This could help avoid partial commits and reduce the complexity around per-stream recovery and cursor management.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -120,102 +118,67 @@ public void commitThread(String threadId, Table table) {
...
+
+        // Check if any WriteResult has delete files
+        boolean hasDeleteFiles = totalDeleteFiles > 0;
+
+        // Create a single transaction to bundle property update + data commit
+        org.apache.iceberg.Transaction transaction = table.newTransaction();
+
+        // 1. Stage Property Update
+        transaction.updateProperties()
+            .set(threadId, "committed")
```
vikaxsh: you can add max cursor here

destination/iceberg/iceberg.go
```diff
@@ -604,6 +630,32 @@ func isUpsertMode(stream types.StreamInterface, backfill bool) bool {
 	return utils.Ternary(stream.Self().StreamMetadata.AppendMode, false, !backfill).(bool)
 }
 
+func (i *Iceberg) ThreadStatus(ctx context.Context, threadID string) (bool, error) {
```
vikaxsh: we can create a function GetTableProperty here

drivers/abstract/backfill.go
vaibhav-datazip: this was similar to what we are doing with processing array , only check the thread id of chunks which has status as preparing.

destination/writers.go
```diff
@@ -83,6 +85,18 @@ func WithThreadID(threadID string) ThreadOptions {
 	}
 }
 
+func WithSyncMode(mode string) ThreadOptions {
+	return func(opt *Options) {
+		opt.SyncMode = mode
+	}
+}
```
vikaxsh: remove this function as stream already have info of sync mode

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -157,55 +165,67 @@ public void commitThread(String threadId, Table table) {
...
+          String key = "olake_2pc_cdc";
+          updateJsonState(table, updateProperties, key, threadId, syncMode, payload);
+      } else if (syncMode.equals("incremental")) {
+          String key = "olake_2pc_incremental";
+          updateJsonState(table, updateProperties, key, threadId, syncMode, payload);
+      }
+      
+      updateProperties.commit();
+
+      // 2. Stage Data Commit
```
vikaxsh, line 181: IMO, we can Simplify state management by treating Java as dumb storage, Store all sync mode state (backfill/cdc/incremental) as a single JSON string under the olake_state property key. Let the Go side own the state schema, marshaling before commit and unmarshaling after fetch, eliminating complex conditional logic

drivers/abstract/backfill.go
```diff
@@ -37,22 +38,75 @@ func (a *AbstractDriver) Backfill(mainCtx context.Context, backfilledStreams cha
...
+
+	checker, err := pool.NewWriter(statusCheckCtx, stream, destination.WithBackfill(true), destination.WithThreadID("status_checker_"+stream.ID()), destination.WithSyncMode("backfill_check"))
+	if err != nil {
+		return fmt.Errorf("failed to create status checker writer: %s", err)
+	}
+	defer checker.Close(statusCheckCtx)
+
+	var chunksToProcess []types.Chunk
+	for _, chunk := range chunks {
+		threadID := getThreadID(chunk)
```
vikaxsh: Why do we pre-filter chunks here by checking commit status before processing? Since we're already iterating chunks in the chunk processor , couldn't we check IsThreadCommitted there and skip processing if already committed? This would eliminate the duplicate iteration.

drivers/abstract/cdc.go
```diff
@@ -147,3 +236,131 @@ func mapChangeKindToOperationType(kind string) string {
 		return "c"
 	}
 }
+
+// cdcRecovery checks processing streams commit status and prepares recovery state.
+func (a *AbstractDriver) cdcRecovery(ctx context.Context, pool *destination.WriterPool, streams []types.StreamInterface) (bool, map[string]bool, error) {
```
vikaxsh: it would be better to create dedicated function for each task instead of single function like checkStreamCommitStatus, handlePerStreamRecovery etc

drivers/abstract/cdc.go
```diff
@@ -147,3 +236,131 @@ func mapChangeKindToOperationType(kind string) string {
...
+	perStream2PC, supportsPerStreamPosition2PC := a.driver.(PerStreamPosition2PC)
+	posAck, supportsPositionAcknowledgment := a.driver.(PositionAcknowledgment)
+
+	// First, check per-stream recovery (MongoDB next_data pattern)
+	// This handles streams that have next_data but may/may not be committed
+	if supportsPerStreamPosition2PC {
+		for _, stream := range streams {
+			if err := perStream2PC.CheckPerStreamRecovery(ctx, pool, stream); err != nil {
+				return false, nil, fmt.Errorf("failed per-stream recovery check for %s: %s", stream.ID(), err)
+			}
```
vikaxsh: why aren't processing of streams parallely?

drivers/abstract/cdc.go
```diff
@@ -147,3 +236,131 @@ func mapChangeKindToOperationType(kind string) string {
...
+			hash := fmt.Sprintf("%x", sha256.Sum256([]byte(startingPosition)))
+			threadID = fmt.Sprintf("%s_%s", streamID, hash)
+		} else {
+			// This shouldn't happen for global position drivers, but handle gracefully
+			logger.Warnf("No starting position for recovery thread ID generation for stream %s", streamID)
+			return false, nil, fmt.Errorf("cannot generate thread ID without starting position")
+		}
+
+		// Create a temporary writer to check commit status
+		writer, err := pool.NewWriter(ctx, stream, destination.WithThreadID(threadID), destination.WithSyncMode("recovery_check"))
```
vikaxsh: what is use of "recovery_check" here

drivers/abstract/cdc.go
```diff
@@ -97,14 +109,67 @@ func (a *AbstractDriver) RunChangeStream(mainCtx context.Context, pool *destinat
...
+	// For global position drivers (MySQL/Postgres), use GetCDCStartPosition
+	// For per-stream drivers (MongoDB), use change.Position per stream
+	var startingPosition string
+	if supportsGlobalPosition2PC {
+		startingPosition = global2PC.GetCDCStartPosition()
+	}
+
+	// Use recovery state prepared by cdcRecovery
+	isRecoveryMode := a.recoveryMode
+	processingSet := a.recoveryProcessingSet
```
vikaxsh: why can't directly use a.recoveryMode and a.recoveryProcessingSet

drivers/mongodb/internal/cdc.go
```diff
@@ -192,16 +199,63 @@ func (m *Mongo) PostCDC(ctx context.Context, streamIndex int) error {
...
+func (m *Mongo) CheckPerStreamRecovery(ctx context.Context, pool *destination.WriterPool, stream types.StreamInterface) error {
+	currentData := m.state.GetCursor(stream.Self(), cdcCursorField)
+
+	streamID := stream.ID()
+	// Generate threadID using _data (the position at which we started the sync)
+	hash := fmt.Sprintf("%x", sha256.Sum256([]byte(currentData.(string))))
+	threadID := fmt.Sprintf("%s_%s", streamID, hash)
+
+	// Create temporary writer to check commit status
+	writer, err := pool.NewWriter(ctx, stream, destination.WithThreadID(threadID))
```
vikaxsh: writer instance should be created only in abstract layer

drivers/mongodb/internal/cdc.go
```diff
@@ -192,16 +199,63 @@ func (m *Mongo) PostCDC(ctx context.Context, streamIndex int) error {
...
+	}
+	if closeErr := writer.Close(ctx); closeErr != nil {
+		logger.Warnf("Failed to close recovery check writer for stream %s: %s", stream.ID(), closeErr)
+	}
+
+	if statePayload != "" {
+		var payloadMap map[string]string
+		if err := json.Unmarshal([]byte(statePayload), &payloadMap); err == nil {
+			if pos, ok := payloadMap["captured_cdc_pos"]; ok && pos != "" {
+				m.state.SetCursor(stream.Self(), cdcCursorField, pos)
```
vikaxsh: see if we can update state in abstract layer itself

drivers/abstract/incremental.go
```diff
@@ -73,22 +82,43 @@ func (a *AbstractDriver) Incremental(mainCtx context.Context, pool *destination.
...
+			hash := fmt.Sprintf("%x", sha256.Sum256([]byte(rawCtx)))
+			threadID := generateThreadID(stream.ID(), hash)
+
+			// Use a map reference for payload so we can update it dynamically inside the loop
+			payload := map[string]string{
+				primaryCursor: fmt.Sprintf("%v", a.FormatCursorValue(maxPrimaryCursorValue)),
+			}
+			if secondaryCursor != "" {
+				payload[secondaryCursor] = fmt.Sprintf("%v", a.FormatCursorValue(maxSecondaryCursorValue))
+			}
```
vikaxsh: IMO, consider creating payload struct for cdc, incremental and full load

drivers/mysql/internal/cdc.go
```diff
@@ -96,12 +104,123 @@ func (m *MySQL) PostCDC(ctx context.Context, _ int) error {
...
+		mysqlState.ServerID = m.BinlogConn.ServerID
+		mysqlState.State = binlog.Binlog{
+			Position: m.BinlogConn.CurrentPos,
+		}
+		// If all streams were successfully committed (processing is empty),
+		// clear next_cdc_pos as we don't need it for recovery
+		if len(mysqlState.Processing) == 0 {
+			mysqlState.NextCDCPos = ""
+		}
+		m.state.SetGlobal(mysqlState)
```
vikaxsh: why do we need this?

drivers/mysql/internal/mysql.go
```diff
@@ -22,25 +22,27 @@ import (
...
 }
 
 // MySQLGlobalState tracks the binlog position and backfilled streams.
 type MySQLGlobalState struct {
-	ServerID uint32        `json:"server_id"`
-	State    binlog.Binlog `json:"state"`
+	ServerID   uint32        `json:"server_id"`
+	State      binlog.Binlog `json:"state"`
+	NextCDCPos string        `json:"next_cdc_pos,omitempty"` // For 2PC recovery - position before writers commit
+	Processing []string      `json:"processing,omitempty"`   // Stream IDs currently being processed in CDC
```
vikaxsh: need to remove these

drivers/postgres/internal/cdc.go
```diff
@@ -85,9 +97,130 @@ func (p *Postgres) PostCDC(ctx context.Context, _ int) error {
...
+		if err := utils.Unmarshal(globalState.State, &walState); err != nil {
+			logger.Warnf("Failed to unmarshal global state for SetCurrentCDCPosition: %s", err)
+		}
+	}
+	walState.LSN = position
+	p.state.SetGlobal(walState)
+	logger.Infof("Set current CDC position (LSN) in state: %s", position)
+}
+
+func (p *Postgres) RemoveProcessingStream(streamID string) {
```
vikaxsh: Could you please remove all outdated and unnecessary implementation functions?

pkg/binlog/filter.go
```diff
@@ -93,10 +93,7 @@ func (f ChangeFilter) FilterRowsEvent(ctx context.Context, e *replication.RowsEv
 			Timestamp: timestamp,
 			Kind:      operationType,
 			Data:      record,
-			ExtraColumns: map[string]any{
-				CDCBinlogFileName: pos.Name,
-				CDCBinlogFilePos:  pos.Pos, // Use the event position
-			},
+			Position:  position,
```
vikaxsh: Why were the extra columns removed?

pkg/binlog/binlog.go
```diff
@@ -63,10 +65,26 @@ func NewConnection(_ context.Context, config *Config, pos mysql.Position, stream
...
+		if err != nil {
+			return fmt.Errorf("failed to parse target position %s: %s", targetPosition, err)
+		}
+		logger.Infof("Using target position for bounded sync: %s:%d", latestBinlogPos.Name, latestBinlogPos.Pos)
+	} else {
+		// Get current latest binlog position
+		latestBinlogPos, err = GetCurrentBinlogPosition(ctx, client)
+		if err != nil {
+			return fmt.Errorf("failed to get latest binlog position: %s", err)
+		}
```
vikaxsh: would be better to keep one function for fetching target position

drivers/abstract/incremental.go
```diff
@@ -73,22 +82,43 @@ func (a *AbstractDriver) Incremental(mainCtx context.Context, pool *destination.
 			incrementalCtx, incrementalCtxCancel := context.WithCancel(gCtx)
 			defer incrementalCtxCancel()
 
-			threadID := generateThreadID(stream.ID(), "")
-			inserter, err := pool.NewWriter(incrementalCtx, stream, destination.WithThreadID(threadID))
+			rawCtx := fmt.Sprintf("primaryCursor_%v_secondaryCursor_%v", a.FormatCursorValue(maxPrimaryCursorValue), a.FormatCursorValue(maxSecondaryCursorValue))
+			hash := fmt.Sprintf("%x", sha256.Sum256([]byte(rawCtx)))
+			threadID := generateThreadID(stream.ID(), hash)
+
```
vikaxsh: It might be better to create a common utility function for thread ID creation and reuse it wherever needed.

types/state.go
```diff
@@ -188,6 +188,54 @@ func (s *State) SetCursor(stream *ConfiguredStream, key string, value any) {
...
+	s.Lock()
+	defer s.Unlock()
+
+	index, contains := utils.ArrayContains(s.Streams, func(elem *StreamState) bool {
+		return elem.Namespace == stream.Namespace() && elem.Stream == stream.Name()
+	})
+
+	if contains {
+		for key, value := range cursors {
+			if key != "" {
```
vikaxsh: are we expecting empty keys? if not nil check not required

drivers/abstract/incremental.go
```diff
@@ -38,16 +44,19 @@ func (a *AbstractDriver) Incremental(mainCtx context.Context, pool *destination.
...
 				logger.Warnf("max primary cursor value is nil for stream: %s", stream.ID())
 			}
 			if secondaryCursor != "" {
-				a.state.SetCursor(stream.Self(), secondaryCursor, a.FormatCursorValue(maxSecondaryCursorValue))
+				cursors[secondaryCursor] = a.FormatCursorValue(maxSecondaryCursorValue)
 				if maxSecondaryCursorValue == nil {
 					logger.Warnf("max secondary cursor value is nil for stream: %s", stream.ID())
 				}
 			}
+			a.state.SetCursors(stream.Self(), cursors)
```
vikaxsh: why?

types/state.go
```diff
@@ -188,6 +188,54 @@ func (s *State) SetCursor(stream *ConfiguredStream, key string, value any) {
 	s.LogState()
 }
 
+// SetCursors sets multiple cursors for a stream in state, making it atomic operation
+func (s *State) SetCursors(stream *ConfiguredStream, cursors map[string]any) {
```
vikaxsh: we can merge setCursors and setCursor loigc

drivers/abstract/cdc.go
```diff
@@ -21,7 +23,17 @@ import (
 //   - Parallel: Process all streams simultaneously after all backfills complete
 //   - Concurrent: Start each stream's CDC immediately after its backfill completes (can overlap)
 func (a *AbstractDriver) RunChangeStream(mainCtx context.Context, pool *destination.WriterPool, streams ...types.StreamInterface) error {
-	// run pre cdc of drivers
+	// This needs to run before PreCDC so that if all streams are committed,
+	// we can update current LSN from next_cdc_pos before driver initializes CDC
+	isRecoveryMode, processingSet, err := a.cdcRecovery(mainCtx, pool, streams)
```
vikaxsh: can we add 2PC support inside existing orchestration?

drivers/abstract/cdc.go
vaibhav-datazip: bounded cdc will run in recovery mode

drivers/abstract/incremental.go
vaibhav-datazip: making cursor update in state atomic , earlier one used to get committed during secondary it might get failed

destination/writers.go
vaibhav-datazip: but during our initial backfill , that will return cdc or incremental instead of backfill right ?

types/state.go
vaibhav-datazip: well, we were doing the same for old set cursor function also , we are doing it in get cursor and delete cursor

drivers/abstract/incremental.go
vaibhav-datazip: generate thread id is common function

drivers/abstract/cdc.go
vaibhav-datazip: changed it to cdc

drivers/mongodb/internal/cdc.go
vaibhav-datazip: that will make it a 2 step process, first fetch thread id from mongo then check recovery

drivers/mongodb/internal/cdc.go
vaibhav-datazip: shouldn't state be handled in respective driver code?

pkg/binlog/binlog.go
vaibhav-datazip: m.GetTargetCDCPosition() does that

drivers/mysql/internal/cdc.go
vaibhav-datazip: old code , removed

drivers/abstract/incremental.go
vaibhav-datazip: there is payload for cdc, we don't need it for backfill

drivers/abstract/cdc.go
vaibhav-datazip: there are dedicated functions which are called inside this cdc recovery

drivers/abstract/backfill.go
```diff
@@ -38,22 +44,51 @@ func (a *AbstractDriver) Backfill(mainCtx context.Context, backfilledStreams cha
 		return typeutils.Compare(chunks[i].Min, chunks[j].Min) < 0
 	})
 
-	logger.Infof("Starting backfill for stream[%s] with %d chunks", stream.GetStream().Name, len(chunks))
```
vikaxsh, line 41: we should keep this log

drivers/abstract/backfill.go
```diff
@@ -84,3 +121,36 @@ func (a *AbstractDriver) Backfill(mainCtx context.Context, backfilledStreams cha
...
+	if err := json.Unmarshal([]byte(statePayload), &state); err != nil {
+		logger.Warnf("Failed to parse commit state JSON, skipping optimization: %s", err)
+		return committed
+	}
+
+	for _, id := range state.FullRefreshCommittedIDs {
+		if id != "" {
+			committed[id] = true
+		}
+	}
```
vikaxsh: you can use set here

drivers/abstract/backfill.go
```diff
@@ -84,3 +121,36 @@ func (a *AbstractDriver) Backfill(mainCtx context.Context, backfilledStreams cha
 	utils.ConcurrentInGroupWithRetry(a.GlobalConnGroup, chunks, a.driver.MaxRetries(), chunkProcessor)
 	return nil
 }
+
+func (a *AbstractDriver) fetchBackfillCommittedThreadIDs(ctx context.Context, pool *destination.WriterPool, stream types.StreamInterface) map[string]bool {
+	committed := make(map[string]bool)
+
+	stateFetcher, err := pool.NewWriter(ctx, stream, destination.WithThreadID("recovery_check"), destination.WithBackfill(true))
```
vikaxsh: can we avoid creating dummy writer instance?

drivers/abstract/cdc.go
```diff
@@ -114,16 +179,40 @@ func (a *AbstractDriver) streamChanges(mainCtx context.Context, pool *destinatio
...
+			// Use startingPosition (global for MySQL/Postgres) or change.Position (per-stream for MongoDB)
+			positionForHash := startingPosition
+			if positionForHash == "" {
+				positionForHash = change.Position // MongoDB per-stream position
+			}
+			hash := fmt.Sprintf("%x", sha256.Sum256([]byte(positionForHash)))
+			threadID := fmt.Sprintf("%s_%s", streamID, hash)
+
+			payload := map[string]string{
+				"captured_cdc_pos": change.Position,
```
vikaxsh: shouldn't we store latest check point here?

drivers/abstract/cdc.go
```diff
@@ -147,3 +236,159 @@ func mapChangeKindToOperationType(kind string) string {
...
+	commitStateLatestThreadIDKey = "latest_threadId"
+	commitStateCapturedCDCPosKey = "captured_cdc_pos"
+)
+
+func cdcThreadID(streamID, positionForHash string) string {
+	hash := fmt.Sprintf("%x", sha256.Sum256([]byte(positionForHash)))
+	return fmt.Sprintf("%s_%s", streamID, hash)
+}
+
+func arrayToSet(values []string) map[string]bool {
```
vikaxsh: use set implementation from types package

drivers/abstract/cdc.go
```diff
@@ -147,3 +236,159 @@ func mapChangeKindToOperationType(kind string) string {
...
+				recoveredGlobalPos = capturedPos
+			}
+			continue
+		}
+
+		uncommittedStreams = append(uncommittedStreams, streamID)
+	}
+
+	// If all streams committed, update source/driver state and proceed normally (no recovery gating).
+	if len(uncommittedStreams) == 0 {
```
vikaxsh: since all stream are committed successfully, why do we need to update in state?

drivers/mssql/internal/cdc.go
```diff
@@ -86,7 +86,7 @@ func (m *MSSQL) PreCDC(ctx context.Context, streams []types.StreamInterface) err
 
 		// Initialize LSN for each stream if not present
 		if lsnVal := m.state.GetCursor(stream.Self(), cdcCursorKey); lsnVal == nil {
-			m.state.SetCursor(stream.Self(), cdcCursorKey, currentLSN)
+			m.state.SetCursors(stream.Self(), map[string]any{cdcCursorKey: currentLSN})
```
vikaxsh: don't we need to send change.Position in mssql cdc?

drivers/abstract/cdc.go
vikaxsh: is it done?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -423,6 +435,50 @@ private PartitionData partitionDataFromTypedValues(PartitionSpec spec,
...
+                  // Merge payload directly into root node
+                  payloadNode.fields().forEachRemaining(entry -> rootNode.set(entry.getKey(), entry.getValue()));
+              }
+          } else {
+              // No payload => backfill/snapshot style: append threadId to full_refresh_committed_ids
+              com.fasterxml.jackson.databind.node.ArrayNode committedIds;
+              if (rootNode.has(STATE_FIELD_FULL_REFRESH_COMMITTED_IDS) && rootNode.get(STATE_FIELD_FULL_REFRESH_COMMITTED_IDS).isArray()) {
+                  committedIds = (com.fasterxml.jackson.databind.node.ArrayNode) rootNode.get(STATE_FIELD_FULL_REFRESH_COMMITTED_IDS);
+              } else {
+                  committedIds = rootNode.putArray(STATE_FIELD_FULL_REFRESH_COMMITTED_IDS);
```
vikaxsh, line 464: open question:
can we avoid merging of thread ids here by keeping state strucre minimal here, for exmaple:
```olake_state: {
    "checkpoint": "primary:12345,secondary:67890",

    "threadid1": "completed",
    "threadid2": "completed",
    "threadid3": "completed"
}```

File: drivers/abstract/cdc.go (hunk 28 above)
Comment (vikaxsh) [reply drivers/abstract/cdc.go]: open question:
Instead of pre-iterating over all streams to decide recovery mode, we can make the process fully dynamic.
With this approach:

- For every thread, Go checks the stored state to determine if that thread is already committed.
- Committed threads are skipped automatically.
- Uncommitted threads are processed normally.
- For uncommitted threads, the target CDC position can be updated dynamically on the Go side as progress is made.
- No separate recovery phase or pre-iteration over streams is required.

File: drivers/abstract/cdc.go (hunk 34 above)
Comment (vaibhav-datazip) [reply drivers/abstract/cdc.go]: cursor value is not yet committed that's why

File: drivers/mssql/internal/cdc.go (hunk 35 above)
Comment (vaibhav-datazip) [reply drivers/mssql/internal/cdc.go]: this driver was added after initial changes were made, will add it

File: drivers/abstract/cdc.go [hunk 37]
```diff
@@ -147,3 +272,47 @@ func mapChangeKindToOperationType(kind string) string {
...
+		return nil
+	}
+	if supportsGlobalPosition2PC {
+		return streams
+	}
+	if streamIndex >= 0 && streamIndex < len(streams) {
+		return streams[streamIndex : streamIndex+1]
+	}
+	return nil
+}
```
vaibhav-datazip: This might return false streams for other drivers for which 2pc is not yet implemented

drivers/abstract/cdc.go
```diff
@@ -133,6 +143,15 @@ func (a *AbstractDriver) streamChanges(mainCtx context.Context, pool *destinatio
...
 		return writer.Push(ctx, types.CreateRawRecord(change.Data, olakeColumns))
 	})
+	if err != nil {
+		return err
+	}
+	if rawMetadata != nil {
+		if m, ok := rawMetadata.(map[string]any); ok {
+			finalMetadataState = m
+		}
+	}
```
hash-data: pass it directly handle cleanup will take care

PR #710, opened 2025-12-29
fix(MySQL): Update CDC Event Enum Index Binlog Issue

### LINKED ISSUE #707: MySQL ENUM datatype doesnt work correctly in CDC mode
## Description
There is  MySQL table which has ENUM column. I configured job to sync this table in "Full refresh + CDC" mode to Iceberg.  If I run sync job 1st time ( full refresh) then I see proper string values in Parquet file. If run sync job 2nd time (CDC) Parquet file contains IDs of those string values 

## Steps to reproduce

- create table 

CREATE TABLE sizes (
    id int,
    size ENUM('x-small', 'small', 'medium', 'large', 'x-large')
);

insert into sizes (id, size) values (76, null);
insert into sizes (id, size) values (77, 'x-small');
insert into sizes (id, size) values (78, 'small');
insert into sizes (id, size) values (79, 'medium');
insert into sizes (id, size) values (80, 'large');
insert into sizes (id, size) values (81, 'x-large');

- run sync
- Parquet file has "x-small, small, large"  values in "**size**" column
- Add new records:

insert into sizes (id, size) values (176, null);
insert into sizes (id, size) values (177, 'x-small');
insert into sizes (id, size) values (178, 'small');
insert into sizes (id, size) values (179, 'medium');
insert into sizes (id, size) values (180, 'large');
insert into sizes (id, size) values (181, 'x-large');
commit;

- run sync job again
- Parquet file has "1,2,3.." values in "**size**" column  

## Expected behavior

In CDC mode Parquet should contain real string values in ENUM column  instead of their IDs ( same behavior as in Full refresh mode)

badalprasadsingh:
# Description

During an update CDC event when the value for an ENUM column type in a particular row is updated, the binlog returns the index in `int64` type, which caused the issue of not able to write int64 into byte array.

This fix extracts the value from the enum using the index and writes that into the parquet.

Fixes #707

#### Step 1
```sql
CREATE TABLE customers_merchant_map (
  id BIGINT PRIMARY KEY AUTO_INCREMENT,
  merchant_id BIGINT NOT NULL,
  phone VARCHAR(50),
  whatsApp_opt_in ENUM('opt_in','require_opt_in','opt_out') DEFAULT NULL,
  loyalty_opt_in ENUM('YES','NO','AWAITING_CUSTOMER_INPUT') DEFAULT NULL,
  rcs_opt_in ENUM('opt_in','require_opt_in','opt_out') DEFAULT 'opt_in',
  split_group ENUM('A','X') DEFAULT NULL,
  is_active TINYINT(1) DEFAULT '1',
  created_on DATETIME DEFAULT CURRENT_TIMESTAMP
);
```
```sql
INSERT INTO customers_merchant_map (id, merchant_id, phone, whatsApp_opt_in, loyalty_opt_in, split_group) 
VALUES 
  (1, 100, '1234567890', NULL, NULL, NULL),
  (2, 100, '1234567891', NULL, NULL, NULL),
  (3, 100, '1234567892', 'opt_in', 'YES', 'A');
```

#### Step 2
Do Full Load

#### Step 3
```sql
UPDATE customers_merchant_map SET whatsApp_opt_in = 'opt_out';
```

#### Step 4
Run CDC

It will fail for parquet writer, and for iceberg writer, it will add numeric values (key indexes)

## Type of change

# How Has This Been Tested?

- [x] Parquet Writer
- [x] Iceberg Writer

### DIFF
--- a/pkg/binlog/filter.go
+++ b/pkg/binlog/filter.go
@@ -106,6 +106,9 @@ func convertRowToMap(row []interface{}, tableMap *replication.TableMapEvent, col
 				// for an update CDC event, the key of enum value is passed in binlog events which is always in int64
 				// during such a case, we need to find out the enum value of it from the index
 				if idx, isInt64 := val.(int64); isInt64 {
+					if idx == 0 {
+						return nil, fmt.Errorf("invalid ENUM value detected for column %s: index 0 indicates an error value (empty string) from MySQL", columns[i])
+					}
 					val = enumValues[idx-1]
 				}
 			}


pkg/binlog/filter.go
```diff
@@ -89,15 +89,28 @@ func (f ChangeFilter) FilterRowsEvent(ctx context.Context, e *replication.RowsEv
...
+
 	// TODO: float values from binlog are not always same as the output of select * from db,
 	// need to typecast it to the datatype of the column for consistency with db.
 	record := make(map[string]interface{})
 	for i, val := range row {
+		if tableMap.IsEnumColumn(i) && val != nil {
+			if enumValues, ok := enumMap[i]; ok {
+				// for an update CDC event, the key of enum value is passed in binlog events which is always in int64
+				// during such a case, we need to find out the enum value of it from the index
+				if idx, isInt64 := val.(int64); isInt64 {
```
vikaxsh, line 108: are we sure here it would be int64 always?

pkg/binlog/filter.go
```diff
@@ -89,15 +89,28 @@ func (f ChangeFilter) FilterRowsEvent(ctx context.Context, e *replication.RowsEv
...
+					val = enumValues[idx-1]
```
vikaxsh, line 112: please check once, is it possible to get index out of bound issue

pkg/binlog/filter.go
badalprasadsingh: yes

please check [go-mysql code snippet](https://github.com/go-mysql-org/go-mysql/blob/a07c974ef5a34a8d0d7dfb543652c4ba2dec90cf/replication/row_event.go#L1393-L1404)

pkg/binlog/filter.go
badalprasadsingh: thanks, nice observation

when strict SQL mode is off (e.g., `SET SESSION sql_mode = '';`), then SQL queries with inserting out of bounds ENUM values will succeed i.e., SQL program won't throw an error. Such a scenario should be handled as well. With the `if idx == 0` case, we are handling this.

PR #713, opened 2025-12-30
fix(arrow): iceberg writer improvements

badalprasadsingh:
# Description

Fixes:
- [x] `now()` in partition transformations 
- [x] `int` to `long` case
- [x] integration test

## Type of change

# How Has This Been Tested?

- [x] REST + Minio + Spark

# Screenshots or Recordings

## Documentation

### DIFF
--- a/destination/iceberg/arrow-writer/writer.go
+++ b/destination/iceberg/arrow-writer/writer.go
@@ -37,7 +37,7 @@ type Writer struct {
 	positionalDeleteWriter *RollingWriter
 	// stores the latest position of olake_id (thread level property)
 	// a data writer might flush multiple data files during cdc, to properly handle duplicate _olake_id values across multiple data files, we only empty it when the thread closes
-	olakeIDPosition   map[string]PositionalDelete
+	olakeIDPosition   map[string]PositionalDelete // should only be emptied while closing the thread
 	data              []types.RawRecord
 	equalityDeletes   []string
 	positionalDeletes []PositionalDelete
@@ -119,25 +119,29 @@ func (w *ArrowWriter) getRecordPartition(record types.RawRecord, olakeTimestamp 
 // getOrCreateWriter retrieves an existing writer for the partition key or creates a new one with all necessary rolling writers.
 func (w *ArrowWriter) getOrCreateWriter(ctx context.Context, pKey string, values []any) (*Writer, error) {
 	writer, exists := w.writers[pKey]
-	if exists {
-		return writer, nil
-	}
-
-	writer = &Writer{
-		olakeIDPosition: make(map[string]PositionalDelete),
+	if !exists {
+		writer = &Writer{
+			olakeIDPosition: make(map[string]PositionalDelete),
+		}
 	}
 
 	var err error
-	if writer.dataWriter, err = w.createWriter(ctx, pKey, values, *w.arrowSchema[fileTypeData], fileTypeData); err != nil {
-		return nil, err
+	if writer.dataWriter == nil {
+		if writer.dataWriter, err = w.createWriter(ctx, pKey, values, *w.arrowSchema[fileTypeData], fileTypeData); err != nil {
+			return nil, err
+		}
 	}
 
 	if w.upsertMode {
-		if writer.equalityDeleteWriter, err = w.createWriter(ctx, pKey, values, *w.arrowSchema[fileTypeEqualityDelete], fileTypeEqualityDelete); err != nil {
-			return nil, err
+		if writer.equalityDeleteWriter == nil {
+			if writer.equalityDeleteWriter, err = w.createWriter(ctx, pKey, values, *w.arrowSchema[fileTypeEqualityDelete], fileTypeEqualityDelete); err != nil {
+				return nil, err
+			}
 		}
-		if writer.positionalDeleteWriter, err = w.createWriter(ctx, pKey, values, *w.arrowSchema[fileTypePositionalDelete], fileTypePositionalDelete); err != nil {
-			return nil, err
+		if writer.positionalDeleteWriter == nil {
+			if writer.positionalDeleteWriter, err = w.createWriter(ctx, pKey, values, *w.arrowSchema[fileTypePositionalDelete], fileTypePositionalDelete); err != nil {
+				return nil, err
+			}
 		}
 	}
 
@@ -310,7 +314,6 @@ func (w *ArrowWriter) EvolveSchema(ctx context.Context, newSchema map[string]str
 	}
 
 	w.schema = newSchema
-	w.writers = make(map[string]*Writer)
 
 	if err := w.initialize(ctx); err != nil {
 		return fmt.Errorf("failed to reinitialize with evolved schema: %s", err)


destination/iceberg/arrow-writer/transforms.go
```diff
@@ -59,43 +61,67 @@ func hashString(s string) uint32 {
...
+	case "long":
+		v := val.(int64)
+		return fmt.Sprintf("%d", v), v, nil
+	case "float":
+		v := val.(float32)
+		return fmt.Sprintf("%g", v), v, nil
+	case "double":
+		v := val.(float64)
+		return fmt.Sprintf("%g", v), v, nil
+	case "string":
```
hash-data, line 81: any reason to typecast and then convert? diff between prev and current aproach ?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -215,7 +252,7 @@ func (w *ArrowWriter) checkAndFlush(ctx context.Context, writer *RollingWriter,
 			return fmt.Errorf("failed to upload parquet during flush: %s", err)
 		}
 
-		w.writers.Delete(fileType + ":" + partitionKey)
+		delete(w.writers, fileType+":"+partitionKey)
```
hash-data: for ":" let us create a function to get fileKey

destination/iceberg/arrow-writer/writer.go
```diff
@@ -124,6 +151,14 @@ func (w *ArrowWriter) extract(records []types.RawRecord) (map[string]*Partitione
 		}
 
 		if data[pKey] == nil {
+			var pValues []any
+			if len(w.partitionInfo) != 0 {
+				values, err := w.getRecordPartitionValues(rec, olakeTimestamp)
+				if err != nil {
+					return nil, nil, err
+				}
```
hash-data: let us discuss the reason for adding it

destination/iceberg/arrow-writer/writer.go
```diff
@@ -262,14 +299,9 @@ func (w *ArrowWriter) Close(ctx context.Context) error {
...
 func (w *ArrowWriter) completeWriters(ctx context.Context) error {
-	var err error
-
-	w.writers.Range(func(key, value interface{}) bool {
-		mapKey := key.(string)
-		writer, _ := value.(*RollingWriter)
-		if closeErr := writer.currentWriter.Close(); closeErr != nil {
-			err = fmt.Errorf("failed to close writer: %s", closeErr)
-			return false
+	for mapKey, writer := range w.writers {
```
hash-data:
```suggestion
	for pKey, writer := range w.writers {
```

destination/iceberg/arrow-writer/transforms.go
```diff
@@ -188,29 +209,68 @@ func ConstructColPath(valueStr, field, transform string) string {
...
+			pv.Value = &proto.ArrowPayload_FileMetadata_PartitionValue_LongValue{LongValue: v}
+		case float32:
+			pv.Value = &proto.ArrowPayload_FileMetadata_PartitionValue_FloatValue{FloatValue: v}
+		case float64:
+			pv.Value = &proto.ArrowPayload_FileMetadata_PartitionValue_DoubleValue{DoubleValue: v}
+		case string:
+			pv.Value = &proto.ArrowPayload_FileMetadata_PartitionValue_StringValue{StringValue: v}
+		case bool:
+			// Booleans stored as string "true"/"false" per Iceberg convention
+			pv.Value = &proto.ArrowPayload_FileMetadata_PartitionValue_StringValue{StringValue: fmt.Sprintf("%t", v)}
```
hash-data: what about timestamp as partition value?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -267,26 +261,19 @@ public void accumulateDataFiles(String threadId, Table table, String filePath,
...
+               String errorMsg = String.format("Thread %s: failed to accumulate data file %s: %s", threadId, filePath,
+                         e.getMessage());
                LOGGER.error(errorMsg, e);
                throw new RuntimeException(e);
           }
      }
 
      public void accumulateDeleteFiles(String threadId, Table table, String filePath, int equalityFieldId,
-               long recordCount, List<String> partitionValues) {
-          if (table == null) {
```
hash-data, line 335: any reason ?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -313,79 +298,31 @@ public void accumulateDeleteFiles(String threadId, Table table, String filePath,
...
-                    }
-               }
-               partitionData.set(i, typedValue);
+               ArrowPayload.FileMetadata.PartitionValue protoValue = partitionValues.get(i);
+               Object value = switch (protoValue.getValueCase()) {
+                    case INT_VALUE -> protoValue.getIntValue();
+                    case LONG_VALUE -> protoValue.getLongValue();
+                    case FLOAT_VALUE -> protoValue.getFloatValue();
+                    case DOUBLE_VALUE -> protoValue.getDoubleValue();
+                    case STRING_VALUE -> protoValue.getStringValue();
```
hash-data, line 419: timestamp not available

utils/testutils/test_utils.go
```diff
@@ -763,11 +792,14 @@ func (cfg *IntegrationTest) TestIntegration(t *testing.T) {
...
-										t.Fatalf("Iceberg Full load + CDC tests failed: %v", err)
-									}
-								})
+								for _, useArrowWriter := range []bool{false, true} {
+									writerType := utils.Ternary(useArrowWriter, "Arrow", "Legacy").(string)
+									t.Run(fmt.Sprintf("Iceberg (%s) Full load + CDC tests", writerType), func(t *testing.T) {
+										if err := cfg.testIcebergWriter(ctx, t, c, currentTestTable, useArrowWriter, cfg.testIcebergFullLoadAndCDC); err != nil {
+											t.Fatalf("Iceberg (%s) Full load + CDC tests failed: %v", writerType, err)
+										}
+									})
```
hash-data: looks little weired can you check other way

utils/typeutils/datatype.go
```diff
@@ -30,10 +30,11 @@ func TypeFromValue(v interface{}) types.DataType {
...
 	case reflect.Bool:
 		return types.Bool
-	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32,
-		reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32:
+	case reflect.Int8, reflect.Int16, reflect.Int32,
+		reflect.Uint8, reflect.Uint16, reflect.Uint32:
 		return types.Int32
-	case reflect.Int64, reflect.Uint64:
+	// on standard 64 bit systems, golang's int type is 64 bits
+	case reflect.Int, reflect.Int64, reflect.Uint, reflect.Uint64:
```
hash-data: how we checked this?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
badalprasadsingh: we don't need to consider the timestamp value as we are adding these as partition values to iceberg and timestamp value gets converted into its required int type based on iceberg transformations like day, hour, etc.

destination/iceberg/arrow-writer/transforms.go
badalprasadsingh: not required.

utils/typeutils/datatype.go
badalprasadsingh: well, according to the test script which produced this error we used to add `99991199` as a value.
```go
a := 99991199 // var a int
r2 := types.CreateRawRecord("2", map[string]any{"name": "rohan", "age": 1, "contact": 99991199, "email": nil}, "c", nil)
```

now, an `int` type can be both `int32` and `int64` in go, but in certain systems having it in `int64` makes it faster. Check this doc on Stackoverflow.

```md
The int, uint, and uintptr types are implementation-specific. On 64-bit systems, int is 64 bits.
```

we didn't face this issue on legacy writer as the java engine there itself used to create parquet files, here we first create arrow memory vector buffers (basically schema) and then write the value. Thus, receiving this error:
```
org.apache.iceberg.arrow.vectorized.VectorizedArrowReader.allocateVectorBasedOnOriginalType(VectorizedArrowReader.java:279)
```

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
badalprasadsingh: well, I couldn't personally think of a case where we will receive the table to null. As this is called during register and commit proto case, and we always either load or create the table before moving with any iceberg writer

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: done.

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: done.

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: from transforms logic, we need two types of transform values:
(a) for generating the partition key for file path, e.g., `2009-11` (kind of human readable string format), and,
(b) partition value to add in the iceberg table, e.g., `478` which goes to the manifests

so, technically, this code snippet helps it to extract the partition values only once per partition

destination/iceberg/arrow-writer/transforms.go
badalprasadsingh: previously we used to type case it towards the java side if you remember in a java function if you remember that's not good

now we do it towards the exact go side in transforms logic and send over proto in its exact type in the java server

utils/testutils/test_utils.go
badalprasadsingh: done.

badalprasadsingh:
Currently in `arrow-writer` both `_olake_timestamp` and `_cdc_timestamp` are in **microsecond** precision.

destination/iceberg/arrow-writer/writer.go
```diff
@@ -60,74 +83,83 @@ func New(ctx context.Context, partitionInfo []internal.PartitionInfo, schema map
...
-			pValues = values
+	for idx, rec := range records {
+		partition, err := w.getRecordPartition(rec, olakeTimestamp)
+		if err != nil {
+			return nil, nil, err
 		}
+		pKey := partition.Key
 
+		// Track deletes for upsert operations (d, u, c all need delete handling)
 		if rec.OperationType == "d" || rec.OperationType == "u" || rec.OperationType == "c" {
```
hash-data: where we are checking if it is append only or not ? as in append only we dont need this check

destination/iceberg/arrow-writer/writer.go
```diff
@@ -60,74 +83,83 @@ func New(ctx context.Context, partitionInfo []internal.PartitionInfo, schema map
...
-					PartitionValues: pValues,
+				deletes[pKey] = &DeleteTracker{
+					Partition:     partition,
+					DupePositions: make(map[string][]int64),
 				}
 			}
-			deletes[pKey].Records = append(deletes[pKey].Records, del)
+
+			dt := deletes[pKey]
+			if dt.DupePositions[rec.OlakeID] == nil {
```
hash-data: can we remove this dupe positions as we are just using to write in eq delete file, if we even add that object multiple time is that going to affect?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -136,88 +168,148 @@ func (w *ArrowWriter) extract(records []types.RawRecord) (map[string]*Partitione
...
+	for pKey, partitioned := range data {
+		if len(partitioned.Records) == 0 {
+			continue
+		}
+
+		record, err := createArrowRecord(partitioned.Records, w.allocator, w.arrowSchema[fileTypeData], w.stream.NormalizationEnabled(), olakeTimestamp)
+		if err != nil {
+			return fmt.Errorf("failed to create arrow record: %s", err)
+		}
+		defer record.Release()
```
hash-data: defer inside for loop is never a good idea. we have to create a function and then use defer

destination/iceberg/arrow-writer/writer.go
```diff
@@ -136,88 +168,148 @@ func (w *ArrowWriter) extract(records []types.RawRecord) (map[string]*Partitione
...
+
+		rw, err := w.getOrCreateWriter(ctx, partitioned.Partition, *record.Schema(), fileTypeData)
+		if err != nil {
+			return fmt.Errorf("failed to get or create writer for data: %s", err)
+		}
+
+		startRowCount := rw.currentRowCount
```
hash-data: do we need to create this variable?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -136,88 +168,148 @@ func (w *ArrowWriter) extract(records []types.RawRecord) (map[string]*Partitione
...
+			continue
+		}
+
+		record, err := createArrowRecord(partitioned.Records, w.allocator, w.arrowSchema[fileTypeData], w.stream.NormalizationEnabled(), olakeTimestamp)
+		if err != nil {
+			return fmt.Errorf("failed to create arrow record: %s", err)
+		}
+		defer record.Release()
+
+		rw, err := w.getOrCreateWriter(ctx, partitioned.Partition, *record.Schema(), fileTypeData)
```
hash-data: use proper names

destination/iceberg/arrow-writer/writer.go
```diff
@@ -136,88 +168,148 @@ func (w *ArrowWriter) extract(records []types.RawRecord) (map[string]*Partitione
...
+		}
+		rw.currentRowCount += record.NumRows()
+
+		// create positional deletes for duplicate _olake_ids
+		if w.upsertMode && deletes[pKey] != nil {
+			if err := w.writePositionalDeletes(ctx, deletes[pKey], rw, startRowCount, partitioned); err != nil {
+				return err
+			}
+		}
+
```
hash-data:
```suggestion
		if err := rw.currentWriter.WriteBuffered(record); err != nil {
			return fmt.Errorf("failed to write data record: %s", err)
		}

		// create positional deletes for duplicate _olake_ids
		if w.upsertMode && deletes[pKey] != nil {
			if err := w.writePositionalDeletes(ctx, deletes[pKey], rw, partitioned); err != nil {
				return err
			}
		}
		rw.currentRowCount += record.NumRows()

```

destination/iceberg/arrow-writer/writer.go
```diff
@@ -136,88 +168,148 @@ func (w *ArrowWriter) extract(records []types.RawRecord) (map[string]*Partitione
...
+		startRowCount := rw.currentRowCount
+
+		if err := rw.currentWriter.WriteBuffered(record); err != nil {
+			return fmt.Errorf("failed to write data record: %s", err)
+		}
+		rw.currentRowCount += record.NumRows()
+
+		// create positional deletes for duplicate _olake_ids
+		if w.upsertMode && deletes[pKey] != nil {
+			if err := w.writePositionalDeletes(ctx, deletes[pKey], rw, startRowCount, partitioned); err != nil {
```
hash-data: try to pass only relevant things

destination/iceberg/arrow-writer/writer.go
```diff
@@ -316,135 +401,198 @@ func (w *ArrowWriter) initialize(ctx context.Context) error {
...
+	if partitionKey == "" {
+		return basePath, nil
+	}
+
+	// Insert partition key into path: dir/file.parquet → dir/partitionKey/file.parquet
+	if idx := strings.LastIndex(basePath, "/"); idx > 0 {
+		return basePath[:idx] + "/" + partitionKey + "/" + basePath[idx+1:], nil
+	}
+
+	return basePath, nil
```
hash-data:
```suggestion
	if partitionKey != "" {
		// Insert partition key into path: dir/file.parquet → dir/partitionKey/file.parquet
        if idx := strings.LastIndex(basePath, "/"); idx > 0 {
            return basePath[:idx] + "/" + partitionKey + "/" + basePath[idx+1:], nil
        }
	}

	return basePath, nil
```

destination/iceberg/arrow-writer/writer.go
```diff
@@ -316,135 +401,198 @@ func (w *ArrowWriter) initialize(ctx context.Context) error {
...
-		partitionValues: partitionValues,
-	}, nil
+	basePath := resp.(*proto.ArrowIngestResponse).GetResult()
+
+	if partitionKey == "" {
+		return basePath, nil
+	}
+
+	// Insert partition key into path: dir/file.parquet → dir/partitionKey/file.parquet
+	if idx := strings.LastIndex(basePath, "/"); idx > 0 {
```
hash-data: is there any case where this will not exist but partition is provided?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -316,135 +401,198 @@ func (w *ArrowWriter) initialize(ctx context.Context) error {
...
+	}, nil)
+
 	return nil
 }
 
-func (w *ArrowWriter) getOrCreateWriter(partitionKey string, schema arrow.Schema, fileType string, partitionValues []string) (*RollingWriter, error) {
-	// differentiating data and delete file writers, "data:pk" : *writer, "delete:pk" : *writer
-	key := fileType + ":" + partitionKey
+func (w *ArrowWriter) getOrCreateWriter(ctx context.Context, partition Partition, schema arrow.Schema, fileType string) (*RollingWriter, error) {
+	key := writerKey(fileType, partition.Key)
```
hash-data: what if we can have partition and writer map where i can store pos, eq and data writer in one single instance?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -316,135 +401,198 @@ func (w *ArrowWriter) initialize(ctx context.Context) error {
...
-	currentBuffer := &bytes.Buffer{}
-	writerProps := parquet.NewWriterProperties(baseProps...)
+	kvMeta := make(metadata.KeyValueMetadata, 0)
+	_ = kvMeta.Append("iceberg.schema", w.fileschemajson[fileType])
 
-	arrowProps := pqarrow.NewArrowWriterProperties(
-		pqarrow.WithStoreSchema(),
-		pqarrow.WithNoMapLogicalType(),
-	)
+	if fileType == fileTypeEqualityDelete {
```
hash-data, line 488: in pos delete we dont need this right?

destination/iceberg/arrow-writer/utils.go
```diff
@@ -56,6 +64,103 @@ func getDefaultWriterProps() []parquet.WriterProperty {
...
+	}
+
+	baseWriter := file.NewParquetWriter(w, pqSchema.Root(),
+		file.WithWriterProps(props),
+		file.WithWriteMetadata(kvMeta))
+
+	return &parquetWriter{
+		wr:     baseWriter,
+		schema: arrSchema,
+		ctx:    pqarrow.NewArrowWriteContext(context.Background(), nil),
```
hash-data: context should be used from the parent.

destination/iceberg/arrow-writer/utils.go
```diff
@@ -56,6 +64,103 @@ func getDefaultWriterProps() []parquet.WriterProperty {
...
+	}
+
+	numRows := int(rec.NumRows())
+	for colIdx := 0; colIdx < int(rec.NumCols()); colIdx++ {
+		col := rec.Column(colIdx)
+		field := fw.schema.Field(colIdx)
+
+		cw, err := fw.rgw.Column(colIdx)
+		if err != nil {
+			return fmt.Errorf("failed to get column writer %d: %w", colIdx, err)
```
hash-data: for each err format use %s insted of %w 

if we are appending an error to other error then we use %w

destination/iceberg/arrow-writer/utils.go
```diff
@@ -56,6 +64,103 @@ func getDefaultWriterProps() []parquet.WriterProperty {
...
+		if field.Nullable {
+			defLevels = make([]int16, numRows)
+			for i := 0; i < numRows; i++ {
+				if col.IsNull(i) {
+					defLevels[i] = 0
+				} else {
+					defLevels[i] = 1
+				}
+			}
+		}
```
hash-data, line 129: need to check this logic and requirement of it

destination/iceberg/arrow-writer/writer.go
```diff
@@ -136,88 +168,148 @@ func (w *ArrowWriter) extract(records []types.RawRecord) (map[string]*Partitione
...
+// checkAndFlush checks if file size threshold is reached and flushes if needed.
+func (w *ArrowWriter) checkAndFlush(ctx context.Context, rw *RollingWriter, partitionKey string, fileType string) error {
 	// logic, sizeSoFar := actual File Size + current compressed data (RowGroupTotalBytesWritten())
 	// we can find out current row group's compressed size even before completing the entire row group
-	sizeSoFar := int64(writer.currentBuffer.Len()) + writer.currentWriter.RowGroupTotalBytesWritten()
-	targetFileSize := utils.Ternary(fileType == fileTypeDelete, targetDeleteFileSize, targetDataFileSize).(int64)
+	sizeSoFar := int64(rw.currentBuffer.Len()) + rw.currentWriter.RowGroupTotalBytesWritten()
+	targetSize := targetDataFileSize
+	if fileType == fileTypeEqualityDelete || fileType == fileTypePositionalDelete {
+		targetSize = targetDeleteFileSize
```
hash-data: use utils ternary at places where possible

destination/iceberg/arrow-writer/writer.go
```diff
@@ -316,135 +401,198 @@ func (w *ArrowWriter) initialize(ctx context.Context) error {
...
+				FileData: uploadData.FileData,
+				FilePath: filePath,
 			},
 		},
 	}
 
-	// Send upload request with timeout
-	uploadCtx, uploadCancel := context.WithTimeout(ctx, 3600*time.Second)
-	defer uploadCancel()
+	uploadCtx, cancel := context.WithTimeout(ctx, 3600*time.Second)
```
hash-data: let us make 3600 * time.second constant as we are using on many places

destination/iceberg/arrow-writer/writer.go
```diff
@@ -316,135 +401,198 @@ func (w *ArrowWriter) initialize(ctx context.Context) error {
...
+
+	pf := w.createdFiles[pKey]
+	switch uploadData.FileType {
+	case fileTypeData:
+		pf.DataFiles = append(pf.DataFiles, fileMeta)
+	case fileTypeEqualityDelete:
+		pf.EqDeleteFiles = append(pf.EqDeleteFiles, fileMeta)
+	case fileTypePositionalDelete:
+		pf.PosDeleteFiles = append(pf.PosDeleteFiles, fileMeta)
+	}
```
hash-data: confusing first you set to different as pointer and then update that below, let us use some simpler code block for this.

destination/iceberg/arrow-writer/writer.go
```diff
@@ -136,88 +168,148 @@ func (w *ArrowWriter) extract(records []types.RawRecord) (map[string]*Partitione
...
-			record.Release()
-			return fmt.Errorf("failed to get or create writer for data: %s", err)
+// suppose _olake_id appears n times, we create positional delete files for n-1 oldest rows, the latest goes for equality delete files
+func (w *ArrowWriter) writePositionalDeletes(ctx context.Context, dt *DeleteTracker, dataWriter *RollingWriter, startRowCount int64, partitioned *PartitionedRecords) error {
+	var posDeletes []PositionalDelete
+
+	for idx, rec := range partitioned.Records {
+		position := startRowCount + int64(idx)
+		if prevPos, exists := dataWriter.olakeIDPosition[rec.OlakeID]; exists {
+			// any previous occurrence becomes a positional delete, latest goes for equality delete
```
hash-data: latest also not goes for positional delete, olake id already gone at first time. comment might confuse contributors

destination/iceberg/arrow-writer/writer.go
hash-data: i checked while write we have upsert check but can we add it here?

badalprasadsingh: done.

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: done.

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: manually doing record.Release() instead of having a defer, as it take unnecessary stack memory allocations.

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: ok.

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: what should be the alternative of getOrCreateWriter?

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: yes

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: updated.

let's just not say the "latest _olake_id goes for equality delete", as it just requires _olake_id value so anyone can go for equality deletes

destination/iceberg/arrow-writer/utils.go
badalprasadsingh: This is required.

In Parquet format, definition levels encode whether a value is null or present at each level of a potentially nested schema.

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: We can. But, the current structure of getting direct access to any type of RollingWriter using this logic `fileType + ":" + partitionKey` works fine I suppose.

Please do tell me if you really want me to change it.

destination/iceberg/arrow-writer/transforms.go
```diff
@@ -188,29 +218,68 @@ func ConstructColPath(valueStr, field, transform string) string {
...
```
hash-data: are we sure to pass boolean as string ?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -60,162 +68,226 @@ func New(ctx context.Context, partitionInfo []internal.PartitionInfo, schema map
...
-		colType, ok := w.schema[field]
+	for _, pInfo := range w.partitionInfo {
+		colType, ok := w.schema[pInfo.Field]
 		if !ok {
-			return "", nil, fmt.Errorf("partition field %q does not exist in schema", field)
+			return "", nil, fmt.Errorf("partition field %q not in schema", pInfo.Field)
 		}
 
-		value, err := TransformValue(record.Data[field], transform, colType)
+		fieldValue := utils.Ternary(pInfo.Field == constants.OlakeTimestamp, olakeTimestamp, record.Data[pInfo.Field])
```
hash-data, line 100: where we are handling now()?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -60,162 +68,226 @@ func New(ctx context.Context, partitionInfo []internal.PartitionInfo, schema map
...
+
+		writer, exists := w.writers[pKey]
+		if !exists {
+			writer = &Writer{
+				olakeIDPosition: make(map[string]PositionalDelete),
 			}
-			pKey = key
-			pValues = values
+			w.writers[pKey] = writer
 		}
```
hash-data: if writer itself not exist why we are not init all instance here at once, if upsert mode enable init eq and pos as well.

destination/iceberg/arrow-writer/writer.go
```diff
@@ -60,162 +68,226 @@ func New(ctx context.Context, partitionInfo []internal.PartitionInfo, schema map
...
+					return err
+				}
+			}
+			if writer.positionalDeleteWriter == nil {
+				if writer.positionalDeleteWriter, err = w.createWriter(ctx, pKey, values, *w.arrowSchema[fileTypePositionalDelete], fileTypePositionalDelete); err != nil {
+					return err
+				}
+			}
+
+			filePosition := writer.dataWriter.currentRowCount + int64(idx)
```
hash-data: the index is of records array not writer data array (which will be written in datafile), is it correct ?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -60,162 +68,226 @@ func New(ctx context.Context, partitionInfo []internal.PartitionInfo, schema map
...
+		return nil, fmt.Errorf("failed to create new rolling writer after flush: %s", err)
+	}
 
-		if err := w.uploadFile(ctx, uploadData); err != nil {
-			return fmt.Errorf("failed to upload parquet during flush: %s", err)
-		}
+	newFilePath, err := w.allocateFilePath(ctx, partitionKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to allocate new file path after flush: %s", err)
+	}
```
hash-data, line 288: new writer means new position delete, means no old olake id should be maintained right?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -60,162 +68,226 @@ func New(ctx context.Context, partitionInfo []internal.PartitionInfo, schema map
...
-		if err != nil {
-			record.Release()
-			return fmt.Errorf("failed to get or create writer for data: %s", err)
-		}
-		if err := writer.currentWriter.WriteBuffered(record); err != nil {
+		if err := writer.dataWriter.currentWriter.WriteBuffered(record); err != nil {
 			record.Release()
-			return fmt.Errorf("failed to write data record: %s", err)
+
+			return fmt.Errorf("failed to write equality delete record: %s", err)
```
hash-data: is it equality delete writter?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -316,135 +389,188 @@ func (w *ArrowWriter) initialize(ctx context.Context) error {
...
+		pf.EqDeleteFiles = append(pf.EqDeleteFiles, fileMeta)
+	case fileTypePositionalDelete:
+		pf.PosDeleteFiles = append(pf.PosDeleteFiles, fileMeta)
+	}
 
 	return nil
 }
 
-func (w *ArrowWriter) fetchAndUpdateIcebergSchema(ctx context.Context) error {
+func (w *ArrowWriter) fetchfileschemajson(ctx context.Context) error {
```
hash-data:
```suggestion
func (w *ArrowWriter) fetchFileSchemaJson(ctx context.Context) error {
```

destination/iceberg/arrow-writer/writer.go
```diff
@@ -235,78 +307,79 @@ func (w *ArrowWriter) EvolveSchema(ctx context.Context, newSchema map[string]str
...
+				}
+				writer.equalityDeleteWriter = nil
+			}
+			if writer.positionalDeleteWriter != nil {
+				if err := w.flush(ctx, writer.positionalDeleteWriter, partitionKey); err != nil {
+					return err
+				}
+				writer.positionalDeleteWriter = nil
+			}
 		}
```
hash-data, line 378: are we not using record count here? can we remove that or any use case?

destination/iceberg/arrow-writer/utils.go
```diff
@@ -238,3 +342,83 @@ func appendValueToBuilder(builder array.Builder, val interface{}) error {
...
+			return nil, fmt.Errorf("failed to convert field %s: %s", field.Name, err)
+		}
+		nodes = append(nodes, node)
+	}
+
+	// creating root group node with name "table"
+	root, err := schema.NewGroupNode("table", parquet.Repetitions.Required, nodes, -1)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create root group node: %s", err)
+	}
```
hash-data, line 361: these function should be in file where we have written for parquet writer as well.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -313,79 +298,66 @@ public void accumulateDeleteFiles(String threadId, Table table, String filePath,
...
+                    deleteFileBuilder.withPartition(partitionData);
+                    LOGGER.debug("Thread {}: positional delete file scoped to partition with {} values", threadId,
+                              partitionValues.size());
+               } else {
+                    LOGGER.debug("Thread {}: positional delete file scoped to global (unpartitioned)", threadId);
+               }
+
+               DeleteFile deleteFile = deleteFileBuilder.build();
+               deleteFiles.add(deleteFile);
+               LOGGER.info("Thread {}: accumulated positional delete file {} (total delete files: {})",
```
hash-data: updates comments as well

destination/iceberg/arrow-writer/utils.go
badalprasadsingh: then should I create a `utils.go` file or `internal` package in `destination` package, right now? or when arrow writer for parquet is being created, then this can be moved out?

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: in case of now(), OLakeTimestamp becomes the partition field

```go
colName := strings.Replace(strings.TrimSpace(strings.Trim(match[1], `'"`)), "now()", constants.OlakeTimestamp, 1)
```

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: yes, you are right.

changed it to,
```go
filePosition := writer.dataWriter.currentRowCount + int64(len(writer.data)-1)
```

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: it is being used.

we pass the writer in flush, and from the rollingwriter struct we can access the filecount while creating the payload filemetadata

destination/iceberg/arrow-writer/transforms.go
badalprasadsingh: we only handle boolean in case of identity transform.

synced a source table with boolean column as the transform col (even with backwardness: legacy sync + arrow sync), it ends up storing boolean as bool only internally, no matter if we pass it as string

### arrow
```json
          "file_path": "s3a://warehouse/dz/arrow/boolean/data/is_active=1/20260120-1-5b1dcef4-46c3-4a18-b1ad-c45c2bb37716-00001.parquet",
          "file_format": "PARQUET",
          "partition": {
               "is_active": 1
          },
          
           "file_path": "s3a://warehouse/dz/arrow/boolean/data/is_active=0/20260120-1-5b1dcef4-46c3-4a18-b1ad-c45c2bb37716-00002.parquet",
          "file_format": "PARQUET",
          "partition": {
               "is_active": 0
          },
```

### legacy
```json
          "file_path": "s3a://warehouse/dz/legacy/boolean/data/is_active=1/20260120-1-6a4d67c3-505a-4ec8-b9ac-62efcf4bc75c-00001.parquet",
          "file_format": "PARQUET",
          "partition": {
               "is_active": 1
          },
          
          "file_path": "s3a://warehouse/dz/legacy/boolean/data/is_active=0/20260120-1-6a4d67c3-505a-4ec8-b9ac-62efcf4bc75c-00002.parquet",
          "file_format": "PARQUET",
          "partition": {
               "is_active": 0
          },
```

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: discussed

destination/iceberg/arrow-writer/writer.go
```diff
@@ -235,78 +318,79 @@ func (w *ArrowWriter) EvolveSchema(ctx context.Context, newSchema map[string]str
...
-	if err == nil {
-		w.writers.Range(func(key, _ interface{}) bool {
-			w.writers.Delete(key)
-			return true
-		})
+		// Clear the writer state but keep the partition entry
+		writer.data = nil
+		writer.equalityDeletes = nil
+		writer.positionalDeletes = nil
+		writer.olakeIDPosition = make(map[string]PositionalDelete)
```
hash-data: as we discussed, if we reinit this then we have to maintain order of files. So can you look into it?

hash-data: also add a note why we are maintaining this for the thread till end.

destination/iceberg/arrow-writer/writer.go
hash-data: for rw not function

badalprasadsingh:
with the introduction of arrow writer integration test, the connections ran parallely for both `legacy-writer` and `arrow-writer`, causing connection exhaustion

updated `docker-compose.yml`, by adding `postgres -c max_connections=300`

and run `spark-iceberg` only when,
```yaml
depends_on:
  minio:
    condition: service_started
  postgres:
    condition: service_healthy
```

PR #715, opened 2025-12-30
feat: add spec for source s3

vikaxsh:
# Description

- Added S3 driver spec with JSON/CSV/Parquet file format support
- Implements schema inference for S3 sources JSON and CSV using typeutils.Resolve instead of custom type inference logic, ensuring consistent type handling

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/pkg/parser/csv_test.go
+++ b/pkg/parser/csv_test.go
@@ -70,15 +70,15 @@ func TestInferColumnType_TimestampDetection(t *testing.T) {
 			description:  "Should fall back to String when not all values are timestamps",
 		},
 		{
-			name: "integer values",
+			name: "integer values treated as float",
 			sampleRows: [][]string{
 				{"123"},
 				{"456"},
 				{"789"},
 			},
 			columnIndex:  0,
-			expectedType: types.Int64,
-			description:  "Should detect Int64 for integer values",
+			expectedType: types.Float64,
+			description:  "Should detect Float64 for integer values when type inference was updated",
 		},
 		{
 			name: "float values",


pkg/parser/json.go
```diff
@@ -58,21 +59,18 @@ func (p *JSONParser) InferSchema(_ context.Context, reader io.Reader) (*types.St
...
 	}
 
-	// Collect all unique field names across samples
-	fieldTypes := make(map[string][]interface{})
-	for _, record := range sampleRecords {
-		for fieldName, value := range record {
-			fieldTypes[fieldName] = append(fieldTypes[fieldName], value)
-		}
+	if err := typeutils.Resolve(p.stream, sampleRecords...); err != nil {
+		return nil, fmt.Errorf("failed to resolve schema: %w", err)
```
hash-data: use proper error formatting

pkg/parser/csv.go
```diff
@@ -80,13 +87,26 @@ func (p *CSVParser) InferSchema(_ context.Context, reader io.Reader) (*types.Str
...
+				convertedValue := convertValueForInference(value)
+				record[headers[j]] = convertedValue
+			}
+		}
+
+		// Resolve schema one record at a time
+		if err := typeutils.Resolve(p.stream, record); err != nil {
+			return nil, fmt.Errorf("failed to resolve schema for record %d: %w", i, err)
+		}
+		recordCount++
```
hash-data: remove the counting logic, make it consistent w.r.t. other drivers

pkg/parser/json.go
```diff
@@ -58,21 +59,14 @@ func (p *JSONParser) InferSchema(_ context.Context, reader io.Reader) (*types.St
...
 
-	// Collect all unique field names across samples
-	fieldTypes := make(map[string][]interface{})
-	for _, record := range sampleRecords {
-		for fieldName, value := range record {
-			fieldTypes[fieldName] = append(fieldTypes[fieldName], value)
+	// Resolve schema one record at a time similar to MongoDB driver
+	for i, record := range sampleRecords {
+		if err := typeutils.Resolve(p.stream, record); err != nil {
+			return nil, fmt.Errorf("failed to resolve schema for record %d: %w", i, err)
```
hash-data: also can you make err format consistent throughout

pkg/parser/csv.go
```diff
@@ -136,9 +156,15 @@ func (p *CSVParser) StreamRecords(ctx context.Context, reader io.Reader, callbac
 				}
 				convertedValue, err := convertValue(value, fieldType)
```
hash-data, line 143: TODO: follow olake arch for type conversion

pkg/parser/csv.go
```diff
@@ -80,13 +87,26 @@ func (p *CSVParser) InferSchema(_ context.Context, reader io.Reader) (*types.Str
...
+			if j < len(headers) {
+				// Try to convert value to appropriate Go type for resolver inference
+				convertedValue := convertValueForInference(value)
+				record[headers[j]] = convertedValue
+			}
+		}
+
+		// Resolve schema one record at a time
+		if err := typeutils.Resolve(p.stream, record); err != nil {
+			return nil, fmt.Errorf("failed to resolve schema for record %d: %w", i, err)
```
hash-data: is it not an irony that is being added, converting values (by knowing the types)and then type detection?

pkg/parser/csv.go
```diff
@@ -136,9 +156,15 @@ func (p *CSVParser) StreamRecords(ctx context.Context, reader io.Reader, callbac
 				}
 				convertedValue, err := convertValue(value, fieldType)
 				if err != nil {
-					return fmt.Errorf("failed to convert value for field %s: %w", headers[i], err)
+					if errors.Is(err, errNullValue) {
+						// errNullValue indicates a valid null/empty value
+						record[headers[i]] = nil
+					} else {
```
hash-data, line 148: why detecting nil ?

pkg/parser/json.go
vikaxsh: could you explain more?

pkg/parser/csv.go
vikaxsh: where can i find the function to convert string to respective datatype in olake?

pkg/parser/csv.go
vikaxsh: i have added comment below, from where i am returning the error, similarly doing in mongo as well

pkg/parser/csv.go
vikaxsh: previously similar string based parsing logic were used for both type and value parsing, resolve function was implemented as per told by you

pkg/parser/csv.go
hash-data: is it required as i can see it is not

pkg/parser/csv.go
hash-data: yes i mentioned to use resolver but that does not mean to do it in such way, we have to make sure we are not repeating something or doing it in difficult way. can think about improving this.

pkg/parser/csv.go
hash-data: will discuss on it

pkg/parser/json.go
hash-data: %w -> %s

pkg/parser/csv.go
vikaxsh: for reading csv currently using encoding/csv libray which return a row as slice of string, we need to do research for csv parser

pkg/parser/csv.go
vikaxsh: let's discuss

PR #716, opened 2025-12-30
fix(MariaDB): add GTID event handling for CDC timestamp

### LINKED ISSUE #711: fix(MariaDB): GTID event structure with microsecond precision for _cdc_timestamp column
## Description

We show `_cdc_timestamp` till microsecond level precision for MySQL using **GTID** based logic. Need to have a similar approach for **MariaDB**

https://github.com/datazip-inc/olake/blob/8be694acf65c1a05fdf1b7aee8a251c570d3cee5/pkg/binlog/binlog.go#L119-L123

majiayu000:
Fixes #711

## Changes
- Add `MariadbGTIDEvent` case handler in binlog processing
- Use event header timestamp (second precision) since MariaDB GTID events don't include microsecond precision timestamps like MySQL

## Notes
Investigated MariaDB GTID event structure - it contains only `GTID`, `Flags`, and `CommitID` fields (no timestamp). The event header timestamp provides second-level precision, which is the best available option for MariaDB CDC timestamp consistency.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/binlog/binlog.go
+++ b/pkg/binlog/binlog.go
@@ -116,7 +116,10 @@ func (c *Connection) StreamMessages(ctx context.Context, client *sqlx.DB, callba
 					c.changeFilter.lastGTIDEvent = time.UnixMicro(int64(e.OriginalCommitTimestamp)) // #nosec G115 - timestamp value is always within int64 range
 				}
 
-				// TODO: Investigate MariaDB GTID event structure for microsecond timestamp support.
+			case *replication.MariadbGTIDEvent:
+				// MariaDB GTID events don't include microsecond precision timestamps like MySQL.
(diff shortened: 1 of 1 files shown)


badalprasadsingh:
Hey @majiayu000, your current implementation is incorrect. We need microsecond level precision for _cdc_timestamp column for MariaDB. Your current implementation only provides second level precision.

Also, when you are done please do an end to end testing and send a video of how you tested it. Please go through [OLake's contributing doc](https://olake.io/docs/community/contributing/) very carefully. Thanks for picking up the issue.

majiayu000:
@badalprasadsingh  Thanks for your review. I will check this pr. Before got a e2e test, I will close this pr.

PR #722, opened 2026-01-02
feat: column selection

### LINKED ISSUE #121: feat: column selection
**Overview**

This feature enables users to select specific columns from a schema during sync configuration. Only these columns are written to the destination table. It reduces cost, improves performance, simplifies downstream usage, and supports governance requirements while ensuring safe incremental and CDC-based ingestion into Apache Iceberg.

**Problem Statement**

Currently, syncs ingest all columns for a selected stream. This leads to:

- Excessive storage and compute costs for wide tables

- Increased compaction and metadata overhead

- Slower query planning and execution

- Compliance risks due to accidental ingestion of sensitive columns

As Iceberg adoption grows for CDC and streaming workloads, lack of column-level control becomes a significant scalability and governance bottleneck.

**Goals**
Enable column-level inclusion/exclusion per stream

**Non-Goals**
Column transformations (masking, hashing, renaming)

**Use Cases**

- As a data engineer, I want to ingest only required columns to reduce cost.

- As a data engineer, I want schema changes in unselected columns to not impact my sync.

- As a compliance owner, I want to ensure PII columns never reach the lake.

- As an analytics engineer, I want destination tables to remain stable and minimal.

**Proposed Solution**

**UI/UX Flow**

Introduce an additional parameter: **“Sync New Columns” toggle.** It will be a Stream Level Property.

Default Behaviour : ON

Toggle ON: Automatically select and sync any newly discovered columns.

Toggle OFF: Discover newly added columns but do not sync them unless the user explicitly enables the columns.

**Scenarios**

| Scenario                   | Toggle: Sync New Columns | Expected Behaviour                                                                 |
|----------------------------|--------------------------|------------------------------------------------------------------------------------|
| Specific Column Selection  | ON                       | Only selected columns sync to the destination;  newly added columns are auto-selected and synced |
| All Columns Selected       | ON                       | All columns are synced, including any newly added columns                          |
| Specific Column Selection  | OFF                      | Only the explicitly selected columns are synced                                    |
| All Columns Selected       | OFF                      | All existing columns are synced; newly added columns are not synced                |

**Edge Cases Handling :** 

All cursor field columns will be mandatorily selected and disabled from editing

All CDC operation columns will be mandatorily selected and disabled from editing

manish-singh-bisht:
# Description

Fixes #121 

## Type of change

# How Has This Been Tested?

ai disclosure
the spark scripts used in the following videos for visualizing the data were completely generated by a llm model

MONGODB
1)parquet-backfill
https://youtu.be/TMIH5RkY0f8
2)cdc icerberg
https://youtu.be/qXD1q6H_-cs
3)incremental sync
https://youtu.be/rQTyRtc4LJI
4)with sync_new_columns as true and false
https://youtu.be/4ndpc5fXG1Q
5) when some columns are selected
https://youtu.be/u4TEOrHf6v0
6) cdc mongodb
https://youtu.be/C92jmU0CFgY

MYSQL
1) CDC iceberg
https://youtu.be/R7w6NBJ2zEE
2) cdc ,backfill, parquet
https://youtu.be/z9XgW-h44Rw

POSTGRES
1) cdc icerberg
https://youtu.be/tpcgyytqrVQ

2) backfill ,cdc parquet
https://youtu.be/3BWLj1Onesg

ORACLE
1) backfill parquet
[ORACLE backfill -parquet.webm](https://github.com/user-attachments/assets/fff388c5-693a-4858-9f29-51554630f395)
2)  icerberg should be similar

S3
1) CDC parquet 
[s3 cdc parquet.webm](https://github.com/user-attachments/assets/7db57a3a-eeed-49a4-9b74-8908e85421e1)
2)  icerberg should be similar

KAFKA

couldn't setup kafka, but it should be similar to s3

## Documentation

## Related PR's (If Any):


nayanj98:
@manish-singh-bisht By when can we expect this PR to be completed?

manish-singh-bisht:
> @manish-singh-bisht By when can we expect this PR to be completed?

Hey @nayanj98 , will open it by Sunday this week

Thanks for the patience

nayanj98:
@manish-singh-bisht Just following up again by when can we expect this PR to be ready for review?

manish-singh-bisht:
> @manish-singh-bisht Just following up again by when can we expect this PR to be ready for review?

i am testing it, had some trouble setting few things, thus late

will open it when testing is done from my side

manish-singh-bisht:
@nayanj98  ready for review

nayanj98:
@manish-singh-bisht Great will assign someone as soon as possible.

types/columns.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+// GetSelectedColumns returns the selected columns
+func (sc *SelectedColumns) GetSelectedColumns() []string {
+	return sc.Columns
+}
+
+// setSelectedColumnsMap sets the selected columns map for the selected columns
+func (sc *SelectedColumns) setSelectedColumnsMap() {
+	sc.Map = make(map[string]struct{})
```
vishalm0509: here, can pre allocate memory since we already know the length of `sc.Columns` ?

types/columns.go
```diff
@@ -0,0 +1,299 @@
...
+	for _, col := range sc.Columns {
+		if _, exists := sc.Map[col]; exists {
```
vishalm0509: When can it exist in the map already ?

types/columns.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+}
+
+// GetSelectedColumnsMap returns the selected columns map
+func (sc *SelectedColumns) GetSelectedColumnsMap() map[string]struct{} {
+	if sc.Map == nil {
+		return nil
+	}
+	return sc.Map
+}
```
vishalm0509: Why can't we just return `sc.Map`

types/columns.go
```diff
@@ -0,0 +1,299 @@
...
+// 3. source defined primary key columns,
+// 4. system generated fields
+func (sc *SelectedColumns) ensureMandatoryColumns(oldStream, newStream *Stream) map[string]struct{} {
+	selectedMap := sc.GetSelectedColumnsMap()
+
+	// if the selected columns map is not set, set it
+	if selectedMap == nil {
+		sc.setSelectedColumnsMap()
+		selectedMap = sc.GetSelectedColumnsMap()
+	}
```
vishalm0509: When can `selectedMap` be nil ?

types/columns.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// Add system generated fields
+	systemFields := []string{constants.OlakeID, constants.OlakeTimestamp, constants.OpType}
```
vishalm0509: Can we add this to constants ?

types/columns.go
```diff
@@ -0,0 +1,299 @@
...
+			sc.Columns = append(sc.Columns, sysField)
+			selectedMap[sysField] = struct{}{}
+		}
+	}
+
+	// Add cursor fields for incremental sync
+	if oldStream.SyncMode == INCREMENTAL && oldStream.CursorField != "" {
+		primaryCursor, secondaryCursor := parseCursorField(oldStream.CursorField)
+		if primaryCursor != "" {
+			if _, exists := selectedMap[primaryCursor]; !exists {
```
vishalm0509: @vikaxsh should we log that we are adding a column that is removed by user notifying this is a mandatory column. Or should we fail the sync if removed. Anyways in UI we should disable unselecting.

types/columns.go
```diff
@@ -0,0 +1,299 @@
...
+	)
+
+	allSelected = true
+
+	newStream.Schema.Properties.Range(func(key, _ interface{}) bool {
+		schemaColumnCount++
+
+		colName, ok := key.(string)
+		if !ok {
+			allSelected = false
```
vishalm0509: since we are returning, why are we setting `allSelected=true`

types/columns.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	// set the selected columns map
+	metadata.SelectedColumns.setSelectedColumnsMap()
+
+	// ensure mandatory columns are included
+	metadata.SelectedColumns.ensureMandatoryColumns(oldStream, newStream)
+
+	// set the all selected flag
+	metadata.SelectedColumns.SetAllSelectedColumnsFlag(newStream)
```
vishalm0509: Since we are calling these three, we can maybe create a helper inside the function only. Not needed as of now, just commenting so that we can decided later. It's a design choice anyway

types/catalog.go
```diff
@@ -243,6 +271,7 @@ func GetStreamsDelta(oldStreams, newStreams *Catalog) *Catalog {
 					(oldMetadata.PartitionRegex != newMetadata.PartitionRegex) ||
 					(oldMetadata.Filter != newMetadata.Filter) ||
 					(oldMetadata.AppendMode != newMetadata.AppendMode) ||
+					(oldMetadata.SyncNewColumns != newMetadata.SyncNewColumns) ||
```
vishalm0509: @vikaxsh not so sure about this, please check. 

Should we clear destination when `selectedCol` change ?

types/type_schema.go
```diff
@@ -145,6 +145,28 @@ func (t *TypeSchema) ToIceberg() []*proto.IcebergPayload_SchemaField {
 
 	return icebergFields
 }
+
+// FilterSchemaBySelectedColumns filters a schema to only include selected columns.
+// Returns filtered schema if some columns are selected, otherwise returns the original schema if all columns are selected or no selection is provided.
+func FilterSchemaBySelectedColumns(schema *TypeSchema, selectedMap map[string]struct{}, allSelected bool) *TypeSchema {
+	if len(selectedMap) == 0 || allSelected {
+		return schema
+	}
```
vishalm0509: `allSelected` is already set based on `len == 0`. Should we check here again ?

vishalm0509:
@manish-singh-bisht quick question we currently apply column filtering at the driver level (Kafka, S3) and again at the destination level (Iceberg, Parquet). Can we centralize this in the abstract layer and remove it from the rest of the pipeline?

 Wanted to check if you’ve already explored or ruled out this approach or does it have flaws in it.

vishalm0509:
@manish-singh-bisht also, we have released 2 new drivers `DB2` and `MSSQL`, please confirm it works fine. Anyways the feature should be driver independent right

manish-singh-bisht:
> @manish-singh-bisht quick question we currently apply column filtering at the driver level (Kafka, S3) and again at the destination level (Iceberg, Parquet). Can we centralize this in the abstract layer and remove it from the rest of the pipeline?
> 
> Wanted to check if you’ve already explored or ruled out this approach or does it have flaws in it.

yes i did explore the central approach. 

**central approach** - get all the data from the source, do the filtering and then pass the data to destination
**driver level approach-** we fetch only the selected column from the source and let the source do all the work/optimization.some source support selection such as pgsql,mysql while source such as kafka,s3 maynot. 


this central approach adds to latency as we do filtering after fetching all data from source. if we go by driver level, the source that support selective selection will optimize the data collection themselves and we can ensure that we add as less latency as possible. however for sources that dont support selective selection such as kaka,s3 we do filtering after fetching all records.

regardless of the strategy above, we need destination level schema filtering to tell the destintaion what columns to create

@vishalm0509

manish-singh-bisht:
> @manish-singh-bisht also, we have released 2 new drivers `DB2` and `MSSQL`, please confirm it works fine. Anyways the feature should be driver independent right

i will test/add them, as the current approach is driver specific 

driver independent? with the current approach, no. if we go by central then yes it will be driver independent.

types/columns.go
manish-singh-bisht: fixed

types/columns.go
manish-singh-bisht: fixed

types/columns.go
manish-singh-bisht: removed, not required,

types/columns.go
manish-singh-bisht: removed allSelected variable

types/columns.go
manish-singh-bisht: hey went ahead with the idea,its better than the current, thanks,

types/type_schema.go
manish-singh-bisht: removed the redundant check

types/columns.go
manish-singh-bisht: there are other places in the codebase where they are used in similar way, also this slice doesnt have cdc timestamp field which is checked later in the method

types/columns.go
manish-singh-bisht: removed from here
added de-duplciation logic at diff place,at source of truth, 

it was protecting against duplicate columns added by user

manish-singh-bisht:
will add for new drivers, once we agree on the approach

vishalm0509:
> will add for new drivers, once we agree on the approach

@manish-singh-bisht  We are discussing about this, will get back once finalized.

vishalm0509:
Hey @manish-singh-bisht 

Flagging an issue we’ve noticed around CDC and `sync_new_columns`.

### What’s happening

If a new column is added **while a sync is running**, it doesn’t show up in CDC events in the same sync run, even when `sync_new_columns=true`.

This is because:
- Schema discovery runs **once at sync start**
- Column selection is finalized at that time
- CDC reads all columns from the binlog, but filters using that initial snapshot

So any column added after sync start gets filtered out.

### When this fails

- Column added during **backfill** → missing from CDC

Columns added before sync or between sync runs work fine.

### Proposed direction

When `sync_new_columns=true`, switch to an **exclusion-based approach**:
- Track `unselected_columns`
- Emit all columns except those explicitly excluded
- New columns are included automatically by default

This avoids relying on a single schema snapshot and lets CDC pick up new columns immediately.

Let me know if you see any edge cases we’re missing.

vishalm0509:
@manish-singh-bisht Right now this PR is applying column selection at the driver level.

We’d like to centralize this logic instead (abstract):
- Drivers should always read all source columns (SQL SELECT *, Mongo without $project, etc.).
- Column selection should happen once in the abstract layer, right before we build RawRecord

types/columns.go
```diff
@@ -0,0 +1,285 @@
...
+// Returns true if all columns are selected or no columns are selected, otherwise returns false
+func (sc *SelectedColumns) checkAllColumnsSelected(newStream *Stream) bool {
+	selectedMap := sc.GetSelectedColumnsMap()
+	if len(selectedMap) == 0 {
+		return true
+	}
+
+	var (
+		schemaColumnCount int
+	)
```
vishalm0509:
```suggestion
	var schemaColumnCount int
```

manish-singh-bisht: done

manish-singh-bisht:
> @manish-singh-bisht Right now this PR is applying column selection at the driver level.
> 
> We’d like to centralize this logic instead (abstract):
> 
>     * Drivers should always read all source columns (SQL SELECT *, Mongo without $project, etc.).
> 
>     * Column selection should happen once in the abstract layer, right before we build RawRecord

changed to central approach, will it be possible to share what went in the discussion to choose this approach would love to hear that!! @vishalm0509

manish-singh-bisht:
> Hey @manish-singh-bisht
> 
> Flagging an issue we’ve noticed around CDC and `sync_new_columns`.
> ### What’s happening
> 
> If a new column is added **while a sync is running**, it doesn’t show up in CDC events in the same sync run, even when `sync_new_columns=true`.
> 
> This is because:
> 
>     * Schema discovery runs **once at sync start**
> 
>     * Column selection is finalized at that time
> 
>     * CDC reads all columns from the binlog, but filters using that initial snapshot
> 
> 
> So any column added after sync start gets filtered out.
> ### When this fails
> 
>     * Column added during **backfill** → missing from CDC
> 
> 
> Columns added before sync or between sync runs work fine.
> ### Proposed direction
> 
> When `sync_new_columns=true`, switch to an **exclusion-based approach**:
> 
>     * Track `unselected_columns`
> 
>     * Emit all columns except those explicitly excluded
> 
>     * New columns are included automatically by default
> 
> 
> This avoids relying on a single schema snapshot and lets CDC pick up new columns immediately.
> 
> Let me know if you see any edge cases we’re missing.


[new col between cdc.webm](https://github.com/user-attachments/assets/f4725ab1-3d3e-4cf3-bbfb-9152b7301c13)

fixed

manish-singh-bisht:
note: new col that are synced when sync_new_columns is set to true, which were not selected by user will not be logged in the selected_columns fields in stream.json catalog automatically

types/columns.go
```diff
@@ -0,0 +1,285 @@
...
+	if allSelected {
+		return data
+	}
+
+	filtered := make(map[string]interface{})
+	if syncNewColumns {
+		// emit all columns except those that are unselected
+		// this ensures all columns that are new are selected by default
+		unSelectedMap := selectedCols.GetUnSelectedColumnsMap()
+		for key, value := range data {
```
vishalm0509: can we name it properly than key, value ?

destination/parquet/parquet.go
```diff
@@ -139,19 +139,20 @@ func (p *Parquet) Setup(_ context.Context, stream types.StreamInterface, schema
 		return p.schema, nil
 	}
 
-	if schema != nil {
-		fields, ok := schema.(typeutils.Fields)
-		if !ok {
-			return nil, fmt.Errorf("failed to typecast schema[%T] into typeutils.Fields", schema)
-		}
+	if fields, ok := schema.(typeutils.Fields); ok {
```
vishalm0509: Proper naming here

vishalm0509:
@manish-singh-bisht right now as I can see if new columns are added to the table they are not added to the `selected_columns` even on discover.

manish-singh-bisht:
> @manish-singh-bisht right now as I can see if new columns are added to the table they are not added to the `selected_columns` even on discover.

Yes that is the current behaviour 

The new columns will not be present in selected_columns automatically even on discover,

Do we want them to be present automatically in the selected_columns?

vishalm0509:
> > @manish-singh-bisht right now as I can see if new columns are added to the table they are not added to the `selected_columns` even on discover.
> 
> Yes that is the current behaviour
> 
> The new columns will not be present in selected_columns automatically even on discover,
> 
> Do we want them to be present automatically in the selected_columns?

Yes, not needed during sync. We need to handle this in the discover when we pass `--streams` as a flag (merge streams) if `sync_new_columns` flag is `true`

destination/parquet/parquet.go
vishalm0509: it's fine but in some places we use meaningful names and `ok` at others. Better to stick to one, let's go with meaningful. What do you say ?

types/columns.go
```diff
@@ -0,0 +1,285 @@
...
+//   - Specific columns selected: Only explicitly selected columns sync
+//   - All columns selected: All existing columns sync; newly added columns are NOT synced
+func FilterDataBySelectedColumns(data map[string]interface{}, stream StreamInterface) map[string]interface{} {
+	selectedCols := stream.Self().StreamMetadata.SelectedColumns
+	syncNewColumns := stream.Self().StreamMetadata.SyncNewColumns
+	allSelected := selectedCols.GetAllSelectedColumnsFlag()
+
+	if allSelected {
+		return data
+	}
```
vishalm0509: - Let’s say at the start of the sync we have selected columns A, B, C, D.

- During the full_refresh phase (which runs under repeatable_read and uses fixed chunk bounds), a new source column E is added. That new column won’t appear in the ongoing full_refresh snapshot, which is expected.

- However, when CDC starts after full_refresh, we’ll see the DDL that adds E. Since all columns were selected when the sync started, selectedCols.GetAllSelectedColumnsFlag() is true, so FilterDataBySelectedColumns returns the entire record (return data) regardless of sync_new_columns.

- This means that E will be included in CDC events even when sync_new_columns=false, whereas we should only bypass filtering like this when sync_new_columns=true.

Right ?

manish-singh-bisht: you are right, bug

vishalm0509:
> > @manish-singh-bisht Right now this PR is applying column selection at the driver level.
> > We’d like to centralize this logic instead (abstract):
> > ```
> > * Drivers should always read all source columns (SQL SELECT *, Mongo without $project, etc.).
> > 
> > * Column selection should happen once in the abstract layer, right before we build RawRecord
> > ```
> 
> changed to central approach, will it be possible to share what went in the discussion to choose this approach would love to hear that!! @vishalm0509

Good question. We did discuss driver-level projection vs a centralized approach.

While applying column selection at the driver level can give a small performance win (less data read from the source), the gain is marginal for most connectors and comes at the cost of higher complexity:

- Each driver needs to implement and maintain its own projection logic (SQL, Mongo, etc.).

- Behavior can easily diverge across drivers, making it harder to reason about correctness.

- Any future change to column selection semantics would need to be replicated in every driver.

By always reading full rows at the driver level and applying column selection once in the abstract layer (right before building RawRecord), we:

- Centralize the logic in one place

- Keep driver implementations simpler and more consistent

- Make future changes easier and safer

For our use case, that trade-off felt more valuable than the relatively small performance gain.

shubham19may:
@manish-singh-bisht the main reason to not reading from source side query is, most databases scan full tables when select with certain columns or all. So only thing we are saving is network bandwidth which doesn't add to a lot of performance gains.

vishalm0509:
@manish-singh-bisht let me know once all the requested changes resolved

manish-singh-bisht:
> @manish-singh-bisht let me know once all the requested changes resolved

yes

types/columns.go
manish-singh-bisht: resolved

destination/parquet/parquet.go
manish-singh-bisht: changed

types/columns.go
manish-singh-bisht: changed

types/columns.go
```diff
@@ -0,0 +1,329 @@
...
+	// automatically add newly discovered columns during discover command to the selected_columns list when sync_new_columns is true
+	if metadata.SyncNewColumns {
+		_, newAddedColumns := getColumnsDelta(oldSchema, newSchema)
+		for _, newCol := range newAddedColumns {
+			if _, exists := preservedSelectedColumnsMap[newCol]; !exists {
+				preservedSelectedColumns = append(preservedSelectedColumns, newCol)
+				preservedSelectedColumnsMap[newCol] = struct{}{}
+			}
+		}
+	}
```
vishalm0509: WE can add this check in the above check itself. 

Here: 
```go
	// when selectedColumns property is present and not empty
	var preservedSelectedColumns []string
	preservedSelectedColumnsMap := make(map[string]struct{})
	for _, previouslySelectedCol := range metadata.SelectedColumns.Columns {
		// check if the column exists in both old and new schemas
		_, existsInOld := oldSchema.Properties.Load(previouslySelectedCol)
		_, existsInNew := newSchema.Properties.Load(previouslySelectedCol)
		if existsInOld && existsInNew {
			// don't add duplicate columns
			if _, exists := preservedSelectedColumnsMap[previouslySelectedCol]; !exists {
				preservedSelectedColumns = append(preservedSelectedColumns, previouslySelectedCol)
				preservedSelectedColumnsMap[previouslySelectedCol] = struct{}{}
			}
		}
	}
```

--- 

also let's do it inline no need a separate getDelta function and also let's maintain a single slice `selected_cols` no `common` and `new`. 

--- 

i believe adding it to the same check will make it simpler

types/columns.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+		finalizeSelectedColumns()
+		return
+	}
+
+	// if the old and new schemas have same columns, so no need to check for presence in both old and new schemas
+	if schemasHaveSameColumns(oldSchema, newSchema) {
+		finalizeSelectedColumns()
+		return
+	}
```
vishalm0509: We can remove this check, the below check will handle this. It's just a minor optimization, we can remove the function as well

types/columns.go
```diff
@@ -0,0 +1,329 @@
...
+
+	if allSelected {
+		if !isCDC {
+			return data
+		}
+
+		if isCDC && syncNewColumns {
+			return data
+		}
+	}
```
vishalm0509, line 62:
```suggestion
	// Skip filtering when all columns are selected and (not CDC, or CDC with sync_new_columns).
	if allSelected && (!isCDC || syncNewColumns) {
		return data
	}
```

types/columns.go
manish-singh-bisht: done

types/columns.go
manish-singh-bisht: done

types/columns.go
manish-singh-bisht: done

types/columns.go
```diff
@@ -0,0 +1,274 @@
+package types
+
+import (
+	"github.com/datazip-inc/olake/constants"
+)
+
```
vikaxsh: since we already have a generic Set[T] implementation in 
types/set.go
 with operations like Intersection, Difference, Union, SubsetOf etc. we could potentially make this even more compact by leveraging it.

types/columns.go
```diff
@@ -0,0 +1,274 @@
...
+type SelectedColumns struct {
+	Columns            []string            `json:"columns"` // all columns that are selected
+	SelectedMap        map[string]struct{} `json:"-"`       // map of columns that are selected
+	UnSelectedMap      map[string]struct{} `json:"-"`       // map of columns that are unselected
+	AllColumnsSelected bool                `json:"-"`       // true if all columns in the schema are selected
+}
```
vikaxsh: for example:

```suggestion
type SelectedColumns struct {
    Columns            []string     `json:"columns"`
    selectedSet        *Set[string] `json:"-"`
    unSelectedSet      *Set[string] `json:"-"`
    AllColumnsSelected bool         `json:"-"`
}
func schemasHaveSameColumns(oldSchema, newSchema *TypeSchema) bool {
    oldSet := schemaColumnSet(oldSchema)
    newSet := schemaColumnSet(newSchema)
    return oldSet.SubsetOf(newSet) && newSet.SubsetOf(oldSet)
}
func getColumnsDelta(oldSchema, newSchema *TypeSchema) (*Set[string], *Set[string]) {
    oldSet := schemaColumnSet(oldSchema)
    newSet := schemaColumnSet(newSchema)
    return newSet.Intersection(oldSet), newSet.Difference(oldSet)
}

```
please check if is it possible to refactor other functions as well

types/columns.go
```diff
@@ -0,0 +1,274 @@
...
+//
+// - sync_new_columns=false:
+//   - Specific columns selected: Only explicitly selected columns sync
+//   - All columns selected: All existing columns sync; newly added columns are NOT synced
+func FilterDataBySelectedColumns(data map[string]interface{}, stream StreamInterface) map[string]interface{} {
+	selectedCols := stream.Self().StreamMetadata.SelectedColumns
+	syncNewColumns := stream.Self().StreamMetadata.SyncNewColumns
+	allSelected := selectedCols.GetAllSelectedColumnsFlag()
+	streamSyncMode := stream.GetSyncMode()
+	isCDC := streamSyncMode == CDC || streamSyncMode == STRICTCDC
```
vikaxsh: how can we we avoid fetching  for each record

types/columns.go
```diff
@@ -0,0 +1,274 @@
...
+
+	// Skip filtering when all columns are selected and (not CDC, or CDC with sync_new_columns).
+	if allSelected && (!isCDC || syncNewColumns) {
+		return data
+	}
+
+	filtered := make(map[string]interface{})
```
vikaxsh: can we consider in place deletion here instead of creating new map

types/catalog.go
```diff
@@ -38,12 +38,14 @@ type StatusRow struct {
...
-	Normalization  bool   `json:"normalization"`
-	Filter         string `json:"filter,omitempty"`
+	ChunkColumn     string           `json:"chunk_column,omitempty"`
+	PartitionRegex  string           `json:"partition_regex"`
+	StreamName      string           `json:"stream_name"`
+	AppendMode      bool             `json:"append_mode,omitempty"`
+	Normalization   bool             `json:"normalization"`
+	Filter          string           `json:"filter,omitempty"`
+	SelectedColumns *SelectedColumns `json:"selected_columns"`
+	SyncNewColumns  bool             `json:"sync_new_columns"`
```
vikaxsh: IMO, we should keep:
Config struct (persisted in catalog): just Columns []string and SyncNewColumns bool - user-provided inputs only

Runtime struct (computed on load): contains the derived selectedSet, unselectedSet

types/catalog.go
vikaxsh: For v0, we should not handle this automatically. If a user selects or changes columns, it should be treated as a conscious decision, and the destination should be cleared manually.

types/columns.go
```diff
@@ -0,0 +1,274 @@
...
+
+	finalizeSelectedColumns()
+}
+
+// ensureMandatoryColumns ensures that mandatory columns are always in SelectedColumns:
+// 1. cursor fields,
+// 2. CDC columns,
+// 3. source defined primary key columns,
+// 4. system generated fields
+func (sc *SelectedColumns) ensureMandatoryColumns(oldStream, newStream *Stream) map[string]struct{} {
```
vikaxsh: We should add a mandatory columns array to the catalog. This would allow the UI to disable editing for these columns, and we can safely assume they'll always be present (no need to ensure/validate)

types/catalog.go
manish-singh-bisht: the mandatory columns array(which will be added in the catalog as per this comment https://github.com/datazip-inc/olake/pull/722#discussion_r2746118024)  will be a part of which struct? since that is computed at runtime and is persisted in catalog @vikaxsh

types/columns.go
manish-singh-bisht: so this array persisting in catalog will be per selected stream right?these three fields are always present in every selected stream and sync mode **olake_timestamp**,**olake_id**,**op_type**  , so should we add them everytime in each selected stream or only those that are specific to that stream (cursor fields or/and primary keys) and sync mode and these three we can hardcode in ui codebase?

types/catalog.go
vikaxsh: it would be stream wise, mandatory column array for each stream parallel to selected column array

types/columns.go
manish-singh-bisht: done

types/catalog.go
manish-singh-bisht: done

types/columns.go
manish-singh-bisht: done

types/columns.go
manish-singh-bisht: done

types/columns.go
manish-singh-bisht: done

types/columns.go
manish-singh-bisht: thanks, done

manish-singh-bisht:
have also accounted for the new driver specific cdc columns that were introduced in a pr merged in staging

types/columns.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+// MergeSelectedColumns merges the selected columns based on the following rules:
+// 1. if selectedColumns property is not present or empty, initialize with columns from new schema
+// 2. if selectedColumns property is present and not empty, filter the selected columns to only those present in both old and new schemas
+// 3. if sync_new_columns is true, add newly discovered columns to the selected columns
+// 4. set the selected columns set
+// 5. set the unselected columns set
+// 6. set the all selected flag
+func MergeSelectedColumns(
```
vikaxsh, line 95: i think we can simplify MergeSelectedColumns with single set lookup
If no columns were previously selected: initializes with all columns from new schema
 Create one set(oldColumnsSet) from previously selected columns
Single iteration over new schema properties check if column not exist in oldColumnsSet then append in current columns array if 
sync_new_columns  bool in enabled
for example
```suggestion
var columns []string

	// no previous selection: initialize with all columns from new schema
	if metadata.SelectedColumns == nil || len(metadata.SelectedColumns.Config.Columns) == 0 {
		columns = collectColumnsFromSchema(newStream.Schema)
	} else {
		previouslySelectedSet := NewSet(metadata.SelectedColumns.Config.Columns...)

		// iterate new schema: retain previously selected columns, add new ones if sync_new_columns enabled
		newStream.Schema.Properties.Range(func(key, _ interface{}) bool {
			col, ok := key.(string)
			if !ok {
				return true
			}
			if previouslySelectedSet.Exists(col) || syncNewColumns {
				columns = append(columns, col)
			}
			return true
		})
	}

	metadata.SelectedColumns = &SelectedColumns{
		Config: Config{
			Columns:        columns,
			SyncNewColumns: syncNewColumns,
		},
	}
```

types/columns.go
```diff
@@ -0,0 +1,269 @@
...
+) {
+	oldSchema := oldStream.Schema
+	newSchema := newStream.Schema
+	oldSchemaColumnsSet := collectColumnsFromSchemaAsSet(oldSchema)
+	newSchemaColumnsSet := collectColumnsFromSchemaAsSet(newSchema)
+
+	finalizeSelectedColumns := func() {
+		metadata.SelectedColumns.Runtime.setSelectedColumnsSet(&metadata.SelectedColumns.Config)
+		metadata.SelectedColumns.Runtime.setUnSelectedColumnsSet(&metadata.SelectedColumns.Config, oldStream)
+		metadata.SelectedColumns.Runtime.SetAllSelectedColumnsFlag(&metadata.SelectedColumns.Config)
```
vikaxsh: what is need of this here?

protocol/sync.go
```diff
@@ -184,6 +184,9 @@ func classifyStreams(catalog *types.Catalog, streams []*types.Stream, state *typ
 				logger.Warnf("Skipping; Configured Stream %s not found in source", elem.ID())
 				return false
 			}
+
+			types.MergeSelectedColumns(&sMetadata, elem.Stream, source)
+
```
vikaxsh: why do we need this?

types/columns.go
```diff
@@ -0,0 +1,269 @@
...
+
+import (
+	"strings"
+
+	"github.com/datazip-inc/olake/constants"
+)
+
+// Config represents the user defined inputs
+// these are persisted in the catalog
+type Config struct {
```
vikaxsh:
```suggestion
type SelectedColumns struct {
```

types/columns.go
```diff
@@ -0,0 +1,269 @@
...
+	Columns        []string `json:"columns"`
+	SyncNewColumns bool     `json:"sync_new_columns"`
+}
+
+// Runtime represents the runtime state which is derived from the config
+type Runtime struct {
```
vikaxsh: can we remove this completely? instead computed on-demand 
can not think of any performance overhead  computed only thrice per stream (merge + filtering + schema creation)
we can keep functions like GetUnSelectedColumnsSet(), GetAllColumnsSelected() etc.

types/columns.go
```diff
@@ -0,0 +1,269 @@
...
+
+	finalizeSelectedColumns()
+}
+
+// ComputeMandatoryColumns computes the mandatory columns for a stream based on:
+// 1. System generated fields (OlakeID, OlakeTimestamp, OpType)
+// 2. Cursor fields (if incremental sync)
+// 3. Driver specific CDC columns (if CDC sync mode) and CDC Timestamp
+// 4. Source defined primary key columns
+func ComputeMandatoryColumns(oldStream, newStream *Stream) []string {
```
vikaxsh: We can remove this function. Earlier I suggested adding a mandatory column array, but after the recent changes, we finalized using the olakeColumn boolean field on each column instead.
The frontend will now rely on this property, so this logic is redundant.
since we are filtering column in memory we can let user unselect them as well

types/stream_configured.go
```diff
@@ -84,6 +85,11 @@ func (s *ConfiguredStream) Cursor() (string, string) {
 	return primaryCursor, secondaryCursor
 }
 
+// returns primary and secondary cursor
+func (s *ConfiguredStream) Cursor() (string, string) {
+	return parseCursorField(s.Stream.CursorField)
+}
+
```
vikaxsh: why do we need this change?

types/type_schema.go
```diff
@@ -165,6 +165,53 @@ func (t *TypeSchema) ToIceberg(onlyOlakeColumns bool) []*proto.IcebergPayload_Sc
...
+
+// FilterSchemaBySelectedColumns filters schema based on the following rules:
+// - sync_new_columns=true:
+//   - Specific schema columns selected: Only selected columns sync; newly added columns are automatically included
+//   - All schema columns selected: All columns sync, including newly added columns.
+//
+// - sync_new_columns=false:
+//   - Specific schema columns selected: Only explicitly selected columns sync
+//   - All schema columns selected: All existing columns sync; newly added columns are NOT synced
+func FilterSchemaBySelectedColumns(stream StreamInterface) *TypeSchema {
```
vikaxsh: can we Replace complex logic with simple array and Remove set operations, flags, and CDC checks
just call `selectedColumns.GetSelectedColumns()` to get pre-computed column list.
Handle all-columns case simply-  If `len(selectedCols) == 0`, return original schema (MergeSelectedColumns already handled the logic).
Direct iteration instead of Range with conditions - Loop through `selectedCols` array and copy properties directly using `schema.Properties.Load()`.

types/columns.go
manish-singh-bisht: removed

types/stream_configured.go
manish-singh-bisht: reverted, was being used in collection of mandatory columns method, since we have now removed the method for calculation of mandatory columns, so reverting this back

types/type_schema.go
manish-singh-bisht: hey @vikaxsh i was doing this suggested method previously, @vishalm0509 pointed out this bug https://github.com/datazip-inc/olake/pull/722#issuecomment-3788768277 after which i switched to current approach and i think we will need cdc checks and the logic unless i misunderstood something

vikaxsh: yeah @manish-singh-bisht , you are right we would track of unselected columns in case ofsync_new_columns=true
but independent of sync mode(incremental, cdc, strict_cdc).

types/columns.go
manish-singh-bisht: was used to set the runtime pre-computed data, removed since we are now computing on demand

types/columns.go
manish-singh-bisht: done

types/columns.go
manish-singh-bisht: removed,computing on demand now

types/columns.go
manish-singh-bisht: so if we do the above, say  we old=[a,b,c] and new as [a,b,c] but the user selected [a], then when sync new col is false , we will sync only [a] but when true we will be syncing [a,b,c] even though the user just selected [a] and there was no new column, so i think we would need the two sets of old and new streams

@vikaxsh

types/type_schema.go
manish-singh-bisht: so you are suggesting to remove this allColumnSelectedFlag completely?? https://github.com/datazip-inc/olake/pull/722#discussion_r2737284486

@vikaxsh

types/columns.go
vikaxsh: @manish-singh-bisht we can compute unselected columns from old catalog and use that in case of sync new columns are true

types/type_schema.go
vikaxsh: @manish-singh-bisht Since we are already providing a toggle to control syncing of new columns, we can determine whether to sync newly added columns based on that flag itself. I don’t see a strong use case for keeping allColumnSelectedFlag. What are your thoughts?

manish-singh-bisht: the allColumnSelectedFlag was used for early returns to skip data filtering when all columns were selected, that could have saved us some processing, but then this requires the cdc checks and other  conditions in the if statement.

@vikaxsh

vikaxsh: @manish-singh-bisht  When **Sync New Columns** is set to `true`, we should filter data only against the unselected columns. Since we are already providing a flag to control syncing of new columns, we should not automatically add new columns to the selected columns list, especially when the user has explicitly set **Sync New Columns** to `false`.

types/columns.go
manish-singh-bisht: hey @vikaxsh , i am not sure how computing unselected columns will help use here,

my logic of having two sets one for old and one for new is that we can get the diff of the schemas and add new columns when sync_new_columns is set to true while preserving previously selected columns, and if its set to false we ignore the new columns while preserving old selection

two sets

manish-singh-bisht: we are updating the selected columns list in case sync_new_columns is set to true, https://github.com/datazip-inc/olake/pull/722#issuecomment-3811952833

vikaxsh: @manish-singh-bisht We can exclude the previously unselected columns from the newly discovered catalog’s new columns.

protocol/sync.go
manish-singh-bisht: was needed previously to set the runtime data, now not needed

manish-singh-bisht:
hey @vikaxsh ,
there is an unwanted column `data` that comes when normalization is set to true,source mongodb ,destination parquet, i have observe this to be present in staging branch too.
can you confirm? thanks

manish-singh-bisht:
i'll fix the tests once we agree on the refactored code

vikaxsh:
> hey @vikaxsh , there is an unwanted column `data` that comes when normalization is set to true,source mongodb ,destination parquet, i have observe this to be present in staging branch too. can you confirm? thanks

Thanks for raising this issue. would be raising a PR to fix it shortly.

protocol/sync.go
```diff
@@ -184,6 +184,7 @@ func classifyStreams(catalog *types.Catalog, streams []*types.Stream, state *typ
 				logger.Warnf("Skipping; Configured Stream %s not found in source", elem.ID())
 				return false
 			}
+
```
vikaxsh: remove this

drivers/abstract/cdc.go
```diff
@@ -131,7 +132,15 @@ func (a *AbstractDriver) streamChanges(mainCtx context.Context, pool *destinatio
...
-		return writer.Push(ctx, types.CreateRawRecord(change.Data, olakeColumns))
+
+		filterDataBySelectedColumnsFn, exists := filterDataBySelectedColumnsFns[change.Stream.ID()]
+		if !exists {
+			filterDataBySelectedColumnsFn = types.FilterDataBySelectedColumns(change.Stream)
+			filterDataBySelectedColumnsFns[change.Stream.ID()] = filterDataBySelectedColumnsFn
+		}
+		filteredData := filterDataBySelectedColumnsFn(change.Data)
+
+		return writer.Push(ctx, types.CreateRawRecord(filteredData, olakeColumns))
```
vikaxsh:
```suggestion
	if _, exists := filterDataBySelectedColumnsFns[change.Stream.ID()]; !exists {
    filterDataBySelectedColumnsFns[change.Stream.ID()] = types.FilterDataBySelectedColumns(change.Stream)
}
filteredData := filterDataBySelectedColumnsFns[change.Stream.ID()](change.Data)
```

types/catalog.go
```diff
@@ -63,10 +64,17 @@ func GetWrappedCatalog(streams []*Stream, driver string) *Catalog {
 			Stream: stream,
 		})
 
+		selectedColumns := collectColumnsFromSchema(stream.Schema)
+		selectedCols := &SelectedColumns{
+			Columns:        selectedColumns,
+			SyncNewColumns: false,
```
vikaxsh: By default, this should be set to true to maintain backward compatibility.

types/columns.go
```diff
@@ -0,0 +1,117 @@
...
+			if previouslySelectedSet.Exists(col) || (metadata.SelectedColumns.SyncNewColumns && !oldSchemaCols.Exists(col)) {
+				columns = append(columns, col)
+			}
+			return true
+		})
+	}
+
+	metadata.SelectedColumns = &SelectedColumns{
+		Columns:        columns,
+		SyncNewColumns: metadata.SelectedColumns.SyncNewColumns,
```
vikaxsh: it should be true in case of 
metadata.SelectedColumns == nil

destination/writers.go
```diff
@@ -161,7 +161,14 @@ func (w *WriterPool) NewWriter(ctx context.Context, stream types.StreamInterface
...
 
-		output, err := writerThread.Setup(ctx, stream, streamArtifact.schema, opts)
+		// filter schema by selected columns if this is the first writer thread for this stream
+		// the first writer thread for this stream will have a TypeSchema
+		filteredSchema := streamArtifact.schema
+		if streamArtifact.schema == nil {
+			filteredSchema = types.FilterSchemaBySelectedColumns(stream)
+		}
+
+		output, err := writerThread.Setup(ctx, stream, filteredSchema, opts)
```
vikaxsh: Rather than modifying the current setup logic at this call site, it would be better to handle column exclusion inside ToParquet and ToIceberg in type_schema.go, which would be a cleaner place to apply this filtering.

types/columns.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+		if syncNewColumns {
+			// emit all columns except those that are unselected
+			// this ensures all columns that are new are selected by default
+			for col := range data {
+				if unselectedColumnsSet.Exists(col) {
+					delete(data, col)
+				}
+			}
```
vikaxsh: Wouldn’t it be better to iterate over the unselected map and delete those entries, instead of iterating over all keys?
```suggestion
		if syncNewColumns {
			// emit all columns except those that are unselected
			// this ensures all columns that are new are selected by default
			unselectedColumnsSet.Range(func(col string) {
				delete(data, col) // no-op if col not present
			})
		}
```

types/catalog.go
manish-singh-bisht: the issue description said to keep it false by default, changing to true  by default

types/columns.go
manish-singh-bisht: the issue description said to keep it false by default, 
changing to true

protocol/sync.go
manish-singh-bisht: removed

types/columns.go
manish-singh-bisht: changed

destination/writers.go
manish-singh-bisht: changed

types/type_schema.go
```diff
@@ -131,29 +131,79 @@ func (t *TypeSchema) ToParquet(onlyOlakeColumns bool) *parquet.Schema {
...
+	if selectedColumns != nil && len(selectedColumns.Columns) > 0 {
+		syncNewColumns := selectedColumns.SyncNewColumns
+
+		if syncNewColumns {
+			unselectedColumnsSet = GetUnSelectedColumnsSet(selectedColumns.Columns, stream.GetStream())
+		} else {
+			selectedColumnsSet = NewSet(selectedColumns.Columns...)
+		}
+	}
+
```
vikaxsh: it would be better to create a reusable function in columns.go can be use in ToIceberg and ToParquet
```
func IsSelectedColumn(stream StreamInterface) func(string) bool {
	selectedColsCfg := stream.Self().StreamMetadata.SelectedColumns
	if selectedColsCfg == nil || len(selectedColsCfg.Columns) == 0 {
		return func(string) bool { return true }
	}

	if selectedColsCfg.SyncNewColumns {
		unselected := GetUnSelectedColumnsSet(
			selectedColsCfg.Columns,
			stream.GetStream(),
		)
		return func(col string) bool {
			return !unselected.Exists(col)
		}
	}

	selected := NewSet(selectedColsCfg.Columns...)
	return func(col string) bool {
		return selected.Exists(col)
	}
}

```

drivers/mongodb/internal/testdata/test_streams.json
```diff
@@ -1 +1 @@
...
..."id_bool","_id","id","id_cursor","id_minkey"],"sync_mode":"cdc","destination_database":"mongodb:olake_mongodb_test","destination_table":"mongodb_test_table_olake","default_stream_properties":{"normalization":false,"append_mode":false}}},{"stream":{"name":"test_collection","namespace":"olake_mongodb_test","type_schema":{"properties":{"_cdc_resume_token":{"type":["string","null"],"destination_column_name":"_cdc_resume_token","olake_column":true},"_cdc_timestamp":{"type":["timestamp_micro","null"],"destination_column_name":"_cdc_timestamp","olake_column":true},"_olake_id":{"type":["string","null"],"destination_column_name":"_olake_id","olake_column":true},"_olake_timestamp":{"type":["timestamp_micro","null"],"destination_column_name":"_olake_timestamp","olake_column":true},"_op_type":{"type":["string","null"],"destination_column_name":"_op_type","olake_column":true}}},"supported_sync_modes":["incremental","cdc","strict_cdc","full_refresh"],"source_defined_primary_key":["_id"],"available_cursor_fields":[],"sync_mode":"cdc","destination_database":"mongodb:olake_mongodb_test","destination_table":"test_collection","default_stream_properties":{"normalization":false,"append_mode":false}}}]}
```
vikaxsh: update these as well with new stream fields

types/columns.go
```diff
@@ -0,0 +1,118 @@
...
+//   - All columns selected: All columns sync, including newly added columns.
+//
+// - sync_new_columns=false:
+//   - Specific columns selected: Only explicitly selected columns sync
+//   - All columns selected: All existing columns sync; newly added columns are NOT synced
+func FilterDataBySelectedColumns(stream StreamInterface) func(map[string]interface{}) map[string]interface{} {
+	selectedColumns := stream.Self().StreamMetadata.SelectedColumns
+	selectedColumnsSet := NewSet(selectedColumns.Columns...)
+	unselectedColumnsSet := GetUnSelectedColumnsSet(selectedColumns.Columns, stream.GetStream())
+	syncNewColumns := selectedColumns.SyncNewColumns
```
vikaxsh, line 66: there could be panicking on a nil pointer here

```suggestion
	selectedColumns := stream.Self().StreamMetadata.SelectedColumns

	// Backward compatibility:
	// Older catalogs (streams.json) may not have the selected_columns field at all.
	// In that case SelectedColumns will be nil after unmarshalling, so we return the data as is.
	if selectedColumns == nil {
		return func(data map[string]interface{}) map[string]interface{} {
			return data
		}
	}

	selectedColumnsSet := NewSet(selectedColumns.Columns...)
	unselectedColumnsSet := GetUnSelectedColumnsSet(selectedColumns.Columns, stream.GetStream())
	syncNewColumns := selectedColumns.SyncNewColumns
```

types/type_schema.go
```diff
@@ -147,17 +153,23 @@ func (t *TypeSchema) ToParquet(defaultColumns bool) *parquet.Schema {
...
 	t.Properties.Range(func(key, value interface{}) bool {
 		prop := value.(*Property)
+		colName := key.(string)
 		// skip non-olake columns if defaultColumns is set to true
 		if defaultColumns && !prop.OlakeColumn {
 			return true
 		}
+		if !isSelected(colName) {
+			return true
+		}
```
hash-data: merge conditions

types/columns.go
```diff
@@ -0,0 +1,148 @@
...
+		syncNewColumns = metadata.SelectedColumns.SyncNewColumns
+	}
+	metadata.SelectedColumns = &SelectedColumns{
+		Columns:        columns,
+		SyncNewColumns: syncNewColumns,
+	}
+}
+
+// IsSelectedColumn returns a function that checks if a column is selected
+func IsSelectedColumn(stream StreamInterface) func(string) bool {
```
hash-data: can be part of stream interface?

types/catalog.go
```diff
@@ -75,29 +83,39 @@ func GetWrappedCatalog(streams []*Stream, driver string) *Catalog {
...
 		selectedStreams := make(map[string][]StreamMetadata)
+
 		for namespace, metadataList := range oldCatalog.SelectedStreams {
 			_ = utils.ForEach(metadataList, func(metadata StreamMetadata) error {
-				_, exists := newStreams[fmt.Sprintf("%s.%s", namespace, metadata.StreamName)]
+				streamID := fmt.Sprintf("%s.%s", namespace, metadata.StreamName)
+				_, exists := newStreams[streamID]
+
 				if exists {
+					oldStream := oldStreams[streamID].Stream
```
hash-data: default columns should be selected always.

types/columns.go
```diff
@@ -0,0 +1,148 @@
...
+		return data
+	}
+}
+
+// MergeSelectedColumns merges the selected columns based on the following rules:
+// if selectedColumns property is not present or empty, initialize with columns from new schema
+// preserve previously selected columns
+// if sync_new_columns is true, add newly discovered columns to the selected columns
+func MergeSelectedColumns(
+	metadata *StreamMetadata,
```
hash-data: metadata of old stream or new stream,

types/columns.go
```diff
@@ -0,0 +1,148 @@
+package types
+
+type SelectedColumns struct {
```
hash-data: we dont have entity as column, we refer property as column, see if we can follow the same.

vikaxsh (approved):
hey @manish-singh-bisht ,
Thanks a lot for the contribution! Column selection was a crucial feature, and your work really helped move it forward.
Also appreciate how quickly you addressed and resolved the PR review comments, great job and thanks again! 🙌

PR #723, opened 2026-01-02
fix(mysql): float values from binlog are not always same as MySQL SELECT output

### LINKED ISSUE #706: fix: float values from binlog are not always same as the output of select * query of db
## Description

In MySQL, there can be a difference between the actual value stored in a column and the value displayed when running a `SELECT * FROM table` query.

For example, consider a `FLOAT` (float32) column in MySQL. A value inserted as `12345.678912345678` may be displayed as `12345.679` when queried using the MySQL CLI. This is due to MySQL’s default display formatting and rounding behavior for floating-point types.

However, when syncing this data using OLake’s Iceberg Writer or Parquet Writer, the value is captured as the actual stored value (`12345.678912345678`) rather than the rounded value shown in the MySQL CLI (`12345.679`).

This behavior should be corrected so that the synced data matches the value displayed by MySQL during a `SELECT` query, ensuring consistency between source reads and downstream data consumers.

---

https://github.com/datazip-inc/olake/blob/765752b2cae164eb5d73cef75365570ea8e22c4a/pkg/binlog/filter.go#L97-L101

---

###### This issue was generated by [todo-issue](https://github.com/DerJuulsn/todo-issue) based on a `TODO` comment in 765752b2cae164eb5d73cef75365570ea8e22c4a. It&#x27;s been assigned to @badalprasadsingh because they committed the code.

vimla01:
## Description

In MySql , `FLOAT` columns are stored as binary float32 values. However, binlog events can expose these values as higher-precision float64 representations, which causes OLake to emit values with extra precision compared to the column’s actual datatype semantics.

This leads to inconsistencies between values produced by OLake (from binlog reads) and values enforced by MySQL for `FLOAT` columns.

This change fixes the issue by enforcing MySQL `FLOAT` (float32) semantics on binlog values.
The implementation casts binlog float values back to float32 (removing the extra precision without performing any manual formatting, rounding, or interpretation)

Fixes #706

## Type of change

## How Has This Been Tested?

- Verified locally by simulating binlog float64 inputs and confirming that casting to float32 removes extra precision and aligns with MySQL `FLOAT` datatype behavior .

 **Testing Video:**  
[Float_mysqll.webm](https://github.com/user-attachments/assets/9cc99300-fafe-425e-9024-b511ee05c2b0)

The issue was reproduced locally to confirm its validity, after which the fix was implemented. Screenshots attached.

before

After fix

image

## Documentation

### DIFF
--- a/utils/typeutils/flatten.go
+++ b/utils/typeutils/flatten.go
@@ -60,7 +60,8 @@ func (f *FlattenerImpl) flatten(key string, value any, destination types.Record)
 
 	// Type switch is faster than reflection for known types
 	switch v := value.(type) {
-	case bool, int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64, float32, float64, string, time.Time:
+	// json.Number is included because Kafka driver uses decoder.UseNumber() for number handling
+	case bool, int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64, float32, float64, string, time.Time, json.Number:
 		destination[reformattedKey] = v
 	case []byte:
 		destination[reformattedKey] = string(v)
--- a/utils/typeutils/flatten_test.go
+++ b/utils/typeutils/flatten_test.go
@@ -1,6 +1,7 @@
 package typeutils
 
 import (
+	"encoding/json"
 	"testing"
 	"time"
 
@@ -498,6 +499,16 @@ func TestFlattenInternal(t *testing.T) {
 			},
 			expected: types.Record{
 				"all_numbers": `{"float32":3.14,"float64":3.14,"int":42,"int16":-42,"int32":423425,"int64":42425463733234,"int8":-42,"uint":42,"uint16":42,"uint32":42,"uint64":42,"uint8":42}`,
+			},
+			expectError: false,
+		},
+		// replicates scenario when Kafka driver parses JSON with decoder.UseNumber() for number handling
+		{
+			name:  "json.Number from Kafka UseNumber",
+			key:   "json_number",
+			value: json.Number("12345678901234567890"),
+			expected: types.Record{
+				"json_number": json.Number("12345678901234567890"),
 			},
 			expectError: false,
 		},


nayanj98:
@vimla01 Please rebase your PR to staging instead of master. Also please add a small video of you showing how you have tested your work, it makes easier for us to know you are in the right direction.

vimla01:
I’ve added a short testing video to the PR description. Please let me know if anything else is needed. @nayanj98 

> @vimla01 Please rebase your PR to staging instead of master. Also please add a small video of you showing how you have tested your work, it makes easier for us to know you are in the right direction.

Akshay-datazip:
hi @vimla01 can you make the video access "anyone with the link" 
As right now it has access issues

vimla01:
> hi @vimla01 can you make the video access "anyone with the link" As right now it has access issues

 I’ve updated the video permissions to “Anyone with the link”
 Please let me know if you’re able to access it now @Akshay-datazip

nayanj98:
@vimla01 Apologies for the delay, just checked your video, it seems you have uploaded a video of running your test file. If you can read the description of the issue and make a video replicating the same workflow with your changes that would be helpful. Also you can check on iceberg by running `SELECT * FROM table` query and check if it is fixed.

vimla01:
> @vimla01 Apologies for the delay, just checked your video, it seems you have uploaded a video of running your test file. If you can read the description of the issue and make a video replicating the same workflow with your changes that would be helpful. Also you can check on iceberg by running `SELECT * FROM table` query and check if it is fixed.

Thanks for the feedback.  @nayanj98 
I’ve uploaded a new video reproducing the exact workflow described in the issue with my changes applied. 
I confirmed the bug locally before fixing it and verified the fix on Iceberg using a 
`SELECT * FROM  table query. `
Please let me know if further changes are required @Akshay-datazip

nayanj98:
@vimla01 I am assigning @ImDoubD-datazip to review your PR.

ImDoubD-datazip:
@vimla01 , Hi can you also provide video for testing this in S3 Parquet as destination?

drivers/mysql/internal/config.go
```diff
@@ -22,6 +22,7 @@ type Config struct {
 	MaxThreads    int              `json:"max_threads"`
 	RetryCount    int              `json:"backoff_retry_count"`
 	SSHConfig     *utils.SSHConfig `json:"ssh_config"`
+	FloatFormat   string           `json:"float_format"`
```
ImDoubD-datazip: we don't want to take an extra input from user regarding this. However it is visible in database, similar to that way it should be visible in the destination

drivers/mysql/internal/config.go
```diff
@@ -88,5 +89,14 @@ func (c *Config) Validate() error {
 		c.RetryCount = constants.DefaultRetryCount // Reasonable default for retries
 	}
 
+	// Default float format
+	if c.FloatFormat == "" {
+		c.FloatFormat = "raw"
+	}
+
+	if c.FloatFormat != "raw" && c.FloatFormat != "display" {
```
ImDoubD-datazip: also how did you come up with these names raw & display? Before doing all these config level changes, we have to clarify with the product teams, which is basically @nayanj98 here. Have you clarified this with him?

drivers/mysql/internal/mysql.go
```diff
@@ -226,6 +226,22 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
...
+            }
+      	f32 := float32(v)
+     	str := strconv.FormatFloat(float64(f32), 'f', -1, 32)
+
+		finalVal, err := strconv.ParseFloat(str, 64)
+           if err != nil {
+               return nil, err
+           }
+		return finalVal, nil
+   	}
```
ImDoubD-datazip: can we do parsing in reformat.go file, because that is where the type formatting takes place. you can see the typeutils.reformatValue is called. check there once. Also make sure this change must not hinder other source drivers behaviour.

docker-compose.yml
```diff
@@ -8,4 +8,25 @@ services:
...
+    healthcheck:
+      test: ["CMD", "mysqladmin", "ping", "-h", "localhost"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    volumes:
+      - mysql-data:/var/lib/mysql
+
+volumes:
+  mysql-data:
```
ImDoubD-datazip: why have you added mysql docker compose in olake's docker compose file? please remove this.

ImDoubD-datazip:
@vimla01 reviewed.

drivers/mysql/internal/config.go
vimla01: I chose the names `raw` and `display` only as a placeholder to distinguish the two behaviors, but I agree that naming and introducing config-level changes should be aligned with product decisions.

I haven’t discussed this with @nayanj98 yet.

docker-compose.yml
vimla01: Sure, I’ll remove it.

drivers/mysql/internal/config.go
vimla01: Understood, I’ll handle it without adding any extra user input.

drivers/mysql/internal/mysql.go
vimla01: sure,I’ll move this parsing logic to reformat.go using typeutils.reformatValue and ensure it doesn’t impact other source drivers.

utils/typeutils/reformat.go
```diff
@@ -476,7 +476,15 @@ func ReformatFloat32(v interface{}) (float32, error) {
...
 	case float64:
-		return float32(v), nil
+    	f32 := float32(v)
+    	str := strconv.FormatFloat(float64(f32), 'f', -1, 32)
+
+    	finalVal, err := strconv.ParseFloat(str, 32)
+    	if err != nil {
+        	return float32(0), err
+    	}
+   		return float32(finalVal), nil
```
ImDoubD-datazip: shouldn't this be specific for mysql, also in mysql only in case where float comes from binlog. only in that case. here it will change for other drivers as well.

drivers/mysql/internal/config.go
```diff
@@ -87,6 +87,6 @@ func (c *Config) Validate() error {
 	if c.RetryCount <= 0 {
 		c.RetryCount = constants.DefaultRetryCount // Reasonable default for retries
 	}
-
+
```
ImDoubD-datazip: please remove extra spaces

drivers/mysql/internal/mysql.go
```diff
@@ -226,7 +226,7 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
 			return typeutils.ReformatGeoType(value)
 		}
 	}
-
+
```
ImDoubD-datazip: please remove extra space

ImDoubD-datazip:
@vimla01 After changes can you provide a video of final testing, by running olake for mysql cdc sync?

vimla01:
> @vimla01 After changes can you provide a video of final testing, by running olake for mysql cdc sync?

@ImDoubD-datazip Sure, after the changes I’ll run OLake for MySQL CDC sync and share a final testing video.

vimla01:
Hi @ImDoubD-datazip 

This is now MySQL CDC (binlog) only and does not affect other drivers.
Verified via MySQL CDC - S3 Parquet sync.

🎥 Video: 
[float.webm](https://github.com/user-attachments/assets/4870efd8-9811-4999-84f8-42ab30084633)

utils/typeutils/reformat.go
```diff
@@ -20,6 +20,7 @@ type StringInterface interface {
 	String() string
 }
 
+var EnableMySQLBinlogFloatFix = false
```
ImDoubD-datazip: this is not required

drivers/mysql/internal/cdc.go
```diff
@@ -66,6 +67,11 @@ func (m *MySQL) PreCDC(ctx context.Context, streams []types.StreamInterface) err
 }
 
 func (m *MySQL) StreamChanges(ctx context.Context, _ types.StreamInterface, OnMessage abstract.CDCMsgFn) error {
+
+	typeutils.EnableMySQLBinlogFloatFix = true
+	defer func() {
+		typeutils.EnableMySQLBinlogFloatFix = false
+	}()
```
ImDoubD-datazip: not required.

docker-compose.yml
```diff
@@ -8,4 +8,4 @@ services:
       --destination /mnt/config/writer.json
       --state /mnt/config/state.json
     volumes:
-      - ./drivers/mongodb/examples:/mnt/config
\ No newline at end of file
+      - ./drivers/mongodb/examples:/mnt/config
```
ImDoubD-datazip: revert changes

drivers/mysql/internal/config.go
```diff
@@ -87,6 +87,5 @@ func (c *Config) Validate() error {
 	if c.RetryCount <= 0 {
 		c.RetryCount = constants.DefaultRetryCount // Reasonable default for retries
 	}
-
```
ImDoubD-datazip, line 90: revert changes

drivers/mysql/internal/mysql.go
```diff
@@ -226,7 +226,6 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
 			return typeutils.ReformatGeoType(value)
 		}
 	}
-
```
ImDoubD-datazip, line 236: revert changes

utils/typeutils/reformat.go
```diff
@@ -462,7 +463,21 @@ func ReformatFloat32(v interface{}) (float32, error) {
...
+		finalVal, err := strconv.ParseFloat(str, 32)
+		if err != nil {
+			return float32(0), err
+		}
+		return float32(finalVal), nil
+	}
+
+	// Default behavior for all other drivers
+	return float32(v), nil
+
```
ImDoubD-datazip: is this really required? since we need the change in binlogs providing float value as float64 in some cases, shouldn't we do changes in filter.go in `/olake/pkg/binlog/filter.go` , and just add this there and converter is called?

```go
if columnTypes[i] == "FLOAT" {
			if f64, ok := convertedVal.(float64); ok {
				convertedVal = float32(f64)
			}
		}

```

Please test thoroughly, provide video for the same if working fine. The float32 must come as float32, float64 based data must come as float64.

vimla01: @ImDoubD-datazip I’ve implemented the changes you suggested.
Here’s the testing video :
https://drive.google.com/file/d/1MvNZg8sPGkvMK02PJAhaLjVtJOOPMtuY/view?usp=sharing

ImDoubD-datazip:
Hey @vimla01 , I tested this. With your previous changes, and current changes as well, issue still persisted.

Database has this:
Screenshot 2026-01-20 at 4 49 56 PM



After sync in Parquet (check value for price_float) : 

```
{"price_float": 12.344342, "_cdc_timestamp": 1768907417332782, "_olake_id": "20", "_olake_timestamp": 1768907436089775, "_op_type": "c", "created_date": 1768559400000000, "created_time": "15:45:30", "created_timestamp": 1768579200000000, "id": 20, "id_bigint": 922337203685477580, "id_bigint_unsigned": 1844674407370955161, "id_bit": 1, "id_int": -123456, "id_int_unsigned": 123456, "id_integer": 1000, "id_integer_unsigned": 2000, "id_mediumint": 50000, "id_mediumint_unsigned": 60000, "id_smallint": -32000, "id_smallint_unsigned": 32000, "id_tinyint": -100, "id_tinyint_unsigned": -56, "is_active": 1, "json_data": "{\"count\":5,\"key1\":\"value1\"}", "long_varchar": "This is a mediumtext column acting as long varchar", "name_bool": 1, "name_char": "Duke", "name_enum": "Medium", "name_longtext": "Long text content stored in longtext column", "name_mediumtext": "Medium text content here", "name_text": "This is a TEXT column", "name_tinytext": "Duke Dhal", "name_varchar": "Sample varchar value", "price_decimal": 9999.99, "price_double": 12345.6789234324, "price_double_precision": 9876.543214234325,  "price_numeric": 4567.89, "price_real": 54321.123}
```

As it comes in full refresh or normal select *, exact same value must come in destination during cdc sync as well. This issue is not solved yet, deeper research is required.

vimla01:
Hey @ImDoubD-datazip 
I dug into this further. The mismatch is expected due to different read paths.

- Full refresh / normal SELECT reads data through MySQL’s SQL layer.
MySQL formats and rounds FLOAT values before showing them in query results.
- CDC (state enabled) reads directly from the MySQL binlog, which contains the raw IEEE-754 stored value for FLOAT (no formatting, no rounding).

This behavior is consistent with other CDC systems (e.g. Debezium / Airbyte CDC), which also emit stored values rather than SELECT display output. Re-applying MySQL’s display formatting inside CDC would mean manually interpreting / rounding the data, which is lossy and changes data semantics.
Could you please advise on what you’d prefer as the next step here?

ImDoubD-datazip:
@vimla01 how did you tested for airbyte/debezium? Like can you provide video around it?

ImDoubD-datazip:
Hi @vimla01 , we have decided to close this issue. We will keep this as a note. Thanks for the contribution. One thing you can do is, wherever this todo is in that file remove the todo, add it as note there.

Thanks.

vimla01:
> @vimla01 how did you tested for airbyte/debezium? Like can you provide video around it?

This was based on their documented CDC behavior :
https://debezium.io/documentation/reference/stable/connectors/mysql.html

vimla01:
> Hi @vimla01 , we have decided to close this issue. We will keep this as a note. Thanks for the contribution. One thing you can do is, wherever this todo is in that file remove the todo, add it as note there.
> 
> Thanks.

Sounds good , I’ll remove the TODO and add it as a note in the file as suggested.

ImDoubD-datazip:
LGTM

PR #743, opened 2026-01-09
fix: evaluate idle checkpoint after every emitted change

vikaxsh:
# Description

Evaluate idle checkpoint after each processed change instead of only on empty change stream iterations, enabling faster and deterministic CDC termination when caught up.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/drivers/mongodb/internal/cdc.go
+++ b/drivers/mongodb/internal/cdc.go
@@ -101,7 +101,7 @@ func (m *Mongo) StreamChanges(ctx context.Context, stream types.StreamInterface,
 		}
 
 		// Check boundary AFTER emitting
-		if err := m.handleIdleCheckpoint(ctx, cursor, stream, lastOplogTime); err != nil {
+		if err := m.handleStreamCatchup(ctx, cursor, stream, lastOplogTime); err != nil {
 			if errors.Is(err, ErrIdleTermination) {
 				// graceful termination requested by helper
 				logger.Infof("change stream %s caught up to cluster opTime; terminating gracefully", stream.ID())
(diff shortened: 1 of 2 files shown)


vishalm0509:
`/drivers/mongodb/internal/cdc.go` line `132`

```go
	// check pointing post batch resume token
	m.cdcCursor.Store(stream.ID(), token)
```

Since we are setting `cdcStore` in `handleChangeDoc` method, do we need to set it in `handleIdleCheckpoint` method as well since `hasNext` is false, then there is no update in cursor value right. 

I think it's sufficient to set only in `handleChangeDoc` method. 

Can you verify once

vikaxsh:
> `/drivers/mongodb/internal/cdc.go` line `132`
> 
> ```go
> 	// check pointing post batch resume token
> 	m.cdcCursor.Store(stream.ID(), token)
> ```
> 
> Since we are setting `cdcStore` in `handleChangeDoc` method, do we need to set it in `handleIdleCheckpoint` method as well since `hasNext` is false, then there is no update in cursor value right.
> 
> I think it's sufficient to set only in `handleChangeDoc` method.
> 
> Can you verify once

in `handleChangeDoc` we updating cursor in case of hasNext true while in `handleIdleCheckpoint` we are updating every time

types/data_types.go
```diff
@@ -71,6 +71,9 @@ type RawRecord struct {
 }
 
 func CreateRawRecord(olakeID string, data map[string]any, operationType string, cdcTimestamp *time.Time) RawRecord {
+	if data == nil {
```
hash-data: let us debug why coming nil this and remove it

drivers/mongodb/internal/cdc.go
```diff
@@ -92,32 +89,35 @@ func (m *Mongo) StreamChanges(ctx context.Context, stream types.StreamInterface,
...
+			if errors.Is(err, ErrIdleTermination) {
+				// graceful termination requested by helper
+				logger.Infof("change stream %s caught up to cluster opTime; terminating gracefully", stream.ID())
+				return nil
+			}
 			return err
 		}
+
+		if !hasNext {
+			// Wait before for a brief pause before the next iteration of the loop
```
hash-data, line 114: add comment about sharding issue

drivers/mongodb/internal/cdc.go
```diff
@@ -92,32 +89,35 @@ func (m *Mongo) StreamChanges(ctx context.Context, stream types.StreamInterface,
...
 				return err
 			}
-			// Wait before for a brief pause before the next iteration of the loop
-			time.Sleep(changeStreamRetryDelay)
-			continue
 		}
 
-		if err := m.handleChangeDoc(ctx, cursor, stream, OnMessage); err != nil {
+		// Check boundary AFTER emitting
+		if err := m.handleIdleCheckpoint(ctx, cursor, stream, lastOplogTime); err != nil {
```
hash-data: not idel now

vishalm0509 (approved):
LGTM

hash-data (approved):

PR #748, opened 2026-01-12
fix(arrow): write positional delete files for duplicate _olake_ids in different destination buffers

badalprasadsingh:
# Description

In case we have multiple cdc events targeting the rows with the same _olake_id across multiple destination buffers, we need to create positional delete files to delete few rows of data from the soft delete data files to avoid duplicate entries.

## Type of change

# How Has This Been Tested?

- [x] Iceberg + Minio + Spark
- [ ] Scenario B

## Documentation

### DIFF
--- a/utils/testutils/test_utils.go
+++ b/utils/testutils/test_utils.go
@@ -693,11 +693,13 @@ func (cfg *IntegrationTest) testParquetFullLoadAndIncremental(
 	}
 
 	// Test cases for incremental sync
+	// Note: All tests use useState=true so they write cursor to state file
+	// This ensures subsequent incremental tests have a valid starting cursor
 	incrementalTestCases := []syncTestCase{
 		{
 			name:      "Full-Refresh",
 			operation: "",
-			useState:  false,
+			useState:  true,
 			opSymbol:  "r",
 			expected:  cfg.ExpectedData,
 		},


destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeArrowIngester.java
```diff
@@ -134,56 +152,42 @@ public void icebergAPI(ArrowPayload request, StreamObserver<RecordIngest.ArrowIn
...
+                              FileFormat fileFormat = IcebergUtil.getTableFileFormat(this.icebergTable);
+                              this.outputFileFactory = IcebergUtil.getTableOutputFileFactory(this.icebergTable,
+                                        fileFormat);
+                         }
+
+                         EncryptedOutputFile encryptedFile = this.outputFileFactory.newOutputFile();
+                         String basePath = encryptedFile.encryptingOutputFile().location();
+
+                         LOGGER.debug("{} Allocated base file path: {}", requestId, basePath);
+                         sendResponse(responseObserver, basePath);
```
hash-data, line 189: are we sure this is going to be data file path?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -305,6 +305,41 @@ public void accumulateDeleteFiles(String threadId, Table table, String filePath,
           }
      }
 
+     public void accumulatePositionalDeleteFiles(String threadId, Table table, String filePath,
```
hash-data, line 308: function name should be registerPosFile

destination/iceberg/arrow-writer/writer.go
```diff
@@ -117,48 +133,51 @@ func (w *ArrowWriter) getRecordPartitionValues(record types.RawRecord, olakeTime
 	return typedValues, nil
 }
 
-func (w *ArrowWriter) extract(records []types.RawRecord, olakeTimestamp time.Time) (map[string]*PartitionedRecords, map[string]*PartitionedRecords, error) {
+type Deletes struct {
+	Partitions      *PartitionedRecords
+	PostitionOffset map[string]*PositionTracker
```
hash-data: add comments what to what mapping it will do?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -117,48 +133,51 @@ func (w *ArrowWriter) getRecordPartitionValues(record types.RawRecord, olakeTime
...
-					PartitionKey:    pKey,
-					PartitionValues: pValues,
+				deletes[pKey] = &Deletes{
+					Partitions: &PartitionedRecords{
+						PartitionKey:    pKey,
+						PartitionValues: pValues,
+					},
+					PostitionOffset: make(map[string]*PositionTracker),
 				}
 			}
```
hash-data, line 170: why we cant add for eq delete here, skip loop at end?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -350,30 +459,61 @@ func (w *ArrowWriter) initialize(ctx context.Context) error {
...
 				}),
 			},
 		}, nil)
+
+		// positional delete schema: file_path (string) and pos (long)
+		// iceberg's reserved field IDs: file_path = MAX_VALUE-101, pos = MAX_VALUE-102
+		const (
+			posDeleteFilePathFieldID = 2147483546 // Integer.MAX_VALUE - 101
+			posDeletePosFieldID      = 2147483545 // Integer.MAX_VALUE - 102
+		)
```
hash-data, line 473: did not get it

destination/iceberg/arrow-writer/writer.go
hash-data: spell mistake

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
badalprasadsingh: should we? the others are also, the others are also like `accumulateDataFiles`

also registering and commit happens at last

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: added, and
fixed

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: for every parquet delete or data file, we store field-id to column mapping in the `metadata` for each parquet file.

since columns of positional delete files, are actually not any column in the iceberg table, we add `Integer.MAX_VALUE - 101` value and `Integer.MAX_VALUE - 102` as the field-ids for `filePath` and `positions`.

destination/iceberg/arrow-writer/writer.go
badalprasadsingh: done.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeArrowIngester.java
badalprasadsingh: this gives you a filePath, that can be used for `datafile`, `equality-delete` file or `positional-delete` file. The same fileFactory will provide same `uuid` in the parquet file name only increasing the number at the end.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
hash-data: but it is not kind of accumlation, we can rename it to appendPosFile kind of

PR #756, opened 2026-01-14
feat: implement filter in cdc and incremental

vikaxsh:
# Description

This PR adds filtering capability for CDC and incremental sync modes, allowing users to selectively sync records based on configurable filter conditions.

Fixes # (issue)

## Type of change

- **Multiple Conditions**: Support for multiple filter conditions with AND/OR logical operators
- **Operators**: Supports `=`, `!=`, `<`, `>`, `<=`, `>=`
- **Type-Aware Filtering**: Automatic type conversion and validation based on schema

# How Has This Been Tested?

Comprehensive test coverage in `destination/filter_test.go` including:
- Legacy filter support
- Multiple operators and data types
- AND/OR logic
- Null value handling
- Type coercion
- Edge cases

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

- [x] Documentation Link: https://github.com/datazip-inc/olake-docs/pull/373
- [] N/A (bug fix, refactor, or test changes only)

### New Filter Format
Filters can be specified in the stream configuration using the new `FilterInput` structure:

```json
{
  "selected_streams": {
    "namespace": [
      {
        "stream_name": "table_1",
        "filter_input": {
          "logical_operator": "AND",
          "conditions": [
            {
              "column": "id",
              "operator": ">",
              "value": 100
            },
            {
              "column": "status",
              "operator": "=",
              "value": "active"
            }
          ]
        }
      }
    ]
  }
}
```

### Legacy Filter Format (Still Supported)
The legacy string-based filter format continues to work:

```json
{
  "filter": "id > 100 AND status = \"active\""
}
```

### DIFF
--- a/.github/workflows/security-ci.yaml
+++ b/.github/workflows/security-ci.yaml
@@ -7,12 +7,14 @@ on:
     paths:
       - '**/*.go'
       - '**/*.java'
+      - '**/go.mod'
   pull_request:
     branches:
       - "*"
     paths:
       - '**/*.go'
       - '**/*.java'
+      - '**/go.mod'
   workflow_dispatch:
     inputs:
       logLevel:
@@ -63,6 +65,7 @@ jobs:
           scan-ref: './'
           scanners: 'vuln'
           severity: 'HIGH,CRITICAL'
+          ignore-unfixed: true
 
   # TODO: Add Java Dependency Trivy (removed because trivy getting stuck or run for hours)
   # trivy-java:
--- a/drivers/abstract/abstract.go
+++ b/drivers/abstract/abstract.go
@@ -237,10 +237,11 @@ func generateThreadID(streamID, hash string) string {
 // The writer parameter can be either:
 //   - *destination.WriterThread for a single writer
 //   - map[string]*destination.WriterThread for multiple writers keyed by stream ID
-//
-// The threadID and closeMessage parameters are optional (empty string means not used) and only apply to single writer cases
-func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *error, writer any, threadID string, mtState any) {
-	// Cancel context if there's an error, so other threads using this context can detect the failure
+func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *error, writer any, threadID string, mtState *any) {
+	if r := recover(); r != nil {
+		*err = utils.Ternary(*err == nil, fmt.Errorf("panic recovered: %v", r), fmt.Errorf("%s: panic recovered: %v", *err, r)).(error)
+	}
+
 	if *err != nil {
 		cancel()
 	}
@@ -248,7 +249,7 @@ func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *er
 	var metadataState any
 	var closeErr error
 	if mtState != nil {
-		metadataState, closeErr = types.SetMetadataState(mtState, threadID)
+		metadataState, closeErr = types.SetMetadataState(*mtState, threadID)
 		if closeErr != nil {
 			closeErr = fmt.Errorf("failed to set metadata state: %s", closeErr)
 		}
@@ -275,12 +276,6 @@ func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *er
 	if closeErr != nil {
 		*err = utils.Ternary(*err == nil, closeErr, fmt.Errorf("%s: prev error: %w", closeErr, *err)).(error)
 	}
-
-	// check for panics before post-processing
-	if r := recover(); r != nil {
-		*err = utils.Ternary(*err == nil, fmt.Errorf("panic recovered: %v", r), fmt.Errorf("%s: prev error: %w", r, *err)).(error)
-	}
-
 	if *err != nil {
 		cancel()
 	}
--- a/drivers/abstract/backfill.go
+++ b/drivers/abstract/backfill.go
@@ -55,20 +55,17 @@ func (a *AbstractDriver) Backfill(mainCtx context.Context, backfilledStreams cha
 		}
 
 		defer func(ctx context.Context) {
-			// Pass nil metadata payload for backfill.
-			// Java 2PC appends threadId into full_refresh_committed_ids when payload is empty.
-			handleWriterCleanup(backfillCtx, backfillCtxCancel, &err, inserter, threadID, nil)
-
 			if ctx.Err() != nil {
 				return
 			}
-
 			chunksLeft := a.state.RemoveChunk(stream.Self(), chunk)
 			if chunksLeft == 0 && backfilledStreams != nil {
 				backfilledStreams <- stream.ID()
 			}
 			logger.Infof("finished chunk min[%v] and max[%v] of stream %s", chunk.Min, chunk.Max, stream.ID())
 		}(backfillCtx)
+
+		defer handleWriterCleanup(backfillCtx, backfillCtxCancel, &err, inserter, threadID, nil)
 
 		if prevMetadataState != nil {
 			if slices.Contains(prevMetadataState.FullRefreshCommittedIDs, threadID) {
--- a/drivers/abstract/cdc.go
+++ b/drivers/abstract/cdc.go
@@ -106,7 +106,7 @@ func (a *AbstractDriver) streamChanges(mainCtx context.Context, pool *destinatio
 
 	defer func() {
 		if postCDCErr := a.driver.PostCDC(cdcCtx, streamIndex); postCDCErr != nil {
-			err = fmt.Errorf("post cdc error: %s", postCDCErr)
+			err = utils.Ternary(err == nil, fmt.Errorf("post cdc error: %s", postCDCErr), fmt.Errorf("%s: post cdc error: %s", err, postCDCErr)).(error)
 		}
 	}()
 
@@ -128,9 +128,7 @@ func (a *AbstractDriver) streamChanges(mainCtx context.Context, pool *destinatio
 		metadataStates[stream.ID()] = writerMetaState
 	}
 
-	defer func() {
-		handleWriterCleanup(cdcCtx, cdcCtxCancel, &err, writers, "", finalMetadataState)
-	}()
+	defer handleWriterCleanup(cdcCtx, cdcCtxCancel, &err, writers, "", &finalMetadataState)
 
 	finalMetadataState, err = a.driver.StreamChanges(cdcCtx, streamIndex, metadataStates, func(ctx context.Context, change CDCChange) error {
 		writer := writers[change.Stream.ID()]
--- a/drivers/abstract/incremental.go
+++ b/drivers/abstract/incremental.go
@@ -84,17 +84,19 @@ func (a *AbstractDriver) Incremental(mainCtx context.Context, pool *destination.
 				if ctx.Err() != nil {
 					return
 				}
-				// Save cursor state on success
 				a.state.SetCursor(stream.Self(), primaryCursor, a.FormatCursorValue(maxPrimaryCursorValue))
 				a.state.SetCursor(stream.Self(), secondaryCursor, a.FormatCursorValue(maxSecondaryCursorValue))
 			}(incrementalCtx)
 
+			var mtState any
+			defer handleWriterCleanup(incrementalCtx, incrementalCtxCancel, &err, inserter, threadID, &mtState)
+
 			defer func() {
-				mtState := map[string]any{primaryCursor: a.FormatCursorValue(maxPrimaryCursorValue)}
+				mtStateMap := map[string]any{primaryCursor: a.FormatCursorValue(maxPrimaryCursorValue)}
 				if secondaryCursor != "" {
-					mtState[secondaryCursor] = a.FormatCursorValue(maxSecondaryCursorValue)
+					mtStateMap[secondaryCursor] = a.FormatCursorValue(maxSecondaryCursorValue)
 				}
-				handleWriterCleanup(incrementalCtx, incrementalCtxCancel, &err, inserter, threadID, mtState)
+				mtState = mtStateMap
 			}()
 
 			// prevMetadataState id will be same as thread id if the state file failed to update
--- a/drivers/db2/go.mod
+++ b/drivers/db2/go.mod
@@ -11,7 +11,7 @@ require (
 )
 
 require (
-	cloud.google.com/go/compute/metadata v0.7.0 // indirect
+	cloud.google.com/go/compute/metadata v0.9.0 // indirect
 	dario.cat/mergo v1.0.1 // indirect
 	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
@@ -34,6 +34,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.34.0 // indirect
 	github.com/aws/smithy-go v1.23.0 // indirect
 	github.com/cenkalti/backoff/v4 v4.2.1 // indirect
+	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/containerd/errdefs v1.0.0 // indirect
 	github.com/containerd/errdefs/pkg v0.3.0 // indirect
 	github.com/containerd/log v0.1.0 // indirect
@@ -113,26 +114,26 @@ require (
 	github.com/yusufpapurcu/wmi v1.2.4 // indirect
 	github.com/zeebo/xxh3 v1.0.2 // indirect
 	go.mongodb.org/mongo-driver v1.17.3 // indirect
-	go.opentelemetry.io/auto/sdk v1.1.0 // indirect
+	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.49.0 // indirect
-	go.opentelemetry.io/otel v1.38.0 // indirect
-	go.opentelemetry.io/otel/metric v1.38.0 // indirect
-	go.opentelemetry.io/otel/trace v1.38.0 // indirect
+	go.opentelemetry.io/otel v1.39.0 // indirect
+	go.opentelemetry.io/otel/metric v1.39.0 // indirect
+	go.opentelemetry.io/otel/trace v1.39.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
-	golang.org/x/crypto v0.45.0 // indirect
+	golang.org/x/crypto v0.46.0 // indirect
 	golang.org/x/exp v0.0.0-20240909161429-701f63a606c0 // indirect
-	golang.org/x/mod v0.29.0 // indirect
-	golang.org/x/net v0.47.0 // indirect
-	golang.org/x/oauth2 v0.30.0 // indirect
-	golang.org/x/sync v0.18.0 // indirect
-	golang.org/x/sys v0.38.0 // indirect
-	golang.org/x/telemetry v0.0.0-20251008203120-078029d740a8 // indirect
-	golang.org/x/text v0.31.0 // indirect
-	golang.org/x/tools v0.38.0 // indirect
+	golang.org/x/mod v0.30.0 // indirect
+	golang.org/x/net v0.48.0 // indirect
+	golang.org/x/oauth2 v0.34.0 // indirect
+	golang.org/x/sync v0.19.0 // indirect
+	golang.org/x/sys v0.39.0 // indirect
+	golang.org/x/telemetry v0.0.0-20251111182119-bc8e575c7b54 // indirect
+	golang.org/x/text v0.32.0 // indirect
+	golang.org/x/tools v0.39.0 // indirect
 	golang.org/x/xerrors v0.0.0-20240903120638-7835f813f4da // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20250728155136-f173205681a0 // indirect
-	google.golang.org/grpc v1.74.2 // indirect
-	google.golang.org/protobuf v1.36.8 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20251202230838-ff82c1b0f217 // indirect
+	google.golang.org/grpc v1.79.3 // indirect
+	google.golang.org/protobuf v1.36.10 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
--- a/drivers/kafka/go.mod
+++ b/drivers/kafka/go.mod
@@ -35,6 +35,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.34.0 // indirect
 	github.com/aws/smithy-go v1.23.0 // indirect
 	github.com/cenkalti/backoff/v4 v4.2.1 // indirect
+	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/containerd/errdefs v1.0.0 // indirect
 	github.com/containerd/errdefs/pkg v0.3.0 // indirect
 	github.com/containerd/log v0.1.0 // indirect
@@ -121,25 +122,25 @@ require (
 	github.com/yusufpapurcu/wmi v1.2.4 // indirect
 	github.com/zeebo/xxh3 v1.0.2 // indirect
 	go.mongodb.org/mongo-driver v1.17.3 // indirect
-	go.opentelemetry.io/auto/sdk v1.1.0 // indirect
+	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.49.0 // indirect
-	go.opentelemetry.io/otel v1.38.0 // indirect
-	go.opentelemetry.io/otel/metric v1.38.0 // indirect
-	go.opentelemetry.io/otel/trace v1.38.0 // indirect
+	go.opentelemetry.io/otel v1.39.0 // indirect
+	go.opentelemetry.io/otel/metric v1.39.0 // indirect
+	go.opentelemetry.io/otel/trace v1.39.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
-	golang.org/x/crypto v0.45.0 // indirect
+	golang.org/x/crypto v0.46.0 // indirect
 	golang.org/x/exp v0.0.0-20240909161429-701f63a606c0 // indirect
-	golang.org/x/mod v0.29.0 // indirect
-	golang.org/x/net v0.47.0 // indirect
-	golang.org/x/sync v0.18.0 // indirect
-	golang.org/x/sys v0.38.0 // indirect
-	golang.org/x/telemetry v0.0.0-20251008203120-078029d740a8 // indirect
-	golang.org/x/text v0.31.0 // indirect
-	golang.org/x/tools v0.38.0 // indirect
+	golang.org/x/mod v0.30.0 // indirect
+	golang.org/x/net v0.48.0 // indirect
+	golang.org/x/sync v0.19.0 // indirect
+	golang.org/x/sys v0.39.0 // indirect
+	golang.org/x/telemetry v0.0.0-20251111182119-bc8e575c7b54 // indirect
+	golang.org/x/text v0.32.0 // indirect
+	golang.org/x/tools v0.39.0 // indirect
 	golang.org/x/xerrors v0.0.0-20240903120638-7835f813f4da // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20250728155136-f173205681a0 // indirect
-	google.golang.org/grpc v1.74.2 // indirect
-	google.golang.org/protobuf v1.36.8 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20251202230838-ff82c1b0f217 // indirect
+	google.golang.org/grpc v1.79.3 // indirect
+	google.golang.org/protobuf v1.36.10 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
--- a/drivers/mongodb/go.mod
+++ b/drivers/mongodb/go.mod
@@ -10,7 +10,7 @@ require (
 )
 
 require (
-	cloud.google.com/go/compute/metadata v0.7.0 // indirect
+	cloud.google.com/go/compute/metadata v0.9.0 // indirect
 	dario.cat/mergo v1.0.1 // indirect
 	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
@@ -33,6 +33,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.34.0 // indirect
 	github.com/aws/smithy-go v1.23.0 // indirect
 	github.com/cenkalti/backoff/v4 v4.2.1 // indirect
+	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/containerd/errdefs v1.0.0 // indirect
 	github.com/containerd/errdefs/pkg v0.3.0 // indirect
 	github.com/containerd/log v0.1.0 // indirect
@@ -58,6 +59,11 @@ require (
 	github.com/google/uuid v1.6.0 // indirect
 	github.com/hashicorp/errwrap v1.1.0 // indirect
 	github.com/hashicorp/go-multierror v1.1.1 // indirect
+	github.com/jackc/chunkreader/v2 v2.0.1 // indirect
+	github.com/jackc/pgconn v1.14.3 // indirect
+	github.com/jackc/pgio v1.0.0 // indirect
+	github.com/jackc/pgproto3/v2 v2.3.3 // indirect
+	github.com/jackc/pgtype v1.14.0 // indirect
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/asmfmt v1.3.2 // indirect
@@ -121,21 +127,21 @@ require (
 	github.com/youmark/pkcs8 v0.0.0-20240726163527-a2c0da244d78 // indirect
 	github.com/yusufpapurcu/wmi v1.2.4 // indirect
 	github.com/zeebo/xxh3 v1.0.2 // indirect
-	go.opentelemetry.io/auto/sdk v1.1.0 // indirect
+	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.49.0 // indirect
-	go.opentelemetry.io/otel v1.38.0 // indirect
-	go.opentelemetry.io/otel/metric v1.38.0 // indirect
-	go.opentelemetry.io/otel/trace v1.38.0 // indirect
+	go.opentelemetry.io/otel v1.39.0 // indirect
+	go.opentelemetry.io/otel/metric v1.39.0 // indirect
+	go.opentelemetry.io/otel/trace v1.39.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	golang.org/x/exp v0.0.0-20240909161429-701f63a606c0 // indirect
-	golang.org/x/mod v0.29.0 // indirect
-	golang.org/x/oauth2 v0.30.0 // indirect
-	golang.org/x/sync v0.18.0 // indirect
-	golang.org/x/telemetry v0.0.0-20251008203120-078029d740a8 // indirect
+	golang.org/x/mod v0.30.0 // indirect
+	golang.org/x/oauth2 v0.34.0 // indirect
+	golang.org/x/sync v0.19.0 // indirect
+	golang.org/x/telemetry v0.0.0-20251111182119-bc8e575c7b54 // indirect
 	golang.org/x/xerrors v0.0.0-20240903120638-7835f813f4da // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20250728155136-f173205681a0 // indirect
-	google.golang.org/grpc v1.74.2 // indirect
-	google.golang.org/protobuf v1.36.8 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20251202230838-ff82c1b0f217 // indirect
+	google.golang.org/grpc v1.79.3 // indirect
+	google.golang.org/protobuf v1.36.10 // indirect
 	gopkg.in/ini.v1 v1.66.6 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
@@ -148,23 +154,18 @@ require (
 	github.com/go-playground/validator/v10 v10.25.0 // indirect
 	github.com/goccy/go-json v0.10.5 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
-	github.com/jackc/chunkreader/v2 v2.0.1 // indirect
-	github.com/jackc/pgconn v1.14.3 // indirect
-	github.com/jackc/pgio v1.0.0 // indirect
 	github.com/jackc/pgpassfile v1.0.0 // indirect
-	github.com/jackc/pgproto3/v2 v2.3.3 // indirect
 	github.com/jackc/pgservicefile v0.0.0-20240606120523-5a60cdf6a761 // indirect
-	github.com/jackc/pgtype v1.14.0 // indirect
 	github.com/leodido/go-urn v1.4.0 // indirect
 	github.com/mitchellh/hashstructure v1.1.0 // indirect
 	github.com/spf13/cobra v1.9.1 // indirect
 	github.com/spf13/pflag v1.0.6 // indirect
 	go.mongodb.org/mongo-driver v1.17.3
-	golang.org/x/crypto v0.45.0
-	golang.org/x/net v0.47.0 // indirect
-	golang.org/x/sys v0.38.0 // indirect
-	golang.org/x/text v0.31.0 // indirect
-	golang.org/x/tools v0.38.0 // indirect
+	golang.org/x/crypto v0.46.0
+	golang.org/x/net v0.48.0 // indirect
+	golang.org/x/sys v0.39.0 // indirect
+	golang.org/x/text v0.32.0 // indirect
+	golang.org/x/tools v0.39.0 // indirect
 )
 
 replace (
--- a/drivers/mongodb/internal/cdc.go
+++ b/drivers/mongodb/internal/cdc.go
@@ -72,12 +72,16 @@ func (m *Mongo) StreamChanges(ctx context.Context, streamIndex int, metadataStat
 	if prevResumeToken == nil {
 		return nil, fmt.Errorf("resume token not found for stream: %s", stream.ID())
 	}
-
-	// check for recoverySync (sync after failiure with broken state)
+	//   metadata > state  →  metadata is further ahead (crash-recovery path: metadata
+	//                         was committed to the destination but state write failed).
+	//                         Use the metadata token so we don't re-read already-written events.
+	//   state >= metadata →  state is current or ahead; read forward normally.
 	if mtState != nil {
-		cmp := typeutils.Compare(prevResumeToken, mtState)
-		if cmp != 0 {
-			logger.Infof("Stream[%s] resume token mismatch, updating resume token in state", stream.ID())
+		// TODO: addition of all the state updations in metadata file even for blank sync scenario
+		// metadata > state → crash-recovery path (metadata committed but state write failed).
+		// state >= metadata → read forward normally.
+		if typeutils.Compare(prevResumeToken, mtState) < 0 {
+			logger.Infof("Stream[%s] metadata ahead of state, using metadata resume token for recovery", stream.ID())
 			prevResumeToken = mtState
 		}
 	}
--- a/drivers/mongodb/internal/mon.go
+++ b/drivers/mongodb/internal/mon.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"sync"
 	"time"
+	"strings"
 
 	"github.com/datazip-inc/olake/constants"
 	"github.com/datazip-inc/olake/drivers/abstract"
@@ -163,6 +164,12 @@ func (m *Mongo) GetStreamNames(ctx context.Context) ([]string, error) {
 		if collectionType, ok := collectionInfo["type"].(string); ok && collectionType == "view" {
 			continue
 		}
+
+		// Skip if collection is system.*
+		if name, ok := collectionInfo["name"].(string); ok && strings.HasPrefix(name, "system.") {
+			continue
+		}
+
 		streamNames = append(streamNames, collectionInfo["name"].(string))
 	}
 	return streamNames, collections.Err()
--- a/drivers/mssql/go.mod
+++ b/drivers/mssql/go.mod
@@ -16,7 +16,7 @@ require (
 )
 
 require (
-	cloud.google.com/go/compute/metadata v0.7.0 // indirect
+	cloud.google.com/go/compute/metadata v0.9.0 // indirect
 	dario.cat/mergo v1.0.1 // indirect
 	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
@@ -39,6 +39,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.34.0 // indirect
 	github.com/aws/smithy-go v1.23.0 // indirect
 	github.com/cenkalti/backoff/v4 v4.2.1 // indirect
+	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/containerd/errdefs v1.0.0 // indirect
 	github.com/containerd/errdefs/pkg v0.3.0 // indirect
 	github.com/containerd/log v0.1.0 // indirect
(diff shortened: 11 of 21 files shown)


constants/constants.go
```diff
@@ -50,6 +50,7 @@ const (
 	MSSQL    DriverType = "mssql"
 )
 
+var FilterInFullRefreshDrivers = []DriverType{S3, Kafka}
```
vaibhav-datazip: Don't other drivers support filter in full refresh ?

destination/filter.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	return filterConcurrently(ctx, records, conditions, filter.LogicalOperator)
+}
+
+// parsedCondition holds a fully-typed condition
+type parsedCondition struct {
+	column   string
+	operator string
+	value    any
+}
```
vaibhav-datazip: move struct above all functions in the file

destination/parquet/parquet.go
```diff
@@ -311,8 +312,18 @@ func (p *Parquet) Close(_ context.Context) error {
...
 func (p *Parquet) FlattenAndCleanData(ctx context.Context, records []types.RawRecord) (bool, []types.RawRecord, any, error) {
-	if !p.stream.NormalizationEnabled() {
-		return false, records, nil, nil
+	filter, isLegacy, err := p.stream.GetFilter()
+	if err != nil {
+		return false, nil, nil, fmt.Errorf("failed to parse stream filter: %s", err)
+	}
+
+	if !p.stream.NormalizationEnabled() && !p.options.Backfill {
+		// Q: do we need to add filter in case of normalization false?
```
vaibhav-datazip: In any case synced value should be filtered only right ?

types/catalog.go
```diff
@@ -43,8 +43,23 @@ type StreamMetadata struct {
...
+	LogicalOperator string            `json:"logical_operator,omitempty"`
+	Conditions      []FilterCondition `json:"conditions,omitempty"`
+}
+
+type FilterCondition struct {
+	Column   string `json:"column,omitempty"`
+	Operator string `json:"operator,omitempty"`
+	Value    any    `json:"value,omitempty"`
+}
+
```
vaibhav-datazip: can you add some documentation or description of how I can test or what all is covered in pr ?

types/stream_configured.go
```diff
@@ -14,19 +14,6 @@ type ConfiguredStream struct {
...
-	Operator string
-	Value    string
-}
-
-// Filter represents the parsed filter
-type Filter struct {
-	Conditions      []Condition // a > b, a < b
-	LogicalOperator string      // condition[0] and/or condition[1], single and/or supported
-}
-
```
vaibhav-datazip, line 29: it was better to be kept in stream_configured, why was this moved to catalog.go

types/stream_configured.go
```diff
@@ -84,10 +71,17 @@ func (s *ConfiguredStream) Cursor() (string, string) {
 	return primaryCursor, secondaryCursor
 }
 
-func (s *ConfiguredStream) GetFilter() (Filter, error) {
+func (s *ConfiguredStream) GetFilter() (FilterInput, bool, error) {
```
vaibhav-datazip: Add a comment on top of this function what it does and what the return values are

drivers/mongodb/internal/backfill.go
```diff
@@ -397,36 +397,69 @@ func buildMongoCondition(cond types.Condition) bson.D {
...
-				return oid
+	mongoOp, ok := opMap[cond.Operator]
+	if !ok {
+		panic(fmt.Sprintf("unsupported operator for mongo: %s", cond.Operator))
+	}
+
+	var value interface{}
+
+	switch v := cond.Value.(type) {
+
```
vaibhav-datazip, line 422: remove extra space

drivers/mongodb/internal/mon_test.go
```diff
@@ -20,6 +20,21 @@ func TestMongodbIntegration(t *testing.T) {
...
+                            "operator": "<",
+                            "value": 239834.89
+                        },
+                        {
+                            "column": "id_timestamp",
+                            "operator": ">=",
+                            "value": "2022-07-01T15:30:00.000+00:00"
+                        }
+                    ]
+                }`,
```
vaibhav-datazip: formatting is not correct

drivers/mongodb/internal/mon_test_util.go
```diff
@@ -104,6 +104,17 @@ func ExecuteQuery(ctx context.Context, t *testing.T, streams []string, operation
 		}
 		_, err := collection.InsertOne(ctx, doc)
 		require.NoError(t, err, "Failed to insert document")
+		// insert a filtered doc
```
vaibhav-datazip: what does inserting a filtered doc mean ? can you make this comment more clear

pkg/jdbc/jdbc.go
```diff
@@ -22,8 +22,10 @@ func QuoteIdentifier(identifier string, driver constants.DriverType) string {
 	switch driver {
 	case constants.MySQL:
 		return fmt.Sprintf("`%s`", identifier) // MySQL uses backticks for quoting identifiers
-	case constants.Postgres, constants.DB2, constants.Oracle:
+	case constants.Postgres, constants.DB2:
 		return fmt.Sprintf("%q", identifier)
+	case constants.Oracle:
+		return fmt.Sprintf("%q", strings.ToUpper(identifier))
```
vaibhav-datazip: What if the oracle column is lowercase ?

pkg/jdbc/jdbc.go
```diff
@@ -1087,8 +1089,12 @@ func IncrementalValueFormatter(ctx context.Context, cursorField, argumentPlaceho
...
 
 // ParseFilter converts a filter string to a valid SQL WHERE condition, also appends the threshold filter if present
-func SQLFilter(stream types.StreamInterface, driver string, thresholdFilter string) (string, error) {
-	buildCondition := func(cond types.Condition, driver string) (string, error) {
+func SQLFilter(
+	stream types.StreamInterface,
+	driver string,
+	thresholdFilter string,
+) (string, error) {
+	buildCondition := func(cond types.FilterCondition, driver string) (string, error) {
```
vaibhav-datazip: you can revert back to single line format.

pkg/jdbc/jdbc.go
```diff
@@ -1102,68 +1108,97 @@ func SQLFilter(stream types.StreamInterface, driver string, thresholdFilter stri
...
-			driverType = constants.Postgres // default fallback
+			driverType = constants.Postgres
 		}
 
 		quotedColumn := QuoteIdentifier(cond.Column, driverType)
 
-		// Handle unquoted null value
-		if cond.Value == "null" {
+		// ---------- NULL handling ----------
+		if cond.Value == nil {
```
vaibhav-datazip: earlier this check was used to filter out null values like field != null , will it work the same way now as well . is it backward compatible.

pkg/jdbc/jdbc.go
```diff
@@ -1102,68 +1108,97 @@ func SQLFilter(stream types.StreamInterface, driver string, thresholdFilter stri
...
+			valueSQL = fmt.Sprintf("'%s'", escaped)
+			// Oracle-specific timestamp handling
+			if driverType == constants.Oracle && strings.Contains(v, "T") && (strings.Contains(v, "Z") || strings.Contains(v, "+") || (strings.Contains(v, "-") && len(v) > 19)) {
+				if t, err := time.Parse(time.RFC3339, v); err == nil {
+					timestampStr := t.Format("2006-01-02 15:04:05.000")
+					valueSQL = fmt.Sprintf(
+						"TO_TIMESTAMP('%s', 'YYYY-MM-DD HH24:MI:SS.FF')",
+						timestampStr,
+					)
+				}
```
vaibhav-datazip: earlier the user had to enter the timestamp format in oracle specific timestamp format for the filter to work , what will this work if user enters the timestamp in that format ?

pkg/jdbc/jdbc.go
```diff
@@ -1102,68 +1108,97 @@ func SQLFilter(stream types.StreamInterface, driver string, thresholdFilter stri
...
-		return fmt.Sprintf("%s %s %s", quotedColumn, cond.Operator, value), nil
+		return fmt.Sprintf("%s %s %s", quotedColumn, cond.Operator, valueSQL), nil
 	}
 
-	filter, err := stream.GetFilter()
+	// ---------- get canonical filter ----------
+	filter, _, err := stream.GetFilter()
 	if err != nil {
-		return "", fmt.Errorf("failed to parse stream filter: %s", err)
+		return "", fmt.Errorf("failed to parse stream filter: %w", err)
```
vaibhav-datazip: change error to %s

pkg/jdbc/jdbc.go
```diff
@@ -1102,68 +1108,97 @@ func SQLFilter(stream types.StreamInterface, driver string, thresholdFilter stri
...
+			escaped := strings.ReplaceAll(fmt.Sprint(v), "'", "''")
+			valueSQL = fmt.Sprintf("'%s'", escaped)
 		}
 
-		return fmt.Sprintf("%s %s %s", quotedColumn, cond.Operator, value), nil
+		return fmt.Sprintf("%s %s %s", quotedColumn, cond.Operator, valueSQL), nil
 	}
 
-	filter, err := stream.GetFilter()
+	// ---------- get canonical filter ----------
```
vaibhav-datazip: can we remove some unnecessary comments.

pkg/jdbc/jdbc.go
```diff
@@ -1102,68 +1108,97 @@ func SQLFilter(stream types.StreamInterface, driver string, thresholdFilter stri
...
 			}
 			conditions = append(conditions, formatted)
-			return nil
-		})
-		finalFilter, filterErr = strings.Join(conditions, fmt.Sprintf(" %s ", filter.LogicalOperator)), err
+		}
+		finalFilter = strings.Join(
+			conditions,
+			fmt.Sprintf(" %s ", strings.ToUpper(filter.LogicalOperator)),
+		)
```
vaibhav-datazip: similarly, bring this in single line

protocol/sync.go
```diff
@@ -142,6 +142,10 @@ var syncCmd = &cobra.Command{
 		}
 		state.LogWithLock()
 		logger.Infof("Total records read: %d", pool.GetStats().ReadCount.Load())
+		filteredRecoredCount := pool.GetStats().RecordsFiltered.Load()
+		if filteredRecoredCount > 0 {
+			logger.Infof("Total records filtered: %d", filteredRecoredCount)
+		}
```
vaibhav-datazip: should we maintain the number of records getting filtered out , or is this the number of records gone through ?

utils/testutils/test_utils.go
```diff
@@ -236,8 +237,15 @@ func updateSelectedStreamsCommand(config TestConfig, namespace, partitionRegex s
 	}
 	condition := strings.Join(streamConditions, " or ")
 	tmpCatalog := fmt.Sprintf("/tmp/%s_%s_streams.json", config.Driver, utils.Ternary(isBackfill, "backfill", "cdc").(string))
+
+	// Default to empty string if filterInput is not provided
+	if filterInput == "" {
+		filterInput = "{}"
+	}
+
```
vaibhav-datazip: this is test file, wont this be handled automatically in code logic ?

types/stream_configured_test.go
```diff
@@ -10,560 +10,162 @@ import (
...
-			name:   "empty filter",
-			filter: "",
-			expectedFilter: Filter{
-				Conditions:      nil,
-				LogicalOperator: "",
-			},
-			expectError: false,
+			name:     "empty filter",
+			filter:   "",
+			expected: FilterInput{},
```
vaibhav-datazip: we shouldn't remove test cases related to string part, as this will help in maintaining backward compatibility. you can add new test cases after these .

drivers/oracle/internal/oracle_test.go
```diff
@@ -20,6 +20,21 @@ func TestOracleIntegration(t *testing.T) {
...
+                    "conditions": [
+                        {
+                            "column": "col_double_precision",
+                            "operator": "<",
+                            "value": 239834.89
+                        },
+                        {
+                            "column": "col_timestamp",
+                            "operator": ">=",
+                            "value": "2022-07-01T15:30:00.000+00:00"
```
vaibhav-datazip, line 34: will this filter work if timestamp is in some different format as well ?

destination/writers.go
```diff
@@ -225,12 +233,12 @@ func (wt *WriterThread) flush(ctx context.Context, buf []types.RawRecord) (err e
 	// create flush context
 	flushCtx, cancel := context.WithCancel(ctx)
 	defer cancel()
-
+	recodsCountBeforeFiltering := len(buf)
```
vaibhav-datazip: it should be `recordsCountBeforeFiltering` , r is missing in records

protocol/sync.go
```diff
@@ -142,6 +142,10 @@ var syncCmd = &cobra.Command{
 		}
 		state.LogWithLock()
 		logger.Infof("Total records read: %d", pool.GetStats().ReadCount.Load())
+		filteredRecoredCount := pool.GetStats().RecordsFiltered.Load()
```
vaibhav-datazip: filterRecoredCount spelling is incorrect

destination/writers.go
```diff
@@ -33,6 +35,7 @@ type (
 	Stats struct {
 		TotalRecordsToSync atomic.Int64 // total record that are required to sync
 		ReadCount          atomic.Int64 // records that got read
+		RecordsFiltered    atomic.Int64 // records that got filtered
```
vaibhav-datazip, line 38: initialization of this was not done below

drivers/mysql/internal/mysql_test.go
```diff
@@ -20,6 +20,21 @@ func TestMySQLIntegration(t *testing.T) {
...
+                            "column": "price_double",
+                            "operator": "<",
+                            "value": 239834.89
+                        },
+                        {
+                            "column": "created_timestamp",
+                            "operator": ">=",
+                            "value": "2022-07-01T15:30:00.000+00:00"
+                        }
+                    ]
```
vaibhav-datazip: can we add other datatypes filter as well ?

constants/constants.go
vikaxsh: This array is derived from the driver filtering logic implemented on the destination side.

destination/parquet/parquet.go
vikaxsh: if it is , we will have to flatten records in normalization false as well

types/catalog.go
vikaxsh: done

types/stream_configured.go
vikaxsh: Earlier, Filter was of type string. Since FilterInput is now a struct, it makes sense to keep both in the same file.

drivers/mongodb/internal/mon_test.go
vikaxsh: fixed

drivers/mongodb/internal/mon_test_util.go
vikaxsh: done

pkg/jdbc/jdbc.go
vikaxsh: done

pkg/jdbc/jdbc.go
vikaxsh: i have tested with putting null value in filter as well, worked fine

protocol/sync.go
vikaxsh: it is count of filtered records after gone through all records

utils/testutils/test_utils.go
vikaxsh: yeah already handled in code

types/stream_configured_test.go
vikaxsh: let's discuss

protocol/sync.go
vikaxsh: fixed

destination/writers.go
vikaxsh: fixed

destination/writers.go
vikaxsh: done

drivers/mysql/internal/mysql_test.go
vikaxsh: it would make integration test run longer, all datatypes are already covered in filter_test.go

drivers/oracle/internal/oracle_test.go
vikaxsh: have tested with this format as well 01-JUL-22 03.30.00.000000 PM, will work in Full load but not in cdc as ReformatValue function only support (YYYY-MM-DD) for parsing timestamp from string

pkg/jdbc/jdbc.go
vikaxsh: have tested with this format as well 01-JUL-22 03.30.00.000000 PM, will work in Full load but not in cdc as ReformatValue function only support (YYYY-MM-DD) for parsing timestamp from string

pkg/jdbc/jdbc.go
vikaxsh: revert the changes

constants/constants.go
vaibhav-datazip: so variable name can be changed according to that , also add a comment regarding this

types/stream_configured.go
vaibhav-datazip: yeah , so why not both in stream_configured ?

pkg/jdbc/jdbc.go
vaibhav-datazip: so wont the timestamp filter work in cdc ?

utils/testutils/test_utils.go
vaibhav-datazip: so this can be removed right ?

drivers/mysql/internal/mysql_test.go
vaibhav-datazip: we can discuss this

constants/constants.go
vikaxsh: done

types/stream_configured.go
vikaxsh: done

pkg/jdbc/jdbc.go
vikaxsh: it will work only if provided in this format (YYYY-MM-DD)

types/stream_configured.go
```diff
@@ -84,10 +82,22 @@ func (s *ConfiguredStream) Cursor() (string, string) {
 	return primaryCursor, secondaryCursor
 }
 
-func (s *ConfiguredStream) GetFilter() (Filter, error) {
+// GetFilter returns the filter input for the stream
+// If the filter input is not nil and has conditions, it returns the filter input
+// If the filter input is nil or has no conditions, it returns the legacy filter input
+// If the legacy filter input is empty, it returns true and nil
+// If the legacy filter input is not empty, it returns false and the error
```
vishalm0509: The comment is misleading I believe. 

```go
// If the legacy filter input is not empty, it returns false and the error
```

We can add something similar to this. 

```go
// GetFilter returns the configured filter for this stream in a normalized form.
//
// The returned FilterInput is always the parsed representation of the filter:
//   - If StreamMetadata.FilterInput is set and has at least one condition,
//     that value is returned as-is (after normalizing the logical operator),
//     and isLegacy is false.
//   - Otherwise, the legacy string-based StreamMetadata.Filter is parsed into
//     a FilterInput using the legacy regex parser. In this case isLegacy is true.
//
// Return values:
//   - FilterInput: parsed filter definition (may be empty if no filter is configured)
//   - isLegacy:   true if the filter came from the legacy string field, false if it
//                 came from the new structured FilterInput field
//   - error:      non-nil only if the legacy string filter is non-empty and fails
//                 to parse; new structured filters do not return parse errors here.
func (s *ConfiguredStream) GetFilter() (FilterInput, bool, error) {
```

types/stream_configured.go
```diff
@@ -84,10 +82,22 @@ func (s *ConfiguredStream) Cursor() (string, string) {
...
+func (s *ConfiguredStream) GetFilter() (FilterInput, bool, error) {
+	//new filter input
+	if s.StreamMetadata.FilterInput != nil && len(s.StreamMetadata.FilterInput.Conditions) > 0 {
+		s.StreamMetadata.FilterInput.LogicalOperator = utils.Reformat(s.StreamMetadata.FilterInput.LogicalOperator)
```
vishalm0509: Do we need to `Reformat` the `LogicalOperator` ? We are using `strings.EqualFold` and `SQLFilter` does `strings.ToUpper` anyways

destination/filter.go
```diff
@@ -0,0 +1,200 @@
...
+
+// FilterRecords applies filtering ONLY for new filters.
+// For legacy filters, records are returned unchanged.
+func FilterRecords(
+	ctx context.Context,
+	records []types.RawRecord,
+	filter types.FilterInput,
+	legacy bool,
+	schema any,
+) ([]types.RawRecord, error) {
```
vaibhav-datazip: change this to single line , also instead of legacy can we rename it to legacyFilter, similarly in other places where it appears

types/stream_configured.go
vaibhav-datazip: no , I meant you can explain just the arguments it expects , and what this function does, this comment can be simplified

pkg/jdbc/jdbc.go
vaibhav-datazip: there are some more line dash comments in the file, can we get rid of those as well ?

destination/writers.go
```diff
@@ -82,6 +85,11 @@ func WithThreadID(threadID string) ThreadOptions {
 		opt.ThreadID = threadID
 	}
 }
+func WithDriverType(driverType constants.DriverType) ThreadOptions {
+	return func(opt *Options) {
+		opt.DriverType = driverType
+	}
+}
```
vishalm0509, line 92:
```suggestion
}

func WithDriverType(driverType constants.DriverType) ThreadOptions {
	return func(opt *Options) {
		opt.DriverType = driverType
	}
}
```

destination/filter.go
```diff
@@ -0,0 +1,200 @@
...
+			filtered = append(filtered, record)
+			mu.Unlock()
+		}
+		return nil
+	})
+	return filtered, err
+}
+
+// matches evaluates AND / OR logic
+func matches(
```
vishalm0509: Can we use a meaningful name here ?

destination/filter.go
```diff
@@ -0,0 +1,200 @@
...
+	record types.RawRecord,
+	conditions []parsedCondition,
+	logicalOp string,
+) bool {
+	isAnd := !strings.EqualFold(strings.TrimSpace(logicalOp), "OR")
```
vishalm0509: the variable name is `isAnd` but we are comparing with `OR`. Why ?

destination/filter.go
```diff
@@ -0,0 +1,200 @@
...
+
+		if !isAnd && ok {
+			return true
+		}
+	}
+	return isAnd
+}
+
+// evaluate compares values using typeutils.Compare
+func evaluate(recordVal, filterVal any, operator string) bool {
```
vishalm0509: Can we use proper name or comment ?

utils/typeutils/filter_test.go
```diff
@@ -0,0 +1,1442 @@
...
+		fields[name] = typeutils.NewField(dt)
+	}
+	return fields
+}
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Test: Legacy Filter (should skip filtering)
+// ─────────────────────────────────────────────────────────────────────────────
+
+func TestFilterRecords_LegacyFilter(t *testing.T) {
```
vishalm0509, line 40: "Delete operation should skip filtering and always be included", should we add a test for this ?

destination/filter.go
vikaxsh: it a utility function

destination/filter.go
vikaxsh: because anything other then or is treated as and

types/stream_configured.go
vikaxsh: already handled

vishalm0509: I am asking if we even need this line ?

```go
		s.StreamMetadata.FilterInput.LogicalOperator = utils.Reformat(s.StreamMetadata.FilterInput.LogicalOperator)
```

isn't it redundant ?

vishalm0509:
```json
  "filter_input": {
                    "logical_operator": "AND",
                    "conditions": [
                        {
                            "column": "isActive",
                            "operator": "=",
                            "value": "false"
                        },
                        {
                            "column": "Category",
                            "operator": "=",
                            "value": "Spaces"
                        }
                    ]
                }
```

For this case the ` "value": "Spaces"` is stored in the database with a whitespace  `"value": "  Spaces  "`. Do we need to handle this case ? 

In MSSQL the comparison support is case-insensitive and for Oracle for `BIT` it supports `0/1/"TRUE"/"FALSE"`, so we need to check if these cases work

destination/filter.go
```diff
@@ -0,0 +1,200 @@
...
+		ok := evaluate(recordVal, cond.value, cond.operator)
+		if isAnd && !ok {
+			return false
+		}
+
+		if !isAnd && ok {
+			return true
+		}
+	}
+	return isAnd
```
vaibhav-datazip: while trying to sync data in incremental mode in oracle db (version used 23)

this was my create query

```
CREATE TABLE classic_types_showcase (
    id                      NUMBER GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
    username                VARCHAR2(50),
    is_active               BOOLEAN,           -- Native to Oracle 23c
    age                     INTEGER,           -- Whole numbers
    wallet_balance          FLOAT,             -- ANSI Float
    scientific_val          BINARY_DOUBLE,     -- High-precision floating point
    join_date               DATE,              -- Date and Time (to the second)
    event_timestamp         TIMESTAMP,         -- Date, Time + Fractional seconds
    global_timestamp        TIMESTAMP WITH TIME ZONE,
    local_timestamp         TIMESTAMP WITH LOCAL TIME ZONE
) TABLESPACE USERS;
```

Insert data with various age value, then add filter as age < 25 . for initial full load you will see the data following the filter , but during the subsequent incremental sync all the data will be visible not respecting the filter .

I am using the older filter format 
```
"filter": "AGE > 25"
```

vikaxsh: it is expected behaviour

vikaxsh:
> ```json
>   "filter_input": {
>                     "logical_operator": "AND",
>                     "conditions": [
>                         {
>                             "column": "isActive",
>                             "operator": "=",
>                             "value": "false"
>                         },
>                         {
>                             "column": "Category",
>                             "operator": "=",
>                             "value": "Spaces"
>                         }
>                     ]
>                 }
> ```
> 
> For this case the ` "value": "Spaces"` is stored in the database with a whitespace `"value": " Spaces "`. Do we need to handle this case ?
> 
> In MSSQL the comparison support is case-insensitive and for Oracle for `BIT` it supports `0/1/"TRUE"/"FALSE"`, so we need to check if these cases work

since we are expecting json input for filter, values should be correct, we should not handle whitespaces

destination/filter.go
vaibhav-datazip: why shouldn't it filter out the data in cdc and incremental as well ? if the reasoning is it will break the current downstream pipeline, the fix can be state versioning .

vikaxsh: how  it would break downstream pipelines? cdc and incremental filter is supposed to work for new jobs only

destination/filter.go
```diff
@@ -0,0 +1,200 @@
...
+		record types.RawRecord,
+		_ int,
+	) error {
+		// Delete operations should always be synced, regardless of filter conditions.
+		// This is because:
+		// 1. If a record was previously synced, we need to delete it
+		// 2. If a record was never synced (filtered out), deleting it is a no-op anyway
+		// 3. Delete operations in MongoDB CDC only contain the document key, not full document fields,
+		//    so filter conditions that require those fields cannot be evaluated
+		if record.OperationType == "d" {
```
vaibhav-datazip: operation type is not defined in raw records

destination/filter.go
```diff
@@ -0,0 +1,200 @@
...
+		if dataType == types.Unknown {
+			return nil, fmt.Errorf("unknown datatype for column [%s]", cond.Column)
+		}
+
+		parsedVal, err := typeutils.ReformatValue(dataType, cond.Value)
+		if err != nil && err != typeutils.ErrNullValue {
+			return nil, fmt.Errorf(
+				"failed to parse filter value for column [%s]: %s",
+				cond.Column,
+				err,
```
vaibhav-datazip: change this to single line, and other places as well where error is being returned

destination/filter.go
```diff
@@ -0,0 +1,200 @@
...
+	return filtered, err
+}
+
+// matches evaluates AND / OR logic from filter conditions
+func matches(
+	record types.RawRecord,
+	conditions []parsedCondition,
+	logicalOp string,
+) bool {
+	isAnd := !strings.EqualFold(strings.TrimSpace(logicalOp), "OR")
```
vaibhav-datazip: the trim space here is repetition we , parsed it during fetching already right ?

destination/filter.go
```diff
@@ -0,0 +1,200 @@
...
+	if recordVal == nil || filterVal == nil {
+		switch operator {
+		case "=":
+			return recordVal == nil && filterVal == nil
+		case "!=":
+			return recordVal != nil || filterVal != nil
+		default:
+			return false
+		}
+	}
```
vaibhav-datazip: shouldn't we sync the column , if the column value in db is nil and our filter is like column < value ?

destination/iceberg/iceberg.go
```diff
@@ -352,8 +357,14 @@ func (i *Iceberg) FlattenAndCleanData(ctx context.Context, records []types.RawRe
...
 	}
+	if i.options.ApplyFilter {
+		records, err = destination.FilterRecords(ctx, records, filter, isLegacy, i.schema)
+		if err != nil {
+			return false, nil, nil, fmt.Errorf("failed to filter records: %s", err)
+		}
+	}
 
-	return schemaDifference, records, recordsSchema, err
+	return schemaDifference, records, recordsSchema, nil
```
vaibhav-datazip, line 380: we can return err as earlier , as it will be nil only

destination/parquet/parquet.go
```diff
@@ -402,9 +407,23 @@ func (p *Parquet) FlattenAndCleanData(ctx context.Context, records []types.RawRe
...
+	if p.options.ApplyFilter {
+		records, err = destination.FilterRecords(ctx, records, filter, isLegacy, p.schema)
+		if err != nil {
+			return false, nil, nil, fmt.Errorf("failed to filter records: %s", err)
+		}
+	}
+	logger.Debugf(
+		"[FlattenAndCleanData] after filtering: records=%d",
+		len(records),
+	)
```
vaibhav-datazip: is this log required ?

drivers/mongodb/internal/backfill.go
```diff
@@ -398,36 +398,65 @@ func buildMongoCondition(cond types.Condition) bson.D {
...
+	case nil:
+		value = nil
+	case primitive.ObjectID:
+		value = v
+	case string:
+		// try ObjectID only if column is _id
+		if cond.Column == "_id" && len(v) == 24 {
+			if oid, err := primitive.ObjectIDFromHex(v); err == nil {
+				value = oid
+				break
```
vaibhav-datazip: in legacy filter we were removing the quotes surrounding value , in new changes we are not doing so, this might create problem of backward compatibility.

pkg/jdbc/jdbc.go
```diff
@@ -1093,68 +1094,92 @@ func SQLFilter(stream types.StreamInterface, driver string, thresholdFilter stri
...
+		// ---------- NULL handling ----------
+		if cond.Value == nil {
 			switch cond.Operator {
 			case "=":
 				return fmt.Sprintf("%s IS NULL", quotedColumn), nil
 			case "!=":
 				return fmt.Sprintf("%s IS NOT NULL", quotedColumn), nil
 			default:
-				return fmt.Sprintf("%s %s NULL", quotedColumn, cond.Operator), nil
+				return "", fmt.Errorf("operator %s not supported with NULL", cond.Operator)
```
vaibhav-datazip, line 1151: the older code replicated how database will handle null value for querying

destination/filter.go
vikaxsh: fixed after resolving conflicts with staging

destination/filter.go
vikaxsh: where?

destination/filter.go
vikaxsh: Only records that match the filter condition should be synced. If the user also wants to include records with nil values, they should explicitly add a condition like columnValue == nil in the filter itself.

destination/iceberg/iceberg.go
vikaxsh: can u explain more about this?

destination/parquet/parquet.go
vikaxsh: removed

drivers/mongodb/internal/backfill.go
vikaxsh: let's discuss

pkg/jdbc/jdbc.go
vikaxsh: let's discuss

vikaxsh: @hash-data do we need to sync record where columnValue == nil, when record not matches condition

drivers/mongodb/internal/backfill.go
vikaxsh: checked with this -> "filter": "\"order_id\" > 1", working fine

destination/iceberg/iceberg.go
vaibhav-datazip: just change nil to err

destination/filter.go
vishalm0509: But what else can we have ? if we are naming it `isAND` we can check equality with `AND` right ?

destination/filter.go
vishalm0509: It's a utility specific to destination filter (not global), so we can name it something meaningful so a new developer need not go to that utility to understand what's happening but understand from the name only.

destination/filter.go
vishalm0509: Same here

vishalm0509:
> ```json
>   "filter_input": {
>                     "logical_operator": "AND",
>                     "conditions": [
>                         {
>                             "column": "isActive",
>                             "operator": "=",
>                             "value": "false"
>                         },
>                         {
>                             "column": "Category",
>                             "operator": "=",
>                             "value": "Spaces"
>                         }
>                     ]
>                 }
> ```
> 
> For this case the ` "value": "Spaces"` is stored in the database with a whitespace `"value": " Spaces "`. Do we need to handle this case ?
> 
> In MSSQL the comparison support is case-insensitive and for Oracle for `BIT` it supports `0/1/"TRUE"/"FALSE"`, so we need to check if these cases work

Confirmed it's a DBMS related feature. So let's not handle it


> - SQL Server (MSSQL): Case sensitivity depends on the database/column collation. Most default installations use case-insensitive (CI) collations, so string comparisons in WHERE clauses are typically case-insensitive unless explicitly configured otherwise.
> - MySQL: Case sensitivity is determined by collation. Common collations like utf8mb4_general_ci are case-insensitive (ci), while cs or bin collations enforce case-sensitive comparisons.
> - PostgreSQL: String comparisons are case-sensitive by default. Case-insensitive searches require ILIKE, LOWER()/UPPER() transformations, or the citext data type.
> - Oracle: Case-sensitive by default. Case-insensitive behavior must be implemented using functions (e.g., LOWER()) or specific collation settings in newer versions.
> - SQLite: Behavior is mixed. Equality (=) is case-sensitive, while LIKE is case-insensitive for ASCII characters unless configured otherwise.

drivers/mongodb/internal/backfill.go
```diff
@@ -398,36 +398,65 @@ func buildMongoCondition(cond types.Condition) bson.D {
...
 
-		if strings.HasPrefix(val, "\"") && strings.HasSuffix(val, "\"") {
-			val = val[1 : len(val)-1]
-		}
-		if field == "_id" && len(val) == 24 {
-			if oid, err := primitive.ObjectIDFromHex(val); err == nil {
-				return oid
+	mongoOp, ok := opMap[cond.Operator]
+	if !ok {
+		panic(fmt.Sprintf("unsupported operator for mongo: %s", cond.Operator))
```
vishalm0509: SHould we return `err` here instead of panic and handle error properly for this function ?

constants/constants.go
```diff
@@ -48,6 +48,8 @@ const (
 	MSSQL    DriverType = "mssql"
 )
 
+// Drivers where filters are applied in memory after full refresh data is read.
+var FullRefreshPostReadFilterDrivers = []DriverType{S3, Kafka}
```
vishalm0509: `"FilterPostBackfill" or "FilterPostFullRefresh"` ?

destination/writers.go
```diff
@@ -225,12 +233,12 @@ func (wt *WriterThread) flush(ctx context.Context, buf []types.RawRecord) (err e
...
 	flushCtx, cancel := context.WithCancel(ctx)
 	defer cancel()
-
+	recordsCountBeforeFiltering := len(buf)
 	evolution, buf, threadSchema, err := wt.writer.FlattenAndCleanData(flushCtx, buf)
 	if err != nil {
 		return fmt.Errorf("failed to flatten and clean data: %s", err)
 	}
-
+	wt.stats.RecordsFiltered.Add(int64(recordsCountBeforeFiltering - len(buf)))
```
vishalm0509, line 246: We are maintaining this only for filtering happening in the Writer side. What about the ones in driver side.

destination/filter.go
```diff
@@ -0,0 +1,197 @@
...
+
+// filterConcurrently evaluates records in parallel
+func filterConcurrently(
+	ctx context.Context,
+	records []types.RawRecord,
+	conditions []parsedCondition,
+	logicalOp string,
+) ([]types.RawRecord, error) {
+	concurrency := runtime.GOMAXPROCS(0) * 16
+	var mu sync.Mutex
```
vishalm0509: Can we move from using mutex locks to channel based approach, or can we add it as an improvement for later.

vishalm0509:
Just noting it here for reference: this case is failing

**filter:**
```
  "filter_input": {
                    "conditions": [
                        {
                            "column": "SubCategory",
                            "operator": "=",
                            "value": "Ω Mega-Byte ⚡"
                        }
                    ]
                }
```

**Logs**

```
2026-02-12T11:18:54Z DEBUG Starting backfill from 1 to <nil> with filter: [SubCategory] = 'Ω Mega-Byte ⚡', args: []
2026-02-12T11:18:54Z DEBUG Starting backfill from <nil> to 1 with filter: [SubCategory] = 'Ω Mega-Byte ⚡', args: []
2026-02-12T11:18:54Z DEBUG Executing chunk query: SELECT * FROM [dbo].[FilterTestBench] WITH (READPAST) WHERE ((([ID] >= '1'))) AND ([SubCategory] = 'Ω Mega-Byte ⚡') ORDER BY [ID]
2026-02-12T11:18:54Z DEBUG Executing chunk query: SELECT * FROM [dbo].[FilterTestBench] WITH (READPAST) WHERE ((([ID] < '1'))) AND ([SubCategory] = 'Ω Mega-Byte ⚡') ORDER BY [ID]
```

vishalm0509:
```
  {
                "partition_regex": "",
                "stream_name": "FilterTestBench",
                "normalization": true,
                "filter": "Category = \"null\" OR SubCategory = \"Electronics\""
}
```

--- 

This particular filter is not working

types/catalog.go
```diff
@@ -43,8 +43,12 @@ type StreamMetadata struct {
 	StreamName     string `json:"stream_name"`
 	AppendMode     bool   `json:"append_mode,omitempty"`
 	Normalization  bool   `json:"normalization"`
-	Filter         string `json:"filter,omitempty"`
+	//legacy filter input
+	Filter string `json:"filter,omitempty"`
+	//new filter input
+	FilterInput *FilterInput `json:"filter_input,omitempty"`
```
vishalm0509: Should we add this change also to clear-destination check ? To this function `GetStreamsDelta()`

vishalm0509:
`{id: 1, status: "active"}` Let's say this record is synced with `sync_mode=full_refresh+cdc`

Now, we update it to `{id: 1, status: "inactive"}` and we get a cdc event, since we have a filter this is not synced. The data in destination is stale now. Is this expected ?

drivers/mongodb/internal/backfill.go
```diff
@@ -389,7 +389,7 @@ func generateMinObjectID(t time.Time) string {
 	return objectID.Hex()
 }
 
-func buildMongoCondition(cond types.Condition) bson.D {
+func buildMongoCondition(cond types.FilterCondition) bson.D {
```
vishalm0509: The `SQLFilter` now supports multiple filter conditions, but here in Mongodb we are restricting to 2. 

Shouldn't it be consistent ?

vishalm0509:
Synced `isActive = 0`, MSSQL Boolean type is `BIT`, so user can very likely set value to be either `0/1`
In destination it is synced as `FALSE` (not an issue)



**Full Refresh**
```
 {
                            "column": "isActive",
                            "operator": "=",
                            "value": 0
}
```

**During CDC**
Updated the value of the synced row from `isActive=0` -> `isActive=1`
```
2026-02-13T03:44:06Z INFO filtering records with filter: {LogicalOperator:or Conditions:[{Column:isActive Operator:= Value:0}]}
2026-02-13T03:44:06Z INFO retry attempt[1], retrying after 60.00 seconds due to err: post cdc error: context canceled: prev error: %!s(<nil>); failed closing writer[dbo.FilterTestBench]: failed to flush data while closing: failed to flatten and clean data: failed to filter records: failed to parse filter value for column [isactive]: found to be boolean, but value is not boolean : 0
```

--- 


No error during full refresh, but failing during CDC.

vikaxsh:
> Just noting it here for reference: this case is failing
> 
> **filter:**
> 
> ```
>   "filter_input": {
>                     "conditions": [
>                         {
>                             "column": "SubCategory",
>                             "operator": "=",
>                             "value": "Ω Mega-Byte ⚡"
>                         }
>                     ]
>                 }
> ```
> 
> **Logs**
> 
> ```
> 2026-02-12T11:18:54Z DEBUG Starting backfill from 1 to <nil> with filter: [SubCategory] = 'Ω Mega-Byte ⚡', args: []
> 2026-02-12T11:18:54Z DEBUG Starting backfill from <nil> to 1 with filter: [SubCategory] = 'Ω Mega-Byte ⚡', args: []
> 2026-02-12T11:18:54Z DEBUG Executing chunk query: SELECT * FROM [dbo].[FilterTestBench] WITH (READPAST) WHERE ((([ID] >= '1'))) AND ([SubCategory] = 'Ω Mega-Byte ⚡') ORDER BY [ID]
> 2026-02-12T11:18:54Z DEBUG Executing chunk query: SELECT * FROM [dbo].[FilterTestBench] WITH (READPAST) WHERE ((([ID] < '1'))) AND ([SubCategory] = 'Ω Mega-Byte ⚡') ORDER BY [ID]
> ```

added a todo

destination/writers.go
vikaxsh: we are not calculating for legacy filter as well, but can be added let's discuss

destination/filter.go
vikaxsh: how channel would enhance performance here?

vikaxsh:
> Synced `isActive = 0`, MSSQL Boolean type is `BIT`, so user can very likely set value to be either `0/1` In destination it is synced as `FALSE` (not an issue)
> 
> **Full Refresh**
> 
> ```
>  {
>                             "column": "isActive",
>                             "operator": "=",
>                             "value": 0
> }
> ```
> 
> **During CDC** Updated the value of the synced row from `isActive=0` -> `isActive=1`
> 
> ```
> 2026-02-13T03:44:06Z INFO filtering records with filter: {LogicalOperator:or Conditions:[{Column:isActive Operator:= Value:0}]}
> 2026-02-13T03:44:06Z INFO retry attempt[1], retrying after 60.00 seconds due to err: post cdc error: context canceled: prev error: %!s(<nil>); failed closing writer[dbo.FilterTestBench]: failed to flush data while closing: failed to flatten and clean data: failed to filter records: failed to parse filter value for column [isactive]: found to be boolean, but value is not boolean : 0
> ```
> 
> No error during full refresh, but failing during CDC.

if it is bool value in source use true and false in value not integer

destination/iceberg/iceberg.go
```diff
@@ -344,6 +344,11 @@ func (i *Iceberg) FlattenAndCleanData(ctx context.Context, records []types.RawRe
 		return diffThreadSchema.Load(), schemaMap, err
 	}
 
+	filter, isLegacy, err := i.stream.GetFilter()
+	if err != nil {
+		return false, nil, nil, fmt.Errorf("failed to parse stream filter: %s", err)
+	}
```
hash-data: move it after line 359

destination/parquet/parquet.go
```diff
@@ -348,8 +348,13 @@ func (p *Parquet) Close(ctx context.Context) error {
 
 // validate schema change & evolution and removes null records
 func (p *Parquet) FlattenAndCleanData(ctx context.Context, records []types.RawRecord) (bool, []types.RawRecord, any, error) {
+	filter, isLegacy, err := p.stream.GetFilter()
+	if err != nil {
+		return false, nil, nil, fmt.Errorf("failed to parse stream filter: %s", err)
+	}
```
hash-data: move it after line 414

destination/filter.go
```diff
@@ -0,0 +1,186 @@
...
+			filtered = append(filtered, record)
+			mu.Unlock()
+			return nil
+		}
+
+		match := matches(record, conditions, logicalOp)
+		if match {
+			mu.Lock()
+			filtered = append(filtered, record)
+			mu.Unlock()
```
hash-data: why not inplace data update?

protocol/sync.go
```diff
@@ -143,6 +143,10 @@ var syncCmd = &cobra.Command{
 		state.LogWithLock()
 		// TODO: record count also contain records which arrived in retry attempts, need to remove them
 		logger.Infof("Total records read: %d", pool.GetStats().ReadCount.Load())
+		filteredRecordsCount := pool.GetStats().RecordsFiltered.Load()
+		if filteredRecordsCount > 0 {
+			logger.Infof("Total records filtered: %d", filteredRecordsCount)
+		}
```
hash-data: do we need it ?

hash-data: and that is after synced records count?

types/catalog.go
hash-data: rename prev one to legacyFilter and create new var as filter

destination/iceberg/iceberg.go
```diff
@@ -353,7 +353,19 @@ func (i *Iceberg) FlattenAndCleanData(ctx context.Context, records []types.RawRe
 		return false, nil, nil, fmt.Errorf("failed to extract schema from records: %s", err)
 	}
 
-	return schemaDifference, records, recordsSchema, err
+	filter, isLegacy, err := i.stream.GetFilter()
+	if err != nil {
+		return false, nil, nil, fmt.Errorf("failed to parse stream filter: %s", err)
+	}
+
+	if i.options.ApplyFilter {
```
hash-data: can we take it once while creating writer instead of everytime passing in batch?

utils/typeutils/filter.go
```diff
@@ -0,0 +1,130 @@
...
+type parsedCondition struct {
+	column   string
+	operator string
+	value    any
+}
+
+// FilterRecords applies filtering ONLY for new filters.
+// For legacy filters, records are returned unchanged.
+// FilterRecords applies filtering ONLY for new filters.
+// For legacy filters, records are returned unchanged.
```
vishalm0509, line 23: Redundant comment

types/catalog.go
vikaxsh: do u mean if user changes the filter?

vishalm0509: Yes, filter is there, but `filter_config` is not there

utils/typeutils/reformat.go
```diff
@@ -276,7 +287,7 @@ func parseStringTimestamp(value string, isTimestampInDB bool) (time.Time, error)
 	if !isTimestampInDB && constants.LoadedStateVersion != 0 {
 		return time.Time{}, fmt.Errorf("failed to parse datetime from available formats: %s", err)
 	}
-
+	logger.Debugf("***Failed to parse datetime from available formats: %s", err)
```
vishalm0509:
```go
logger.Debugf("Failed to parse datetime from available formats: %s", err)
```

protocol/sync.go
```diff
@@ -190,6 +193,31 @@ func classifyStreams(catalog *types.Catalog, streams []*types.Stream, state *typ
...
 				return false
 			}
+			// TODO: move filter validation to validate method in types package
+			filter, isLegacy, err := elem.GetFilter()
+			if err != nil {
+				logger.Warnf("Skipping; Configured Stream %s failed to get filter: %s", elem.ID(), err)
+				return false
+			}
+			if !isLegacy {
+				for _, cond := range filter.Conditions {
```
vaibhav-datazip, line 207: why are we not validating here only and throwing error if filter conditions are greater than 0 instead of adding switch cases when len(filter.Conditions) > 2:

utils/typeutils/reformat.go
```diff
@@ -120,6 +120,18 @@ func ReformatValue(dataType types.DataType, v any) (any, error) {
 	}
 }
 
+// ParseFilterValue parses v into dataType strictly and returns the typed value.
+// For timestamp types, an unparseable string always returns an error (unlike
```
vaibhav-datazip: looks like an incomplete comment

types/stream_configured.go
```diff
@@ -206,11 +229,11 @@ func (s *ConfiguredStream) GetFilter() (Filter, error) {
 	var FilterRegex = regexp.MustCompile(`^(?:"([^"]*)"|(\w+))\s*(>=|<=|!=|>|<|=)\s*((?:"[^"]*"|-?\d+\.\d+|-?\d+|\.\d+|\w+))\s*(?:((?i:and|or))\s*(?:"([^"]*)"|(\w+))\s*(>=|<=|!=|>|<|=)\s*((?:"[^"]*"|-?\d+\.\d+|-?\d+|\.\d+|\w+)))?\s*$`)
```
vaibhav-datazip, line 229: can we take this out so that it wont be compiled every time we call this function

types/stream_configured.go
```diff
@@ -183,10 +181,35 @@ func (s *ConfiguredStream) Cursor() (string, string) {
...
+//
+// Return values:
+//   - FilterConfig: parsed filter definition (may be empty if no filter is configured)
+//   - isLegacy:   true if the filter came from the legacy string field, false if it
+//     came from the new structured FilterConfig field
+//   - error:      non-nil only if the legacy string filter is non-empty and fails
+//     to parse; new structured filters do not return parse errors here.
+func (s *ConfiguredStream) GetFilter() (FilterConfig, bool, error) {
+	//new filter input — only apply structured filter_config when normalization is enabled
+	if s.StreamMetadata.Normalization && s.StreamMetadata.FilterConfig != nil && len(s.StreamMetadata.FilterConfig.Conditions) > 0 {
```
vaibhav-datazip, line 201: can we add the test case functions which detects race condition here in the unit tests as well.

vikaxsh: no need for that even before this fix the race condition wasn't affecting any use case

types/stream_configured.go
vikaxsh: it is legacy code and won't be used further so no need for that

PR #768, opened 2026-01-16
fix: update performance test CI to use destination as iceberg_destination.json

vishalm0509:
# Description

Updated the `"Create destination"` step to write `iceberg_destination.json` instead of `destination.json`, matching the test configuration in `utils/testutils/test_utils.go.`

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -7,10 +7,6 @@ on:
     paths:
     - '**/*.go'
     - '**/*.java'
-  # TODO: remove after testing
-  pull_request:
-    branches:
-    - "*"
 
 permissions:
   actions: read


.github/workflows/performance-test.yml
```diff
@@ -65,6 +69,16 @@ jobs:
         sudo resolvectl dns tun0 ${{ secrets.VPN_DNS_SERVER }}
         sudo resolvectl domain tun0 ~private.postgres.database.azure.com
 
+    - name: Wake up server
+      if: matrix.driver != 'oracle'
+      run: |
+        mkdir -p ~/.ssh
+        echo "${{ secrets.TESTING_SERVER_SSH_KEY }}" | base64 --decode > ~/.ssh/key.pem
+        chmod 600 ~/.ssh/key.pem
+        ssh -o StrictHostKeyChecking=no -i ~/.ssh/key.pem testing-server-1@172.31.32.11 "wake-up ${{ matrix.instance_id }}"
```
vikaxsh: we should not hardcode server ip

.github/workflows/performance-test.yml
```diff
@@ -65,6 +69,16 @@ jobs:
         sudo resolvectl dns tun0 ${{ secrets.VPN_DNS_SERVER }}
         sudo resolvectl domain tun0 ~private.postgres.database.azure.com
 
+    - name: Wake up server
+      if: matrix.driver != 'oracle'
+      run: |
+        mkdir -p ~/.ssh
+        echo "${{ secrets.TESTING_SERVER_SSH_KEY }}" | base64 --decode > ~/.ssh/key.pem
```
vikaxsh, line 76: shouldn't we delete private key after use

.github/workflows/performance-test.yml
vikaxsh: do we need to add ssh time out here?

.github/workflows/performance-test.yml
```diff
@@ -123,6 +137,15 @@ jobs:
         path: ./drivers/${{ matrix.driver }}/internal/testdata/benchmarks.json
         retention-days: 90
    
+    - name: Sleep server
+      if: always() && matrix.driver != 'oracle'
+      run: |
+        mkdir -p ~/.ssh
+        echo "${{ secrets.TESTING_SERVER_SSH_KEY }}" | base64 --decode > ~/.ssh/key.pem
+        chmod 600 ~/.ssh/key.pem
+        ssh -o StrictHostKeyChecking=no -i ~/.ssh/key.pem testing-server-1@172.31.32.11 "sleep-off ${{ matrix.instance_id }}"
```
schitizsharma: Try to use absolute path when doing a remote execution `/usr/local/bin/sleep-off` this is more common pattern in System Administration.

.github/workflows/performance-test.yml
vikaxsh: how are we ensuring that it succeeds?

schitizsharma: Try to use absolute path when doing a remote execution /usr/local/bin/wake-up this is more common pattern in System Administration.

.github/workflows/performance-test.yml
vishalm0509: Github runners are ephemeral. So after the test the VM is destroyed

.github/workflows/performance-test.yml
vishalm0509: Done

.github/workflows/performance-test.yml
vishalm0509: Done

vishalm0509: Added to secrets

vishalm0509: It's async, once triggered the `wake-up` and `sleep` happens async. If that fails, sync will fail

vishalm0509: Done

vikaxsh (approved):

PR #770, opened 2026-01-16
Change binlog check from metadata to image

Vladislaff:
# Description

The [documentation](https://olake.io/docs/connectors/mysql/setup/generic/) specifies that binlog-row-metadata is optional and binlog-row-image is required to be FULL, but the check is done incorrectly. Right now IsCDCSupported checks for binlog-row-metadata instead of binlog-row-image

## Type of change


nayanj98:
The doc mentions the wrong info will update the doc and hence closing this PR

PR #774, opened 2026-01-19
fix: fixed parquet hardcoded path issue

### LINKED ISSUE #736: issue: parquet test connection happens on different folder
## Description
currently parquet test connection happens on different s3 folder. 
in some organization permission are folder level so the test connection must happen on same folder that user provided in s3 path field.

## Steps to reproduce
-> run OLake with parquet writer or run check connection command

## Expected behavior
It must create test folder inside the folder that user have provided in s3_path field.

saksham-datazip:
# Description
This PR fixes a bug in the Parquet destination where the connection check would ignore the configured `s3_path` (Prefix) and always write the test file to a hardcoded `olake_writer_test/` directory at the bucket root.
This caused connection checks to fail for users who only have permissions to write to specific subdirectories (e.g., `restricted-folder/`).
The fix ensures that if `s3_path` is configured, the test file key is constructed as: `{s3_path}/olake_writer_test/{timestamp}.txt`
Fixes #736
## Type of change
# How Has This Been Tested?
I verified this change by reproducing the failure locally with MinIO and confirming the fix with the OLake CLI.
1. I set my store location to be `working-temp-file`.
2. **Before Fix:** I observed that it was incorrectly creating the file in `olake_writer_test/` (ignoring the restricted folder).
3. **Fix:** I edited [destination/parquet/parquet.go](cci:7://file:///Users/apple/olake/destination/parquet/parquet.go:0:0-0:0) to prepend the configured prefix.
4. **After Fix:** I ran the check command again and confirmed it worked correctly, creating the file inside the restricted folder.
# Screenshots or Recordings

before changes->
image

after changes->
image

## Documentation

## Related PR's (If Any):**

### DIFF
--- a/pkg/binlog/filter.go
+++ b/pkg/binlog/filter.go
@@ -19,6 +19,11 @@ const (
 
 )
 
+// TableMapEvent wraps replication.TableMapEvent so we can define receiver methods (unsignedMap, isNumericColumn).
+type TableMapEvent struct {
+	*replication.TableMapEvent
+}
+
 // ChangeFilter filters binlog events based on the specified streams.
 type ChangeFilter struct {
 	streams       map[string]types.StreamInterface // Keyed by "schema.table"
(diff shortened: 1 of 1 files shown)


destination/parquet/parquet.go
```diff
@@ -213,7 +213,7 @@ func (p *Parquet) Check(_ context.Context) error {
 	}
 	// test for s3 permissions
 	if p.s3Client != nil {
-		testKey := fmt.Sprintf("olake_writer_test/%s", utils.TimestampedFileName(".txt"))
+		testKey := filepath.Join(p.config.Path, "olake_writer_test/%s", utils.TimestampedFileName(".txt"))
```
vikaxsh: can you verify is it correct path for s3?

vikaxsh (approved):

PR #788, opened 2026-01-21
fix: issue with dedup when schema evolution happens

hash-data:
# Description

When schema evolution occurs in the Iceberg writer and the same record gets updated simultaneously, queries return duplicate records. This data inconsistency occurs because:

- All equality delete files were being committed first, followed by all data files
- This incorrect ordering leads to improper sequence number mapping in Iceberg manifests
- As a result, delete operations are not correctly applied to their corresponding data files, causing duplicates to appear in query results

The previous implementation committed files in batches:
1. All delete files from all writer completions were added first
2. Then all data files from all writer completions were added
3. This broke the relationship between data files and delete files from the same writer completion

Example scenario:
- Writer completion 1: `e1` (data), `d1` (delete)
- Writer completion 2: `e2` (data), `d2` (delete)
- Previous behavior: `e1, e2`, d1, d2 → incorrect sequence, causing `d1` to potentially delete rows from `e2`
- Expected behavior: `e1, d1, e2, d2` → correct sequence where each delete applies to its corresponding data

## Solution

This PR implements a single atomic transaction that maintains the correct ordering of equality delete files and data files according to their writer completion:

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Tested inserting and updating the same record along with schema evolution.
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
+++ b/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
@@ -188,12 +188,13 @@ public class IcebergTableOperator {
           }
   
           RowDelta delta = txn.newRowDelta();
-  
-          if (eqDeletes != null) {
-            eqDeletes.forEach(delta::addDeletes);
-          }
+
           if (dataFiles != null) {
             dataFiles.forEach(delta::addRows);
+          }
+
+          if (eqDeletes != null) {
+            eqDeletes.forEach(delta::addDeletes);
           }
   
           delta.commit();
@@ -314,9 +315,13 @@ public class IcebergTableOperator {
                }
 
                DataFile dataFile = dataFileBuilder.build();
-               filesToCommit.get(filesToCommit.size() - 1).second().add(dataFile);
+               if (filesToCommit.size() > 0) {
+                filesToCommit.get(0).second().add(dataFile);
+               } else {
+                filesToCommit.add(Pair.of(new ArrayList<DeleteFile>(), new ArrayList<>(Arrays.asList(dataFile))));
+               }
                LOGGER.info("Thread {}: accumulated data file {} (total: {})", threadId, filePath,
-                         filesToCommit.get(filesToCommit.size() - 1).second().size());
+                         filesToCommit.get(0).second().size());
           } catch (Exception e) {
                String errorMsg = String.format("Thread %s: failed to accumulate data file %s: %s", threadId,
                          filePath, e.getMessage());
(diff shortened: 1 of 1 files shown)


destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -266,9 +314,9 @@ public void accumulateDataFiles(String threadId, Table table, String filePath,
                }
 
                DataFile dataFile = dataFileBuilder.build();
-               dataFiles.add(dataFile);
+               filesToCommit.get(filesToCommit.size() - 1).second().add(dataFile);
```
shubham19may: Arrow path can crash: filesToCommit.get(filesToCommit.size() - 1) on empty list

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -118,55 +119,101 @@ public void commitThread(String threadId, Table table) {
...
-        // Refresh table before committing
-        table.refresh();
-
-        // Check if any WriteResult has delete files
-        boolean hasDeleteFiles = totalDeleteFiles > 0;
-
-        if (hasDeleteFiles) {
-          RowDelta rowDelta = table.newRowDelta();
-          // Add all data and delete files from all WriteResults
-          dataFiles.forEach(rowDelta::addRows);
```
shubham19may, line 149: we first used to add data then deletes, but now its first deletes and data
```
          if (eqDeletes != null) {
            eqDeletes.forEach(delta::addDeletes);
          }
          if (dataFiles != null) {
            dataFiles.forEach(delta::addRows);
          }
```

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
hash-data: arrow i have not tested, will test it

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
hash-data: Each file have same seq number when we add in row delta, so don't think it matterss which we add first, only matter is row delta commit sequence

hash-data: But let me change the order to keep it same

shubham19may: Lets keep it as it was previously

PR #789, opened 2026-01-21
fix:  mssql cdc capture-instance selection for schema evolution

vishalm0509:
# Description

Fixes # (issue)

## Type of change

Update MSSQL CDC streaming to correctly handle multiple capture instances created during schema evolution.
Select the newest capture instance whose `start_lsn ≤ the current fromLSN` to continue reading from a valid instance.

When a newer capture instance exists, clamp `targetLSN` to the newer instance’s `start_lsn` so rows that conceptually belong to the new schema are not read from the old capture instance.

**Note**: we expect column-level data loss (e.g., new columns missing) in the LSN range between the DDL and when the new capture instance becomes active.

# How Has This Been Tested?
- Tested ALTER (new column and delete)
- Tested data type change

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/drivers/mssql/internal/cdc.go
+++ b/drivers/mssql/internal/cdc.go
@@ -70,15 +70,15 @@ func (m *MSSQL) PreCDC(ctx context.Context, streams []types.StreamInterface) err
 		return fmt.Errorf("failed to get MSSQL current max LSN: %w", err)
 	}
 
+	// check if CDC is enabled for each stream
 	for _, stream := range streams {
-		// check if CDC is enabled for each stream
 		enabled, err := m.validateCDCStream(ctx, stream.Namespace(), stream.Name())
 		if err != nil {
 			return fmt.Errorf("failed to validate CDC for stream %s.%s: %s", stream.Namespace(), stream.Name(), err)
 		}
 
 		if !enabled {
-			return fmt.Errorf("CDC is not enabled for table %s.%s", stream.Namespace(), stream.Name())
+			return fmt.Errorf("CDC is not enabled for stream %s.%s", stream.Namespace(), stream.Name())
 		}
 
 		// Initialize LSN for each stream if not present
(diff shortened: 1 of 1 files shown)


drivers/mssql/internal/cdc.go
```diff
@@ -131,16 +131,32 @@ func (m *MSSQL) StreamChanges(ctx context.Context, stream types.StreamInterface,
...
+
 	for i := len(captures) - 1; i >= 0; i-- {
 		// LSNs are fixed-length hex strings and can be compared lexicographically
 		if captures[i].startLSN <= fromLSN {
 			selectedCapture = &captures[i]
+
+			// If a newer capture instance exists, restrict the upper bound of this
+			// read to that newer instance's start LSN.
+			nextIdx := i + 1
+			if nextIdx < len(captures) {
```
vikaxsh: are we sure about order of capture instances?

drivers/mssql/internal/cdc.go
```diff
@@ -131,16 +131,32 @@ func (m *MSSQL) StreamChanges(ctx context.Context, stream types.StreamInterface,
...
+				newerCapture := captures[nextIdx]
```
vikaxsh: shouldn't we add a warn log about update of capture instance

drivers/mssql/internal/cdc.go
```diff
@@ -131,16 +131,32 @@ func (m *MSSQL) StreamChanges(ctx context.Context, stream types.StreamInterface,
...
+				if targetLSN > newerCapture.startLSN {
```
vikaxsh: we can remove nested  checks as well

drivers/mssql/internal/cdc.go
vishalm0509: Yes, we are using ORDER BY in `start_lsn` in the query

```sql
ORDER BY
    s.name ASC,
    t.name ASC,
    c.start_lsn ASC;
```

drivers/mssql/internal/cdc.go
vishalm0509: Right, done

drivers/mssql/internal/cdc.go
vishalm0509: Done, I think prev one was more readable. What do you think

vikaxsh:
lgtm

drivers/mssql/internal/cdc.go
```diff
@@ -133,16 +133,36 @@ func (m *MSSQL) StreamChanges(ctx context.Context, stream types.StreamInterface,
...
+		// Skip if this capture started after fromLSN
+		if captures[i].startLSN > fromLSN {
+			continue
 		}
+
+		// Select the capture instance
+		selectedCapture = &captures[i]
+
+		// If a newer capture instance exists, restrict the targetLSN to the newer instance's startLSN
+		nextIdx := i + 1
```
ImDoubD-datazip: please use proper naming, change i to something meaningful

drivers/mssql/internal/cdc.go
```diff
@@ -133,16 +133,36 @@ func (m *MSSQL) StreamChanges(ctx context.Context, stream types.StreamInterface,
...
+		if nextIdx < len(captures) && targetLSN > captures[nextIdx].startLSN {
+			newerCapture := captures[nextIdx]
```
ImDoubD-datazip: directly use it.

ImDoubD-datazip:
LGTM

nayanj98:
Tested schema evolution by adding new columns, dropping columns and also data type changes, approved from my end.

PR #798, opened 2026-01-28
feat: kafka schema registry introduced for avro based data format

ImDoubD-datazip:
# Description

Schema registry feature has been added (using schema registry api) for kafka sync, specifically for avro based data format. 

Currently supported schema registry is specifically `Confluent Schema Registry`. For this, the data must be in `Confluent Wire Format` for confluent schema registry to work, else the fallback is JSON.

## Type of change

# How Has This Been Tested?

This has been tested on both json message, avro message (with schema provided), json message (with json schema provided). In case of multiple types present for key in avro data format, changes in avro.go file is done.

to handle this, where the emp_id allows both int, string values :
```
decoded Avro: map[emp_id:map[string:EMP-2009] emp_name:User_9 emp_tenure1:1.2857152 emp_tenure2:1.2857152857142857 event_name:event_9 event_triggered_tm:2026-02-01T11:50:14.076Z event_type:UI_CLICK is_recovered:false log:Volume test request_ip:192.168.9.0 server_type:PROD]
```

it should come as this:
```
decoded Avro record: map[emp_id:EMP-2009 emp_name:User_9 emp_tenure1:1.2857152 emp_tenure2:1.2857152857142857 event_name:event_9 event_triggered_tm:2026-02-01T11:50:14.076Z event_type:UI_CLICK is_recovered:false log:Volume test request_ip:192.168.9.0 server_type:PROD]
```

# Screenshots or Recordings

Normalization on: 

- Avro based topic
Screenshot 2026-01-28 at 12 00 31 PM

Screenshot 2026-01-28 at 11 59 37 AM

Screenshot 2026-01-28 at 11 59 09 AM

- JSON based topic
Screenshot 2026-01-28 at 12 05 45 PM

Screenshot 2026-01-28 at 12 06 07 PM

Normalization off:

- Avro based topic
Screenshot 2026-01-28 at 12 03 26 PM

- JSON based topic
Screenshot 2026-01-28 at 12 03 57 PM

## Documentation

## Related PR's (If Any):


shubham19may:
gpt-5.1-codex-max

Findings
- High: Avro/JSON payloads that are valid primitives (e.g., string, int, union with null) will be dropped. `parseKafkaData` requires the decoded value to be a `map[string]interface{}`, otherwise it returns an error and the record is skipped. Confluent Avro allows primitive schemas; those will now never be ingested. Example guard that rejects anything but map:  
```261:272:drivers/kafka/internal/cdc.go
if message.Value != nil {
    valDecoded, err := parseData(message.Value)
    if err != nil {
        return nil, "", err
    }
    if vm, ok := valDecoded.(map[string]interface{}); ok {
        messageValue = vm
    } else {
        return nil, "", fmt.Errorf("expected map[string]interface{} for message value, got %T", valDecoded)
    }
}
```
- Medium: Confluent-wire-format messages without a configured registry are parsed as JSON after stripping the 5-byte header. For Avro data this will consistently fail JSON decoding and silently skip the record instead of surfacing a configuration error or falling back to raw bytes. This makes Avro topics unreadable unless registry is configured, even though the fallback behavior is advertised as JSON when the data isn’t Confluent wire format.  
```221:225:drivers/kafka/internal/cdc.go
if isConfluentWireFormat(data) {
    if k.schemaRegistryClient == nil {
        return decodeJSONMessage(data[5:])
    }
```
- Low: Root `go.mod` drops the `aws-sdk-go-v2/service/s3` dependency and related transitive modules, unrelated to the Kafka change. If the root module still builds the S3 driver, this can break the build; consider confirming this removal or isolating it to a separate tidy.  
```395:418:go.mod
-    github.com/aws/aws-sdk-go-v2/service/s3 v1.88.4
...
-    github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream v1.5.1 // indirect
-    github.com/aws/aws-sdk-go-v2/internal/v4a v1.2.3 // indirect
-    github.com/aws/aws-sdk-go-v2/service/internal/checksum v1.2.3 // indirect
-    github.com/aws/aws-sdk-go-v2/service/internal/s3shared v1.16.3 // indirect
```

Tests
- Not run. Recommend covering Avro record and primitive payloads (value and key) with/without schema registry to verify decoding paths and ensure non-Avro messages still pass through.

shubham19may:
Gemini-3-pro

The PR adds support for **Confluent Schema Registry** to the Kafka source, specifically targeting **Avro** serialization. The implementation correctly handles the Confluent Wire Format (Magic Byte + Schema ID), fetches schemas by ID, and falls back to JSON if the format is not detected.

Here is the review of the changes:

### **Summary of Changes**
-   **Core Logic**: Added `SchemaRegistryClient` in `pkg/kafka` to fetch and cache schemas.
-   **Driver**: Updated `drivers/kafka` to detect Confluent Wire Format in `CDC` and decode Avro messages using `goavro`.
-   **Configuration**: Added `schema_registry` config (Endpoint, Username, Password) and updated UI Schema.
-   **Utilities**: Added `typeutils.ConvertAvroRecord` to recursively convert Avro records to JSON-compatible maps.

### **Code Review**

#### **1. Binary Data Handling in Avro Conversion**
In `utils/typeutils/avro.go`, there is a conversion from `[]byte` to `string`:

```go:utils/typeutils/avro.go
	case []byte:
		return string(val)
```

**Observation**: This assumes that all `bytes` or `fixed` types in the Avro schema contain valid UTF-8 strings. If the Avro message contains actual binary data (e.g., an image, a protobuf payload, or encrypted data), casting it to `string` may result in invalid UTF-8 sequences. When this is later marshaled to JSON (which seems to be the downstream intent), `json.Marshal` may error or replace invalid bytes, corrupting the data.

**Recommendation**: Standard `encoding/json` automatically handles `[]byte` by encoding it as a **Base64 string**. It is generally safer to return the `[]byte` as-is and let the JSON marshaler handle the Base64 conversion, unless you are certain that `bytes` fields will always be text.

```go
    // Suggestion: Remove this case or ensure it's safe
	// case []byte:
	// 	return string(val) 
```

#### **2. Schema Registry Client**
The implementation in `pkg/kafka/schema_registry.go` looks solid:
-   It correctly defaults to `AVRO` if `schemaType` is missing in the response (compliant with Confluent API).
-   It implements caching using `sync.Map`, which is good for performance given schemas are immutable by ID.
-   **Nit**: The `http.Client` is created with a 30s timeout, which is reasonable.

#### **3. Kafka Driver Integration**
-   `drivers/kafka/internal/cdc.go`: The `parseKafkaData` logic efficiently checks for the wire format (`isConfluentWireFormat`) before attempting schema registry lookups. The fallback to `decodeJSONMessage` ensures backward compatibility for topics that mix formats or don't use the registry.
-   `drivers/kafka/internal/kafka.go`: The `extractSchemaID` correctly reads the big-endian uint32.

### **Conclusion**
The implementation is robust and follows the specifications well. The only significant point to verify is the `[]byte` to `string` conversion in `utils/typeutils/avro.go` to avoid data corruption for binary fields.

**Action**: I'll create a todo to verify the `[]byte` conversion behavior or bring it to your attention as a comment if I were posting this on GitHub. Since I am the assistant, I will proceed assuming this behavior is either intended (legacy reasons) or I should fix it if you want me to.

For now, I will mark the review as **Approved with a note on binary data**.

Do you want me to attempt to fix the `[]byte` conversion in `utils/typeutils/avro.go` to be safer (i.e., remove the manual string conversion)?

shubham19may:
gpt-5.2-high


Plan: I’ll first diff your current branch against `staging` to see exactly what changed, then walk the Kafka ingest path end-to-end (message decode → schema registry lookup/cache → fallback JSON → typing), and finally give you a PR-style review (correctness, edge cases, performance, security, tests/docs).

[2 tools called]

Nice—diff is scoped and clearly centered on Kafka ingest plus new `pkg/kafka/schema_registry.go` + `utils/typeutils/avro.go`. Next I’ll inspect the actual diffs for the Kafka pipeline and the new schema registry client/decoder, then I’ll validate the config/spec/UI schema changes and call out any unrelated edits (e.g., MSSQL CDC) for PR hygiene.

[2 tools called]

Good—core change is a new “decode path” in Kafka CDC that detects Confluent wire format, fetches schema by ID, and decodes Avro via `goavro`, with a schema cache in `SchemaRegistryClient`. Next I’ll review the remaining diffs (types/spec/go.mod) and then I’ll call out a few correctness/UX edge cases I’m already spotting around fallback behavior and non-record payloads.

[8 tools called]

### PR review: Kafka source — Confluent Schema Registry (Avro/JSON) support

## Summary
- **What’s added**: Confluent Schema Registry lookup-by-ID with caching (`pkg/kafka/schema_registry.go`) + Confluent wire-format detection (`isConfluentWireFormat` / `extractSchemaID`) and Avro decode via `goavro`, wired into Kafka CDC parsing (`drivers/kafka/internal/cdc.go`).
- **Config/UI**: New `schema_registry` config section added to Kafka spec + UI schema (with password widget).
- **Behavior**: If payload is Confluent wire format, it attempts schema-registry-based decode; otherwise it treats data as JSON.

## High-signal issues / changes requested
- **Wire-format + no schema registry currently falls back to JSON incorrectly**
  - In `parseKafkaData`, when `isConfluentWireFormat(data)` and `k.schemaRegistryClient == nil`, it calls `decodeJSONMessage(data[5:])`.
  - For **Confluent Avro wire format**, `data[5:]` is **binary Avro**, not JSON—this will fail and you’ll drop messages (warn + continue) even though the message is correctly serialized for SR.
  - **Recommendation**: if wire format is detected and SR isn’t configured, return a clear error like “schema registry required for Confluent wire format”, or treat it as raw bytes (but don’t assume JSON).

- **Value decoding assumes a JSON object / Avro record**
  - `decodeJSONMessage` returns `map[string]interface{}` only; arrays/primitives will fail.
  - `parseKafkaData` also errors if decoded **value** isn’t `map[string]interface{}` (even if Avro schema is primitive/array).
  - **Recommendation**: either (a) explicitly document “Kafka message value must be a record/object”, or (b) support non-object payloads by wrapping, e.g. `{ "_value": decoded }`.

- **Avro logical types / bytes conversion risks data loss**
  - `utils/typeutils/avro.go`:
    - `*big.Rat` decimals are converted to `float64` → **precision loss**.
    - `[]byte` converts to `string` → can corrupt non-UTF8 bytes; also loses the “binary” signal.
  - **Recommendation**:
    - For decimals, serialize as **string** (preserve exact value).
    - For bytes, keep as `[]byte` and let JSON marshalling base64 it (or explicitly base64).

## Design / implementation notes (looks good)
- **Schema cache**: `sync.Map` caching by schema ID is a good first cut.
- **SR API usage**: `/schemas/ids/{id}` + `schemaType` defaulting to Avro matches Confluent behavior.
- **Topic discover**: Skipping internal topics (`topic.Internal` + `_schemas`, `__amazon_msk_canary`) is a nice UX improvement.
- **Build sanity**: `go test ./...` passes on this diff.

## PR hygiene / scope concerns
- **Unrelated changes bundled in**
  - `drivers/mssql/internal/cdc.go` has behavior changes (not just nits)—this should be **a separate PR**.
  - `go.mod` and `drivers/kafka/go.mod` have significant dependency churn (AWS/grpc/protobuf versions, removal of `service/s3`, etc.). Even if it’s from `go mod tidy`, it’s noisy and complicates review.
  - **Recommendation**: split/revert unrelated dependency updates and MSSQL CDC changes from this PR.

## Tests I’d want before approving
- **Unit tests**:
  - `extractSchemaID` / `isConfluentWireFormat` (valid/invalid inputs).
  - SR client caching (two decodes with same ID makes 1 HTTP call).
  - Avro decode conversion (union, decimal, bytes).
- **Integration-ish** (optional): consume a known Confluent Avro-serialized message and assert decoded map output.

## Config/spec feedback
- **Spec/UI** additions look consistent:
  - `schema_registry.endpoint/username/password` in `drivers/kafka/resources/spec.json`
  - UI grid + password widget in `utils/spec/uischema.go`

If you want, I can also propose a small patch to fix the “wire format + SR missing” behavior and make value parsing tolerant (without changing downstream schema resolution expectations).

go.mod
```diff
@@ -38,6 +37,7 @@ require (
 	github.com/golang/snappy v0.0.4 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/asmfmt v1.3.2 // indirect
+	github.com/linkedin/goavro/v2 v2.15.0 // indirect
```
shubham19may: why the base go.mod needs to be changed?

pkg/kafka/types.go
```diff
@@ -34,3 +37,12 @@ type CustomGroupBalancer struct {
 	readerIndex         int
 	partitionIndex      map[string]types.PartitionMetaData
 }
+
+// SchemaRegistryClient holds the schema registry client information
+type SchemaRegistryClient struct {
+	endpoint   string
```
shubham19may: SchemaRegistryConfig reuse this

pkg/kafka/schema_registry.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// fetch schema from registry
+	url := fmt.Sprintf("%s/schemas/ids/%d", c.endpoint, schemaID)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	// set basic auth if credentials provided
+	if c.username != "" && c.password != "" {
```
shubham19may: basic auth is going to be the least preferred way for enterprises. We need to support all the ways for auth for schema registry if there are other ways.

drivers/kafka/internal/kafka.go
```diff
@@ -395,3 +413,20 @@ func (k *Kafka) getReaderAssignedPartitions(ctx context.Context, readerIndex int
 
 	return assigned, nil
 }
+
+// extractSchemaID extracts schema ID from Confluent wire format message
+// Wire format: [magic byte (0x00)] [4-byte schema ID (big-endian)] [payload]
+func extractSchemaID(data []byte) (uint32, error) {
+	if len(data) < 5 {
```
shubham19may: these are already covered in the above method. isConfluentWireFormat. so not needed

go.mod
shubham19may: @hash-data we need to refactor the whole setup to not include libraries like these in the base go.mod. 

This will explode the size of overall go executable

utils/typeutils/avro.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"math/big"
+)
+
+// convertAvroRecord recursively converts Avro record to JSON-compatible map
+func ConvertAvroRecord(record map[string]interface{}) map[string]interface{} {
+	result := make(map[string]interface{}, len(record))
+	for k, v := range record {
+		result[k] = ConvertAvroValue(v)
```
shubham19may: can't we use a generic types function we already have to convert any interface into its respective value?

drivers/kafka/internal/cdc.go
```diff
@@ -215,3 +214,116 @@ func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kafka.Reader,
...
+func decodeAvroMessage(data []byte, codec *goavro.Codec) (interface{}, error) {
+	nativeDatum, _, err := codec.NativeFromBinary(data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode Avro: %s", err)
+	}
+
+	if record, ok := nativeDatum.(map[string]interface{}); ok {
+		return typeutils.ConvertAvroRecord(record), nil
+	}
+	return nativeDatum, nil
```
shubham19may: if its not a map[string]interface{} then we just return it?

should't we stringify it?

utils/typeutils/avro.go
shubham19may: ex, datatype.go TypeFromValue function?

we can add rat type over there?

drivers/kafka/internal/cdc.go
```diff
@@ -215,3 +214,116 @@ func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kafka.Reader,
...
+			case types.SchemaTypeAvro:
+				decoded, err := decodeAvroMessage(data[5:], schema.Codec)
+				if err != nil {
+					return nil, err
+				}
+				if record, ok := decoded.(map[string]interface{}); ok {
+					return record, nil
+				}
+				// For keys or non-record values, it might be primitive
+				return decoded, nil
```
shubham19may:
```suggestion
				return decodeAvroMessage(data[5:], schema.Codec)
```

go.mod
ImDoubD-datazip: I was running a script in the base olake directory, that script required this. Removing this.

pkg/kafka/schema_registry.go
ImDoubD-datazip: i checked airbyte, estuary, and fivetran. They ask for schema registry endpoint url and username, password for auth

utils/typeutils/avro.go
ImDoubD-datazip: in case of avro, goavro returns Avro unions as single-key maps (e.g., {"string": "hello"} for a nullable string) and decimals as *big.Rat, neither of which TypeFromValue or ReformatValue handle, TypeFromValue gets datatype in datatype format, then it maps it to types.String and further types. 

for example

- Given this goavro-decoded Avro record:
```
avroRecord := map[string]interface{}{
    "name":    map[string]interface{}{"string": "Alice"},  // Avro union ["null", "string"]
    "balance": big.NewRat(12345, 100),                     // Avro decimal → *big.Rat (123.45)
}
```

if used TypeFromValue:
```
TypeFromValue(avroRecord["name"])     // → types.Object (sees the wrapper map, not the actual string)
TypeFromValue(avroRecord["balance"])  // → types.Unknown (doesn't recognize *big.Rat as of now)
```

when used ConvertAvroRecord(avroRecord) does value transformation:
```
map[string]interface{}{
    "name":    "Alice",   // Unwrapped from {"string": "Alice"}
    "balance": 123.45,    // Converted from *big.Rat to float64
}
```

drivers/kafka/internal/cdc.go
ImDoubD-datazip: changes done

drivers/kafka/internal/kafka.go
ImDoubD-datazip: removed this

pkg/kafka/types.go
ImDoubD-datazip: SchemaRegistryClient reused at config as well

constants/constants.go
```diff
@@ -58,3 +58,6 @@ var SkipCDCDrivers = []DriverType{Oracle, DB2}
 
 // DriversRequiringIncrementalFormatter are drivers that require special formatting for incremental value
 var DriversRequiringIncrementalFormatter = []DriverType{Oracle, DB2, MSSQL}
+
+// InternalKafkaTopics are internal kafka topics (created due to external services) to be skipped
```
hash-data: driver specific constants need to be in drivers, not global

drivers/kafka/internal/cdc.go
```diff
@@ -215,3 +215,106 @@ func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kafka.Reader,
...
+	// use json.Number to avoid automatic conversion of numbers to float64
+	decoder.UseNumber()
+
+	if err := decoder.Decode(&data); err != nil {
+		return nil, err
+	}
+	return data, nil
+}
+
+// TODO: we need to handle when multiple type are present in avro schema, need to find the super type in driver level (through LCA)
```
hash-data: why to do that we can directly map gotypes from avro types, let us say big.Rat -> Int64 kind of, all the super type can be detected while writing ?

drivers/kafka/internal/config.go
```diff
@@ -56,6 +58,12 @@ func (c *Config) Validate() error {
 		}
 	}
 
+	if c.SchemaRegistry != nil {
+		if c.SchemaRegistry.Endpoint == "" {
```
hash-data, line 62: is this tested from ui side ? putting it and leaving empty.

drivers/kafka/internal/kafka.go
```diff
@@ -100,6 +102,14 @@ func (k *Kafka) Setup(ctx context.Context) error {
 	}
 
 	k.dialer = dialer
+
+	// initialize confluent schema registry client if configured
+	if k.config.SchemaRegistry != nil && k.config.SchemaRegistry.Endpoint != "" {
```
hash-data: is it going to reach empty as we have added a check in start?

drivers/kafka/internal/kafka.go
```diff
@@ -119,6 +129,10 @@ func (k *Kafka) GetStreamNames(ctx context.Context) ([]string, error) {
 
 	var topicNames []string
 	for _, topic := range resp.Topics {
+		// skip internal topics
+		if topic.Internal || slices.Contains(constants.InternalKafkaTopics, topic.Name) {
```
hash-data: prev we not had it, any backward compatibility issues? that we might face.

drivers/kafka/internal/kafka.go
```diff
@@ -206,6 +222,11 @@ func (k *Kafka) ProduceSchema(ctx context.Context, streamName string) (*types.St
 		return nil, fmt.Errorf("failed to fetch schema for topic %s: %s", streamName, err)
 	}
 
+	// check if all messages failed to parse
+	if parsedMessageCount == 0 {
```
hash-data: can we use bool ?

drivers/kafka/internal/kafka.go
```diff
@@ -395,3 +416,9 @@ func (k *Kafka) getReaderAssignedPartitions(ctx context.Context, readerIndex int
 
 	return assigned, nil
 }
+
+// isConfluentWireFormat checks if data starts with Confluent wire format magic byte
+// Wire format: [magic byte (0x00)] [4-byte schema ID (big-endian)] [payload]
+func isConfluentWireFormat(data []byte) bool {
+	return len(data) >= 5 && data[0] == 0x00
```
hash-data: make magic byte constant

drivers/kafka/resources/spec.json
```diff
@@ -183,13 +182,124 @@
             "title": "Threads Equal Total Partitions",
             "description": "If true, number of threads will be set equal to total partitions (overrides max_threads)",
             "default": true
+        },
+        "use_schema_registry": {
```
hash-data, line 186: can you add a ui screenshot ?

pkg/kafka/schema_registry.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+}
+
+// TODO: fetch schema by subject strategy if needed (e.g. latest, version specific)
+// currently we only fetch by ID which is sufficient for deserialization of consumed messages
+func (c *SchemaRegistryClient) FetchSchema(schemaID uint32) (*types.RegisteredSchema, error) {
+	// if schema exists previously
+	if schema, ok := c.schemaMap.Load(schemaID); ok {
+		return schema.(*types.RegisteredSchema), nil
```
hash-data: is any way typecasting can fail?

pkg/kafka/schema_registry.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	// fetch schema from registry
+	url := fmt.Sprintf("%s/schemas/ids/%d", c.Endpoint, schemaID)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
```
hash-data: use proper error formatting %s

pkg/kafka/schema_registry.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	// Set authentication headers (if bearer token is present, it takes priority over basic auth)
+	if c.BearerToken != "" {
+		req.Header.Set("Authorization", "Bearer "+c.BearerToken)
```
hash-data:
```suggestion
		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.BearerToken))
```

utils/typeutils/avro.go
```diff
@@ -0,0 +1,44 @@
...
+		if len(val) == 1 {
+			for _, unionVal := range val {
+				return ConvertAvroValue(unionVal)
+			}
+		}
+		return ConvertAvroRecord(val)
+	case []interface{}:
+		result := make([]interface{}, len(val))
+		for i, elem := range val {
+			result[i] = ConvertAvroValue(elem)
```
hash-data: do we need to go on deeper levels ?

hash-data: Also my question is: do we receive data types as well from schema ? and if yes where we are using it? or we are just using codec to decode bytes?

types/kafka_types.go
```diff
@@ -21,3 +32,10 @@ type KafkaRecord struct {
 	Data    map[string]interface{}
 	Message kafka.Message
 }
+
+// RegisteredSchema holds the schema information
+type RegisteredSchema struct {
+	SchemaType SchemaType
+	Schema     string
```
hash-data: where we are using this?

drivers/kafka/internal/cdc.go
ImDoubD-datazip: multiple type can occur for a single key in avro.
for example:

```
const valueSchemaStr = `{
  "type": "record",
  "name": "EventPayload",
  "namespace": "com.example.events",
  "fields": [
    {"name": "event_name", "type": "string"},
    {"name": "event_type", "type": "string"},
    {"name": "event_triggered_tm", "type": "string"},
    {"name": "client_key", "type": "string"},
    {"name": "cm_id", "type": "int"},
    {"name": "emp_id", "type": ["int", "string"]},
    {"name": "emp_name", "type": "string"},
    {"name": "server_type", "type": "string"},
    {"name": "request_ip", "type": "string"},
    {"name": "log", "type": "string"},
    {"name": "is_recovered", "type": "boolean"},
    {"name": "emp_tenure1", "type": "float"},
    {"name": "emp_tenure2", "type": "double"}
  ]
}`
```

In this emp_id can have multiple type for its values.

drivers/kafka/internal/config.go
ImDoubD-datazip: will do after final approve. schema registry working has been checked in ui, and the basic original json working as been tested.

drivers/kafka/internal/kafka.go
ImDoubD-datazip: not backward compatibility, just from now onwards kafka internal topics won't be present in discover. schema registry introduces one internal topic called _schema which is needed to be ignored.

drivers/kafka/internal/kafka.go
ImDoubD-datazip: i have added this check as I want to initialize the schema registry client only when both are present.

drivers/kafka/resources/spec.json
ImDoubD-datazip: added

pkg/kafka/schema_registry.go
ImDoubD-datazip: Bearer token is already a string in nature. this is not required as there is no inner formatting involved and also there are no multiple variables. there is only BearerToken.

types/kafka_types.go
ImDoubD-datazip: we are not usnig this, just for debug purpose I kept it to check what schema is received in go from goavro once avro schema is passed in its codec.

I have removed it now.

utils/typeutils/avro.go
ImDoubD-datazip: Yes we do need to go into deeper level, when for a key, multiple types are given. 

```
const valueSchemaStr = `{
  "type": "record",
  "name": "EventPayload",
  "namespace": "com.example.events",
  "fields": [
    {"name": "event_name", "type": "string"},
    {"name": "event_type", "type": "string"},
    {"name": "event_triggered_tm", "type": "string"},
    {"name": "client_key", "type": "string"},
    {"name": "cm_id", "type": "int"},
    {"name": "emp_id", "type": ["int", "string"]},
    {"name": "emp_name", "type": "string"},
    {"name": "server_type", "type": "string"},
    {"name": "request_ip", "type": "string"},
    {"name": "log", "type": "string"},
    {"name": "is_recovered", "type": "boolean"},
    {"name": "emp_tenure1", "type": "float"},
    {"name": "emp_tenure2", "type": "double"}
  ]
}`
```

Like in above case, if we don't go into deeper levels, we get the value in a map i.e. like this:

```
decoded Avro: map[client_key:client_110 cm_id:110 emp_id:map[string:EMP-2009] emp_name:User_9 emp_tenure1:1.2857152 emp_tenure2:1.2857152857142857 event_name:event_9 event_triggered_tm:2026-02-01T11:50:14.076Z event_type:UI_CLICK is_recovered:false log:Volume test request_ip:192.168.9.0 server_type:PROD]
```
 
emp_id has int, string as type. after decoding, it comes in map format (given by goavro) which will at the end be given in object format. but we want the value directly after decode.

we want in this format:

```
decoded Avro record: map[client_key:client_110 cm_id:110 emp_id:EMP-2009 emp_name:User_9 emp_tenure1:1.2857152 emp_tenure2:1.2857152857142857 event_name:event_9 event_triggered_tm:2026-02-01T11:50:14.076Z event_type:UI_CLICK is_recovered:false log:Volume test request_ip:192.168.9.0 server_type:PROD]
```

we are just using the codec to decode bytes (but in the codec the avro schema is passed which we get from the schema registry as per schemaID) and then inferring types from the decoded Go values. similar to JSON, I am inferring the types from 10000 messages I receive from the decoded bytes.

drivers/kafka/internal/kafka.go
ImDoubD-datazip: changes done. if all messages failed to parse, then skipping that topic entirely.

drivers/abstract/abstract.go
```diff
@@ -75,6 +75,10 @@ func (a *AbstractDriver) Discover(ctx context.Context) ([]*types.Stream, error)
 		if err != nil {
 			return fmt.Errorf("%w: failed to produce schema for stream %s: %s", constants.ErrNonRetryable, stream, err)
 		}
+		// skip streams (case: when all messages failed to parse in kafka)
+		if streamSchema == nil {
```
hash-data: it should be non nil here always

drivers/kafka/internal/kafka.go
hash-data: empty one

drivers/kafka/internal/kafka.go
hash-data: current all default one are of avro so current user are not syncing these topics

drivers/kafka/internal/kafka.go
hash-data: removal of it

utils/typeutils/avro.go
```diff
@@ -0,0 +1,44 @@
...
+func ConvertAvroRecord(record map[string]interface{}) map[string]interface{} {
+	result := make(map[string]interface{}, len(record))
+	for k, v := range record {
+		result[k] = ConvertAvroValue(v)
+	}
+	return result
+}
+
+// convertAvroValue converts Avro-decoded values to JSON-compatible types
+func ConvertAvroValue(v interface{}) interface{} {
```
hash-data: ExtractAvroValue

drivers/abstract/abstract.go
ImDoubD-datazip: kafka topics with all message parsing which failed will be seen in streams with only kafka columns, nothing from data.

pkg/kafka/schema_registry.go
ImDoubD-datazip: doing this change

drivers/kafka/internal/cdc.go
```diff
@@ -215,3 +215,105 @@ func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kafka.Reader,
...
+	var messageValue map[string]interface{}
+	if message.Value != nil {
+		valDecoded, err := parseData(message.Value)
+		if err != nil {
+			return nil, "", err
+		}
+		if vm, ok := valDecoded.(map[string]interface{}); ok {
+			messageValue = vm
+		} else {
+			return nil, "", fmt.Errorf("expected format for message value is not supported, got %s", valDecoded)
```
hash-data: add type here which we got where it failed

drivers/kafka/internal/kafka.go
```diff
@@ -185,6 +206,7 @@ func (k *Kafka) ProduceSchema(ctx context.Context, streamName string) (*types.St
 		fetchCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
 		defer cancel()
 		_ = k.processKafkaMessages(fetchCtx, reader, func(record types.KafkaRecord) (bool, error) {
```
hash-data, line 211: why err is neglected here?

hash-data:
Also auth check in check connection

drivers/kafka/internal/kafka.go
ImDoubD-datazip: because context timeout is expected here, if returned error then it will give context deadline exceeded error. we only want parse errors, which are logged inside this function.

drivers/kafka/internal/kafka.go
```diff
@@ -185,6 +209,7 @@ func (k *Kafka) ProduceSchema(ctx context.Context, streamName string) (*types.St
...
+			messageCount++
```
hash-data, line 212: so we are counting all iterated message now.

drivers/kafka/internal/config.go
```diff
@@ -4,16 +4,18 @@ import (
 	"fmt"
 
 	"github.com/datazip-inc/olake/constants"
+	kafkapkg "github.com/datazip-inc/olake/pkg/kafka"
```
hash-data: any reason to use alias?

hash-data (dismissed):

drivers/kafka/internal/kafka.go
hash-data: but we can add ctx deadline check and skip that err?

drivers/kafka/internal/kafka.go
ImDoubD-datazip: yes, we are counting all iterated messages, stop as it reaches 10K.

drivers/kafka/internal/config.go
ImDoubD-datazip: schema registry client struct in pkg. shubham bhaiya said to use one thing at pkg and config only.

drivers/kafka/internal/kafka.go
ImDoubD-datazip: we can but there won;t be any other error that will be encountered here. al of those errors will be logged inside that processKafkaMessage function.

drivers/kafka/internal/config.go
hash-data: No, you have used alias instead of pkg name 
Like kafkapkg where you can directly use kafka

ImDoubD-datazip: ok understood, changing.

ImDoubD-datazip: changes done.

hash-data (approved):

PR #815, opened 2026-02-05
fix: align CDC and full refresh TIMESTAMP

### LINKED ISSUE #811: TIMESTAMP column treated incorrectly in CDC mode if MySQL server timezone is different
## Description
I have MySQL server running in "US/Eastern" timezone and table with TIMESTAMP column. 
OLake saves into Parquet file different values for the same row in Initial Load and CDC modes. In initial it respects  timezone and save into Parquet proper value, in CDC mode it doesnt respect timezone and save value in UTC. 
Which looks like a different value even if I didnt change it.

## Steps to reproduce
1. Run MySQL 8.14 with  "--default-time-zone=US/Eastern" argument
2.  Check that time zone was changed
```
SELECT @@GLOBAL.time_zone, @@SESSION.time_zone;
+--------------------+---------------------+
| @@GLOBAL.time_zone | @@SESSION.time_zone |
+--------------------+---------------------+
| US/Eastern         | US/Eastern          |
+--------------------+---------------------+
```
3. Run these SQL commands
```
create database test;

use test;

CREATE TABLE items4 (
  id int NOT NULL AUTO_INCREMENT,
  name varchar(60) NOT NULL,
  created timestamp NULL DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3;

SET autocommit=0;

START TRANSACTION;

insert into items4 (name, created) values ('Name A', '2026-01-30 12:34:56');

COMMIT;
```
4. Run Olake and do Initial Data load
5. Update row in MySQL but dont change TIMESTAMP column
```
START TRANSACTION;
update items4 set name='Name B' where id=LAST_INSERT_ID();
COMMIT;
```
6. Run Olake and do CDC load

## Expected behavior
Values in TIMESTAMP column should be the same

## Actual behavior
Values in TIMESTAMP column are different

|        _cdc_timestamp         | _olake_id |       _olake_timestamp        | _op_type |         created          |  id   |  name   |
|---|---|---|---|---|---|---|
| 1970-01-01 00:00:00+00        | 1         | 2026-02-03 05:51:32.906918+00 | r        | 2026-01-30 **12:34:56+00**   |     1 | Name A  |
| 2026-02-03 05:52:34.542397+00 | 1         | 2026-02-03 05:52:52.780196+00 | u        | 2026-01-30 **17:34:56+00**   |     1 | Name B  |

## Environment

- **Olake version**:  0.3.15
- **MySQL version**:  8.0.44 with US/Eastern global timezone

vaibhav-datazip:
# Description

Fixes incorrect TIMESTAMP handling in MySQL CDC. Data is read from MySQL in the database (server) timezone and should be written to the destination in that same DB timezone. Initial load did this correctly.
 CDC was writing TIMESTAMP in local instead, so the same row showed different values in full refresh vs CDC (e.g. `12:34:56` vs `17:34:56` for the same instant in US/Eastern). This PR resolves the effective timezone from MySQL (session → global → system) or from `jdbc_url_params.time_zone`, passes it to the CDC binlog syncer as `TimestampStringLocation`.

Fixes #811

## Type of change

# How Has This Been Tested?

- [x] **CDC vs initial load (US/Eastern):** MySQL with `--default-time-zone=US/Eastern`; initial load then CDC update without changing TIMESTAMP column; same TIMESTAMP value in both modes (no 12:34 vs 17:34 mismatch).
- [x] **Other timezones:** Asia/Kolkata, Asia/Tokyo, UTC — same TIMESTAMP value in initial load and CDC.
- [x] **Config override:** `jdbc_url_params.time_zone` set (e.g. `"Asia/Tokyo"` or `"'Asia/Tokyo'"`); timezone detection query skipped; CDC uses configured zone.

# Screenshots or Recordings

## Documentation


pkg/binlog/binlog.go
```diff
@@ -29,16 +29,17 @@ type Connection struct {
...
+		Flavor:                  config.Flavor,
+		Host:                    config.Host,
+		Port:                    config.Port,
+		User:                    config.User,
+		Password:                config.Password,
+		Charset:                 config.Charset,
+		VerifyChecksum:          config.VerifyChecksum,
+		HeartbeatPeriod:         config.HeartbeatPeriod,
+		TLSConfig:               config.TLSConfig,
+		TimestampStringLocation: config.TimestampStringLocation,
```
vishalm0509:
```	// If ParseTime is false, convert TIMESTAMP into this specified timezone. If
	// ParseTime is true, this option will have no effect and TIMESTAMP data will
	// be parsed into the local timezone and a full time.Time struct will be
	// returned.
```

This is mentioned in the library ? Can you check if we need to set `parseTime` explicilty to `true` or leave it to the default value ?

drivers/mysql/internal/mysql.go
```diff
@@ -306,3 +328,38 @@ func (m *MySQL) IsCDCSupported(ctx context.Context) (bool, error) {
...
+		if err != nil {
+			return time.UTC
+		}
+		return loc
+	}
+
+	if candidate != "" {
+		return fromNamed(candidate)
+	}
+	return fromNamed(system)
```
vishalm0509:
```go
func resolveMySQLTimeZone(sessionTimezone, globalTimezone, systemTimezone string) *time.Location {
    normalize := func(s string) string {
        s = strings.TrimSpace(s)
        return strings.Trim(s, `'"`)
    }
    
    loadLocation := func(name string) *time.Location {
        loc, err := time.LoadLocation(name)
        if err != nil {
            return time.UTC
        }
        return loc
    }
    
    session := normalize(sessionTimezone)
    if session != "" && !strings.EqualFold(session, SystemTimezone) {
        return loadLocation(session)
    }
    
    global := normalize(globalTimezone)
    if global != "" && !strings.EqualFold(global, SystemTimezone) {
        return loadLocation(global)
    }
    
    system := normalize(systemTimezone)
    return loadLocation(system)
}
```

--- 


Can we do something like this ? I think this is easier to read

drivers/mysql/internal/mysql.go
```diff
@@ -306,3 +328,38 @@ func (m *MySQL) IsCDCSupported(ctx context.Context) (bool, error) {
 
 	return true, nil
 }
+
+// resolveMySQLTimeZone picks an effective timezone from MySQL vars and returns a *time.Location
+// for interpreting TIMESTAMP values (e.g. CDC binlog). Precedence: session, then global, then
+// system abbreviation; otherwise UTC.
+func resolveMySQLTimeZone(sessionTimezone, globalTimezone, systemTimezone string) *time.Location {
```
vishalm0509, line 336: Does this problem exists for other connectors as well? if yes, we need to either solve it or add a `TODO`

drivers/mysql/internal/mysql.go
```diff
@@ -113,6 +116,25 @@ func (m *MySQL) Setup(ctx context.Context) error {
 	if err := client.PingContext(ctx); err != nil {
 		return fmt.Errorf("failed to ping database: %s", err)
 	}
+
+	var resolved *time.Location
+	if m.config.JDBCURLParams["time_zone"] != "" {
+		if tzOverride := strings.TrimSpace(m.config.JDBCURLParams["time_zone"]); tzOverride != "" {
```
vishalm0509: `resolveMySQLTimeZone` already does `strings.TrimSpace`

pkg/binlog/binlog.go
vaibhav-datazip: no , If we will set parse time as true , we will start receiving data as timestamp instead of string , because of that we will get timezone as well which will get converted to utc later . which we don't want

drivers/mysql/internal/mysql.go
vaibhav-datazip: yes

pkg/binlog/binlog.go
vishalm0509: I meant `false`

vaibhav-datazip: by default its false only

drivers/mysql/internal/mysql.go
```diff
@@ -306,3 +328,33 @@ func (m *MySQL) IsCDCSupported(ctx context.Context) (bool, error) {
...
+	case session != "" && !strings.EqualFold(session, "SYSTEM"):
+		name = session
+	case global != "" && !strings.EqualFold(global, "SYSTEM"):
+		name = global
+	default:
+		name = system
+	}
+
+	loc, err := time.LoadLocation(name)
+	if name == "" || err != nil {
```
vishalm0509: `time.LoadLocation` returns `UTC` if `name === nil`

Here,
```go
func LoadLocation(name string) (*Location, error) {
	if name == "" || name == "UTC" {
		return UTC, nil
	}
    ...
```

So do we need to handle it explicitly

drivers/mysql/internal/mysql.go
```diff
@@ -306,3 +328,33 @@ func (m *MySQL) IsCDCSupported(ctx context.Context) (bool, error) {
...
+func resolveMySQLTimeZone(sessionTimezone, globalTimezone, systemTimezone string) *time.Location {
+	// Strip surrounding quotes so "'Asia/Tokyo'" or "\"UTC\"" from jdbc params become valid IANA names.
+	normalize := func(s string) string {
+		return strings.Trim(strings.TrimSpace(s), `'"`)
+	}
+	session := normalize(sessionTimezone)
+	global := normalize(globalTimezone)
+	system := normalize(systemTimezone)
+
+	var name string
```
vishalm0509, line 346:
```suggestion
	}
	
	session := normalize(sessionTimezone)
	global := normalize(globalTimezone)
	system := normalize(systemTimezone)
```

drivers/mysql/internal/mysql.go
```diff
@@ -113,6 +116,25 @@ func (m *MySQL) Setup(ctx context.Context) error {
...
 		return fmt.Errorf("failed to ping database: %s", err)
 	}
+
+	var resolved *time.Location
+	if m.config.JDBCURLParams["time_zone"] != "" {
+		if tzOverride := strings.TrimSpace(m.config.JDBCURLParams["time_zone"]); tzOverride != "" {
+			resolved = resolveMySQLTimeZone(tzOverride, tzOverride, tzOverride)
+		}
+	}
+	if resolved == nil {
```
vishalm0509, line 124: Why are we not using this in `else` block

drivers/mysql/internal/mysql.go
```diff
@@ -113,6 +116,25 @@ func (m *MySQL) Setup(ctx context.Context) error {
 	if err := client.PingContext(ctx); err != nil {
 		return fmt.Errorf("failed to ping database: %s", err)
 	}
+
+	var resolved *time.Location
+	if m.config.JDBCURLParams["time_zone"] != "" {
+		if tzOverride := strings.TrimSpace(m.config.JDBCURLParams["time_zone"]); tzOverride != "" {
+			resolved = resolveMySQLTimeZone(tzOverride, tzOverride, tzOverride)
+		}
+	}
```
vishalm0509, line 123:
```go
if tz := strings.TrimSpace(m.config.JDBCURLParams["time_zone"]); tz != "" {
    resolved = resolveMySQLTimeZone(tz, tz, tz)
}
```

One map scan ?

constants/state_version.go
vishalm0509: Should it say current version ?

pkg/binlog/binlog.go
```diff
@@ -29,17 +30,19 @@ type Connection struct {
...
+		Host:            config.Host,
+		Port:            config.Port,
+		User:            config.User,
+		Password:        config.Password,
+		Charset:         config.Charset,
+		VerifyChecksum:  config.VerifyChecksum,
+		HeartbeatPeriod: config.HeartbeatPeriod,
+		TLSConfig:       config.TLSConfig,
+	}
+	if constants.LoadedStateVersion > 1 {
```
vishalm0509, line 47: It seems like a magic number, should we make it a constant mentioning what the number `1` is for, or should this be fine. wdyt ?

constants/state_version.go
```diff
@@ -17,10 +17,14 @@ package constants
...
+//     * This prevents discrepancies where CDC timestamps could differ from Full Refresh data.
 
 const (
-	LatestStateVersion = 1
+	LatestStateVersion = 2
 )
 
 // Used as the current version of the state when the program is running
-var LoadedStateVersion = 1
+var LoadedStateVersion = 2
```
vishalm0509: If we are updating this for every version and we are sure that the default is always `LatestStateVersion` can we do this ?

```go
var LoadedStateVersion = LatestStateVersion
```

pkg/binlog/binlog.go
```diff
@@ -29,17 +30,19 @@ type Connection struct {
...
+		syncerConfig.TimestampStringLocation = config.TimestampStringLocation
```
vishalm0509, line 48:
```go
// can we add a comment here to why we are doing this.
	if constants.LoadedStateVersion > 1 {
		syncerConfig.TimestampStringLocation = config.TimestampStringLocation
	}
```

pkg/binlog/binlog.go
vaibhav-datazip: this is the version , can you suggest any other way we can define it ?

constants/state_version.go
vaibhav-datazip: forgot changing this

PR #826, opened 2026-02-12
fix: extra column appearing in parquet schema when normalization is enabled

vikaxsh:
# Description

Issue: When normalization was enabled in parquet output, an extra column (data) was incorrectly appearing in the schema due to conditional addition logic.
Fix: Initialize default olake columns upfront; data column now only appears when normalization is disabled.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/destination/iceberg/config.go
+++ b/destination/iceberg/config.go
@@ -42,6 +42,14 @@ type Config struct {
 	CatalogType CatalogType `json:"catalog_type,omitempty"`
 	CatalogName string      `json:"catalog_name,omitempty"`
 
+	// Glue catalog optional overrides
+	UseGlueAdditionalConfig bool   `json:"glue_additional_config,omitempty"`
+	GlueEndpoint            string `json:"glue_endpoint,omitempty"`
+	GlueAccessKey           string `json:"glue_access_key,omitempty"`
+	GlueSecretKey           string `json:"glue_secret_key,omitempty"`
+	GlueRegion              string `json:"glue_region,omitempty"`
+	GlueCatalogID           string `json:"glue_catalog_id,omitempty"`
+
 	// JDBC specific configuration
 	JDBCUrl      string `json:"jdbc_url,omitempty"`
 	JDBCUsername string `json:"jdbc_username,omitempty"`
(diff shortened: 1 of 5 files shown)


types/type_schema.go
```diff
@@ -124,13 +124,7 @@ func (t *TypeSchema) GetProperty(column string) (bool, *Property) {
...
 func (t *TypeSchema) ToParquet(onlyOlakeColumns bool) *parquet.Schema {
 	// keeping default columns parquet schema for backward compatibility for olake columns
-	groupNode := parquet.Group{
-		constants.StringifiedData: parquet.JSON(),
-		constants.OlakeID:         parquet.String(),
-		constants.OlakeTimestamp:  parquet.Timestamp(parquet.Microsecond),
-		constants.OpType:          parquet.String(),
-		constants.CdcTimestamp:    parquet.Optional(parquet.Timestamp(parquet.Microsecond)),
-	}
+	groupNode := parquet.Group{}
```
hash-data: will not work if user not done discover think again once

types/type_schema.go
```diff
@@ -124,7 +124,12 @@ func (t *TypeSchema) GetProperty(column string) (bool, *Property) {
 
 func (t *TypeSchema) ToParquet(onlyOlakeColumns bool) *parquet.Schema {
 	// keeping default columns parquet schema for backward compatibility for olake columns
-	groupNode := parquet.Group{}
+	groupNode := parquet.Group{
```
hash-data, line 127: what about data column now ? let us say normalization is off, and user not edited, in that case only these column will be created in new parquet file and data column will not be added

vikaxsh: in case of normalization false onlyOlakeColumns bool would be true so below i have added condition to add that column in schema similar to iceberg

hash-data: oh i saw now, most probably i not refreshed it

types/type_schema.go
```diff
@@ -140,6 +139,9 @@ func (t *TypeSchema) ToParquet(onlyOlakeColumns bool) *parquet.Schema {
 		groupNode[prop.getDestinationColumnName(key.(string))] = prop.DataType().ToNewParquet()
 		return true
 	})
+	if onlyOlakeColumns {
```
hash-data: can you rename this variable as `normalization` only,  onlyOlakeColumns looks confusing w.r.t. use case

vikaxsh: The purpose of this variable, as its name suggests, is to add only OLake generated columns when it is set to true. Renaming it to “normalization” would not be self-explanatory.

hash-data: Can you tell is data an Olake Column?

vikaxsh: change to defaultColumns

PR #827, opened 2026-02-12
fix: parse timezone offset correctly

vaibhav-datazip:
# Description

Adds support for MySQL timezone **offsets** (e.g. `+05:30`, `-08:00`, `0:00`) in addition to IANA names. Previously, when session/global timezone or `jdbc_url_params.time_zone` was set as an offset, it was not parsed and the driver fell back to UTC. Now offsets are parsed via a regex and used with `time.FixedZone` so CDC and full refresh use the correct timezone.

- **Parsing:** New `parseMySQLOffset()` with regex `^([+-]?)([0-1]?\d|2[0-3]):([0-5]\d)$` to handle `+05:30`, `+5:30`, `-12:00`, `0:00`, etc.
- **State version:** Offset handling is enabled only when `LoadedStateVersion > 2` (state version 3) for backward compatibility.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] **Session/global in offset format:** Set MySQL `@@session.time_zone` and/or `@@global.time_zone` to an offset (e.g. `SET time_zone = '+05:30'`). Run sync (full refresh and/or CDC). Verify timestamps are interpreted in that timezone (e.g. IST), not UTC.
- [x] **jdbc_url_params.time_zone:** Set `jdbc_url_params.time_zone` to an offset in config (e.g. `"'+05:30'"`). Run sync. Verify connection uses that timezone and that CDC/full refresh timestamps match (no fallback to UTC).
- [x] **tested by passing jdbc url params in olake ui**: passed the jdbc url parameter as time_zone '-04:00' , to test if the quote format is supported from ui side as well
# Screenshots or Recordings
N/A

## Documentation

## Related PR's (If Any):
None

### DIFF
--- a/drivers/mysql/internal/mysql.go
+++ b/drivers/mysql/internal/mysql.go
@@ -362,7 +362,7 @@ func resolveMySQLTimeZone(sessionTimezone, globalTimezone, systemTimezone string
 
 	loc, err := time.LoadLocation(name)
 	if err != nil {
-		logger.Warnf("failed to load mysql timezone location %s, falling back to UTC: %s", name, err)
+		logger.Warnf("failed to load mysql timezone location %s, falling back to UTC. Set jdbc_url_params.time_zone to override: %s", name, err)
 		return time.UTC
 	}
 	return loc


constants/state_version.go
```diff
@@ -18,12 +18,16 @@ package constants
...
 //     * This prevents data corruption by failing fast on invalid date strings
 //
-//   - Version 2: Current Version (Introduces consistent timezone handling between MySQL Full Refresh and CDC.)
+//   - Version 2: Introduces consistent timezone handling between MySQL Full Refresh and CDC.
 //     * Binlog CDC now uses TimestampStringLocation to align with the connection's timezone configuration.
 //     * This prevents discrepancies where CDC timestamps could differ from Full Refresh data.
+//
+//   - Version 3: Current Version (Parses the timezone offset for MySQL correctly)
+//     * Earlier if the session timezone or global was set in offset format or was passed through jdbc_url_params.time_zone, it was not parsed correctly and used to fallback to UTC.
+//     * Now it parses the offset correctly and uses the timezone offset to set the timezone for the connection.
```
vishalm0509, line 27:
```suggestion
//   - Version 3: Current Version (Parses the timezone offset for MySQL correctly)
//     * Earlier if the session timezone or global was set in offset format, it was not parsed correctly and used to fallback to UTC.
//     * Now it parses the offset correctly and uses the timezone offset to set the timezone for the connection.
```

drivers/mysql/internal/mysql.go
```diff
@@ -353,9 +355,36 @@ func resolveMySQLTimeZone(sessionTimezone, globalTimezone, systemTimezone string
...
+
 	loc, err := time.LoadLocation(name)
 	if err != nil {
 		return time.UTC
 	}
 	return loc
 }
+
+// parseMySQLOffset parses MySQL-style timezone offset. Returns offset in seconds and true, or 0, false.
+func parseMySQLOffset(s string) (offsetSeconds int, ok bool) {
```
vishalm0509: `parseMySQLTimezoneOffset` ?

drivers/mysql/internal/mysql.go
```diff
@@ -353,9 +355,36 @@ func resolveMySQLTimeZone(sessionTimezone, globalTimezone, systemTimezone string
...
+	// MySql supports offsets ranging from -13:59 to +14:00
```
vishalm0509, line 373: Here, the range mentioned is different. 

Image

drivers/mysql/internal/mysql.go
```diff
@@ -353,9 +355,36 @@ func resolveMySQLTimeZone(sessionTimezone, globalTimezone, systemTimezone string
...
+	var mysqlOffsetRegex = regexp.MustCompile(`^([+-])(0?\d|1[0-4]):([0-5]\d)$`)
```
vishalm0509:
```
+05:30, +5:30, -12:00, 0:00, etc.
```

These are the examples mentioned in the description. It contains invalid offset. We should fail in that case right ?

vishalm0509: Can we move this to constants

drivers/mysql/internal/mysql.go
vishalm0509: Can you check this refactored version ?

```go
func parseMySQLOffset(s string) (offsetSeconds int, ok bool) {
	s = strings.TrimSpace(s)
	matches := mysqlOffsetRegex.FindStringSubmatch(s)
	if matches == nil {
		return 0, false
	}

	signStr, hourStr, minuteStr := matches[1], matches[2], matches[3]
	
	// Regex guarantees these conversions succeed
	hours, _ := strconv.Atoi(hourStr)
	minutes, _ := strconv.Atoi(minuteStr)

	// MySQL allows +14:00 as maximum, but not +14:01 or -14:00
	if hours == 14 && (minutes > 0 || signStr == "-") {
		return 0, false
	}

	offsetSeconds = hours*3600 + minutes*60
	if signStr == "-" {
		offsetSeconds = -offsetSeconds
	}

	return offsetSeconds, true
}
```

constants/state_version.go
vaibhav-datazip: even for jdbc url params it used to fall back

drivers/mysql/internal/mysql.go
vaibhav-datazip: Screenshot 2026-02-13 at 6 52 21 PM
  

I have took mysql doc as reference here is the [link](https://dev.mysql.com/doc/refman/8.4/en/time-zone-support.html)

drivers/mysql/internal/mysql.go
vaibhav-datazip: No mysql allows missing the preceding value, 
Screenshot 2026-02-13 at 6 57 50 PM

I have took mysql doc as reference here is the [link](https://dev.mysql.com/doc/refman/8.4/en/time-zone-support.html)

vaibhav-datazip: this is only used here , I can change this from var to const

drivers/mysql/internal/mysql.go
vaibhav-datazip: sure will add the string variable line fetched from regex

drivers/mysql/internal/mysql.go
```diff
@@ -353,9 +355,34 @@ func resolveMySQLTimeZone(sessionTimezone, globalTimezone, systemTimezone string
...
 
+	offsetSeconds, ok := parseMySQLTimeZoneOffset(name)
+	if constants.LoadedStateVersion > 2 && ok {
+		return time.FixedZone(name, offsetSeconds)
+	}
+
 	loc, err := time.LoadLocation(name)
 	if err != nil {
 		return time.UTC
 	}
```
vishalm0509, line 367: Here, if the timezone isn't in the desired format `IANA` in our case shouldn't we fail it and ask for proper format ?

Or atleast we need to `log` what's the error and mention that we are falling back to `UTC`

vaibhav-datazip: Yeah correct, should we fail it or should we just return the log that we will be using UTC

vaibhav-datazip: Added a warning log , if the timezone parsing fails it will output falling back to utc

vishalm0509:
```go
	minutes, err2 := strconv.Atoi(minuteStr)
	if err1 != nil || err2 != nil || (hours == 14 && minutes > 0) || (signStr == "-" && hours == 14) {
		return 0, false
	}
```

--- 

We are missing this errors if any occurs. But discussed with Vaibhav, error will not occur

drivers/mysql/internal/mysql.go
```diff
@@ -353,9 +355,36 @@ func resolveMySQLTimeZone(sessionTimezone, globalTimezone, systemTimezone string
...
+
+	s = strings.TrimSpace(s)
+	matches := mysqlOffsetRegex.FindStringSubmatch(s)
+	if matches == nil {
+		return 0, false
+	}
+	signStr, hourStr, minuteStr := matches[1], matches[2], matches[3]
+
+	hours, err1 := strconv.Atoi(hourStr)
+	minutes, err2 := strconv.Atoi(minuteStr)
```
hash-data, line 384: Is there no package that does this?

drivers/mysql/internal/mysql.go
```diff
@@ -353,9 +355,36 @@ func resolveMySQLTimeZone(sessionTimezone, globalTimezone, systemTimezone string
...
 	}
 
+	offsetSeconds, ok := parseMySQLTimeZoneOffset(name)
+	if constants.LoadedStateVersion > 2 && ok {
+		return time.FixedZone(name, offsetSeconds)
+	}
+
 	loc, err := time.LoadLocation(name)
 	if err != nil {
+		logger.Warnf("failed to load mysql timezone location %s, falling back to UTC: %s", name, err)
```
hash-data: also add to avoid use jdbc params

vishalm0509 (approved):

PR #834, opened 2026-02-18
fix: hex-encode physloc chunk values to avoid invalid UTF-8 in gRPC

vishalm0509:
# Description

 MSSQL `%%physloc%%` returns binary data that, when stored as raw strings in chunk boundaries, produces invalid UTF-8. This causes gRPC marshaling failures (string field contains invalid UTF-8) since protobuf string fields require valid UTF-8.

**Fix**: hex-encode physloc values at chunk creation time. MSSQL natively supports hex literals for binary comparison, so downstream SQL queries work without changes. Existing state files with old-format chunks are handled by a fallback re-encoding path in `formatPhysLocValue`.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

Screenshot 2026-02-18 at 20 50 57

Screenshot 2026-02-18 at 20 52 45

## Documentation

### DIFF
--- a/pkg/jdbc/jdbc.go
+++ b/pkg/jdbc/jdbc.go
@@ -877,7 +877,7 @@ func MSSQLPhysLocChunkScanQuery(stream types.StreamInterface, chunk types.Chunk,
 			return value
 		}
 
-		return fmt.Sprintf("%s", val)
+		return fmt.Sprintf("%v", val)
 	}
 
 	var chunkCond string


vishalm0509:
The `0x` prefix is T-SQL syntax that tells the parser "this is binary data."  Without it, the SQL tries to parse it to numeric type

```bash
2026-02-19T06:55:16Z DEBUG Starting backfill from 4888000001000000 to <nil> with filter: , args: []
2026-02-19T06:55:16Z DEBUG Executing chunk query: SELECT * FROM [dbo].[HeapTable2] WITH (READPAST) WHERE %%physloc%% > 4888000001000000 ORDER BY %%physloc%%
2026-02-19T06:55:16Z INFO Thread[dbo.HeapTable2_01KHTAWV2CTTD7VCWWKQ8NEY52_min[4888000001000000]-max[<nil>]]: shutting down Iceberg server on port 50053
2026-02-19T06:55:16Z INFO retry attempt[1], retrying after 60.00 seconds due to err: thread[dbo.HeapTable2_01KHTAWV2CTTD7VCWWKQ8NEY52_min[4888000001000000]-max[<nil>]]: context canceled: prev error: failed to close writer: failed to close writer: context canceled: prev error: mssql: Arithmetic overflow error converting varbinary to data type numeric.
```

drivers/mssql/internal/backfill.go
```diff
@@ -213,51 +214,44 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
 			// Iteratively find chunk boundaries until we reach the end of the table
 			for {
-				var next any
+				var next []byte
 				// This gives us the next chunk boundary, ensuring each chunk has ~chunkSize rows
-				query := jdbc.MSSQLPhysLocNextChunkEndQuery(stream, chunkSize)
+				query := jdbc.MSSQLPhysLocNextChunkEndQuery(stream, chunkSize, currentHex)
 
-				err := tx.QueryRowContext(ctx, query, current).Scan(&next)
+				err := tx.QueryRowContext(ctx, query).Scan(&next)
```
vikaxsh: shouldn't we hex encode only when inserting to chunks?

vishalm0509: We are doing that only. Also the query supports having hex value. We can change it but we need to maintain the binary as well. Should we do it ?

vikaxsh:
lgtm

PR #835, opened 2026-02-18
feat: addition of max-parallel-threads flag for schema discovery

vaibhav-datazip:
# Description

- **max-discover-threads:** A new flag `--max-discover-threads` has been added. It can be used during schema discovery. When provided, it overrides the thread count from the source config file (e.g. `max_threads`) or the default (e.g. `DefaultThreadCount`), this enables how many streams are discovered in parallel without changing config. It only takes affect during discover command , if left blank it uses default value as 50.

- **Discover timeout handling:** When the discover context times out (e.g. 5 min), drivers no longer panic with a nil pointer. MongoDB, Postgres, MySQL, and MSSQL now return `context.DeadlineExceeded`.
Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Run discover with a short timeout  on MongoDB (or other driver) and confirm it fails with "context deadline exceeded" instead of a panic.
- [x] Run discover with `--max-discover-threads N` and confirm schema discovery respects the concurrency limit.

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/drivers/mssql/internal/backfill.go
+++ b/drivers/mssql/internal/backfill.go
@@ -197,6 +197,8 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 	// %%physloc%% returns the physical location (file_id, page_id, slot_id) of a row as binary.
 	// We iteratively find chunk boundaries by querying for the N-th row (N = chunkSize) where
 	// physloc > current, creating evenly-sized chunks: [nil, min], [min, next1], ..., [last, nil]
+	//
+	// All physloc values are hex-encoded before storing in chunks to ensure valid UTF-8 chunk values.
 	splitViaPhysLoc := func(stream types.StreamInterface, chunks *types.Set[types.Chunk]) error {
 		// SQL Server doesn't support read-only transactions
 		// Use repeatable read isolation without read-only flag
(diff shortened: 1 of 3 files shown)


protocol/root.go
```diff
@@ -96,6 +97,7 @@ func init() {
 	RootCmd.PersistentFlags().StringVarP(&streamsPath, "streams", "", "", "Path to the streams file for the connector")
 	RootCmd.PersistentFlags().StringVarP(&statePath, "state", "", "", "(Required) State for connector")
 	RootCmd.PersistentFlags().Int64VarP(&batchSize, "destination-buffer-size", "", 10000, "(Optional) Batch size for destination")
+	RootCmd.PersistentFlags().IntVarP(&maxParallelThreads, "max-parallel-threads", "", 0, "(Optional) Max parallel threads for discover")
```
vikaxsh: The current flag name isn’t self-explanatory. Can we rename it  like --discover-concurrency or --discover-max-parallel-threads?

protocol/discover.go
```diff
@@ -53,7 +53,7 @@ var discoverCmd = &cobra.Command{
 		discoverCtx, cancel := context.WithTimeout(cmd.Context(), discoverTimeout)
 		defer cancel()
 
-		streams, err := connector.Discover(discoverCtx)
+		streams, err := connector.Discover(discoverCtx, maxParallelThreads)
```
vikaxsh: Can we make this configurable via a global flag variable similar to the other flags in root.go instead of passing it directly here?

protocol/sync.go
```diff
@@ -86,8 +86,8 @@ var syncCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		// Get Source Streams
-		streams, err := connector.Discover(cmd.Context())
+		// Get Source Streams, sending 0 max parallel threads to discover
```
vikaxsh: why aren't we using discover paralleism here? if we are taking that as a job level input

vikaxsh:
lgtm

drivers/mongodb/internal/mon.go
```diff
@@ -168,6 +168,7 @@ func (m *Mongo) GetStreamNames(ctx context.Context) ([]string, error) {
 	return streamNames, collections.Err()
 }
 
+// TODO: remove sampling of table in produce schema when sync command is used
```
hash-data: remove produce schema only in sync running

PR #843, opened 2026-02-23
fix: prevent duplicate rows during MSSQL CDC backfill by waiting for the async capture to agent catch-up

vishalm0509:
# Description

- MSSQL's CDC capture agent runs asynchronously, so `sys.fn_cdc_get_max_lsn()` can lag behind the
  transaction log at the start of a sync. This causes the backfill to read rows that later appear
  again in the CDC change stream, producing duplicates.
- Before recording the initial CDC cursor, we now poll `sys.dm_cdc_log_scan_sessions` until the
  agent completes a non-throttled scan (tran_count < maxtrans), ensuring the CDC max LSN reflects
  all committed transactions.
- Gracefully degrades: if `VIEW DATABASE STATE` / `VIEW DATABASE PERFORMANCE STATE` permission is
  missing, falls back to the previous behavior with a warning.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

- [x] Documentation Link: https://github.com/datazip-inc/olake-docs/pull/377

## Related PR's (If Any):

### DIFF
--- a/.cursor/rules/olake.mdc
+++ b/.cursor/rules/olake.mdc
@@ -10,7 +10,7 @@ alwaysApply: true
 
 OLake is an open-source, high-performance **EL (Extract-Load)** tool that replicates data from databases, Kafka, and S3 into **Apache Iceberg** tables or **plain Parquet** files.
 
-- **Languages:** Go 1.24 (primary) + Java 17 (Iceberg writer)
+- **Languages:** Go 1.25.8 (primary) + Java 17 (Iceberg writer)
 - **Go workspace:** `go.work` links the root module with 8 driver sub-modules
 - **CLI:** Cobra-based commands: `spec`, `check`, `discover`, `sync`, `clear`
 - **Sources:** Postgres, MongoDB, MySQL, Oracle, MSSQL, DB2, Kafka, S3
--- a/.github/workflows/golang-ci.yml
+++ b/.github/workflows/golang-ci.yml
@@ -19,23 +19,22 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v4
         with:
-          check-latest: "true"
-          go-version: "1.24.x"
+          go-version-file: "go.mod"
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
+        uses: golangci/golangci-lint-action@v6
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
           version: latest
+          install-mode: goinstall
   build-check:
     name: Build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v4
         with:
-          check-latest: "true"
-          go-version: "1.24.x"
+          go-version-file: "go.mod"
       - name: Run build
         run: make build
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: '1.24.x'
+          go-version-file: "go.mod"
 
       - name: Set up Java for Maven
         uses: actions/setup-java@v3
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -42,7 +42,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v4
       with:
-        go-version: '1.24.2'
+        go-version-file: "go.mod"
 
     - name: Set up Java for Maven
       uses: actions/setup-java@v3
--- a/.github/workflows/security-ci.yaml
+++ b/.github/workflows/security-ci.yaml
@@ -26,10 +26,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v4
         with:
-          check-latest: "true"
-          go-version: "1.24.x"
+          go-version-file: "go.mod"
       - name: Install govulncheck
         run: go install golang.org/x/vuln/cmd/govulncheck@latest
       - name: Run vulnerability checks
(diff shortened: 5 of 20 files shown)


drivers/mssql/internal/cdc.go
```diff
@@ -429,3 +441,96 @@ func operationTypeFromCDCCode(code int32) string {
...
+// `maxtrans` transactions per session (default 500), a session that finishes
+// with fewer transactions indicates it reached the end of the transaction log
+// and is fully caught up at that moment.
+func (m *MSSQL) waitForCDCAgentCatchUp(ctx context.Context) (string, error) {
+	var (
+		maxTrans         int
+		pollingIntervalS int
+	)
+	err := m.client.QueryRowContext(ctx, jdbc.MSSQLCDCCaptureJobConfigQuery()).Scan(&maxTrans, &pollingIntervalS)
+	if err != nil {
```
vikaxsh, line 488: shouldn't we return error instead of logging?

drivers/mssql/internal/cdc.go
```diff
@@ -429,3 +441,96 @@ func operationTypeFromCDCCode(code int32) string {
...
+	ticker := time.NewTicker(cdcAgentPollInterval) // poll every 2 seconds
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return "", ctx.Err()
+		case <-deadline:
+			logger.Warnf("CDC agent did not catch up within %s, the CDC agent may not be running", catchUpTimeout)
+			return m.currentMaxLSN(ctx) // fallback to current max LSN
```
vikaxsh, line 517: why do we need fallback here?

drivers/mssql/internal/cdc.go
```diff
@@ -429,3 +441,96 @@ func operationTypeFromCDCCode(code int32) string {
...
+		maxTrans         int
+		pollingIntervalS int
+	)
+	err := m.client.QueryRowContext(ctx, jdbc.MSSQLCDCCaptureJobConfigQuery()).Scan(&maxTrans, &pollingIntervalS)
+	if err != nil {
+		logger.Errorf("unable to query CDC capture job config: %s; using default max trans %d", err, defaultCDCAgentMaxTrans)
+		maxTrans = defaultCDCAgentMaxTrans
+	}
+
+	catchUpTimeout := max(defaultCDCAgentCatchUpTimeout, time.Duration(catchUpTimeoutPollMultiplier*pollingIntervalS)*time.Second)
```
vikaxsh, line 492: can we think about adding a max cap here

drivers/mssql/internal/cdc.go
```diff
@@ -429,3 +441,96 @@ func operationTypeFromCDCCode(code int32) string {
...
+
+	// Record the baseSession scan session before we start waiting.
+	baseSession, err := m.latestCDCScanSession(ctx)
+	if err != nil {
+		return "", fmt.Errorf("failed to query CDC scan sessions: %s", err)
+	}
+
+	logger.Infof("waiting for CDC capture agent to catch up (baseline end_time=%s, maxtrans=%d, pollinginterval=%ds, timeout=%s)", baseSession.endTime.Format(time.RFC3339), maxTrans, pollingIntervalS, catchUpTimeout)
+
+	deadline := time.After(catchUpTimeout)
```
vikaxsh: time.After internally allocates a time.Timer that cannot be garbage collected, use timer.C instead of deadline

drivers/mssql/internal/cdc.go
```diff
@@ -429,3 +441,96 @@ func operationTypeFromCDCCode(code int32) string {
...
+				logger.Infof("CDC agent caught up (session_id=%d, end_time=%s, tran_count=%d)", session.sessionID, session.endTime.Format(time.RFC3339), session.tranCount)
+				return m.currentMaxLSN(ctx)
+			}
+		}
+	}
+}
+
+// latestCDCScanSession returns the most recent completed CDC log scan session.
+func (m *MSSQL) latestCDCScanSession(ctx context.Context) (session cdcScanSession, err error) {
+	err = m.client.QueryRowContext(ctx, jdbc.MSSQLCDCLatestScanSessionQuery()).Scan(&session.sessionID, &session.endTime, &session.tranCount)
```
vikaxsh, line 541: What happens if a newly created instance hasn't completed even a single session yet? Would it fail?

drivers/mssql/internal/cdc.go
vishalm0509: Other tool was adding fallback, so added. Removed it as we have the required permission to query it.

drivers/mssql/internal/cdc.go
vishalm0509: After deadline, if all the lag is not caught if we fail then sync will fail. Instead we can resume from the current maxLSN. Uninterupted sync

drivers/mssql/internal/cdc.go
vishalm0509: The max value that is supported is `86399s (~24h)`, and most prod DB's can usually in rare cases can configure to `5-60s`, very unlikely to go above that

drivers/mssql/internal/cdc.go
vishalm0509: Yes it will fail now

vishalm0509: It is very unlikely to happen, but I have added a check for it now

drivers/mssql/internal/cdc.go
```diff
@@ -429,3 +442,108 @@ func operationTypeFromCDCCode(code int32) string {
...
+func (m *MSSQL) resolveInitialLSN(ctx context.Context) (string, error) {
+	var hasPermission bool
+	err := m.client.QueryRowContext(ctx, jdbc.MSSQLViewDatabaseStatePermissionQuery()).Scan(&hasPermission)
+	if err != nil {
+		return "", fmt.Errorf("failed to check VIEW DATABASE STATE permission: %s", err)
+	}
+
+	if !hasPermission {
+		logger.Warnf("VIEW DATABASE STATE permission not granted; LSN may be lagging behind the transaction log")
+		return m.currentMaxLSN(ctx)
```
vikaxsh, line 471: can we make this split fallback logic at one place

drivers/mssql/internal/cdc.go
vikaxsh: Could this cause issues in existing syncs?

vishalm0509: There are a set of permissions that user needs to grant if they do that properly sync won't fail. Else if we want to be a little lenient we can apply fallbacks `(maxTrans=500, pollingInterval=5s)`

PR #846, opened 2026-02-24
fix: unsigned integer overflow

### LINKED ISSUE #842: Incorrect data type mapping for MySQL for INT UNSGINED
## Description

## Expected behavior

saksham-datazip:
# Description

This PR fixes incorrect handling of UNSIGNED numeric columns in MySQL, which was causing integer overflow issues.

The go-mysql binlog parser returns all integers as signed types, leading to overflow and incorrect (often negative) values for large UNSIGNED data. This change reinterprets such values as their correct unsigned types before processing, ensuring accurate CDC output.

Fixes #842 

## Type of change

# How Has This Been Tested?

I created a table containing different datatype and for unsigned int32 ,the data parsed was negative.
- [x] Scenario A: Created a table with multiple data types, including UNSIGNED INT32. Performed initial sync, then applied updates on the table and triggered CDC. Verified that previously negative values are now correctly interpreted and CDC output is accurate after the fix.

- [x] Scenario B: Performed initial sync using the previous version (where the issue existed), then upgraded to the new version and applied further updates. Verified that new CDC events correctly reflect unsigned values, ensuring backward compatibility for existing pipelines.

# Screenshots or Recordings
Input table format :-
image
Recreated Issue after versioning and cdc update:-
image

## Documentation

## Related PR's (If Any):
N/A

### DIFF
--- a/drivers/mysql/internal/mysql_test_util.go
+++ b/drivers/mysql/internal/mysql_test_util.go
@@ -224,9 +224,9 @@ func insertTestData(t *testing.T, ctx context.Context, db *sqlx.DB, tableName st
 var ExpectedMySQLData = map[string]interface{}{
 	"id_bigint":              int64(123456789012345),
 	"id_int":                 int32(100),
-	"id_int_unsigned":        int32(101),
+	"id_int_unsigned":        int64(101),
 	"id_integer":             int32(102),
-	"id_integer_unsigned":    int32(103),
+	"id_integer_unsigned":    int64(103),
 	"id_mediumint":           int32(5001),
 	"id_mediumint_unsigned":  int32(5002),
 	"id_smallint":            int32(101),
@@ -258,9 +258,9 @@ var ExpectedMySQLData = map[string]interface{}{
 var ExpectedUpdatedData = map[string]interface{}{
 	"id_bigint":              int64(987654321098765),
 	"id_int":                 int64(200),
-	"id_int_unsigned":        int32(201),
+	"id_int_unsigned":        int64(201),
 	"id_integer":             int32(202),
-	"id_integer_unsigned":    int32(203),
+	"id_integer_unsigned":    int64(203),
 	"id_mediumint":           int32(6001),
 	"id_mediumint_unsigned":  int32(6002),
 	"id_smallint":            int32(201),
@@ -290,12 +290,12 @@ var ExpectedUpdatedData = map[string]interface{}{
 }
 
 var MySQLToDestinationSchema = map[string]string{
-	"id":                     "unsigned int",
+	"id":                     "bigint",
 	"id_bigint":              "bigint",
 	"id_int":                 "int",
-	"id_int_unsigned":        "unsigned int",
+	"id_int_unsigned":        "bigint",
 	"id_integer":             "int",
-	"id_integer_unsigned":    "unsigned int",
+	"id_integer_unsigned":    "bigint",
 	"id_mediumint":           "mediumint",
 	"id_mediumint_unsigned":  "unsigned mediumint",
 	"id_smallint":            "smallint",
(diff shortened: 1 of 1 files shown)


drivers/mysql/internal/datatype_conversion.go
```diff
@@ -10,10 +10,11 @@ var mysqlTypeToDataTypes = map[string]types.DataType{
 	"mediumint":          types.Int32,
 	"int":                types.Int32,
 	"integer":            types.Int32,
-	"unsigned int":       types.Int32,
+	"unsigned int":       types.Int64,
 	"unsigned smallint":  types.Int32,
 	"unsigned tinyint":   types.Int32,
 	"unsigned mediumint": types.Int32,
+	"unsigned bigint":    types.Int64,
```
vikaxsh, line 18: how int64 will handle this value?
Signed max = 2^63-1
Unsigned max = 2^64-1

pkg/binlog/filter.go
```diff
@@ -140,7 +141,22 @@ func convertRowToMap(row []interface{}, tableMap *replication.TableMapEvent, col
...
+			return "UNSIGNED TINYINT"
+		case mysql.MYSQL_TYPE_SHORT:
+			return "UNSIGNED SMALLINT"
+		case mysql.MYSQL_TYPE_INT24:
+			return "UNSIGNED MEDIUMINT"
+		case mysql.MYSQL_TYPE_LONG:
+			return "UNSIGNED INT"
+		case mysql.MYSQL_TYPE_LONGLONG:
+			return "UNSIGNED BIGINT"
+		}
```
vikaxsh: consider moving unsigned check inside existing condiitons

drivers/mysql/internal/datatype_conversion.go
saksham-datazip: it has been discussed we are not handling this case

drivers/mysql/internal/mysql.go
```diff
@@ -225,6 +225,10 @@ func (m *MySQL) ProduceSchema(ctx context.Context, streamName string) (*types.St
 			}
 			stream.WithCursorField(columnName)
 			datatype := types.Unknown
+			isInteger := strings.Contains(strings.ToLower(dataType), "int") || strings.EqualFold(dataType, "integer")
```
vishalm0509: Here, `strings.Contains` will handle the case for `strings.EqualFold(dataType, "integer")` as well. 

Do we need both ?

drivers/mysql/internal/mysql.go
```diff
@@ -271,6 +275,36 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
...
+			if v, ok := value.(int64); ok {
+				value = uint64(v)
+			}
+		}
+	} else {
+		// Version < 4 legacy behavior: map all unsigned types to types.Int32
+		if strings.Contains(strings.ToLower(columnType), "unsigned") {
+			columnType = "int"
+		}
+	}
```
vishalm0509, line 301: Earlier, `unsigned bigint` were processed as `BIGINT`, now we are mapping to `INT`. 

Instead of mapping to int, can we just remove the prefix "unsigned "

```go
if strings.HasPrefix(columnType, "unsigned ") {
	columnType = strings.TrimSpace(strings.TrimPrefix(columnType, "unsigned "))
}
```

--- 

In `filter.go`, Now we are setting `UNSIGNED BIGINT` if unsigned is `true` now as that passed through the datatype converted for old jobs, and they are mapped to `int`, but earlier they were mapped to bigint.

pkg/binlog/filter.go
saksham-datazip: solved

drivers/mysql/internal/mysql.go
saksham-datazip: solved

drivers/mysql/internal/mysql.go
```diff
@@ -271,6 +271,35 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
...
+		switch strings.ToLower(columnType) {
+		case "unsigned tinyint":
+			if v, ok := value.(int8); ok {
+				value = uint8(v)
+			}
+		case "unsigned smallint":
+			if v, ok := value.(int16); ok {
+				value = uint16(v)
+			}
+		case "unsigned mediumint", "unsigned int":
```
vishalm0509: add `unsigned_integer` here

PR #854, opened 2026-03-03
fix(postgres): reverse write ordering in PreCDC/PostCDC to prevent LSN mismatch

### LINKED ISSUE #853: Bug: PostCDC write ordering causes irrecoverable LSN mismatch between OLake state and PostgreSQL replication slot
## Bug Description

There is a write-ordering issue in `PostCDC` and `PreCDC` that causes OLake's persisted cursor (LSN) to advance **ahead** of the PostgreSQL replication slot's `confirmed_flush_lsn`. Once this happens, the `validateGlobalState` check triggers a non-retryable error requiring a full "clear destination" — which is destructive and operationally painful for large datasets.

## Root Cause

In [`drivers/postgres/internal/cdc.go:81-91`](https://github.com/datazip-inc/olake/blob/main/drivers/postgres/internal/cdc.go#L81-L91):

```go
func (p *Postgres) PostCDC(ctx context.Context, _ int) error {
    defer waljs.Cleanup(ctx, p.replicator.Socket())
    select {
    case <-ctx.Done():
        return ctx.Err()
    default:
        socket := p.replicator.Socket()
        p.state.SetGlobal(waljs.WALState{LSN: socket.ClientXLogPos.String()})  // ① writes to disk
        return waljs.AcknowledgeLSN(ctx, p.client, socket, false)              // ② tells PostgreSQL
    }
}
```

`SetGlobal()` calls `LogState()` which **synchronously writes the state file to disk** before `AcknowledgeLSN()` sends the `StandbyStatusUpdate` to PostgreSQL. If the process crashes, gets OOM-killed, or AcknowledgeLSN fails/times out between steps ① and ②:

- OLake's state file has LSN X (already flushed to disk)
- PostgreSQL's replication slot still has the old LSN Y (where Y < X)
- On restart, `validateGlobalState` detects the mismatch and returns a **non-retryable error**

The same pattern exists in `PreCDC` on first run ([`cdc.go:44-47`](https://github.com/datazip-inc/olake/blob/main/drivers/postgres/internal/cdc.go#L44-L47)):

```go
p.state.SetGlobal(waljs.WALState{LSN: slot.CurrentLSN.String()})  // ① writes to disk
return waljs.AdvanceLSN(ctx, p.client, p.cdcConfig.ReplicationSlot, slot.CurrentLSN.String())  // ② can fail
```

## Failure Scenarios

| Scenario | Result |
|----------|--------|
| Pod killed/OOMKilled between SetGlobal and AcknowledgeLSN | State file persists, PG slot never updated |
| AcknowledgeLSN times out (5-min poll) | State already written, no rollback |
| Network partition during PostCDC | Local write succeeds, PG unreachable |
| PG replication slot recreated externally | State file untouched, slot reset |

## Proposed Fix

Reverse the ordering: **acknowledge to PostgreSQL first, then persist state.**

```go
// PostCDC - safe ordering
err := waljs.AcknowledgeLSN(ctx, p.client, socket, false)  // ① PG confirms first
if err != nil {
    return err  // don't persist if PG didn't confirm
}
p.state.SetGlobal(waljs.WALState{LSN: socket.ClientXLogPos.String()})  // ② persist only after PG confirmed
```

This way, if the process crashes after step ① but before step ②, on restart OLake's state file will have an **older** LSN than the PG slot. OLake would re-process some changes (at-least-once delivery, which is safe and expected in CDC) rather than being stuck in the current unrecoverable state.

Same fix applies to `PreCDC`:
```go
err := waljs.AdvanceLSN(ctx, p.client, p.cdcConfig.ReplicationSlot, slot.CurrentLSN.String())
if err != nil {
    return err
}
p.state.SetGlobal(waljs.WALState{LSN: slot.CurrentLSN.String()})
p.state.ResetStreams()
```

## Related

- #321 — discusses user-facing behavior on mismatch (fail vs re-sync), but this issue addresses the root cause of **why** the mismatch occurs in the first place

osalloum:
## Summary

Fixes #853

The write ordering in `PreCDC` and `PostCDC` causes an irrecoverable LSN mismatch between OLake's persisted state and the PostgreSQL replication slot's `confirmed_flush_lsn`.

**Before (broken):** `SetGlobal` (writes state to disk) → `AcknowledgeLSN`/`AdvanceLSN` (tells PostgreSQL). If the process crashes or the PG call fails between these two steps, OLake's cursor is ahead of the slot — triggering a non-retryable error on restart that requires "clear destination".

**After (fixed):** `AcknowledgeLSN`/`AdvanceLSN` (confirms with PostgreSQL first) → `SetGlobal` (persists state only after PG confirmed). A crash between the two steps now means OLake's cursor is *behind* the slot — resulting in at-least-once reprocessing, which is safe and recoverable.

### Changes
- `PreCDC`: Call `AdvanceLSN` before `SetGlobal` + `ResetStreams`
- `PostCDC`: Call `AcknowledgeLSN` before `SetGlobal`

### Tests added

**Unit tests** (`cdc_test.go`) for `validateGlobalState`:
- Matching LSN passes validation
- OLake cursor ahead of slot → non-retryable error
- Slot ahead of OLake cursor → non-retryable error
- Empty LSN → non-retryable error
- Invalid LSN format → error

**Integration tests** (`cdc_integration_test.go`) using `embedded-postgres` (PG 16, `wal_level=logical`):
- `TestPreCDC_LSNConsistency`: Verifies that after `PreCDC`, OLake state LSN exactly equals the replication slot's `confirmed_flush_lsn`
- `TestPreCDC_ExistingState_ValidatesLSN`: Verifies that re-running PreCDC with matching state succeeds, and simulating the old bug (state LSN ahead of slot) correctly triggers a mismatch error

## Test plan
- [x] Unit tests for `validateGlobalState` pass
- [x] Integration tests with embedded PostgreSQL pass
- [ ] Existing CDC integration tests pass (requires Docker setup)
- [ ] Verify normal CDC flow still produces correct state file and slot position

### DIFF
--- a/drivers/postgres/go.mod
+++ b/drivers/postgres/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/apache/thrift v0.21.0 // indirect
 	github.com/aws/aws-sdk-go v1.55.6 // indirect
 	github.com/felixge/fgprof v0.9.5 // indirect
+	github.com/fergusstrange/embedded-postgres v1.33.0 // indirect
 	github.com/fsnotify/fsnotify v1.8.0 // indirect
 	github.com/gabriel-vasile/mimetype v1.4.8 // indirect
 	github.com/go-playground/locales v0.14.1 // indirect
@@ -55,6 +56,7 @@ require (
 	github.com/spf13/cast v1.7.1 // indirect
 	github.com/spf13/viper v1.20.1 // indirect
 	github.com/subosito/gotenv v1.6.0 // indirect
+	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
 	github.com/xitongsys/parquet-go v1.6.2 // indirect
 	github.com/xitongsys/parquet-go-source v0.0.0-20241021075129-b732d2ac9c9b // indirect
 	go.mongodb.org/mongo-driver v1.17.3 // indirect
--- a/drivers/postgres/internal/cdc_integration_test.go
+++ b/drivers/postgres/internal/cdc_integration_test.go
@@ -1,0 +1,190 @@
+package driver
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+	"testing"
+
+	"github.com/datazip-inc/olake/constants"
+	"github.com/datazip-inc/olake/pkg/waljs"
+	"github.com/datazip-inc/olake/types"
+	"github.com/datazip-inc/olake/utils"
+	embeddedpostgres "github.com/fergusstrange/embedded-postgres"
+	"github.com/jackc/pglogrepl"
+	"github.com/jmoiron/sqlx"
+	_ "github.com/lib/pq"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	testSlot        = "olake_test_slot"
+	testPublication = "olake_test_pub"
+	testTable       = "public.cdc_test"
+	testPort        = 15432
+)
+
+// startEmbeddedPostgres starts an embedded PostgreSQL instance with logical replication enabled.
+func startEmbeddedPostgres(t *testing.T) *embeddedpostgres.EmbeddedPostgres {
+	t.Helper()
+
+	pg := embeddedpostgres.NewDatabase(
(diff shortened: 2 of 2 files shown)


nayanj98:
Closing this PR as the issue got fixed with the release of OLake v0.5.0 which has 2PC feature.

PR #860, opened 2026-03-09
fix: vulnerability check failure by upgrading go lang version

vaibhav-datazip:
# Description

Go Upgrade: Updated go.mod, go.work, and Dockerfile to use Go 1.25.8.

Dependency Isolation: Added replace directives for google.golang.org/genproto and cloud.google.com/go/compute to the root go.mod and all 8 driver modules. This ensures that every driver (MySQL, Postgres, S3, etc.) can now be built and tidied in isolation without hitting "ambiguous import" errors.

Linting Modernization: Updated .golangci.yml to remove deprecated configuration options (skip-dirs, gci local prefixes, etc.) and rebuilt the golangci-lint binary with Go 1.25.8 to prevent version mismatch errors during commit.

Toolchain Cleanup: Removed stale toolchain directives and invalid comments from various go.mod files to maintain a clean workspace.

Forcing golangci-lint to be built with the project’s Go version, and updates deprecated `.golangci.yml` options to remove config warnings.

## Type of change

# How Has This Been Tested?

- [x] Ran go mod tidy at the root workspace.
- [x] Verified isolated go mod tidy in all 8 driver directories.
- [x] Verified golangci-lint execution locally.

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/drivers/mssql/internal/cdc.go
+++ b/drivers/mssql/internal/cdc.go
@@ -2,7 +2,9 @@ package driver
 
 import (
 	"context"
+	"database/sql"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"time"
 
@@ -17,6 +19,10 @@ const (
 
 	CDCStartLSN = "_cdc_start_lsn" // MSSQL start LSN
 	CDCSeqVal   = "_cdc_seqval"    // MSSQL seqval
+
+	cdcAgentPollInterval          = 2 * time.Second // interval between scan session checks
+	defaultCDCAgentCatchUpTimeout = 15 * time.Minute
+	catchUpTimeoutPollMultiplier  = 3
 )
 
 // CDC capture instance for a table
@@ -27,6 +33,13 @@ type captureInstance struct {
 	startLSN     string
 }
 
+// cdcScanSession represents a completed CDC log scan session from sys.dm_cdc_log_scan_sessions.
+type cdcScanSession struct {
+	sessionID int
+	endTime   time.Time
+	tranCount int
+}
+
 func (m *MSSQL) ChangeStreamConfig() (bool, bool, bool) {
 	return false, false, true // concurrent change streams supported, stream can start after finishing full load
 }
@@ -35,12 +48,6 @@ func (m *MSSQL) ChangeStreamConfig() (bool, bool, bool) {
 func (m *MSSQL) PreCDC(ctx context.Context, streams []types.StreamInterface) error {
 	if !m.cdcSupported {
 		return fmt.Errorf("invalid call; %s not running in CDC mode", m.Type())
-	}
-
-	// Get current max LSN to use as start point for new streams
-	currentLSN, err := m.currentMaxLSN(ctx)
-	if err != nil {
-		return fmt.Errorf("failed to get MSSQL current max LSN: %w", err)
 	}
 
 	streamIDs := make([]string, len(streams))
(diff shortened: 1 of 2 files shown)


.github/workflows/golang-ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           check-latest: "true"
-          go-version: "1.24.x"
+          go-version: "1.25.x"
```
vishalm0509:
```yml
       with:
          check-latest: "true"
          go-version-file: "server/go.mod"
```

--- 

Can we use this instead of hardcoding the version, it will use the version specified in the `go.mod` file. If using, please update in all the relevant places

.github/workflows/golang-ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
...
       - name: golangci-lint
```
saksham-datazip, line 25: why are we not updating golang version in the below files:-
```
integration-tests.yml
performance-test.yml
unit-tests.yml
olake.mdc
test_utils.go

File: drivers/kafka/go.mod [hunk 3]
```diff
@@ -1,8 +1,6 @@
 module github.com/datazip-inc/olake/drivers/kafka
 
-go 1.24.0
-
-toolchain go1.24.8
+go 1.25.8
 
 replace github.com/datazip-inc/olake => ../../
```
vishalm0509, line 5: Do we need this  since we have `go.work`. Can you check once

```
replace github.com/datazip-inc/olake => ../../
```

.golangci.yml
```diff
@@ -143,13 +144,13 @@ issues:
     - linters:
         - gocritic
       text: "unnecessaryDefer:"
-
-run:
-  skip-dirs:
+  exclude-dirs:
```
vishalm0509, line 117: Why changes in this file ?

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build Stage
-FROM golang:1.24-bookworm AS builder
+FROM golang:1.25-bookworm AS builder
```
vishalm0509: Do we need to pin it to `1.25.8`

Makefile
```diff
@@ -4,7 +4,7 @@ gomod:
 	find . -name go.mod -execdir go mod tidy \;
 
 golangci:
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest;
+	GOTOOLCHAIN=$$(go env GOVERSION) go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest;
```
vishalm0509: this used the go version of the environment ( I have 1.26.1 ) but I think we can use the prject version. What do you think ?

.github/workflows/golang-ci.yml
vishalm0509: we can this property 

```yml
install-mode: goinstall
```

this uses go to install golangci and uses the verion from out `go.mod` if we use

```yml
go-version-file: "server/go.mod"
```

--- 

This will fix the current failure of CI as well

.golangci.yml
vaibhav-datazip: removing the deprecated changes

drivers/kafka/go.mod
vaibhav-datazip: individual go.mod should also be able to run right ? like it works for db2 and mssql that's why added these

Dockerfile
vaibhav-datazip: that can be done, kept it for consistency by this way the latest patch was being used

utils/testutils/test_utils.go
```diff
@@ -776,7 +776,7 @@ func (cfg *IntegrationTest) TestIntegration(t *testing.T) {
 
 	t.Run("Discover", func(t *testing.T) {
 		req := testcontainers.ContainerRequest{
-			Image:         "golang:1.24.0",
+			Image:         "golang:1.25.8-bookworm",
```
vaibhav-datazip, line 779: using this because

golang:1.25.8 image uses Debian Trixie, which no longer provides openjdk-17-jre-headless
 
Debian Bookworm still provides openjdk-17-jre-headless, so the existing installCmd (and the rest of the test setup) keeps working without changing the install command or moving to Java 21.

drivers/db2/go.mod
saksham-datazip: why are you not upgrading its version like the way you did in other files

drivers/db2/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/datazip-inc/olake/drivers/db2
 
-go 1.24.0
+go 1.25.8
 
 require (
 	github.com/apache/arrow-go/v18 v18.2.0
```
saksham-datazip, line 6: why are you not updating arrow-go version as well , also look for other driver where we have not upgraded the version in some places while made changes in other places

drivers/db2/go.mod
saksham-datazip: crypto is not upgraded as well so just look for all the packages where version upgraded is needed

vaibhav-datazip: I did go mod tidy , is it breaking something without upgrade ?

vaibhav-datazip: how is this affecting current code behaviour, there are some version mismatch from older code as well

PR #866, opened 2026-03-13
feat: addition of new chunking strategies for Oracle driver

vaibhav-datazip:
# Description

The older strategy used to fail for read replicas or Oracle Active Data Guard standby database, with highly restrictive locks. This PR adds strategies which will allow user to change the chunking strategy based on the permissions they are able to provide.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/destination/parquet/parquet.go
+++ b/destination/parquet/parquet.go
@@ -213,7 +213,7 @@ func (p *Parquet) Check(_ context.Context) error {
 	}
 	// test for s3 permissions
 	if p.s3Client != nil {
-		testKey := filepath.Join(p.config.Prefix, "olake_writer_test/%s", utils.TimestampedFileName(".txt"))
+		testKey := filepath.Join(p.config.Prefix, "olake_writer_test", utils.TimestampedFileName(".txt"))
 		// Try to upload a small test file
 		_, err = p.s3Client.PutObject(&s3.PutObjectInput{
 			Bucket: aws.String(p.config.Bucket),


pkg/jdbc/jdbc.go
```diff
@@ -932,6 +932,46 @@ func OracleChunkScanQuery(stream types.StreamInterface, chunk types.Chunk, filte
 		quotedTable, chunkMin, filterClause)
 }
 
+func OracleObjectIDQuery() string {
```
saksham-datazip: Add a comment expain what these functions are doing

saksham-datazip: Add a comment explaining what this function is doing and also for other

pkg/jdbc/jdbc.go
saksham-datazip: this function comment and OracleTableSizeQuery function comment are same so change the comment for OracleTableSizeQuery

drivers/oracle/internal/backfill.go
```diff
@@ -61,83 +61,218 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+	}
+}
 
-		// TODO: Research about filteration during chunk creation and CREATE_CHUNKS_BY_SQL strategy
-		query = jdbc.OracleChunkCreationQuery(stream, blocksPerChunk, taskName)
-		_, err = o.client.ExecContext(ctx, query)
-		if err != nil {
-			return nil, fmt.Errorf("failed to create chunks: %s", err)
-		}
+func (o *Oracle) splitViaExtents(ctx context.Context, stream types.StreamInterface, blocksPerChunk int64) (*types.Set[types.Chunk], error) {
```
saksham-datazip: in other drivers like mysql and postgres  and previously even in this driver we were using inline function so why we are not using it now?

saksham-datazip: also add a comment above this function

drivers/oracle/internal/backfill.go
```diff
@@ -61,83 +61,218 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+	for idx, startRowID := range startRowIDs {
+		var maxRowID interface{}
+
+		if idx < len(startRowIDs)-1 {
+			maxRowID = startRowIDs[idx+1]
+		} else {
+			maxRowID = nil
+		}
+
+		chunks.Insert(types.Chunk{
```
saksham-datazip, line 203: in other driver ,we were using one more chunk [nil,min) like in mysql and postgres so can we do it here as well for consistency ?

drivers/oracle/internal/backfill.go
```diff
@@ -61,83 +61,218 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+	rows, err := o.client.QueryContext(ctx, jdbc.OracleExtentsQuery(), stream.Namespace(), stream.Name())
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch extents: %s", err)
+	}
+	defer rows.Close()
+
+	var startRowIDs []string
+	var currentBlocks int64 = 0
+	var isFirstChunk = true
+
```
saksham-datazip: can we write it like this ?

```
	var(
		startRowIDs []string
		currentBlocks int64 = 0
		isFirstChunk = true
	)

File: drivers/oracle/internal/backfill.go [hunk 5]
```diff
@@ -61,83 +61,218 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+	// 2. Iterate through extents and group them into chunks
+	for rows.Next() {
+		var fileID, blockID, blocks, objectID int64
+		if err := rows.Scan(&fileID, &blockID, &blocks, &objectID); err != nil {
```
saksham-datazip: can we write it like this ?no need to recreate error variable again?
```
err = rows.Scan(&fileID, &blockID, &blocks, &objectID); 
if err != nil {
	return nil, fmt.Errorf("failed to scan extents: %s", err)
}

File: drivers/oracle/internal/backfill.go [hunk 6]
```diff
@@ -61,83 +61,218 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
-		for rows.Next() {
-			var chunkID int
-			var startRowID, endRowID string
-			err := rows.Scan(&chunkID, &startRowID, &endRowID)
+
+		// If this is the start of a new chunk, generate its starting ROWID
+		if isFirstChunk || currentBlocks == 0 {
+			var startRowID string
+			// Uses DBMS_ROWID.ROWID_CREATE to convert file/block to ROWID
+			err := o.client.QueryRowContext(ctx, jdbc.OracleRowIDCreateQuery(), objectID, fileID, blockID).Scan(&startRowID)
```
saksham-datazip: why we are recreating this err variable again ?

drivers/oracle/internal/backfill.go
```diff
@@ -61,83 +61,218 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
-			} else {
-				maxRowID = nil
-			}
+		// If we hit our block limit, reset the counter so the next loop starts a new chunk
+		if currentBlocks >= blocksPerChunk {
+			currentBlocks = 0
+		}
+	}
+
+	// 3. Format the startRowIDs into the Min/Max Chunk structure (just like your original code)
```
saksham-datazip: remove this comment part and also can we also create a chunk from [nil,min) just like we create in other drivers
```
(just like your original code)

File: drivers/oracle/internal/backfill.go [hunk 8]
```diff
@@ -61,83 +61,218 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+
+// splitViaTableIteration chunks a table by sequentially scanning and counting rows
+func (o *Oracle) splitViaTableIteration(ctx context.Context, stream types.StreamInterface, approxRowCount int64) (*types.Set[types.Chunk], error) {
+	var minRowId, maxRowId string
+
+	// 1. Get the absolute boundaries and total row count for the table
+	query := jdbc.OracleMinMaxCountQuery(stream)
+	err := o.client.QueryRowContext(ctx, query).Scan(&minRowId, &maxRowId)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get min-max row id and total rows: %w", err)
```
saksham-datazip: Use %s here instead

pkg/jdbc/jdbc.go
```diff
@@ -932,6 +932,46 @@ func OracleChunkScanQuery(stream types.StreamInterface, chunk types.Chunk, filte
...
+	AND e.SEGMENT_TYPE IN ('TABLE', 'TABLE PARTITION', 'TABLE SUBPARTITION')
+	AND o.DATA_OBJECT_ID IS NOT NULL
+  ORDER BY o.DATA_OBJECT_ID, e.RELATIVE_FNO, e.BLOCK_ID;`
+}
+
+func OracleRowIDCreateQuery() string {
+	return `SELECT DBMS_ROWID.ROWID_CREATE(1, :1, :2, :3, 0) FROM DUAL`
+}
+
+func OracleMinMaxCountQuery(stream types.StreamInterface) string {
```
saksham-datazip: here we are fetching min and max boundary so can we change name of this function?

pkg/jdbc/jdbc.go
```diff
@@ -932,6 +932,46 @@ func OracleChunkScanQuery(stream types.StreamInterface, chunk types.Chunk, filte
...
+	return fmt.Sprintf(`SELECT MIN(ROWID) AS minRowId, MAX(ROWID) AS maxRowId AS totalRows FROM %q.%q`, stream.Namespace(), stream.Name())
+}
+
+// NextRowIDQuery returns the query to fetch the next max row id
+func NextRowIDQuery(stream types.StreamInterface, ROWID string, chunkSize int64) string {
+	return fmt.Sprintf("SELECT MAX(ROWID),COUNT(*) AS row_count FROM(SELECT ROWID FROM %q.%q AS WHERE ROWID >= '%s' ORDER BY ROWID FETCH FIRST %d ROWS ONLY)", stream.Namespace(), stream.Name(), ROWID, chunkSize)
```
saksham-datazip: is this querry corrrect as after AS their is no alias ?

pkg/jdbc/jdbc.go
```diff
@@ -932,6 +932,46 @@ func OracleChunkScanQuery(stream types.StreamInterface, chunk types.Chunk, filte
...
+   AND e.SEGMENT_NAME = o.OBJECT_NAME
+   AND NVL(e.PARTITION_NAME, 'NONE') = NVL(o.SUBOBJECT_NAME, 'NONE')
+  WHERE e.OWNER = :1
+	AND e.SEGMENT_NAME = :2
+	AND e.SEGMENT_TYPE IN ('TABLE', 'TABLE PARTITION', 'TABLE SUBPARTITION')
+	AND o.DATA_OBJECT_ID IS NOT NULL
+  ORDER BY o.DATA_OBJECT_ID, e.RELATIVE_FNO, e.BLOCK_ID;`
+}
+
+func OracleRowIDCreateQuery() string {
```
saksham-datazip, line 991: where we have used this function ?

drivers/oracle/internal/backfill.go
```diff
@@ -61,83 +61,218 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
-			return nil, fmt.Errorf("failed to create task: %s", err)
-		}
-		defer func(taskName string) {
-			stmt := jdbc.OracleChunkTaskCleanerQuery(taskName)
-			_, err := o.client.ExecContext(ctx, stmt)
-			if err != nil {
-				logger.Warnf("failed to clean up chunk task: %s", err)
-			}
-		}(taskName)
+	switch o.config.ChunkingStrategy {
```
saksham-datazip: can we make splitViaTableIteration as our fallback strategy when splitViaExtents fails due to access problem?

drivers/oracle/internal/backfill.go
```diff
@@ -61,83 +61,218 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+	case "extents":
```
saksham-datazip: can we use tolower function for user ease and also add a log if o.config.ChunkingStrategy is present but wrongly typed or something?

drivers/oracle/internal/backfill.go
vaibhav-datazip: separate functions makes it easy for developing unit tests in future

drivers/oracle/internal/backfill.go
vaibhav-datazip: well, we had not implemented it because the data for that particular table started from the row id, anyways chunk nil to min will be empty

pkg/jdbc/jdbc.go
vaibhav-datazip: count(*) will be as row_count ?

drivers/oracle/internal/backfill.go
vaibhav-datazip: will have to discuss this , according to our previous discussions the original strategy will be split via row id

drivers/oracle/internal/backfill.go
vaibhav-datazip: no we will fallback to original strategy. plus user should add the strategy name according to docs only right ?

pkg/jdbc/jdbc.go
vaibhav-datazip:
```
err := o.client.QueryRowContext(ctx, jdbc.OracleRowIDCreateQuery(), objectID, fileID, blockID).Scan(&startRowID)
```

saksham-datazip: This function i meant ,where we have used this function ?
```
OracleObjectIDQuery

File: pkg/jdbc/jdbc.go (hunk 10 above)
Comment (saksham-datazip) [reply pkg/jdbc/jdbc.go]: i was asking for second 'AS' , as this querry has two 'AS'
```
 FROM %q.%q AS WHERE ROWID
```

File: drivers/oracle/internal/backfill.go (hunk 13 above)
Comment (saksham-datazip) [reply drivers/oracle/internal/backfill.go]: than atleast a log specifying o.config.ChunkingStrategy is present but is not in correct form as intended so falling back to default

File: drivers/oracle/internal/backfill.go (hunk 12 above)
Comment (saksham-datazip) [reply drivers/oracle/internal/backfill.go]: can we keep plit via row id as default but if access fails we should atleast add a log saying u can use this method if u want to or provide access ? or directly fall back to splitViaTableIteration no need to change case structure ?

File: drivers/oracle/internal/backfill.go (hunk 3 above)
Comment (saksham-datazip) [reply drivers/oracle/internal/backfill.go:203]: for the same sake of consistency with other driver i was asking as other drivers use it

File: drivers/oracle/internal/backfill.go (hunk 13 above)
Comment (vaibhav-datazip) [reply drivers/oracle/internal/backfill.go]: we are logging what strategy is being used, will write in docs as well which strategy will be used by default

File: drivers/oracle/internal/backfill.go (hunk 12 above)
Comment (vaibhav-datazip) [reply drivers/oracle/internal/backfill.go]: yeah that was discussed earlier , to keep default as split via rowid only , but we can have a discussion regarding this as well with @hash-data

File: pkg/jdbc/jdbc.go (hunk 11 above)
Comment (vaibhav-datazip) [reply pkg/jdbc/jdbc.go:991]: forgot to remove this query this has been merged with `OracleExtentsQuery`

File: pkg/jdbc/jdbc.go (hunk 10 above)
Comment (vaibhav-datazip) [reply pkg/jdbc/jdbc.go]: fixed this

File: drivers/oracle/internal/backfill.go [hunk 14]
```diff
@@ -61,83 +61,237 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+	defer rows.Close()
+
+	var (
+		startRowIDs   []string
+		currentBlocks int64 = 0
+		isFirstChunk        = true
+	)
+
+	// 2. Iterate through extents and group them into chunks
+	for rows.Next() {
```
vikaxsh: for each chunk , making seperate db calls for ROWID_CREATE,
can we use one single SQL query using Common Table Expressionsfor?
and replace db call in loop with one query
example query
`WITH extents_data AS (...),
     extents_with_sums AS (...),
     extents_with_buckets AS (...),
     selected_extents AS (...),
     expanded_extents AS (...),
     answer AS (...)
SELECT ...`

drivers/oracle/internal/backfill.go
```diff
@@ -61,83 +61,237 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+
+		// If we hit our block limit, reset the counter so the next loop starts a new chunk
+		if currentBlocks >= blocksPerChunk {
+			currentBlocks = 0
+		}
+	}
+
+	chunks.Insert(types.Chunk{
+		Min: nil,
+		Max: startRowIDs[0],
```
vikaxsh: don't we require any check weather 0th element exist or not?

drivers/oracle/internal/backfill.go
```diff
@@ -61,83 +61,237 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
 			if err != nil {
-				logger.Warnf("failed to clean up chunk task: %s", err)
+				return nil, err
 			}
-		}(taskName)
+			startRowIDs = append(startRowIDs, startRowID)
+			isFirstChunk = false
+		}
+
+		currentBlocks += blocks
```
vikaxsh: is there possibility of uneven chunk distribution?
what if single extent value is a lot bigeger
see if we can use width_bucket SQL window function for this
doc:
https://docs.oracle.com/en/database/oracle/oracle-database/19/cncpt/logical-storage-structures.html#GUID-38EC39DC-D9A2-41AD-8AD4-0E1A8A91FFED:~:text=For%20uniform%20extents%2C%20you%20can%20specify%20an%20extent%20size%20or%20use%20the%20default%20size%20of%201%20MB.%20All%20extents%20in%20the%20tablespace%20are%20of%20this%20size.%20Locally%20managed%20temporary%20tablespaces%20can%20only%20use%20this%20type%20of%20allocation.

pkg/jdbc/jdbc.go
```diff
@@ -932,12 +939,51 @@ func OracleChunkScanQuery(stream types.StreamInterface, chunk types.Chunk, filte
...
+	return fmt.Sprintf(`SELECT MIN(ROWID) AS minRowId, MAX(ROWID) AS maxRowId FROM %q.%q`, stream.Namespace(), stream.Name())
+}
+
+// NextRowIDQuery returns the query to fetch the next max row id
+func NextRowIDQuery(stream types.StreamInterface, ROWID string, chunkSize int64) string {
+	return fmt.Sprintf("SELECT MAX(ROWID),COUNT(*) AS row_count FROM(SELECT ROWID FROM %q.%q WHERE ROWID >= '%s' ORDER BY ROWID FETCH FIRST %d ROWS ONLY)", stream.Namespace(), stream.Name(), ROWID, chunkSize)
+}
+
+// OracleTableSizeQuery returns the query to fetch the size of a table in bytes in OracleDB
+func OracleTableSizeQuery(stream types.StreamInterface) string {
```
vikaxsh: don't we need to add schema or name space here?

drivers/oracle/internal/backfill.go
```diff
@@ -61,83 +61,237 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+	query := jdbc.OracleMinMaxRowIDQuery(stream)
+	err := o.client.QueryRowContext(ctx, query).Scan(&minRowId, &maxRowId)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get min-max row id and total rows: %s", err)
+	}
+
+	chunks := types.NewSet[types.Chunk]()
+
+	// Handle empty tables gracefully
+	if approxRowCount == 0 || minRowId == "" {
```
vikaxsh: Does the user need to run a table warm-up query to ensure the correct statistics are populated in Oracle?

drivers/oracle/internal/config.go
```diff
@@ -21,6 +21,7 @@ type Config struct {
 	SSLConfiguration *utils.SSLConfig  `json:"ssl"`
 	JDBCURLParams    map[string]string `json:"jdbc_url_params"`
 	SSHConfig        *utils.SSHConfig  `json:"ssh_config"`
+	ChunkingStrategy string            `json:"chunking_strategy"` // rowid, extents, iteration
```
vikaxsh: don't we need to update ui schema as well?

drivers/oracle/internal/backfill.go
vaibhav-datazip: Commented out the extents strategy and added a todo for later implementation

drivers/oracle/internal/config.go
vaibhav-datazip: we wont need this because if the rowid strategy fails then we will fallback to iteration strategy automatically

pkg/jdbc/jdbc.go
```diff
@@ -948,11 +948,18 @@ func OraclePrimaryKeyColummsQuery(schemaName, tableName string) string {
...
 
+	if chunk.Min == nil {
+		if chunk.Max != nil {
+			return fmt.Sprintf("SELECT * FROM %s WHERE ROWID < '%v' %s", quotedTable, chunk.Max.(string), filterClause)
+		}
+		logger.Warnf("Nil to nil chunk received for table %s", stream.Name())
+		return ""
+	}
+
+	chunkMin := chunk.Min.(string)
```
vikaxsh: can we use switch case here similar to other drivers

for example

```suggestion
 var chunkCond string
    switch {
    case chunk.Min != nil && chunk.Max != nil:
        chunkCond = fmt.Sprintf("ROWID >= '%v' AND ROWID < '%v'", chunk.Min.(string), chunk.Max.(string))
    case chunk.Min != nil:
        chunkCond = fmt.Sprintf("ROWID >= '%v'", chunk.Min.(string))
    case chunk.Max != nil:
        chunkCond = fmt.Sprintf("ROWID < '%v'", chunk.Max.(string))
    default:
        logger.Warnf("Nil to nil chunk received for table %s", stream.Name())
        chunkCond = "1 = 1"
    }

    if filter != "" {
        return fmt.Sprintf("SELECT * FROM %s WHERE (%s) AND (%s)", quotedTable, chunkCond, filter)
    }
```

pkg/jdbc/jdbc.go
```diff
@@ -962,12 +969,49 @@ func OracleChunkScanQuery(stream types.StreamInterface, chunk types.Chunk, filte
...
+}
+*/
+
+// OracleMinMaxRowIDQuery returns the query to fetch the min and max row id of a table in OracleDB
+func OracleMinMaxRowIDQuery(stream types.StreamInterface) string {
+	return fmt.Sprintf(`SELECT MIN(ROWID) AS minRowId, MAX(ROWID) AS maxRowId FROM %q.%q`, stream.Namespace(), stream.Name())
+}
+
+// NextRowIDQuery returns the query to fetch the next max row id
+func NextRowIDQuery(stream types.StreamInterface, ROWID string, chunkSize int64) string {
```
vikaxsh, line 1002: The current fallback chunking repeatedly executes ORDER BY ROWID FETCH FIRST N ,A better approach is to compute chunk boundaries once using ROW_NUMBER() OVER (ORDER BY ROWID)
can we explore this query or for now can add todo as well
```SELECT rowid
FROM (
    SELECT
        rowid,
        ROW_NUMBER() OVER (ORDER BY rowid) rn
    FROM schema.table
)
WHERE MOD(rn, :rows_per_chunk) = 0
ORDER BY rowid;```

File: drivers/oracle/internal/backfill.go [hunk 22]
```diff
@@ -54,90 +54,229 @@ func (o *Oracle) ChunkIterator(ctx context.Context, stream types.StreamInterface
...
+	blocksPerChunk := int64(math.Ceil(float64(constants.EffectiveParquetSize) / float64(blockSize)))
+
+	logger.Debugf("Chunking via DBMS_PARALLEL_EXECUTE for %s.%s", stream.Namespace(), stream.Name())
+	chunks, err := o.splitViaRowId(ctx, stream, blocksPerChunk)
+
+	// If the fast RowID strategy fails for ANY reason (Read-Only, Permissions, etc.) fallback to Table Iteration strategy
+	if err != nil {
+		logger.Debugf("DBMS Parallel Execute failed (%v). Automatically falling back to Table Iteration strategy for stream [%s.%s]", err, stream.Namespace(), stream.Name())
+
+		return o.splitViaTableIteration(ctx, stream, avgRowSize)
```
vikaxsh, line 85: we should add fallback only if get such errors insufficient privileges, database open for read-only access etc
not in case of network error, database down

drivers/oracle/internal/backfill.go
```diff
@@ -54,90 +54,229 @@ func (o *Oracle) ChunkIterator(ctx context.Context, stream types.StreamInterface
...
+
+	query = jdbc.OracleBlockSizeQuery()
+	var blockSize int64
+	err = o.client.QueryRowContext(ctx, query).Scan(&blockSize)
+	if err != nil || blockSize == 0 {
+		logger.Warnf("failed to get block size from query, switching to default block size value 8192")
+		blockSize = 8192
+	}
+	blocksPerChunk := int64(math.Ceil(float64(constants.EffectiveParquetSize) / float64(blockSize)))
+
```
vikaxsh: move this to row id function

drivers/oracle/internal/backfill.go
```diff
@@ -54,90 +54,229 @@ func (o *Oracle) ChunkIterator(ctx context.Context, stream types.StreamInterface
...
+func (o *Oracle) splitViaTableIteration(ctx context.Context, stream types.StreamInterface, avgRowSize float64) (*types.Set[types.Chunk], error) {
+	chunks := types.NewSet[types.Chunk]()
+
+	var minRowId, maxRowId string
+	query := jdbc.OracleMinMaxRowIDQuery(stream)
+	err := o.client.QueryRowContext(ctx, query).Scan(&minRowId, &maxRowId)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get min-max row id: %s", err)
+	}
+	rowsPerChunk := int64(math.Ceil(float64(constants.EffectiveParquetSize) / avgRowSize))
```
vikaxsh, line 100: add a avgRowSize non zero check

drivers/oracle/internal/backfill.go
```diff
@@ -54,90 +54,229 @@ func (o *Oracle) ChunkIterator(ctx context.Context, stream types.StreamInterface
...
 		if err != nil {
-			return nil, fmt.Errorf("failed to create chunks: %s", err)
+			return nil, fmt.Errorf("failed to scan chunk %d: %s", chunkID, err)
 		}
+		startRowIDs = append(startRowIDs, startRowID)
+	}
+
+	chunks.Insert(types.Chunk{
+		Min: nil,
+		Max: startRowIDs[0],
```
vikaxsh, line 191: add a check here for len>0

drivers/oracle/internal/backfill.go
```diff
@@ -54,90 +54,229 @@ func (o *Oracle) ChunkIterator(ctx context.Context, stream types.StreamInterface
...
+		err = rows.Scan(&chunkID, &startRowID, &endRowID)
 		if err != nil {
-			return nil, fmt.Errorf("failed to create chunks: %s", err)
+			return nil, fmt.Errorf("failed to scan chunk %d: %s", chunkID, err)
 		}
+		startRowIDs = append(startRowIDs, startRowID)
+	}
+
+	chunks.Insert(types.Chunk{
+		Min: nil,
```
vikaxsh, line 190: move this out of loop, can take referece from mysql splitViaPrimaryKey function

drivers/oracle/internal/backfill.go
vaibhav-datazip: that wont be needed it will return at least one row id always

pkg/jdbc/jdbc.go
vaibhav-datazip: shouldn't we throw error if default is reached ?

pkg/jdbc/jdbc.go
vaibhav-datazip: will add a todo for it

drivers/oracle/internal/backfill.go
vaibhav-datazip: yeah but that's not necessary we have fallbacks

pkg/jdbc/jdbc.go
vaibhav-datazip: changed this query, it wasn't required

drivers/oracle/internal/backfill.go
vaibhav-datazip: but that can't be zero right ? if there are no rows it will do early exit other wise if table is not populated we return 2048

drivers/oracle/internal/backfill.go
vaibhav-datazip:
```
chunks.Insert(types.Chunk{
		Min: nil,
		Max: startRowIDs[0],
	})

	for idx, startRowID := range startRowIDs {
```

for the rowid strategy its outside for loop already

drivers/oracle/internal/backfill.go
vaibhav-datazip: The first function in this strategy will throw the error , if there is some network error we will get that from second strategy as well .

should we still add a string comparison for error of what error code is being returned ?

vikaxsh: we should be aware about exact error codes, when we wants to run fallback, not always

drivers/oracle/internal/backfill.go
vikaxsh: better to have a non zero check

drivers/oracle/internal/backfill.go
vikaxsh: commented for splitViaTableIteration

pkg/jdbc/jdbc.go
vaibhav-datazip: made this clear

drivers/oracle/internal/backfill.go
vaibhav-datazip: had discussed this, we can fallback to iteration strategy for all errors

PR #868, opened 2026-03-14
fix: fixed signedness bitmap

### LINKED ISSUE #842: Incorrect data type mapping for MySQL for INT UNSGINED

saksham-datazip:
# Description

This PR fixes an issue in the handling of the signedness bitmap for MySQL numeric columns.

Previously, the implementation relied on the UnsignedMap function from the MySQL driver to determine whether numeric columns were signed or unsigned. However, the driver's internal isNumeric logic had a mismatch with MySQL server behavior.

Specifically, the driver’s isNumeric function did not include the YEAR column type as a numeric type, while the MySQL server does include YEAR when generating the signedness bitmap in the binlog TABLE_MAP_EVENT.

Because of this mismatch:

MySQL writes a signedness bit for the YEAR column.

The driver skips YEAR when iterating numeric columns.

This causes a bit offset in the signedness bitmap parsing.

As a result, all numeric columns appearing after a YEAR column may be interpreted with incorrect signedness, which can lead to incorrect values being decoded during CDC processing.

This issue was introduced when switching to the UnsignedMap function from the MySQL driver in the following PR:

https://github.com/datazip-inc/olake/pull/846

This PR resolves the issue by aligning the numeric type handling with MySQL server behavior and ensuring that YEAR is treated consistently when parsing the signedness bitmap.

Relevant MySQL source references for signedness metadata:

https://github.com/mysql/mysql-server/blob/trunk/sql/field_common_properties.h#L11238

https://github.com/mysql/mysql-server/blob/trunk/sql/field_common_properties.h#L204

These references show how MySQL determines whether a column type contains signedness information when generating binlog metadata.

Fixes #842

## Type of change

# How Has This Been Tested?

The fix was validated using CDC pipelines with both Iceberg and Parquet destinations.

A table containing a YEAR column followed by numeric columns was created. CDC replication was executed and the decoded values were verified to ensure that unsigned numeric columns appearing after the YEAR column are interpreted correctly.

Test procedure:

Create a table with a YEAR column and numeric columns following it.

Insert values into unsigned numeric columns.

Run CDC replication.

Verify decoded values in the destination.

- [x] Scenario A — CDC replication to Iceberg destination
- [x] Scenario B — CDC replication to Parquet destination

# Screenshots or Recordings
Output in iceberg:-
image

Output in parquet :-
image

## Documentation

## Related PR's (If Any):
https://github.com/datazip-inc/olake/pull/846

### DIFF
--- a/pkg/binlog/filter.go
+++ b/pkg/binlog/filter.go
@@ -59,10 +59,24 @@ func (f ChangeFilter) FilterRowsEvent(ctx context.Context, e *replication.RowsEv
 		return nil
 	}
 
-	unsignedMap := e.Table.UnsignedMap()
+	bitmapIdx := 0
+	bitMask := byte(0x80)
+	// Iterate through column types and determine if the column is unsigned using SignednessBitmap,
+	// which stores one bit per "signedness" column (integer, YEAR, DECIMAL).
 	columnTypes := make([]string, len(e.Table.ColumnType))
 	for i, ct := range e.Table.ColumnType {
-		columnTypes[i] = mysqlTypeName(ct, unsignedMap != nil && unsignedMap[i])
+		unsigned := false
+		if hasSignednessInBinlog(ct) && len(e.Table.SignednessBitmap) > 0 {
+			if bitmapIdx < len(e.Table.SignednessBitmap) {
+				unsigned = (e.Table.SignednessBitmap[bitmapIdx] & bitMask) != 0
+				bitMask >>= 1
+				if bitMask == 0 {
+					bitMask = 0x80
+					bitmapIdx++
+				}
+			}
+		}
+		columnTypes[i] = mysqlTypeName(ct, unsigned)
 	}
 
 	var rowsToProcess [][]interface{}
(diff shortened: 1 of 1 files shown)


pkg/binlog/filter.go
```diff
@@ -59,10 +59,24 @@ func (f ChangeFilter) FilterRowsEvent(ctx context.Context, e *replication.RowsEv
...
+			if bitmapIdx < len(e.Table.SignednessBitmap) {
+				unsigned = (e.Table.SignednessBitmap[bitmapIdx] & bitMask) != 0
+				bitMask >>= 1
+				if bitMask == 0 {
+					bitMask = 0x80
+					bitmapIdx++
+				}
+			}
+		}
+		columnTypes[i] = mysqlTypeName(ct, unsigned)
```
vikaxsh, line 79: Earlier in our discussion, we decided to keep the `UnsignedMap` function code as it is and only modify the `IsNumericFunction` by adding `mysql.MYSQL_TYPE_YEAR`. Why was this implemented later?

PR #869, opened 2026-03-14
fix: fixed signedness bitmap

### LINKED ISSUE #842: Incorrect data type mapping for MySQL for INT UNSGINED

saksham-datazip:
# Description

This PR fixes an issue in the handling of the signedness bitmap for MySQL numeric columns.

Previously, the implementation relied on the UnsignedMap function from the MySQL driver to determine whether numeric columns were signed or unsigned. However, the driver's internal isNumeric logic had a mismatch with MySQL server behavior.

Specifically, the driver’s isNumeric function did not include the YEAR column type as a numeric type, while the MySQL server does include YEAR when generating the signedness bitmap in the binlog TABLE_MAP_EVENT.

Because of this mismatch:

MySQL writes a signedness bit for the YEAR column.

The driver skips YEAR when iterating numeric columns.

This causes a bit offset in the signedness bitmap parsing.

As a result, all numeric columns appearing after a YEAR column may be interpreted with incorrect signedness, which can lead to incorrect values being decoded during CDC processing.

This issue was introduced when switching to the UnsignedMap function from the MySQL driver in the following PR:

https://github.com/datazip-inc/olake/pull/846

This PR resolves the issue by aligning the numeric type handling with MySQL server behavior and ensuring that YEAR is treated consistently when parsing the signedness bitmap.

Relevant MySQL source references for signedness metadata:

https://github.com/mysql/mysql-server/blob/trunk/sql/field_common_properties.h#L11238

https://github.com/mysql/mysql-server/blob/trunk/sql/field_common_properties.h#L204

These references show how MySQL determines whether a column type contains signedness information when generating binlog metadata.

Fixes #842

## Type of change

# How Has This Been Tested?

The fix was validated using CDC pipelines with both Iceberg and Parquet destinations.

A table containing a YEAR column followed by numeric columns was created. CDC replication was executed and the decoded values were verified to ensure that unsigned numeric columns appearing after the YEAR column are interpreted correctly.

Test procedure:

Create a table with a YEAR column and numeric columns following it.

Insert values into unsigned numeric columns.

Run CDC replication.

Verify decoded values in the destination.

- [x] Scenario A — CDC replication to Iceberg destination
- [x] Scenario B — CDC replication to Parquet destination

# Screenshots or Recordings
Output in iceberg:-
image

Output in parquet :-
image

## Documentation

## Related PR's (If Any):
https://github.com/datazip-inc/olake/pull/846

### DIFF
--- a/.cursor/rules/olake.mdc
+++ b/.cursor/rules/olake.mdc
@@ -10,7 +10,7 @@ alwaysApply: true
 
 OLake is an open-source, high-performance **EL (Extract-Load)** tool that replicates data from databases, Kafka, and S3 into **Apache Iceberg** tables or **plain Parquet** files.
 
-- **Languages:** Go 1.24 (primary) + Java 17 (Iceberg writer)
+- **Languages:** Go 1.25.8 (primary) + Java 17 (Iceberg writer)
 - **Go workspace:** `go.work` links the root module with 8 driver sub-modules
 - **CLI:** Cobra-based commands: `spec`, `check`, `discover`, `sync`, `clear`
 - **Sources:** Postgres, MongoDB, MySQL, Oracle, MSSQL, DB2, Kafka, S3
--- a/.github/workflows/golang-ci.yml
+++ b/.github/workflows/golang-ci.yml
@@ -19,23 +19,22 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v4
         with:
-          check-latest: "true"
-          go-version: "1.24.x"
+          go-version-file: "go.mod"
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
+        uses: golangci/golangci-lint-action@v6
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
           version: latest
+          install-mode: goinstall
   build-check:
     name: Build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v4
         with:
-          check-latest: "true"
-          go-version: "1.24.x"
+          go-version-file: "go.mod"
       - name: Run build
         run: make build
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: '1.24.x'
+          go-version-file: "go.mod"
 
       - name: Set up Java for Maven
         uses: actions/setup-java@v3
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -42,7 +42,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v4
       with:
-        go-version: '1.24.2'
+        go-version-file: "go.mod"
 
     - name: Set up Java for Maven
       uses: actions/setup-java@v3
--- a/.github/workflows/security-ci.yaml
+++ b/.github/workflows/security-ci.yaml
@@ -26,10 +26,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v4
         with:
-          check-latest: "true"
-          go-version: "1.24.x"
+          go-version-file: "go.mod"
       - name: Install govulncheck
         run: go install golang.org/x/vuln/cmd/govulncheck@latest
       - name: Run vulnerability checks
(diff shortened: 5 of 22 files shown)


pkg/binlog/filter.go
```diff
@@ -59,10 +59,10 @@ func (f ChangeFilter) FilterRowsEvent(ctx context.Context, e *replication.RowsEv
 		return nil
 	}
 
-	unsignedMap := e.Table.UnsignedMap()
+	unsignedMap := UnsignedMap(e.Table)
 	columnTypes := make([]string, len(e.Table.ColumnType))
 	for i, ct := range e.Table.ColumnType {
-		columnTypes[i] = mysqlTypeName(ct, unsignedMap != nil && unsignedMap[i])
+		columnTypes[i] = mysqlTypeName(ct, unsignedMap[i])
```
saksham-datazip: add unsignedMap != nil this as well

pkg/binlog/filter.go
```diff
@@ -215,3 +215,45 @@ func mysqlTypeName(t byte, unsigned bool) string {
...
+// Note that only columns with signedness information will be returned.
+// nil is returned if not available or no signedness columns at all.
+func UnsignedMap(tableMap *replication.TableMapEvent) map[int]bool {
+	if len(tableMap.SignednessBitmap) == 0 {
+		return nil
+	}
+	ret := make(map[int]bool)
+	i := 0
+	for _, field := range tableMap.SignednessBitmap {
+		for c := byte(0x80); c != 0; {
```
saksham-datazip: use 0x80 similar to driver code

pkg/binlog/filter.go
```diff
@@ -215,3 +215,45 @@ func mysqlTypeName(t byte, unsigned bool) string {
...
+			if IsNumericColumn(i, tableMap) {
+				ret[i] = field&byte(c) != 0
+				c >>= 1
+			}
+			i++
+			if i >= len(tableMap.ColumnType) {
+				break
```
saksham-datazip: use return ret similar to mysql-driver

pkg/binlog/filter.go
```diff
@@ -215,3 +215,45 @@ func mysqlTypeName(t byte, unsigned bool) string {
...
+	ret := make(map[int]bool)
+	i := 0
+	for _, field := range tableMap.SignednessBitmap {
+		for c := byte(0x80); c != 0; {
+			if IsNumericColumn(i, tableMap) {
+				ret[i] = field&byte(c) != 0
+				c >>= 1
+			}
+			i++
+			if i >= len(tableMap.ColumnType) {
```
saksham-datazip: change len(tableMap.ColumnType) to int(tableMap.ColumnCount)
similar to mysql-drivers code

saksham-datazip: image

pkg/binlog/filter.go
saksham-datazip: solved

pkg/binlog/filter.go
saksham-datazip: solved

pkg/binlog/filter.go
saksham-datazip: solved

pkg/binlog/filter.go
```diff
@@ -215,3 +215,45 @@ func mysqlTypeName(t byte, unsigned bool) string {
 		return fmt.Sprintf("UNKNOWN_TYPE: %d", t)
 	}
 }
+
+// UnsignedMap returns a map: column index -> unsigned.
+// Note that only columns with signedness information will be returned.
+// nil is returned if not available or no signedness columns at all.
+func UnsignedMap(tableMap *replication.TableMapEvent) map[int]bool {
```
vikaxsh: Make TableMapEvent the receiver for UnsignedMap and IsNumericColumn

vikaxsh: since these are from external library, can't extend directly
rename these to unsignedMap and isNumericColumn

pkg/binlog/filter.go
```diff
@@ -215,3 +215,45 @@ func mysqlTypeName(t byte, unsigned bool) string {
...
+	ret := make(map[int]bool)
+	i := 0
+	for _, field := range tableMap.SignednessBitmap {
+		for c := 0x80; c != 0; {
+			if IsNumericColumn(i, tableMap) {
+				ret[i] = field&byte(c) != 0
+				c >>= 1
+			}
+			i++
+			if i >= len(tableMap.ColumnType) {
```
vikaxsh: Once verified that this is an equivalent substitute, it shouldn’t introduce any breaking changes.

saksham-datazip: here columnCount and columnType are defined and used and clearly columnType is defined as biteslice exactly of length columnCount
Line number :- 118 to 123 
https://github.dev/go-mysql-org/go-mysql/replication/row_event.go

pkg/binlog/filter.go
saksham-datazip: changed

vikaxsh (approved):

PR #870, opened 2026-03-14
fix: mysql cdc sync stuck

vaibhav-datazip:
# Description

The Iceberg Java writer does not update the `olake_2pc` table property when there are no files to commit (`filesToCommit.isEmpty()` → early return). On blank CDC syncs (no DML), state and slot advance but metadata does not → state ends up ahead of metadata. A symmetric recovery check (`state != metadata`) then triggers the wrong path on every sync.

---

## MySQL

**Problem:** State (live binlog position from `SHOW MASTER STATUS`) advanced on every sync; metadata only on commit. Blank syncs caused state > metadata → wrong recovery reverted state to metadata → infinite loop (blank sync → recovery → blank sync).

**Fix:**
- **Position normalization:** After `StreamMessages` returns, set `CurrentPos = currentBinlogPos` and return `currentBinlogPos` as metadata so state and metadata stay aligned.
- **Asymmetric recovery:** Trigger recovery only when **metadata > state** (crash path). When **state ≥ metadata** (blank sync), do not trigger recovery; stream forward from state.

---

## Postgres

**Symptom:** After blank syncs, the next CDC sync hung ~5 minutes then failed with `post cdc error: context canceled`.

**Cause:** Symmetric check triggered recovery when state > metadata. We set `recoveryLSN = metadata.LSN` (older). Slot was already at the newer LSN → sync exited immediately → PostCDC tried to acknowledge the older LSN. Postgres only advances `confirmed_flush_lsn`, never rewinds → ack ignored → wait loop (`slot.LSN == walPosition`) never succeeded → timeout.and then fail with:

```
FATAL error occurred while reading records: error occurred while waiting for connections: post cdc error: context canceled
```

**Fix:**
- **Asymmetric recovery:** Trigger recovery only when **metadata > state**; when state > metadata, stream forward from state.
- **AcknowledgeLSN guard:** Treat success when `slot.LSN >= walPosition` instead of `==` so we do not wait forever if the slot is already ahead.

---

## MSSQL

`cdc.fn_cdc_get_max_lsn()` advances for internal transactions even with no user DML. Blank sync → state > metadata. Symmetric check caused every sync to early-return with old metadata LSN → state pinned to metadata forever, no progress.

**Asymmetric recovery** — trigger recovery (reset to metadata, early-return) only when **metadata > state**. When state ≥ metadata, stream forward from state. LSN comparison uses string `>` (safe: 20-char lowercase hex from `hex.EncodeToString` = big-endian numeric order).

---

## MongoDB

**No asymmetric check.** Resume tokens can advance on blank syncs (cluster activity), but wrong recovery only causes re-reads of already-committed events — no hang or fatal. An asymmetric check would require decoding opaque resume tokens (parsing complexity, version sensitivity) for a non-critical optimization. We keep the symmetric check (`prevResumeToken = mtState` on any mismatch) and occasional re-reads on blank syncs can occur due to this.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Blank sync:
Run CDC sync with no DML on the source; confirm state and destination metadata positions match after sync and that the next sync does not trigger incorrect recovery or “no new data.”

- [x] Sync with DML:
Run CDC sync with inserts/updates/deletes; confirm records are replicated and state/metadata both advance to the same position after commit.

- [x] Idle then DML:
Leave source idle long enough for binlog to advance (heartbeats), then run a sync this will update the state and state updating will occur.
Run the sync again this time recovery sync will run the new data wont be synced but state will have same position as metadata.
from next sync onwards the data will start syncing.

- [x] Sync after alter table in mysql:
Add a new column in mysql table that will update the position in state after sync but on later on syncs it will fallback to original state and updated records will be fetched easily.

Postgres CDC was validated with the following scenarios after the asymmetric recovery check and `AcknowledgeLSN` `>=` guard:
- [x] **No state → full refresh + CDC:** Sync run without existing state; full refresh then CDC. State and metadata align after first CDC run.
- [x] **CDC with one record update:** One row updated on source; CDC sync runs; updated record is synced; state and metadata advance together.
- [x] **Blank syncs (LSN advances in state only):** Multiple CDC syncs with no DML. State and slot advance; metadata unchanged (no files to commit). Next sync does **not** trigger recovery and does **not** hang (previously: 5‑minute hang → fatal).
- [x] **Update then CDC after blank syncs:** After blank syncs, update a row and run CDC; record is synced and state/metadata realign.
- [x] **State update failure (crash after commit, before PostCDC):** Metadata ahead of state. Next sync runs recovery (metadata > state), completes without duplicate rows, then state/slot catch up.
- [x] **Insert after state-failure sync:** With metadata > state, insert one row, run sync (recovery completes; new row may not appear in this run). Next sync runs normally and the recently added record is synced.

MSSQL
- [x] ran full refresh , break sync in between and run with state
- [x] cdc completes normally
- [x] run blank syncs and run cdc for few tables and break for other, then run sync again for failed tables, add some more data in the table , run sync with state it will run bounded sync and not sync the new data at first. when you run command again it will sync that data 

Mongo 
- [x] ran full refresh , break sync in between and run with state
- [x] cdc completes normally
- [x] run blank syncs and run cdc for few tables and break for other, then run sync again for failed tables, add some more data in the table , run sync with state it will run bounded sync and not sync the new data at first. when you run command again it will sync that data 
- [x] incremental sync run fine , even with filter. 

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/.github/workflows/security-ci.yaml
+++ b/.github/workflows/security-ci.yaml
@@ -7,12 +7,14 @@ on:
     paths:
       - '**/*.go'
       - '**/*.java'
+      - '**/go.mod'
   pull_request:
     branches:
       - "*"
     paths:
       - '**/*.go'
       - '**/*.java'
+      - '**/go.mod'
   workflow_dispatch:
     inputs:
       logLevel:
@@ -63,6 +65,7 @@ jobs:
           scan-ref: './'
           scanners: 'vuln'
           severity: 'HIGH,CRITICAL'
+          ignore-unfixed: true
 
   # TODO: Add Java Dependency Trivy (removed because trivy getting stuck or run for hours)
   # trivy-java:
--- a/drivers/db2/go.mod
+++ b/drivers/db2/go.mod
@@ -11,7 +11,7 @@ require (
 )
 
 require (
-	cloud.google.com/go/compute/metadata v0.7.0 // indirect
+	cloud.google.com/go/compute/metadata v0.9.0 // indirect
 	dario.cat/mergo v1.0.1 // indirect
 	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
@@ -34,6 +34,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.34.0 // indirect
 	github.com/aws/smithy-go v1.23.0 // indirect
 	github.com/cenkalti/backoff/v4 v4.2.1 // indirect
+	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/containerd/errdefs v1.0.0 // indirect
 	github.com/containerd/errdefs/pkg v0.3.0 // indirect
 	github.com/containerd/log v0.1.0 // indirect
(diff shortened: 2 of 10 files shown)


pkg/waljs/replicator.go
```diff
@@ -175,7 +175,9 @@ func AcknowledgeLSN(ctx context.Context, db *sqlx.DB, socket *Socket, fakeAck bo
 				return fmt.Errorf("failed to get slot position: %s", err)
 			}
 
-			if slot.LSN == walPosition {
+			// Use >= instead of == so that if the slot is already ahead of the
+			// position we sent (e.g. from a previous sync), we don't wait forever.
+			if slot.LSN >= walPosition {
```
hash-data: why?

hash-data: if we are already ahead then dont go for ack

drivers/mongodb/internal/cdc.go
```diff
@@ -73,13 +72,32 @@ func (m *Mongo) StreamChanges(ctx context.Context, streamIndex int, metadataStat
...
+		if !ok {
+			return nil, fmt.Errorf("failed to typecast state resume token of type[%T] to string", prevResumeToken)
+		}
+
+		cmp, err := compareResumeTokens(prevStateStr, mtStateStr)
+		if err != nil {
+			// Tokens could not be decoded — fall back to metadata for safety
+			logger.Warnf("Stream[%s] failed to compare resume tokens (%s), using metadata token", stream.ID(), err)
+			prevResumeToken = mtStateStr
+		} else if cmp < 0 {
```
hash-data: let us compare strings instead of writing this much logic

drivers/mysql/internal/cdc.go
```diff
@@ -134,7 +139,10 @@ func (m *MySQL) StreamChanges(ctx context.Context, streamIndex int, metadataStat
 	if err != nil {
 		return nil, err
 	}
-	return binlog.Binlog{Position: m.BinlogConn.CurrentPos}, nil
+	// Incase of recovery, the current binlog position is updated to the metadata state
+	// so that the next sync starts from the correct position
+	m.BinlogConn.CurrentPos = currentBinlogPos
```
hash-data: in postgres we check, is it not required here?

Image

drivers/mongodb/internal/cdc.go
```diff
@@ -72,13 +72,26 @@ func (m *Mongo) StreamChanges(ctx context.Context, streamIndex int, metadataStat
...
+		if !ok {
+			return nil, fmt.Errorf("failed to typecast metadata state of type[%T] to string", mtState)
+		}
+		prevStateStr, ok := prevResumeToken.(string)
+		if !ok {
+			return nil, fmt.Errorf("failed to typecast state resume token of type[%T] to string", prevResumeToken)
+		}
+
+		// TODO: addition of all the state updations in metadata file even for blank sync scenario
+		// metadata > state → crash-recovery path (metadata committed but state write failed).
```
hash-data: use typeutils.Compare

PR #875, opened 2026-03-18
fix: correct incremental cursor timezone handling for TZ-naive TIMESTAMP columns

### LINKED ISSUE #867: Incremental jobs from Oracle to Iceberg do not pick up changes
## Environment

- **Deploy method**: Kubernetes with Helm chart `0.0.14`
- **Olake version**:  `0.5.0`
- **OS**: N/A
- **Cloud Provider**: `GCP`
- **Params**: chart-provided NFS and Postgres, increased memory requests/limits

## Description

We are trying to set up an Oracle to Iceberg job with incremental updates. We set the cursor to a datetime field that updates when a row is updated. However, OLake doesn't seem to copy the updated data, and no Iceberg snapshot is created. New rows do not seem to get picked up either. The cursor seems stuck on the initial highest value of the datetime field.

## Steps to reproduce

* Set up a table in an Oracle database with ID (pk), datetime field, and two text fields
* Set up the Oracle to Iceberg job, cursor set to datetime field, and perform initial copy
* Try to modify some records, making sure that the datetime field is changed
* Run job again

## Expected behavior

* A new Iceberg snapshot appears in catalog
* New parquet files are created on destination

## Actual behavior

* Zero rows ingested, no new snapshot, no new parquet files

vikaxsh:
# Description

- Oracle's plain TIMESTAMP and DATE columns store wall-clock time with no timezone, but go-ora attaches the session-timezone offset when returning values to Go , causing the saved cursor to be shifted by the UTC offset and re-fetching already-synced records on every incremental run.
- Fixed by stripping the session-TZ offset in dataTypeConverter record path) and FormatCursorValue (SQL MAX path)

Fixes #867 

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/destination/iceberg/iceberg.go
+++ b/destination/iceberg/iceberg.go
@@ -365,12 +365,11 @@ func (i *Iceberg) FlattenAndCleanData(ctx context.Context, records []types.RawRe
 		return false, nil, nil, fmt.Errorf("failed to extract schema from records: %s", err)
 	}
 
-	filter, isLegacy, err := i.stream.GetFilter()
-	if err != nil {
-		return false, nil, nil, fmt.Errorf("failed to parse stream filter: %s", err)
-	}
-
 	if i.options.ApplyFilter {
+		filter, isLegacy, filterErr := i.stream.GetFilter()
+		if filterErr != nil {
+			return false, nil, nil, fmt.Errorf("failed to parse stream filter: %s", filterErr)
+		}
 		records, err = typeutils.FilterRecords(ctx, records, filter, isLegacy, recordsSchema)
 		if err != nil {
 			return false, nil, nil, fmt.Errorf("failed to filter records: %s", err)
--- a/destination/parquet/parquet.go
+++ b/destination/parquet/parquet.go
@@ -408,11 +408,11 @@ func (p *Parquet) FlattenAndCleanData(ctx context.Context, records []types.RawRe
 	}); err != nil {
 		return false, nil, nil, fmt.Errorf("failed to reformat records: %s", err)
 	}
-	filter, isLegacy, err := p.stream.GetFilter()
-	if err != nil {
-		return false, nil, nil, fmt.Errorf("failed to parse stream filter: %s", err)
-	}
 	if p.options.ApplyFilter {
+		filter, isLegacy, filterErr := p.stream.GetFilter()
+		if filterErr != nil {
+			return false, nil, nil, fmt.Errorf("failed to parse stream filter: %s", filterErr)
+		}
 		records, err = typeutils.FilterRecords(ctx, records, filter, isLegacy, p.schema)
 		if err != nil {
 			return false, nil, nil, fmt.Errorf("failed to filter records: %s", err)
--- a/utils/typeutils/filter.go
+++ b/utils/typeutils/filter.go
@@ -22,12 +22,12 @@ type parsedCondition struct {
 // FilterRecords applies filtering ONLY for new filters.
 // For legacy filters, records are returned unchanged.
 func FilterRecords(ctx context.Context, records []types.RawRecord, filter types.FilterConfig, isLegacy bool, schema any) ([]types.RawRecord, error) {
-	if isLegacy {
-		logger.Warnf("legacy filter detected, skipping destination filtering")
+	if len(filter.Conditions) == 0 {
 		return records, nil
 	}
-	if len(filter.Conditions) == 0 {
-		logger.Debugf("no filter conditions, returning records as-is")
+	if isLegacy {
+		// Legacy filters are pushed down to SQL (see pkg/jdbc SQLFilter); skip re-filtering here.
+		logger.Warnf("legacy filter detected, skipping filtering records")
 		return records, nil
 	}
 	logger.Infof("filtering records with filter: %+v", filter)


drivers/oracle/internal/oracle.go
```diff
@@ -211,10 +211,34 @@ func (o *Oracle) ProduceSchema(ctx context.Context, streamName string) (*types.S
...
+//	DATE (12)          → "DATE"         — Oracle DATE columns
+//	TimeStampDTY (180) → "TimeStampDTY" — TIMESTAMP columns in regular SELECT results
+//	TIMESTAMP (187)    → "TIMESTAMP"    — TIMESTAMP in RETURNING INTO / PL/SQL OUT params
+//
+// Strip the attached offset to preserve wall-clock digits consistently as UTC.
+var tzNaiveOracleTypes = map[string]bool{
+	"date":         true,
+	"timestampdty": true,
+	"timestamp":    true,
+}
```
vaibhav-datazip, line 225: instead of map you can use a set

drivers/oracle/internal/incremental.go
```diff
@@ -43,9 +46,41 @@ func (o *Oracle) StreamIncrementalChanges(ctx context.Context, stream types.Stre
...
+			stream.Namespace(), stream.Name(), cursorField, err)
+		return value
+	}
+	// GetMaxCursorValues bypasses dataTypeConverter, so go-ora attaches the session timezone
+	// to TZ-naive columns and the original timezone to TZ-aware columns. Normalize here so
+	// the cursor values entering FormatCursorValue are already plain UTC.
+	upper := strings.ToUpper(dataType)
+	if strings.Contains(upper, "WITH TIME ZONE") || strings.Contains(upper, "WITH LOCAL TIME ZONE") {
+		// TZ-aware: convert to the correct UTC instant.
+		return t.UTC()
```
vaibhav-datazip, line 82: do we need this, isn't the timestamp already correct for this datatype ?
Also this might affect backward compatibility and data might get skipped or duplicated.

drivers/oracle/internal/incremental.go
```diff
@@ -43,9 +46,41 @@ func (o *Oracle) StreamIncrementalChanges(ctx context.Context, stream types.Stre
...
+		return value
+	}
+	var dataType string
+	query := jdbc.OracleColumnDataTypeQuery(stream.Namespace(), stream.Name(), strings.ToUpper(cursorField))
+	if err := o.client.QueryRowContext(ctx, query).Scan(&dataType); err != nil {
+		logger.Warnf("normalizeCursorValue: failed to get DATA_TYPE for %s.%s.%s, cursor may be incorrect: %s",
+			stream.Namespace(), stream.Name(), cursorField, err)
+		return value
+	}
+	// GetMaxCursorValues bypasses dataTypeConverter, so go-ora attaches the session timezone
```
vaibhav-datazip: The comment and function name are different

drivers/oracle/internal/incremental.go
```diff
@@ -43,9 +46,41 @@ func (o *Oracle) StreamIncrementalChanges(ctx context.Context, stream types.Stre
...
+	maxPrimary = o.normalizeCursorValue(ctx, stream, primaryCursor, maxPrimary)
+	if secondaryCursor != "" {
+		maxSecondary = o.normalizeCursorValue(ctx, stream, secondaryCursor, maxSecondary)
+	}
+	return maxPrimary, maxSecondary, nil
+}
+
+// normalizeCursorTime converts a cursor time.Time from GetMaxCursorValues to plain UTC,
+// applying wall-clock strip for TZ-naive Oracle columns and .UTC() for TZ-aware ones.
+func (o *Oracle) normalizeCursorValue(ctx context.Context, stream types.StreamInterface, cursorField string, value any) any {
```
vaibhav-datazip, line 64: instead of this shouldn't we just use explicit UTC while connecting to oracle , it wont have any affect on TSTZ and so much handling will be reduced

vikaxsh: this is the issue we are solving in this pr, timestamp without timezone is getting converted into utc and leading to dupicate data

drivers/oracle/internal/oracle.go
vikaxsh: why?

drivers/oracle/internal/incremental.go
vikaxsh: let's discuss

drivers/oracle/internal/oracle.go
```diff
@@ -211,10 +211,34 @@ func (o *Oracle) ProduceSchema(ctx context.Context, streamName string) (*types.S
...
+//
+//	DATE (12)          → "DATE"         — Oracle DATE columns
+//	TimeStampDTY (180) → "TimeStampDTY" — TIMESTAMP columns in regular SELECT results
+//	TIMESTAMP (187)    → "TIMESTAMP"    — TIMESTAMP in RETURNING INTO / PL/SQL OUT params
+//
+// Strip the attached offset to preserve wall-clock digits consistently as UTC.
+var tzNaiveOracleTypes = map[string]bool{
+	"date":         true,
+	"timestampdty": true,
+	"timestamp":    true,
```
vaibhav-datazip: timestamp isn't present in datatype conversion map so this wont get triggered. either you need to add this in that map as well

pkg/jdbc/jdbc.go
```diff
@@ -941,6 +941,11 @@ func OracleTableDetailsQuery(schemaName, tableName string) string {
 	return fmt.Sprintf("SELECT column_name, data_type, nullable, data_precision, data_scale FROM all_tab_columns WHERE owner = '%s' AND table_name = '%s'", schemaName, tableName)
 }
 
+// OracleColumnDataTypeQuery returns the query to fetch the data type of a column in OracleDB
+func OracleColumnDataTypeQuery(schemaName, tableName, columnName string) string {
+	return fmt.Sprintf("SELECT DATA_TYPE FROM ALL_TAB_COLUMNS WHERE OWNER = '%s' AND TABLE_NAME = '%s' AND COLUMN_NAME = '%s'", schemaName, tableName, columnName)
+}
+
```
vaibhav-datazip, line 948: we are using this same function inside `IncrementalValueFormatter` function can this query call be reduced somehow or can we abstract that query and use this function only instead ?

PR #879, opened 2026-03-19
fix: sync invalid ENUM values as empty string

### LINKED ISSUE #876: Empty ENUM fields are not handled properly in MySQL CDC
## Description
MySQL allows to have empty string in ENUM columns even if empty string was not in the list of possible values.
From MySQL doc
 https://dev.mysql.com/doc/refman/8.4/en/enum.html
```
An enumeration value can also be the empty string ('') or NULL under certain circumstances:

If you insert an invalid value into an ENUM (that is, a string not present in the list of permitted values), the empty string is inserted instead as a special error value. This string can be distinguished from a “normal” empty string by the fact that this string has the numeric value 0. See [Index Values for Enumeration Literals](https://dev.mysql.com/doc/refman/8.4/en/enum.html#enum-indexes) for details about the numeric indexes for the enumeration values.
If strict SQL mode is enabled, attempts to insert invalid ENUM values result in an error.

```
OLake catches this situation but generates this error and fails:
"invalid ENUM value detected for column source: index 0 indicates an error value (empty string) from MySQL"
 

## Steps to reproduce
```
CREATE TABLE `enum_test` (
  `id` int unsigned NOT NULL AUTO_INCREMENT,
  `source` enum('OFFLINE','AVATAX','BUFFER','SKIPPED') DEFAULT NULL,
  PRIMARY KEY (`id`),
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3;

SET SQL_MODE ='';

INSERT enum_test VALUES (1, 'OFFLINE');
INSERT enum_test VALUES (2, 'XYZ');
COMMIT;

#Run Olake here

INSERT enum_test VALUES (3, 'AVATAX');
INSERT enum_test VALUES (4, 'XYZ2');
COMMIT;

#Run Olake here with CDC

INSERT enum_test VALUES (5, 'BUFFER');
INSERT enum_test VALUES (6, NULL);
COMMIT;
```

## Expected behavior
Since its allowed behavior in MySQL OLake should put empty string into ENUM column and continue to read other records.

## Actual behavior
Fails with error
"invalid ENUM value detected for column source: index 0 indicates an error value (empty string) from MySQL"

vishalm0509:
# Description

sync invalid `ENUM` values as empty string

Fixes #876 

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/drivers/abstract/abstract.go
+++ b/drivers/abstract/abstract.go
@@ -237,10 +237,11 @@ func generateThreadID(streamID, hash string) string {
 // The writer parameter can be either:
 //   - *destination.WriterThread for a single writer
 //   - map[string]*destination.WriterThread for multiple writers keyed by stream ID
-//
-// The threadID and closeMessage parameters are optional (empty string means not used) and only apply to single writer cases
-func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *error, writer any, threadID string, mtState any) {
-	// Cancel context if there's an error, so other threads using this context can detect the failure
+func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *error, writer any, threadID string, mtState *any) {
+	if r := recover(); r != nil {
+		*err = utils.Ternary(*err == nil, fmt.Errorf("panic recovered: %v", r), fmt.Errorf("%s: panic recovered: %v", *err, r)).(error)
+	}
+
 	if *err != nil {
 		cancel()
 	}
@@ -248,7 +249,7 @@ func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *er
 	var metadataState any
 	var closeErr error
 	if mtState != nil {
-		metadataState, closeErr = types.SetMetadataState(mtState, threadID)
+		metadataState, closeErr = types.SetMetadataState(*mtState, threadID)
 		if closeErr != nil {
 			closeErr = fmt.Errorf("failed to set metadata state: %s", closeErr)
 		}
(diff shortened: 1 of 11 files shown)


pkg/binlog/filter.go
```diff
@@ -129,10 +129,12 @@ func convertRowToMap(row []interface{}, tableMap *replication.TableMapEvent, col
 				// for an update CDC event, the key of enum value is passed in binlog events which is always in int64
 				// during such a case, we need to find out the enum value of it from the index
 				if idx, isInt64 := val.(int64); isInt64 {
+					// MySQL stores invalid ENUM inserts as index 0 (special error value), which maps to empty string.
 					if idx == 0 {
-						return nil, fmt.Errorf("invalid ENUM value detected for column %s: index 0 indicates an error value (empty string) from MySQL", columns[i])
+						val = ""
+					} else {
+						val = enumValues[idx-1]
```
vikaxsh, line 135: can we remove if else here?
example:
```
if tableMap.IsEnumColumn(i) && val != nil {
			if enumValues, ok := enumMap[i]; ok {
				indexedEnumValues := append([]string{""}, enumValues...)
				// Binlog update events pass ENUM values as int64 indices.
				if idx, isInt64 := val.(int64); isInt64 && idx >= 0 && idx < int64(len(indexedEnumValues)) {
					val = indexedEnumValues[idx]
				}
			}
		}
```

PR #903, opened 2026-04-01
feat: skip ProduceSchema for all the drivers during sync

### LINKED ISSUE #838: omit usage of ProduceSchema when sync command is used
There is a produce schema function in mongodb , it mainly used to sample the table data and discover the schema structure of mongo db . But it also runs during full refresh currently which might not be required as it kills the speed of data transfer as well in some cases. The goal is to not trigger produce schema during syncing of data.

https://github.com/datazip-inc/olake/blob/c4ec012fef473e5c15c0077f3b5cdb4493f49125/drivers/mongodb/internal/mon.go#L171-L174

---

###### This issue was generated by [todo-issue](https://github.com/DerJuulsn/todo-issue) based on a `TODO` comment in c4ec012fef473e5c15c0077f3b5cdb4493f49125. It&#x27;s been assigned to @vaibhav-datazip because they committed the code.

vaibhav-datazip:
# Description

During `sync`, OLake still calls `Discover` → `ProduceSchema` so the connector can validate the live source against `streams.json`. The same `ProduceSchema` path used by `discover` was doing **expensive work on every visible stream**, even though the catalog already holds the full schema from prior `discover` command run in `streams.json`.

**MongoDB:** schema inference used two collection scans with `SetLimit(10000)` each (natural ascending and descending), i.e. **up to ~20,000 documents** per collection.

**Kafka:** `ProduceSchema` performed message-based schema discovery (readers, offsets, decoding).

This PR reduces sync-time cost in two ways:

1. **`SyncContext` on the sync path** — `sync` attaches `constants.SyncContext` to the request context. Drivers skip heavy inference in `ProduceSchema` when it is set, and rely on the catalog for column/schema details. Lighter steps still run where needed (e.g. primary-key / index metadata for validation).
2. **Selected streams only during sync** — `AbstractDriver.Discover` accepts optional stream name filters built from `catalog.SelectedStreams`, so `ProduceSchema` runs only for streams the user selected, not for every table/collection/topic returned by `GetStreamNames`. The `discover` command is unchanged (no filter, no `SyncContext`).

`ConfiguredStream.Validate` was adjusted so incremental **cursor** validation is skipped, because it can be fetched only via sampling in mongo.

Fixes #838

## Type of change

# How Has This Been Tested?

- [x] ** Discover unchanged:** Run `olake discover` with a source config; confirm full catalog output and that `streams.json` still looks correct (all expected streams).
- [x] **Sync faster path:** Run `olake sync` with an existing `streams.json` and a **subset** of `selected_streams`; confirm sync completes, record counts match expectations (done for mongo and mysql).

- [x] monogdb sync , along with schema evolution
- [x] postgres sync , along with schema evolution
- [x] kafka sync, both json and Avro , along with new field introduction
   

# Screenshots or Recordings

N/A (behavioral / performance change)

## Documentation


vikaxsh:
can we think of creating a `DiscoverForSync`  fucnction which would call `ProduceSchemaLite `instead of ProduceSchema for selected streams only during sync
`ProduceSchemaLite` = cheap part only (PKs + sync modes)
`ProduceSchema` = calls `ProduceSchemaLite `internally + adds the expensive schema inference on top

drivers/mssql/internal/incremental.go
```diff
@@ -50,4 +50,4 @@ func (m *MSSQL) FetchMaxCursorValues(ctx context.Context, stream types.StreamInt
 		return nil, nil, err
 	}
 	return maxPrimaryCursorValue, maxSecondaryCursorValue, nil
-}
\ No newline at end of file
+}
```
vaibhav-datazip: remove unnecessary changes

vikaxsh: remove this

drivers/abstract/abstract.go
```diff
@@ -71,6 +71,16 @@ func (a *AbstractDriver) Discover(ctx context.Context, maxDiscoverThreads int) (
 	if err != nil {
 		return nil, fmt.Errorf("failed to get stream names: %s", err)
 	}
```
vikaxsh: can we remove this as well in case of sync comand

protocol/sync.go
vikaxsh, line 138: streams would be always nil, so the metric would be wrong every time

vikaxsh (commented):
sync no longer skips stale or catalog entries; it will now try to run them and can fail the whole sync instead of warning and skipping the bad stream.

drivers/abstract/abstract.go
vaibhav-datazip: no its needed for s3 , as explained in the comment of skip as well. we are storing the files needed to sync in this step only

vikaxsh: image
this part

protocol/sync.go
vaibhav-datazip: as the catalog is already being passed here, we can directly use length of catalog.streams for total number of streams instead of passing streams here

drivers/abstract/abstract.go
vaibhav-datazip: this is also used for validation if maxDiscoverThreads passed by user is 0 we fallback to config threads

protocol/sync.go
vikaxsh: can we move filter validation outside this if block? as it only uses elem (schema + config from catalog)

vikaxsh (commented):
lgtm

vikaxsh (approved):

PR #904, opened 2026-04-03
test: kafka integration test

saksham-datazip:
# Description

This PR introduces an integration test for the Kafka driver to validate end-to-end data flow and ensure correctness during sync operations.

Along with the test addition, two bugs were identified and fixed:

transformer.go: Fixed an issue related to parsing logic which could lead to incorrect data transformation under certain scenarios.
compare.go: Resolved a bug in comparison logic that was causing mismatches during validation.

These changes improve the reliability of Kafka integrations and ensure more accurate test validation.

No additional dependencies are required for this change.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

All tests were executed locally and passed successfully.

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings
N/A

## Documentation

## Related PR's (If Any):


.github/workflows/integration-tests.yml
```diff
@@ -100,6 +101,24 @@ jobs:
...
+            docker exec kafka kafka-topics --bootstrap-server localhost:9092 --list
+
+      - name: Wait for Kafka2
+        uses: nick-fields/retry@v2
+        with:
+          timeout_minutes: 5
+          max_attempts: 30
+          retry_wait_seconds: 5
+          command: |
+            docker exec kafka2 kafka-topics --bootstrap-server localhost:9092 --list
```
ImDoubD-datazip: why a separate checkup is there? Only single checkup for kafka cluster should be present.

drivers/kafka/internal/testdata/json/iceberg_destination.json
ImDoubD-datazip, line 1: no need for this. can we somehow use the above one itself?

drivers/kafka/internal/testdata/json/parquet_destination.json
ImDoubD-datazip, line 1: no need for this. can we somehow use the above one itself?

drivers/kafka/docker-compose.yml
```diff
@@ -0,0 +1,90 @@
...
+      KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
+      KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
+    healthcheck:
+      test: ["CMD-SHELL", "kafka-broker-api-versions --bootstrap-server localhost:9092 || exit 1"]
+      interval: 10s
+      timeout: 10s
+      retries: 5
+      start_period: 30s
+    volumes:
+      - kafka2_storage:/var/lib/kafka/data
```
ImDoubD-datazip: no need for 2 clusters.

drivers/kafka/internal/kafka_test.go
```diff
@@ -0,0 +1,84 @@
...
+				TestConfig:                       testutils.GetTestConfig(string(constants.Kafka), "avro"),
+				Namespace:                        "topics",
+				ExpectedData:                     ExpectedKafkaAvroData,
+				ExpectedUpdatedData:              ExpectedKafkaUpdatedAvroData,
+				DestinationDataTypeSchema:        KafkaToDestinationAvroSchema,
+				UpdatedDestinationDataTypeSchema: EvolvedKafkaToDestinationAvroSchema,
+				DefaultCDCColumnsSchema:          ExpectedKafkaDefaultCDCColumnsSchema,
+				ExecuteQuery:                     ExecuteQueryForAvro,
+				DestinationDB:                    "kafka_topics",
+				PartitionRegex:                   "/{int_value,identity}",
```
ImDoubD-datazip: change the partition regex to partition on timestamp based key.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+	"time"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/datazip-inc/olake/utils"
+	"github.com/linkedin/goavro/v2"
+	"github.com/segmentio/kafka-go"
+	"github.com/stretchr/testify/require"
+)
+
+var (
```
ImDoubD-datazip: do the values here change? I mean shouldn't it be const?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+	// Message key and value for JSON and Avro
+	jsonKey          = []byte("json-key")
+	avroKey          = []byte("avro-key")
+	jsonValue        = []byte(`{"int_value": 100,"float_value": 99.99,"boolean_true": true,"boolean_false": false,"timestamp_value": "2026-03-22T14:30:00Z","string_value": "test_string"}`)
+	jsonEvolvedValue = []byte(`{"int_value": 100,"float_value": 99.99,"boolean_true": true,"boolean_false": false,"timestamp_value": "2026-03-22T14:30:00Z","string_value": "test_string", "id_int": 101}`)
+	filterValue1     = []byte(`{"string_value": "","float_value": 99.99}`)
+	filterValue2     = []byte(`{"string_value": "filter_string","float_value": 100.00}`)
```
ImDoubD-datazip: why are 2 filter values present?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+		"string_value":    "test_string",
+		"id_int":          int64(101),
+	}
+)
+
+// ExecuteQueryForJson executes Kafka queries for testing based on the operation type
+func ExecuteQueryForJson(ctx context.Context, t *testing.T, streams []string, operation string, fileConfig bool) {
+	t.Helper()
+
+	broker := "127.0.0.1:29092"
```
ImDoubD-datazip: put it in const block above

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+		AllowAutoTopicCreation: false,
+	}
+	defer writer.Close()
+
+	switch operation {
+	case "create":
+		createKafkaTopic(ctx, t, broker, streams[0])
+	case "clean":
+		deleteKafkaTopic(ctx, t, broker, streams[0])
+		createKafkaTopic(ctx, t, broker, streams[0])
```
ImDoubD-datazip: why are creating the kafka topic again, in clean?

utils/testutils/test_utils.go
```diff
@@ -481,12 +483,31 @@ func (cfg *IntegrationTest) testIcebergFullLoadAndCDC(
...
+	kafkaTestCases := []syncTestCase{
+		{
+			name:      "CDC - strict - insert",
+			operation: "",
+			useState:  false,
+			opSymbol:  "c",
+			expected:  cfg.ExpectedData,
+		},
+		{
+			name:      "CDC - update",
```
ImDoubD-datazip: CDC - strict - update

utils/testutils/test_utils.go
```diff
@@ -481,12 +483,31 @@ func (cfg *IntegrationTest) testIcebergFullLoadAndCDC(
...
+
+	testCases = utils.Ternary(slices.Contains(constants.OnlyStrictCDCDriver, constants.DriverType(cfg.TestConfig.Driver)), kafkaTestCases, testCases).([]syncTestCase)
+
 	// Run each test case
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			// schema evolution
 			if tc.operation == "update" {
-				if cfg.TestConfig.Driver != "mongodb" && cfg.TestConfig.Driver != "mssql" {
+				if cfg.TestConfig.Driver != "mongodb" && cfg.TestConfig.Driver != "mssql" && cfg.TestConfig.Driver != "kafka" {
```
ImDoubD-datazip, line 515: why are we not evolving the schema when driver is kafka?

utils/testutils/test_utils.go
```diff
@@ -561,12 +582,31 @@ func (cfg *IntegrationTest) testParquetFullLoadAndCDC(
...
```
ImDoubD-datazip, line 614: same here?

.github/workflows/integration-tests.yml
```diff
@@ -143,7 +162,7 @@ jobs:
 
       - name: Run Integration Tests
         run: |  
-          go test -v -p 6 ./drivers/mysql/internal/... ./drivers/postgres/internal/... ./drivers/mongodb/internal/... ./drivers/oracle/internal/... ./drivers/db2/internal/... ./drivers/mssql/internal/... -timeout 0 -run 'Integration'
+          go test -v -p 8 ./drivers/kafka/internal/... ./drivers/mysql/internal/... ./drivers/postgres/internal/... ./drivers/mongodb/internal/... ./drivers/oracle/internal/... ./drivers/db2/internal/... ./drivers/mssql/internal/... -timeout 0 -run 'Integration'
```
ImDoubD-datazip: why 8, there is only one integration test right for kafka? it should be 7.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+	filterValue2     = []byte(`{"string_value": "filter_string","float_value": 100.00}`)
+	partitionCount   = 5
+
+	// Base Avro schema
+	avroSchema = `{
+		"type":"record",
+		"name":"test",
+		"fields":[
+			{"name":"int32_value","type":"int","default":0},
+			{"name":"int64_value","type":"long","default":0},
```
ImDoubD-datazip: why is default added in every field, isn't it required only for those fields which we will be evolving?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+	}`
+
+	// Evolved Avro schema
+	updatedAvroSchema = `{
+		"type":"record",
+		"name":"test",
+		"fields":[
+			{"name":"int32_value","type":"long","default":0},
+			{"name":"int64_value","type":"long","default":0},
+			{"name":"int_value","type":"long","default":0},
```
ImDoubD-datazip: remove default fro every field, only keep in field where it is required for schema evolution.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+			{"name":"timestamp_value","type":{"type":"long","logicalType":"timestamp-micros"},"default":0},
+			{"name":"string_value","type":"string","default":""},
+			{"name":"id_int","type":"long","default":0}
+		]
+	}`
+
+	avroValue = map[string]interface{}{
+		"int32_value":     int32(132),
+		"int64_value":     int64(6400000000),
+		"int_value":       int64(101),
```
ImDoubD-datazip: this not needed.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+		"float32_value":   float32(32.5),
+		"float64_value":   float64(64.6464),
+		"float_value":     float64(66.6666),
```
ImDoubD-datazip: this not needed.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+		"boolean_true":    true,
```
ImDoubD-datazip: boolean value only single needed. change this value in update.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+		createKafkaTopic(ctx, t, broker, streams[0])
+	case "clean":
+		deleteKafkaTopic(ctx, t, broker, streams[0])
+		createKafkaTopic(ctx, t, broker, streams[0])
+	case "drop":
+		deleteKafkaTopic(ctx, t, broker, streams[0])
+	case "add":
+		for partition := 0; partition < partitionCount; partition++ {
+			writeMessagesWithRetry(ctx, t, writer, kafka.Message{Key: jsonKey, Value: jsonValue, Partition: partition})
+		}
```
ImDoubD-datazip: use utils.Foreach here.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+		writeMessagesWithRetry(ctx, t, writer, kafka.Message{Key: jsonKey, Value: filterValue1})
+		writeMessagesWithRetry(ctx, t, writer, kafka.Message{Key: jsonKey, Value: filterValue2})
```
ImDoubD-datazip: do we need 2 filter values?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+		t.Logf("Added 7 messages to topic '%s' (one per partition and two for filters)", streams[0])
+	case "update":
+		for partition := 0; partition < partitionCount; partition++ {
```
ImDoubD-datazip: use for each here. check utils.go.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+	case "drop":
+		deleteKafkaTopic(ctx, t, broker, streams[0])
+	case "add":
+		for partition := 0; partition < partitionCount; partition++ {
+			writeMessagesWithRetry(ctx, t, writer, kafka.Message{Key: jsonKey, Value: jsonValue, Partition: partition})
+		}
+		writeMessagesWithRetry(ctx, t, writer, kafka.Message{Key: jsonKey, Value: filterValue1})
+		writeMessagesWithRetry(ctx, t, writer, kafka.Message{Key: jsonKey, Value: filterValue2})
+		t.Logf("Added 7 messages to topic '%s' (one per partition and two for filters)", streams[0])
+	case "update":
```
ImDoubD-datazip, line 153: also in update, we are writing 5 new messages? not 1?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+}
+
+// ExecuteQueryForAvro executes Kafka queries for testing based on the operation type
+func ExecuteQueryForAvro(ctx context.Context, t *testing.T, streams []string, operation string, fileConfig bool) {
+	t.Helper()
+
+	var config Config
+	require.NoError(t, utils.UnmarshalFile("./testdata/avro/source.json", &config, false), "failed to unmarshal kafka test source config")
+	broker := "127.0.0.1:29192"
```
ImDoubD-datazip: put it in const block.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+	err := conn.DeleteTopics(topic)
+	require.NoError(t, err, "failed to delete topic '%s'", topic)
+	time.Sleep(5 * time.Second)
+}
+
+// createTopic creates the test topic with a fixed partition count and replication factor 1.
+func createKafkaTopic(ctx context.Context, t *testing.T, broker, topic string) {
+	t.Helper()
+	conn := dialKafkaAdminConn(ctx, t, broker)
+	defer conn.Close()
```
ImDoubD-datazip: these 2 lines, is it required in every function block? Can't I just call them in executeQuery function above?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+	for {
+		err := writer.WriteMessages(ctx, msg)
+		if err == nil {
+			return
+		}
+
+		if ctx.Err() != nil {
+			require.NoError(t, err, "failed to write seed kafka message (topic=%q partition=%d)", writer.Topic, msg.Partition)
```
ImDoubD-datazip: what does seed mean?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+func writeMessagesWithRetry(ctx context.Context, t *testing.T, writer *kafka.Writer, msg kafka.Message) {
+	t.Helper()
+
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+	for {
+		err := writer.WriteMessages(ctx, msg)
+		if err == nil {
```
ImDoubD-datazip, line 258: we are also not returning error? why?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+func registerSchemaWithRetry(t *testing.T, url, topic, schema string) uint32 {
+	body, err := json.Marshal(map[string]string{"schema": schema})
+	require.NoError(t, err)
+
+	client := &http.Client{Timeout: 10 * time.Second}
+	delay := 2 * time.Second
+
+	for i := 0; i < 5; i++ {
```
ImDoubD-datazip: please use range or for each.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+
+	require.FailNow(t, "failed to register schema")
+	return 0
+}
+
+// encodeAndWriteAvro encodes the Avro value and writes it to the Kafka topic
+func encodeAndWriteAvro(ctx context.Context, t *testing.T, writer *kafka.Writer, codec *goavro.Codec, schemaID uint32, key []byte, value map[string]interface{}) {
+	t.Helper()
+	binaryData, err := codec.BinaryFromNative(nil, value)
+	require.NoError(t, err)
```
ImDoubD-datazip: write meaningful errors.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+	msg := make([]byte, 5+len(binaryData))
+	msg[0] = 0x00
```
ImDoubD-datazip, line 319: explain in comment why added?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+				err = json.NewDecoder(resp.Body).Decode(&res)
+				resp.Body.Close()
+				require.NoError(t, err)
+				return res.ID
+			}
+			resp.Body.Close()
+		}
+
+		time.Sleep(delay)
+		delay *= 2
```
ImDoubD-datazip: can we make this code block a bit simple? i mean you are retrying right, you can use RetryOnBackoff function we have.

something like this (need to test):

```
func registerSchemaWithRetry(t *testing.T, url, topic, schema string) uint32 {
	t.Helper()

	body, err := json.Marshal(map[string]string{"schema": schema})
	require.NoError(t, err)

	client := &http.Client{Timeout: 10 * time.Second}
	var schemaID uint32

	err = utils.RetryOnBackoff(context.Background(), 5, 2*time.Second, func(_ context.Context) error {
		resp, err := client.Post(
			fmt.Sprintf("%s/subjects/%s-value/versions", url, topic),
			"application/vnd.schemaregistry.v1+json",
			bytes.NewReader(body),
		)
		if err != nil {
			return err
		}
		defer resp.Body.Close()

		if resp.StatusCode != http.StatusOK {
			return fmt.Errorf("unexpected status: %d", resp.StatusCode)
		}

		var res struct {
			ID uint32 `json:"id"`
		}
		if err := json.NewDecoder(resp.Body).Decode(&res); err != nil {
			return err
		}

		schemaID = res.ID
		return nil
	})

	require.NoError(t, err, "failed to register schema")
	return schemaID
}
```

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,372 @@
...
+	"boolean_false":   "boolean",
+	"timestamp_value": "timestamp",
+	"string_value":    "string",
+}
+
+var EvolvedKafkaToDestinationJSONSchema = map[string]string{
+	"int_value":       "bigint",
+	"float_value":     "double",
+	"boolean_true":    "boolean",
+	"boolean_false":   "boolean",
```
ImDoubD-datazip: remove this double boolean from every where.

utils/testutils/test_utils.go
```diff
@@ -779,6 +819,9 @@ func (cfg *IntegrationTest) TestIntegration(t *testing.T) {
 	t.Logf("Root Project directory: %s", cfg.TestConfig.HostRootPath)
 	t.Logf("Test data directory: %s", cfg.TestConfig.HostTestDataPath)
 	currentTestTable := fmt.Sprintf("%s_test_table_olake", cfg.TestConfig.Driver)
+	if cfg.TestConfig.DataFormat != "" {
+		currentTestTable = fmt.Sprintf("%s_%s_test_table_olake", cfg.TestConfig.Driver, cfg.TestConfig.DataFormat)
+	}
```
ImDoubD-datazip: use ternary

utils/testutils/test_utils.go
```diff
@@ -561,12 +582,31 @@ func (cfg *IntegrationTest) testParquetFullLoadAndCDC(
...
+		{
+			name:      "CDC - update",
+			operation: "update",
+			useState:  true,
+			opSymbol:  "c",
+			expected:  cfg.ExpectedUpdatedData,
+		},
+	}
+
+	testCases = utils.Ternary(slices.Contains(constants.OnlyStrictCDCDriver, constants.DriverType(cfg.TestConfig.Driver)), kafkaTestCases, testCases).([]syncTestCase)
```
ImDoubD-datazip: direct write in ternary, if driver is kafka, then use kafkatestcase.

utils/testutils/test_utils.go
ImDoubD-datazip: rename this to dbTestCases

utils/testutils/test_utils.go
```diff
@@ -922,20 +965,22 @@ func (cfg *IntegrationTest) TestIntegration(t *testing.T) {
 								})
 							}
 
-							for _, wt := range writerTypes {
-								t.Run(fmt.Sprintf("Iceberg (%s) Full load + Incremental tests", wt.name), func(t *testing.T) {
-									if err := cfg.testIcebergWriter(ctx, t, c, currentTestTable, wt.useArrow, cfg.testIcebergFullLoadAndIncremental); err != nil {
-										t.Fatalf("Iceberg (%s) Full load + Incremental tests failed: %v", wt.name, err)
+							if !slices.Contains(constants.OnlyStrictCDCDriver, constants.DriverType(cfg.TestConfig.Driver)) {
```
ImDoubD-datazip: write a comment above for this check, its purpose.

utils/testutils/test_utils.go
```diff
@@ -1005,6 +1050,11 @@ func VerifyIcebergSync(t *testing.T, tableName, icebergDB string, datatypeSchema
 		"SELECT * FROM %s WHERE _op_type = '%s'",
 		fullTableName, opSymbol,
 	)
+	if slices.Contains(constants.OnlyStrictCDCDriver, constants.DriverType(driver)) {
+		if _, ok := schema["id_int"]; ok {
+			selectQuery += " AND id_int IS NOT NULL"
+		}
+	}
```
ImDoubD-datazip: why is this required?

utils/typeutils/compare.go
```diff
@@ -65,6 +66,15 @@ func Compare(a, b any) int {
...
 		return 0
+	case json.Number:
+		// Try int64 first, then fall back to float64.
+		if aInt, err := aVal.Int64(); err == nil {
+			return Compare(aInt, b)
+		}
+		if aFloat, err := aVal.Float64(); err == nil {
+			return Compare(aFloat, b)
+		}
+		return strings.Compare(aVal.String(), fmt.Sprintf("%v", b))
```
ImDoubD-datazip, line 81: for this we have add test cases in compare unit test file as well.

.github/workflows/integration-tests.yml
saksham-datazip: Because we need two kafka sources for running avro and json in parallel to avoid race conditions :-
Issues-> 
1->We need two expectedDataTypes one for json and one for avro 
2->Thats, why we will need to run them saperately now to run those tests in parallel,we can have same streams.json but it will cause race condition as one will override others streams.json while running discover command for them saperately.
3->so we kept a saperate folder for each of them.
4->Now the problem came that if we keep source same ,one kafka topic containing json and other avro ,but still each one will try to sync others topic data and to avoid that we will be needing to use jq command so that we can remove topics from selected streams.
5->and even after jq command ,during discover their might be a possibility that create runs for one datatype not for other thus failing the discover command,so we need to saperate the sources as well.

drivers/kafka/docker-compose.yml
saksham-datazip: To avoid race condition we need two clusters

utils/testutils/test_utils.go
saksham-datazip: this querry is for drivers like mysql postgres where we have to define schema and evolve it ,so just like mongodb we can skip it here

utils/testutils/test_utils.go
saksham-datazip: this querry is for drivers like mysql postgres where we have to define schema and evolve it ,so just like mongodb we can skip it here

.github/workflows/integration-tests.yml
saksham-datazip: Changed

utils/testutils/test_utils.go
saksham-datazip: so in other drivers we run querry on basis of op_type but in kafka we have just one op_type 'c' so we check using id_int if schem is evolved or not

drivers/kafka/internal/testdata/json/iceberg_destination.json
saksham-datazip: It would require multiple changes in testutils/GetTestConfig function which would be driver specific and it would also not be consistent with other drivers as in other driver destination is along with config.json

drivers/kafka/internal/testdata/json/parquet_destination.json
saksham-datazip: It would require multiple changes in testutils/GetTestConfig function which would be driver specific and it would also not be consistent with other drivers as in other driver destination is along with config.json

drivers/kafka/internal/kafka_test.go
saksham-datazip: in all other drivers we used PartitionRegex to be identity so why do we need to change it just for kafka ?

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: This is not possible in go syntax as const can only store compile time values while map, byteSlice are runtime values which will cause syntax error issues

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Removed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Clean means removing all the data present in topics, and in other drivers, we are using truncate equivalent functions but kafka doesn't have its equivalent so we will have to delete topic first and than recreate the topic again

utils/testutils/test_utils.go
saksham-datazip: Changed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Changed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Changed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Changed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: removed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: removed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: removed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: removed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip:
```
drivers/s3/internal/edge_cases_test.go:L534

/Users/apple/Desktop/olake/drivers/s3/internal/edge_cases_test.go:L534,L543
```
in this file alos we are using simple loop and in some other parts of code as well, so why to change from simple loop and also foreach needs first parameter as slice so we will have to make a slice with number 0,1,2,3,4? but changed to using range

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: in this file alos we are using simple loop and in some other parts of code as well, so why to change from simple and also foreach needs first parameter as slice so we will have to make a slice with number 0,1,2,3,4? but changed to using range
```
drivers/s3/internal/edge_cases_test.go:L534
/Users/apple/Desktop/olake/drivers/s3/internal/edge_cases_test.go:L534,L543
```

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: removed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: so changed to adding one message in case of update

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Yes, we can put it above but here only in two cases for delete and update topic we need this connection but if we put it in executeQuery function it will be called everytime even when this is not necessary a connection would be setup which is redundant.

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Changed to this log 
```
require.NoError(t, err, "timed out writing kafka message after retries (topic=%q partition=%d): %v",writer.Topic, msg.Partition, err)

File: utils/testutils/test_utils.go (hunk 32 above)
Comment (saksham-datazip) [reply utils/testutils/test_utils.go]: added

File: drivers/kafka/internal/kafka_test_utils.go (hunk 6 above)
Comment (saksham-datazip) [reply drivers/kafka/internal/kafka_test_utils.go]: added

File: drivers/kafka/internal/kafka_test_utils.go (hunk 26 above)
Comment (saksham-datazip) [reply drivers/kafka/internal/kafka_test_utils.go]: added this 
```
	require.NoError(t, err, "encode Avro value to binary (topic=%q, schema_id=%d)", writer.Topic, schemaID)

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: added

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: changed

utils/testutils/test_utils.go
saksham-datazip: changed

utils/testutils/test_utils.go
saksham-datazip: changed

utils/testutils/test_utils.go
saksham-datazip: changed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: because in github logs, log related to it was for failing on an attempt was not needed and it was clustered thats why but if after timeout write message fails it returns with a log.
For more reference check this integration run :-
https://github.com/datazip-inc/olake/actions/runs/24128894250/job/70400262796#logs

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: changed this part of code as per the changes in function u provided

.github/workflows/integration-tests.yml
ImDoubD-datazip: there is no race condition, only in order to keep the testutils file less complex we are doing this.

drivers/kafka/internal/testdata/json/iceberg_destination.json
ImDoubD-datazip: it would only require to change the path, thats all. but for now we can skip it.

drivers/kafka/internal/testdata/json/parquet_destination.json
ImDoubD-datazip: it would only require to change the path, thats all. but for now we can skip it. we need to make this thing perfect in future

drivers/kafka/docker-compose.yml
ImDoubD-datazip: there is no race condition, we are just trying to keep things simple in test utils we are doing this. we can use one cluster and still run the sync, but we have to create 2 topics for avro and json separate and do changes in via jq command to make things work or either add extra code in test utils. we are just avoiding that.

drivers/kafka/internal/kafka_test.go
ImDoubD-datazip: because we need to test other cases as well, many users do partition on the basis of timestamp.

drivers/kafka/internal/kafka_test_utils.go
ImDoubD-datazip: so create a separate const block and keep the constant in that block.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+			{"name":"excludedColumn","type":"int"}
+		]
+	}`
+
+	// Evolved Avro schema
+	updatedAvroSchema = `{
+		"type":"record",
+		"name":"test",
+		"fields":[
+			{"name":"int32_value","type":"long"},
```
ImDoubD-datazip: how is the schema evolving here, because in for int32_value you are making the type to be long from int but in avro schema evolution does not take place with default right? then how is the schema getting evolve here?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+		t.Fatalf("unsupported operation: %s", operation)
+	}
+}
+
+// ExecuteQueryForAvro executes Kafka queries for testing based on the operation type
+func ExecuteQueryForAvro(ctx context.Context, t *testing.T, streams []string, operation string, fileConfig bool) {
+	t.Helper()
+
+	var config Config
+	require.NoError(t, utils.UnmarshalFile("./testdata/avro/source.json", &config, false), "failed to unmarshal kafka test source config")
```
ImDoubD-datazip: we have done this in case of avro, but not json why?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+		deleteKafkaTopic(ctx, t, kafkaAvroBroker, streams[0])
+		createKafkaTopic(ctx, t, kafkaAvroBroker, streams[0])
+	case "drop":
+		deleteKafkaTopic(ctx, t, kafkaAvroBroker, streams[0])
+	case "add":
+		codec, err := goavro.NewCodec(avroSchema)
+		require.NoError(t, err)
+		schemaID := registerSchemaWithRetry(t, registryURL, streams[0], avroSchema)
+		encodeAndWriteAvro(ctx, t, writer, codec, schemaID, avroKey, avroValue)
+		encodeAndWriteAvro(ctx, t, writer, codec, schemaID, avroKey, avroFilterValue)
```
ImDoubD-datazip: here can you explain how you are writing 5 avro messages here?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+		require.NoError(t, err, "failed to create topic '%s' explicitly", topic)
+	}
+}
+
+func dialKafkaAdminConn(ctx context.Context, t *testing.T, broker string) *kafka.Conn {
+	t.Helper()
+	conn, err := kafka.DialContext(ctx, "tcp", broker)
+	require.NoError(t, err, "failed to dial kafka broker")
+	_, err = conn.ReadPartitions()
+	require.NoError(t, err, "failed to read kafka partitions metadata")
```
ImDoubD-datazip: what is the need of these lines?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+	avroFilterValue = map[string]interface{}{
+		"int32_value":     int32(132),
+		"int64_value":     int64(6400000000),
+		"float32_value":   float32(32.5),
+		"float64_value":   float64(64.6464),
+		"boolean":         true,
+		"timestamp_value": int64(time.Date(2026, 3, 22, 14, 30, 0, 0, time.UTC).UnixNano() / int64(time.Microsecond)),
+		"string_value":    "",
+		"excludedColumn":  int32(101),
```
ImDoubD-datazip: isse col_excluded krde.

drivers/kafka/internal/kafka_test.go
```diff
@@ -0,0 +1,86 @@
...
+				Namespace:                        "topics",
+				ExpectedData:                     ExpectedKafkaJSONData,
+				ExpectedUpdatedData:              ExpectedKafkaUpdatedJSONData,
+				DestinationDataTypeSchema:        KafkaToDestinationJSONSchema,
+				UpdatedDestinationDataTypeSchema: EvolvedKafkaToDestinationJSONSchema,
+				DefaultCDCColumnsSchema:          ExpectedKafkaDefaultCDCColumnsSchema,
+				ExecuteQuery:                     ExecuteQueryForJson,
+				DestinationDB:                    "kafka_topics",
+				PartitionRegex:                   "/{int_value,identity}",
+				ColumnToExclude:                  "excludedColumn",
```
ImDoubD-datazip: here make it col_excluded

drivers/kafka/internal/kafka_test_utils.go
ImDoubD-datazip: ok

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+		}
+		if ctx.Err() != nil {
+			require.NoError(t, err, "timed out writing kafka message after retries (topic=%q partition=%d): %v", writer.Topic, msg.Partition, err)
+			return
+		}
+		time.Sleep(200 * time.Millisecond)
+	}
+}
+
+func registerSchemaWithRetry(t *testing.T, url, topic, schema string) uint32 {
```
ImDoubD-datazip, line 270: add comment for this function.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+func dialKafkaAdminConn(ctx context.Context, t *testing.T, broker string) *kafka.Conn {
+	t.Helper()
+	conn, err := kafka.DialContext(ctx, "tcp", broker)
+	require.NoError(t, err, "failed to dial kafka broker")
+	_, err = conn.ReadPartitions()
+	require.NoError(t, err, "failed to read kafka partitions metadata")
+	return conn
+}
+
+func writeMessagesWithRetry(ctx context.Context, t *testing.T, writer *kafka.Writer, msg kafka.Message) {
```
ImDoubD-datazip, line 250: add comment for this function

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+	t.Helper()
+	conn := dialKafkaAdminConn(ctx, t, broker)
+	defer conn.Close()
+	err := conn.CreateTopics(kafka.TopicConfig{Topic: topic, NumPartitions: partitionCount, ReplicationFactor: 1})
+	if err != nil && err != kafka.TopicAlreadyExists {
+		require.NoError(t, err, "failed to create topic '%s' explicitly", topic)
+	}
+}
+
+func dialKafkaAdminConn(ctx context.Context, t *testing.T, broker string) *kafka.Conn {
```
ImDoubD-datazip: add comment for this function.

utils/testutils/test_utils.go
```diff
@@ -1007,6 +1051,12 @@ func VerifyIcebergSync(t *testing.T, tableName, icebergDB string, datatypeSchema
 		"SELECT * FROM %s WHERE _op_type = '%s'",
 		fullTableName, opSymbol,
 	)
+	// For strict CDC drivers, filter by the new column since _op_type is always 'c'
+	if slices.Contains(constants.OnlyStrictCDCDriver, constants.DriverType(driver)) {
```
ImDoubD-datazip: remove this OnlyStrictCDCDriver, because in every driver you can do strict cdc. in kafka, you can only do strict cdc. directly add check on kafka. if any driver in future comes similar to kafka, then we can make a slice for them and update the check. for now check as per kafka.

utils/testutils/test_utils.go
ImDoubD-datazip: ok

utils/testutils/test_utils.go
```diff
@@ -1007,6 +1051,12 @@ func VerifyIcebergSync(t *testing.T, tableName, icebergDB string, datatypeSchema
 		"SELECT * FROM %s WHERE _op_type = '%s'",
 		fullTableName, opSymbol,
 	)
+	// For strict CDC drivers, filter by the new column since _op_type is always 'c'
```
ImDoubD-datazip: in comment you have written what is happening, also write the problem due to it and what you are doing to solve this case.

.github/workflows/integration-tests.yml
```diff
@@ -100,6 +101,16 @@ jobs:
...
 
+      - name: Wait for Kafka
+        uses: nick-fields/retry@v2
+        with:
+          timeout_minutes: 5
+          max_attempts: 30
+          retry_wait_seconds: 5
+          command: |
+            docker exec kafkaJson kafka-topics --bootstrap-server localhost:9092 --list
+            docker exec kafkaAvro kafka-topics --bootstrap-server localhost:9092 --list
```
ImDoubD-datazip, line 112: for avro, also add check if schema registry is online.

drivers/kafka/internal/kafka_test_utils.go
ImDoubD-datazip: range is fine. i was asking to check whether for each is applicable here, if yes then use it. but range is also fine.

drivers/kafka/internal/kafka_test_utils.go
ImDoubD-datazip: in delete and create topic function it is ok, but we can directly call the service in these function, without requiring extra dialKafkaAdminConn function?
just do:
```
	conn, err := kafka.DialContext(ctx, "tcp", broker)
	require.NoError(t, err, "failed to dial kafka broker")
```

drivers/kafka/internal/kafka_test_utils.go
ImDoubD-datazip: i assume you have tested this retry thing after your changes.

drivers/kafka/internal/kafka_test_utils.go
ImDoubD-datazip: range is fine.

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: changed

utils/typeutils/compare.go
saksham-datazip: Added cases in unit test file

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Because json dont need schema registry and config is needed for schema registry

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Changed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Removed

utils/testutils/test_utils.go
saksham-datazip: Changed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Added

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Added

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Function Removed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Changed

drivers/kafka/internal/kafka_test.go
saksham-datazip: Changed

.github/workflows/integration-tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
   integration-tests:
     environment: integration_tests
     runs-on: 32gb-runner
-    timeout-minutes: 45
+    timeout-minutes: 50
```
vaibhav-datazip: what's the need of this change. ?

utils/typeutils/compare_test.go
```diff
@@ -100,6 +102,18 @@ func TestCompare(t *testing.T) {
 		{"numeric_string_lex_order_2", "2", "100", 1},
 		{"unicode_comparison_less", "α", "β", -1},
 
+		// json.Number
+		{"json_number_int_equal", json.Number("123"), int64(123), 0},
+		{"json_number_float_equal", json.Number("123.45"), float64(123.45), 0},
+		{"json_number_int_less", json.Number("100"), int64(200), -1},
```
vaibhav-datazip, line 108: add some test cases like where both values are json number 

- 100 100.0
- 281  354.5
- 96 53.2

utils/testutils/test_utils.go
```diff
@@ -1081,7 +1132,7 @@ func VerifyIcebergSync(t *testing.T, tableName, icebergDB string, datatypeSchema
 			for key := range defaultCDCColumnsSchema {
 				icebergValue, ok := icebergMap[key]
 				require.Truef(t, ok, "Row %d: missing column %q in Iceberg result", rowIdx, key)
-				require.NotEmpty(t, icebergValue, "Row %d: expected column %q to be non-empty, got %#v", rowIdx, key, icebergValue)
+				require.NotNil(t, icebergValue, "Row %d: expected column %q to be non-empty, got %#v", rowIdx, key, icebergValue)
```
vaibhav-datazip: this weakens the assertion, as it will pass even well strings are empty

utils/testutils/test_utils.go
```diff
@@ -172,7 +173,7 @@ func (s *benchmarkStore) stats(
 }
 
 // GetTestConfig returns the test config for the given driver
-func GetTestConfig(driver string) *TestConfig {
+func GetTestConfig(driver string, dataFormat string) *TestConfig {
```
vaibhav-datazip: use a variadic parameter here instead, this will remove the need of passing empty string from other drivers than kafka

```
func GetTestConfig(driver string, dataFormat ...string) *TestConfig {
 format := ""
    if len(dataFormat) > 0 {
        format = dataFormat[0]
    }
```

destination/iceberg/arrow-writer/transforms.go
```diff
@@ -64,25 +64,44 @@ func hashString(s string) uint32 {
...
 		return fmt.Sprintf("%g", v), v, nil
 	case "double":
-		v := val.(float64)
+		v, err := typeutils.ReformatFloat64(val)
+		if err != nil {
+			return "", nil, err
+		}
 		return fmt.Sprintf("%g", v), v, nil
 	case "string":
 		s := val.(string)
```
vaibhav-datazip: shouldn't we throw an error here as well if parsing fails , like we are doing in other column datatypes.

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,389 @@
...
+		"boolean":         true,
+		"timestamp_value": int64(time.Date(2026, 3, 22, 14, 30, 0, 0, time.UTC).UnixNano() / int64(time.Microsecond)),
+		"string_value":    "test_string",
+		"col_excluded":    int32(101),
+		"col_included":    int32(102),
+	}
+)
+
+// ExecuteQueryJSON executes Kafka queries for testing based on the operation type
+func ExecuteQueryJSON(ctx context.Context, t *testing.T, streams []string, operation string, fileConfig bool) {
```
vaibhav-datazip, line 113: file config is not used remove it from Avro one as well

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -0,0 +1,389 @@
...
+		"float32_value":   float32(32.5),
+		"float64_value":   float64(64.6464),
+		"boolean":         true,
+		"timestamp_value": int64(time.Date(2026, 3, 22, 14, 30, 0, 0, time.UTC).UnixNano() / int64(time.Microsecond)),
+		"string_value":    "",
+		"col_excluded":    int32(101),
+	}
+
+	avroUpdatedValue = map[string]interface{}{
+		"int32_value":     int64(132),
```
vaibhav-datazip: don't change existing field instead introduce a new field for updating

.github/workflows/integration-tests.yml
saksham-datazip: Changed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Changed

utils/testutils/test_utils.go
saksham-datazip: Changed

utils/typeutils/compare_test.go
saksham-datazip: Solved

utils/testutils/test_utils.go
saksham-datazip: Resolved

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: it is not of any use here but i followed a pattern which other drivers are following as well, so added this block for fileconfig

destination/iceberg/arrow-writer/transforms.go
saksham-datazip: Added a error handling case

ImDoubD-datazip (approved):
LGTM

vaibhav-datazip (approved):

PR #906, opened 2026-04-04
fix: handle multi-level partitioned tables in backfill chunking

vikaxsh:
# Description

Fixes silent backfill skip on multi-level partitioned PostgreSQL tables. The partition page
discovery query only looked one level deep via `pg_inherits` — intermediate partitions
return `pg_relation_size = 0`, causing the chunk loop to never execute and the table to be
skipped entirely with no error.

- PG 12+: uses `pg_partition_tree(...) WHERE isleaf = true`
- PG < 12: uses a recursive CTE over `pg_inherits` with leaf filter
- Stale `pg_class` stats (ANALYZE not run) now return a clear error instead of silent skip

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Normal (non-partitioned) table — full refresh, verified row count matches
- [x] 2-level partitioned table — full refresh, verified row count matches
- [x] 3-level partitioned table — previously skipped, now syncs correctly
- [x] 4-level partitioned table — previously skipped, now syncs correctly

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/.govulnignore
+++ b/.govulnignore
@@ -1,0 +1,10 @@
+# GO-2026-4887 / CVE-2026-34040: Moby AuthZ plugin bypass via oversized request body.
+# GO-2026-4883 / CVE-2026-33997: Moby off-by-one error in plugin privilege validation.
+# Root cause: github.com/docker/docker has no fixed version (fixes are only in
+# github.com/moby/moby/v2 >= v2.0.0-beta.8, a different module path).
+# Affected code path: testcontainers-go (test-only dep) -> docker/docker client.
+# Real-world risk: Both CVEs target Docker daemon internals (AuthZ plugins, plugin install).
+# Olake only uses the Docker client library; daemon-side code is never invoked.
+# TODO: Remove after testcontainers-go v0.42.0 ships (PR #3591 merged 2026-04-03).
+GO-2026-4887
+GO-2026-4883
--- a/drivers/postgres/internal/backfill.go
+++ b/drivers/postgres/internal/backfill.go
@@ -219,16 +219,32 @@ type PartitionPage struct {
 	Pages uint32
 }
 
-// loadPartitionPages fetches partition-level relpages using PostgresPartitionPages query.
+// postgresMinVersionPG12 is the server_version_num for PostgreSQL 12.0.
+// pg_partition_tree() (PG 12+) is preferred; older versions use a recursive CTE.
+const postgresMinVersionPG12 = 120000
+
 func loadPartitionPages(ctx context.Context, db *sql.DB, stream types.StreamInterface) ([]PartitionPage, uint32, error) {
-	query := jdbc.PostgresPartitionPages(stream)
+	var serverVersionNum int
+	if err := db.QueryRowContext(ctx, jdbc.PostgresServerVersionNum()).Scan(&serverVersionNum); err != nil {
+		return nil, 0, fmt.Errorf("failed to detect postgres server version: %s", err)
+	}
+
+	var query string
+	if serverVersionNum >= postgresMinVersionPG12 {
+		logger.Debugf("using pg_partition_tree for partition page discovery (server_version_num=%d)", serverVersionNum)
+		query = jdbc.PostgresPartitionPagesPG12(stream)
+	} else {
+		logger.Debugf("using recursive CTE for partition page discovery (server_version_num=%d)", serverVersionNum)
+		query = jdbc.PostgresPartitionPages(stream)
+	}
+
 	rows, err := db.QueryContext(ctx, query)
 	if err != nil {
 		return nil, 0, fmt.Errorf("failed to load partition pages: %s", err)
 	}
-	var maxPageCountAcrossPartitions uint32
 	defer rows.Close()
 
+	var maxPageCountAcrossPartitions uint32
 	var partitions []PartitionPage
 	for rows.Next() {
 		var p PartitionPage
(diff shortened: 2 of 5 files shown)


drivers/postgres/internal/backfill.go
```diff
@@ -238,10 +254,28 @@ func loadPartitionPages(ctx context.Context, db *sql.DB, stream types.StreamInte
...
+		return partitions, maxPageCountAcrossPartitions, nil
 	}
+
+	if maxPageCountAcrossPartitions == 0 {
+		var hasRows bool
+		if err := db.QueryRowContext(ctx, jdbc.PostgresTableExistsQuery(stream)).Scan(&hasRows); err != nil {
+			return nil, 0, fmt.Errorf("failed to check if table has rows: %s", err)
+		}
+		if hasRows {
+			return nil, 0, fmt.Errorf("stats not populated for table[%s]. Please run ANALYZE on the table and its partitions to update statistics", stream.ID())
```
vaibhav-datazip: if page count has 0 partitions , and rows are still there, it should've had partitions I don't think its dependent on stats table. For this case we should fallback to other strategy

drivers/postgres/internal/backfill.go
```diff
@@ -238,10 +254,28 @@ func loadPartitionPages(ctx context.Context, db *sql.DB, stream types.StreamInte
...
 		partitions = append(partitions, p)
 	}
+	if err := rows.Err(); err != nil {
+		return nil, 0, fmt.Errorf("failed to iterate partition pages: %s", err)
+	}
 
 	if len(partitions) == 0 {
 		partitions = append(partitions, PartitionPage{Name: stream.Name(), Pages: 1})
+		maxPageCountAcrossPartitions = 1
+		return partitions, maxPageCountAcrossPartitions, nil
```
vaibhav-datazip: you can directly pass 1

drivers/postgres/internal/backfill.go
```diff
@@ -238,10 +254,28 @@ func loadPartitionPages(ctx context.Context, db *sql.DB, stream types.StreamInte
...
 	}
```
vaibhav-datazip: do we need partition len check, wont the maxPageCountAcrossPartitions be able to do the same thing, as for partition len 0 we will have  `maxPageCountAcrossPartitions` also as 0

drivers/postgres/internal/backfill.go
vaibhav-datazip, line 251: In the queries we are returning the pages in bigint in that case we should store the variable in a int64 instead

pkg/jdbc/jdbc.go
```diff
@@ -134,45 +134,62 @@ func PostgresBlockSizeQuery() string {
...
+    `,
+		stream.Namespace(),
+		stream.Name(),
+	)
+}
+
+// PostgresPartitionPages returns leaf-partition page counts using a recursive CTE over
+// pg_inherits. Works on all Postgres versions (10+); used as fallback for PG < 12.
 func PostgresPartitionPages(stream types.StreamInterface) string {
 	return fmt.Sprintf(`
```
vaibhav-datazip, line 162: please check this query might be more optimized 
```
WITH RECURSIVE
  blksz(b) AS (
    SELECT current_setting('block_size')::int
  ),
  partition_tree AS (
    SELECT c.oid, c.relname AS name
    FROM   pg_class c
    JOIN   pg_namespace n ON n.oid = c.relnamespace
    WHERE  n.nspname = '%s'
      AND  c.relname = '%s'

    UNION ALL

    SELECT child.oid, child.relname
    FROM   pg_inherits i
    JOIN   pg_class child  ON child.oid = i.inhrelid
    JOIN   partition_tree pt ON pt.oid  = i.inhparent
  )
SELECT
  pt.name,
  CEIL(1.05 * pg_relation_size(pt.oid) / b.b::numeric)::bigint AS pages
FROM   partition_tree pt
JOIN   blksz b ON true
-- LEFT JOIN replaces the correlated NOT EXISTS — much cheaper
LEFT   JOIN pg_inherits i ON i.inhparent = pt.oid
WHERE  i.inhparent IS NULL         
ORDER  BY pages DESC;
```

pkg/jdbc/jdbc.go
```diff
@@ -134,45 +134,62 @@ func PostgresBlockSizeQuery() string {
...
+	return `SELECT current_setting('server_version_num')::int`
+}
+
+// PostgresPartitionPagesPG12 returns leaf-partition page counts using pg_partition_tree (PG 12+).
+// Intermediate partitions are excluded via isleaf=true; works for any partition depth.
+func PostgresPartitionPagesPG12(stream types.StreamInterface) string {
+	return fmt.Sprintf(`
+        SELECT
+            pt.relid::text AS name,
+            CEIL(1.05 * (pg_relation_size(pt.relid::oid) / current_setting('block_size')::int))::bigint AS pages
```
vaibhav-datazip, line 148: redid is already mid, but here `(pg_relation_size(pt.relid::oid) / current_setting('block_size')::int)` we are doing int/int , it will be better to typecast to  `block_size` to float

vikaxsh: image

pkg/jdbc/jdbc.go
vikaxsh: added an todo

.govulnignore
```diff
@@ -0,0 +1,10 @@
+# GO-2026-4887 / CVE-2026-34040: Moby AuthZ plugin bypass via oversized request body.
+# GO-2026-4883 / CVE-2026-33997: Moby off-by-one error in plugin privilege validation.
+# Root cause: github.com/docker/docker has no fixed version (fixes are only in
+# github.com/moby/moby/v2 >= v2.0.0-beta.8, a different module path).
+# Affected code path: testcontainers-go (test-only dep) -> docker/docker client.
+# Real-world risk: Both CVEs target Docker daemon internals (AuthZ plugins, plugin install).
+# Olake only uses the Docker client library; daemon-side code is never invoked.
+# TODO: Remove after testcontainers-go v0.42.0 ships (PR #3591 merged 2026-04-03).
```
vaibhav-datazip, line 8: Remove duplicate todo, this will create two issues

vikaxsh: govulncheck and trivy are different

vaibhav-datazip (approved):

PR #907, opened 2026-04-06
fix: normalize incremental cursor values in drivers before persisting state

vishalm0509:
# Description

The shared JDBC helper no longer converts scanned cursor values (e.g. []byte → string). It returns raw values; each source normalizes before writing state.

**Postgres**, **MySQL**, **DB2**: normalize by turning []byte cursors into strings.
**MSSQL**: normalize using column type (existing normalizeMSSQLValueForState) so incremental state matches what the backfill path uses.

**Why**
Central conversion was wrong or insufficient for some drivers/types; normalization must be driver- and type-aware (especially MSSQL).

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):


pkg/jdbc/jdbc.go
```diff
@@ -1338,15 +1338,6 @@ func GetMaxCursorValues(ctx context.Context, client *sqlx.DB, driverType constan
 
 	var maxPrimaryCursorValue, maxSecondaryCursorValue any
 
-	bytesConverter := func(value any) any {
```
vikaxsh, line 1349: For now, instead of removing this entirely, let’s add an MSSQL-specific check; otherwise, we’d need to test all other drivers.

drivers/mssql/internal/backfill.go
```diff
@@ -300,6 +300,11 @@ func normalizeBoundaryValue(value any, pkCols []string, columnType string) strin
 		return utils.ConvertToString(value)
 	}
 
+	value = normalizeMSSQLValueForState(value, columnType)
+	return utils.ConvertToString(value)
+}
+
+func normalizeMSSQLValueForState(value any, columnType string) any {
```
vikaxsh:
```suggestion
func normalizeMSSQLValue(value any, columnType string) any {
```

drivers/mssql/internal/incremental.go
```diff
@@ -49,5 +49,31 @@ func (m *MSSQL) FetchMaxCursorValues(ctx context.Context, stream types.StreamInt
...
+		}
+	}
+
 	return maxPrimaryCursorValue, maxSecondaryCursorValue, nil
-}
\ No newline at end of file
+}
+
+func (m *MSSQL) normalizeCursorValue(ctx context.Context, stream types.StreamInterface, cursorField string, value any) (any, error) {
+	if cursorField == "" || value == nil {
```
vikaxsh: do we need this check as cursorField is never going to be empty

drivers/mssql/internal/backfill.go
vishalm0509: Done

drivers/mssql/internal/incremental.go
vishalm0509: No need to check cursorField

pkg/jdbc/jdbc.go
vishalm0509: It will not break anything. As there is no change in the code flow. I am doing the same thing now but in a different place

vishalm0509: I have also verified with 3 models

vikaxsh: IMO, for now we should implement for mssql only

vishalm0509: Okay. Done

drivers/mssql/internal/incremental.go
```diff
@@ -49,5 +49,31 @@ func (m *MSSQL) FetchMaxCursorValues(ctx context.Context, stream types.StreamInt
...
+}
+
+func (m *MSSQL) normalizeCursorValue(ctx context.Context, stream types.StreamInterface, cursorField string, value any) (any, error) {
+	if value == nil {
+		return value, nil
+	}
+
+	var dataType string
+	if err := m.client.QueryRowContext(ctx, jdbc.MSSQLColumnTypeQuery(), stream.Namespace(), stream.Name(), cursorField).Scan(&dataType); err != nil {
+		return nil, fmt.Errorf("failed to get MSSQL cursor type: %s", err)
```
vikaxsh, line 75: we already have a function in backfill for same
 getColumnTypeMSSQL

nayanj98:
we are working on adding support to Iceberg Binary datatype (hence this PR's can be closed as new PR will be raised)

PR #914, opened 2026-04-09
fix(oracle): optimize splitViaTableIteration to use single boundary query

### LINKED ISSUE #872: Make this function more efficient by using a single query to get all the rowids in the table using one query
Currently this function is called every time to fetch the boundary, instead of this a change is expected here that can fetch all the rowids at once, this might significantly improve the query performance and reduce repeated db calls.

https://github.com/datazip-inc/olake/blob/880451956d0d13b332c4d5fc3bc7066365c64efd/drivers/oracle/internal/backfill.go#L90-L95

---

###### This issue was generated by [todo-issue](https://github.com/DerJuulsn/todo-issue) based on a `TODO` comment in 880451956d0d13b332c4d5fc3bc7066365c64efd. It&#x27;s been assigned to @vaibhav-datazip because they committed the code.

mvanhorn:
# Description

The `splitViaTableIteration` function made N+1 database round trips to compute chunk boundaries. For large tables, this meant hundreds of sequential queries.

This replaces the loop with a single NTILE-based query that fetches all chunk boundary ROWIDs at once. Falls back to the original loop when table stats are unavailable.

Also extracts `buildChunksFromStartRowIDs` as a shared helper with `splitViaRowId` to remove duplicate chunk construction logic.

Fixes #872

## Type of change

# How Has This Been Tested?

- [x] Verified Go compilation passes for both `drivers/oracle/...` and `pkg/jdbc/...`
- [x] Reviewed NTILE query produces correct bucket boundaries matching the original loop output
- [x] Fallback path preserves original behavior when stats are unavailable

# Screenshots or Recordings
N/A - backend optimization

## Documentation

---

This contribution was developed with AI assistance (Codex).


nayanj98:
@mvanhorn Please raise your PR on top of staging branch instead of master. Also you will have to sign the CLA first. Also can you please comment on the issue so that I can assign the issue to you.

mvanhorn:
Retargeted to staging. Commented on #872 for assignment. Will sign the CLA now.

drivers/oracle/internal/backfill.go
```diff
@@ -87,8 +87,58 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
 	return chunks, nil
 }
 
-// TODO: Make this function more efficient by using a single query to get all the rowids in the table using one query
 func (o *Oracle) splitViaTableIteration(ctx context.Context, stream types.StreamInterface, avgRowSize float64) (*types.Set[types.Chunk], error) {
+	rowsPerChunk := int64(math.Ceil(float64(constants.EffectiveParquetSize) / avgRowSize))
+	if rowsPerChunk <= 0 {
+		rowsPerChunk = 1
+	}
```
vaibhav-datazip: utils.Ternary can be used here.

drivers/oracle/internal/backfill.go
```diff
@@ -87,8 +87,58 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+
+	var numRows int64
+	err := o.client.QueryRowContext(ctx, jdbc.OracleTableRowStatsQuery(), stream.Namespace(), stream.Name()).Scan(&numRows, new(float64))
```
vaibhav-datazip: Instead of fetching NumRows here again, you can pass down `approxRowCount` to this function and add checks for that

drivers/oracle/internal/backfill.go
```diff
@@ -87,8 +87,58 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+		} else {
+			logger.Debugf("Oracle table stats returned num_rows=%d for [%s.%s], falling back to row-by-row chunk iteration", numRows, stream.Namespace(), stream.Name())
+		}
+		return o.splitViaTableIterationLoop(ctx, stream, rowsPerChunk)
+	}
+
+	numChunks := int64(math.Ceil(float64(numRows) / float64(rowsPerChunk)))
+	if numChunks <= 0 {
+		numChunks = 1
+	}
```
vaibhav-datazip: this can be renamed as `numberOfChunks` and same utils.Ternary can be used here

pkg/jdbc/jdbc.go
```diff
@@ -1026,6 +1026,25 @@ func NextRowIDQuery(stream types.StreamInterface, ROWID string, chunkSize int64)
...
+		)
+		SELECT
+			bucket_id,
+			MIN(row_id) AS min_row_id,
+			MAX(row_id) AS max_row_id
+		FROM bucketed_rows
+		GROUP BY bucket_id
+		ORDER BY bucket_id
+	`, numChunks, stream.Namespace(), stream.Name())
+}
```
vaibhav-datazip: This query might consume heavy memory on the db side , for large tables such as 2 Billion rows, can we optimize that to reduce pressure on the db.

This might also cause ORA 1652

mvanhorn:
@vaibhav-datazip addressed all four in 57ceb8f0.

Items 1-3 (backfill.go): `splitViaTableIteration` now takes `approxRowCount` from the caller, no more duplicate `NumRows` fetch. `rowsPerChunk` and the renamed `numberOfChunks` both use `utils.Ternary`.

Item 4 (ORA-1652 risk): Went with a row-count threshold fallback rather than rewriting the NTILE query. Added `oracleNTILERowLimit = 500_000_000` at package scope in `backfill.go`. Tables above that cap skip the NTILE path and reuse `splitViaTableIterationLoop`, which bounds memory per query.

Reasoning: the loop fallback already exists and already runs when stats are missing or the boundary query fails. Gating by row count is the smallest change that removes the 2B-row scenario you flagged without adding a new query path.

Two alternatives if the threshold feels too conservative:
1. `SAMPLE(p)` inside the NTILE CTE, bounded memory but boundaries drift from uniform
2. Coarse outer range split by extent, then NTILE per range, more uniform but needs a new loop

Open question: should the 500M cap stay hardcoded, or wired to a config field? No existing oracle driver config to hook it to, so I left it as a package constant.

vaibhav-datazip:
Hi @mvanhorn,
Our primary goal here is to minimize the load on the database, while shifting as much computation as possible to the OLake runtime. As well as enable fast full loads even for very large datasets (billions of rows).
 
The TODO mainly focuses on improving the existing query by either:
- reducing the number of network calls, or
- using a more efficient query that operates with minimal database permissions.
 
Regarding the NTILE approach, it would be great if you can research for alternative approaches that:
- scale reliably regardless of table size, 
- and provide measurable performance improvements.
 
You can also take a look at the commented-out DB extents query.
Additionally, it would be useful to evaluate and quantify the expected performance gains from any proposed approach.

mvanhorn:
@vaibhav-datazip researched the alternatives you asked about. Here's what I found.

**Strategies evaluated**

| Strategy | DB Load | Scales to 1B+ | Permissions | Network Calls |
|---|---|---|---|---|
| NTILE (current, <500M) | High (full sort) | No (ORA-1652) | SELECT | 1 |
| Iterative loop (old fallback) | Medium | Yes | SELECT | O(N) |
| DBMS_PARALLEL_EXECUTE (primary) | Low (block-based) | Yes | EXECUTE on pkg | 3 |
| DBA_EXTENTS (commented out) | Very low (dict only) | Yes | DBA/SELECT_CATALOG | 2 |
| SAMPLE BLOCK (new, b74bfe65) | Low (~0.1% of blocks) | Yes | SELECT | 1 |

**What I implemented in b74bfe65**

Replaced the iterative loop fallback for tables >500M rows with `SAMPLE BLOCK` estimation (`splitViaTableIterationSample` in `backfill.go:193`).

How it works: `SELECT ROWID FROM schema.table SAMPLE BLOCK(P) ORDER BY ROWID` reads P% of data blocks (not rows), sorts the small sample, picks evenly-spaced boundaries. For a 2B-row table needing 1000 chunks, P is ~0.001% - reads ~20K rows instead of 2B. One query, SELECT-only permissions, no temp tablespace pressure.

Falls back to the iterative loop if the sample returns too few rows.

**About the commented-out DBA_EXTENTS approach (`jdbc.go:993`)**

The extents strategy reads physical file/block layout from `DBA_EXTENTS` and converts to ROWIDs via `DBMS_ROWID.ROWID_CREATE`. Zero table scan, purely dictionary-based. Fastest possible option. The blocker is permissions: `DBA_EXTENTS` requires DBA privilege or SELECT_CATALOG_ROLE. Most OLake users connecting with minimal read-only credentials won't have this.

If you want to support it for users who do have DBA access, the code is already written (`backfill.go:274-342`). Could be gated behind a config flag. `USER_EXTENTS` works for tables owned by the connected user but not for cross-schema reads.

**Performance comparison (estimated for 2B-row table)**

| Strategy | Temp Space | Query Time | Round Trips |
|---|---|---|---|
| NTILE | ~20GB+ sort | Minutes (risks ORA-1652) | 1 |
| Loop (N=1000 chunks) | None | ~10-30s per iteration | 1000+ |
| SAMPLE BLOCK (0.001%) | <1MB sort | <1s | 1 |
| DBA_EXTENTS | None | <100ms (dict lookup) | 2 |

**Current priority order in the code**

1. `splitViaRowId` (DBMS_PARALLEL_EXECUTE) - tried first
2. `splitViaTableIteration` (NTILE for <500M, SAMPLE BLOCK for >500M) - fallback
3. `splitViaTableIterationLoop` - last resort if both NTILE and SAMPLE fail

Let me know if you'd like me to also uncomment and wire up the extents path behind a config flag, or if the SAMPLE BLOCK approach covers the use case.

vaibhav-datazip:
Thanks for the detailed breakdown @mvanhorn , I’ll go through the changes and add a review soon.

One thing to note is a small gap in the permissions table: for a normal Oracle DB (read-write), DBMS_PARALLEL_EXECUTE can still work even if the user only has read permissions. However, in read-only environments like Oracle ADG, it fails with ORA-16000 since it performs internal metadata writes.
Since we rely on fallback strategies in such cases, we should ensure they work in these environments.

I’ll take a closer look at your changes and follow up with a detailed review soon.

mvanhorn:
Good catch on the ADG nuance @vaibhav-datazip, that belongs in the table:

| Strategy | Works in ADG / read-only? | Why |
|---|---|---|
| `DBMS_PARALLEL_EXECUTE` (primary) | No | Internal metadata writes fail with ORA-16000 on standby |
| NTILE | Yes | Pure `SELECT` + sort in temp tablespace |
| `SAMPLE BLOCK` (new fallback) | Yes | Pure `SELECT` with sampling clause, no writes |
| Iterative loop (last resort) | Yes | Pure `SELECT` bounded-range queries |
| `DBA_EXTENTS` | Yes | Dictionary-only lookup, no writes |

So the fallback chain for a 2B-row table in an ADG replica should degrade as: `splitViaRowId` fails with ORA-16000 -> `splitViaTableIteration` handles it (NTILE for <500M rows, SAMPLE BLOCK for larger) -> `splitViaTableIterationLoop` if both fail. All three fallback paths are SELECT-only so none of them should hit the ORA-16000 problem.

I haven't actually exercised that chain against a real ADG replica end-to-end, though. Happy to add an integration test or a note in the docs once you've had a chance to review b74bfe65.

pkg/jdbc/jdbc.go
```diff
@@ -1026,6 +1026,33 @@ func NextRowIDQuery(stream types.StreamInterface, ROWID string, chunkSize int64)
...
+		WITH bucketed_rows AS (
+			SELECT
+				ROWID AS row_id,
+				NTILE(%d) OVER (ORDER BY ROWID) AS bucket_id
+			FROM %q.%q
+		)
+		SELECT
+			bucket_id,
+			MIN(row_id) AS min_row_id,
+			MAX(row_id) AS max_row_id
```
vaibhav-datazip: you can remove max_row_id from this query as this is not being used in chunk generation

drivers/oracle/internal/backfill.go
```diff
@@ -82,13 +88,69 @@ func (o *Oracle) GetOrSplitChunks(ctx context.Context, pool *destination.WriterP
...
+		logger.Debugf("Oracle table [%s.%s] has approxRowCount=%d exceeding NTILE limit=%d, using SAMPLE BLOCK estimation", stream.Namespace(), stream.Name(), approxRowCount, oracleNTILERowLimit)
+		chunks, err := o.splitViaTableIterationSample(ctx, stream, approxRowCount, numberOfChunks)
+		if err != nil {
+			logger.Debugf("SAMPLE BLOCK estimation failed for [%s.%s]: %v, falling back to iterative loop", stream.Namespace(), stream.Name(), err)
+			return o.splitViaTableIterationLoop(ctx, stream, rowsPerChunk)
+		}
+		return chunks, nil
+	}
+
+	rows, err := o.client.QueryContext(ctx, jdbc.AllChunkBoundaryRowIDsQuery(stream, numberOfChunks))
```
vaibhav-datazip: Won't it be better to use just the `splitViaTableIterationLoop` instead of `NTILE query`(removing ntile completely), as a fallback to sampling query.
 As it might happen that the table stats are not updated and NTILE runs for 2B rows table, this will cause high memory usage in the user db.

drivers/oracle/internal/backfill.go
```diff
@@ -134,6 +195,59 @@ func (o *Oracle) splitViaTableIteration(ctx context.Context, stream types.Stream
...
+	minSampleRows := numberOfChunks * 10
+	samplePercent := float64(minSampleRows) / float64(approxRowCount) * 100.0
+
+	// Clamp: Oracle SAMPLE requires (0.000001, 100). Floor at 0.001 for practical minimum.
+	if samplePercent < 0.001 {
+		samplePercent = 0.001
+	}
+	if samplePercent > 50 {
+		samplePercent = 50
+	}
```
vaibhav-datazip: this can be simplified to this
```suggestion
const samplePercentMin = 0.001
const samplePercentMax = 50.0
samplePercent = math.Max(samplePercentMin, math.Min(samplePercentMax, samplePercent))
```

drivers/oracle/internal/backfill.go
```diff
@@ -134,6 +195,59 @@ func (o *Oracle) splitViaTableIteration(ctx context.Context, stream types.Stream
...
 
+// splitViaTableIterationSample uses Oracle's SAMPLE BLOCK clause to estimate chunk
+// boundaries without a full table scan. Reads a small percentage of data blocks,
+// sorts the sampled ROWIDs, and picks evenly-spaced boundaries.
+func (o *Oracle) splitViaTableIterationSample(ctx context.Context, stream types.StreamInterface, approxRowCount int64, numberOfChunks int64) (*types.Set[types.Chunk], error) {
+	// We want at least 10x numberOfChunks sample rows for reasonable boundary accuracy.
+	minSampleRows := numberOfChunks * 10
+	samplePercent := float64(minSampleRows) / float64(approxRowCount) * 100.0
+
+	// Clamp: Oracle SAMPLE requires (0.000001, 100). Floor at 0.001 for practical minimum.
```
vaibhav-datazip: can you add comments , explaining the basis of number being used here , eg why multiplication factor of 10 is being used for min sample rows. why min clamp is 0.001 and max is 50 etc.

mvanhorn:
@vaibhav-datazip thanks, addressed in 29a3b402:

- Dropped the NTILE path entirely; `SAMPLE BLOCK` is now the primary strategy and `splitViaTableIterationLoop` is the single fallback when sampling fails or stats are unavailable. That removes the 2B-row ORA-1652 risk you flagged.
- Removed `AllChunkBoundaryRowIDsQuery` from `pkg/jdbc/jdbc.go` since nothing calls it anymore (and `max_row_id` was unused regardless).
- Extracted `sampleBlockPercentMin` / `sampleBlockPercentMax` / `sampleRowsPerChunkMultiplier` as named constants with comments explaining the rationale.
- Replaced the two-step clamp with the `math.Max(min, math.Min(max, x))` form you suggested.

The `numberOfChunks` / `utils.Ternary` pattern was already in place for that variable in the current code.

vaibhav-datazip:
LGTM, @vikaxsh will be doing 2nd review.

pkg/jdbc/jdbc.go
```diff
@@ -1026,6 +1026,14 @@ func NextRowIDQuery(stream types.StreamInterface, ROWID string, chunkSize int64)
 	return fmt.Sprintf("SELECT MAX(ROWID),COUNT(*) AS row_count FROM(SELECT ROWID FROM %q.%q WHERE ROWID >= '%s' ORDER BY ROWID FETCH FIRST %d ROWS ONLY)", stream.Namespace(), stream.Name(), ROWID, chunkSize)
 }
 
+// SampleBlockBoundaryQuery returns a query that uses Oracle's SAMPLE BLOCK clause
+// to read a small percentage of data blocks and return sorted ROWIDs. This avoids
+// the full-table sort that NTILE requires, making it safe for tables with billions of rows.
+func SampleBlockBoundaryQuery(stream types.StreamInterface, samplePercent float64) string {
```
vikaxsh, line 1070: Could you please confirm whether this might cause any issues
https://dba.stackexchange.com/questions/166495/why-does-sample-block-often-return-zero

mvanhorn: Valid concern from that DBA SE thread - Oracle's SAMPLE BLOCK can return zero rows on small tables or when block selection doesn't intersect any allocated blocks. The code handles this in two layers:

1. `splitViaTableIterationSample` (drivers/oracle/internal/backfill.go:206) explicitly errors when the sample returns fewer rows than `numberOfChunks` (which covers zero).
2. The caller at backfill.go:122-126 wraps that call and falls back to `splitViaTableIterationLoop` (the per-row ROWID walk) on ANY error from the sample path. So a zero-row SAMPLE BLOCK doesn't fail the backfill - it just reverts to the safe iterator.

Belt-and-suspenders: the small tables where SAMPLE BLOCK is most likely to return zero are the same tables where the iterative loop is cheap anyway, so the fallback cost is minimal.

vikaxsh:
hi @mvanhorn, We are currently validating performance improvements on several large tables in our staging environment. If all goes well, we will proceed with this PR. Thank you again for your contribution.

drivers/oracle/internal/backfill.go
```diff
@@ -16,6 +16,20 @@ import (
 	"github.com/datazip-inc/olake/utils/logger"
 )
 
+// Oracle SAMPLE BLOCK clamp bounds for boundary estimation. Oracle allows
+// (0.000001, 100) but 0.001 is the practical floor where the sampler still
+// produces useful output on very large tables, and 50 caps worst-case block
+// scanning so a misconfigured stats row count cannot escalate to a full table
+// scan equivalent.
+const (
+	sampleBlockPercentMin = 0.001
```
vikaxsh: Consider increasing it to 0.01. It improves chunk quality, and in our testing on a ~500 GB table it only took a couple of seconds. I don’t expect much performance impact, but what do you think?

mvanhorn: Bumped to 0.01 in 9b58433. Updated the leading comment to credit your staging numbers (~500 GB, couple of seconds) since that's the data point worth carrying.

vikaxsh (approved):

vaibhav-datazip:
@mvanhorn , if you are interested in contributing more, you can check this issue #871.

Adding extents based strategy might help in optimizing the oracle chunking strategy further.

PR #915, opened 2026-04-09
fix: mssql boolean filter literal handling

vishalm0509:
# Description

This PR fixes filter SQL generation for MSSQL by converting boolean values to `1/0` instead of `TRUE/FALSE`.
The update is applied consistently for both structured (filter_config) and legacy string filter paths.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):


pkg/jdbc/jdbc.go
```diff
@@ -1225,8 +1234,7 @@ func SQLFilter(stream types.StreamInterface, driver string, thresholdFilter stri
 			}
 
 		case bool:
-			// SQL standard boolean
-			valueSQL = utils.Ternary(v, "TRUE", "FALSE").(string)
+			valueSQL = formatFilterBoolValue(driverType, v)
```
hash-data, line 1237: why we are not using reformat ?

pkg/jdbc/jdbc.go
```diff
@@ -1128,6 +1128,13 @@ func IncrementalValueFormatter(ctx context.Context, cursorField, argumentPlaceho
 	return fmt.Sprintf("%s %s %s", quotedCol, operator, argumentPlaceholder), formattedValue, nil
 }
 
+func formatFilterBoolValue(driverType constants.DriverType, value bool) string {
+	if driverType == constants.MSSQL {
```
vikaxsh: can we check for oracle as well

pkg/jdbc/jdbc.go
```diff
@@ -1128,6 +1128,13 @@ func IncrementalValueFormatter(ctx context.Context, cursorField, argumentPlaceho
 	return fmt.Sprintf("%s %s %s", quotedCol, operator, argumentPlaceholder), formattedValue, nil
 }
 
+func formatFilterBoolValue(driverType constants.DriverType, value bool) string {
```
vikaxsh: the file follows a query-builder pattern, formatFilterBoolValue is only called within SQLFilter and has no other consumers , move it as a local closure

pkg/jdbc/jdbc.go
vishalm0509: In **Oracle** there is no `bit` or `bool` datatype

pkg/jdbc/jdbc.go
vishalm0509: Done

pkg/jdbc/jdbc.go
vishalm0509: `ReformatValue` is used for internal purpose, and here we are formatting the value to be used in the SQL Query. 


For example, we perform driver specific query supported formatting, which is not done in the reformatValue
```go
	// Driver-specific timestamp handling for ISO 8601 / RFC3339 strings.
			isISO8601 := strings.Contains(v, "T") && (strings.Contains(v, "Z") || strings.Contains(v, "+") || (strings.Contains(v, "-") && len(v) > 19))
			if isISO8601 {
				if t, err := time.Parse(time.RFC3339, v); err == nil {
					switch driverType {
					case constants.Oracle:
						valueSQL = fmt.Sprintf("TO_TIMESTAMP('%s', 'YYYY-MM-DD HH24:MI:SS.FF')", t.UTC().Format("2006-01-02 15:04:05.000"))
					case constants.DB2:
						// DB2 TIMESTAMP() scalar accepts 'YYYY-MM-DD HH:MM:SS.ffffff'
						valueSQL = fmt.Sprintf("TIMESTAMP('%s')", t.UTC().Format("2006-01-02 15:04:05.000000"))
					}
				}
			}
````

vikaxsh (approved):

PR #922, opened 2026-04-19
fix: last message parsing issue in kafka sync

ImDoubD-datazip:
# Description

If the last message in partition is not json or avro, that means it is not parsable, the sync pauses at that point until the context gets cancelled itself. The fix is to skip the last message, mark in kafka but not write in destination.

## Type of change

# How Has This Been Tested?

Tested in json and avro topics, by inserting non-parseable messages at beginning, in middle and at the end of the topics with multiple partitions.

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):


drivers/kafka/internal/cdc.go
```diff
@@ -79,28 +79,26 @@ func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates
...
-		err := processFn(ctx, abstract.CDCChange{
-			Stream:    currentPartitionMeta.Stream,
-			Timestamp: record.Message.Time,
-			Kind:      "create",
-			Data:      record.Data,
-		})
-		if err != nil {
-			return false, err
+		// record nil is safety check, should not happen
+		if !shouldSkip && record.Data != nil {
```
saksham-datazip: can u tell me a condition why record.Data alone is not sufficient except from the case when message is null ? and in which case we will be needing this variable?

drivers/kafka/internal/cdc.go
```diff
@@ -185,31 +183,36 @@ func (k *Kafka) PostCDC(ctx context.Context, readerIdx int) error {
...
 		message, err := reader.FetchMessage(ctx)
 		if err != nil {
 			return fmt.Errorf("error reading message in Kafka CDC sync: %s", err)
 		}
 
-		var data map[string]interface{}
+		var (
+			key        string
+			data       map[string]interface{}
+			shouldSkip bool
```
saksham-datazip: Can we rename shouldSkip to shouldProcessChange and align the logic accordingly? The current negative naming makes the condition harder to read  and we can use this variable directly ????

drivers/kafka/internal/cdc.go
```diff
@@ -185,31 +183,36 @@ func (k *Kafka) PostCDC(ctx context.Context, readerIdx int) error {
...
-			data[Partition] = message.Partition
-			data[Offset] = message.Offset
-			data[Key] = key
-			data[KafkaTimestamp], err = typeutils.ReformatDate(message.Time, true)
-			if err != nil {
-				return fmt.Errorf("failed to reformat date: %s", err)
+				logger.Warnf("failed to parse message of topic: %s, partition: %d, offset %d, error: %s", message.Topic, message.Partition, message.Offset, err)
+				shouldSkip = true
+			} else {
+				data[Partition] = message.Partition
```
saksham-datazip: also their is no error handling for the case when message is null causing panic 
Logs for the following :-
```
2026-04-19T20:51:31Z INFO Sync completed, wait 5 seconds cleanup in progress...
2026-04-19T20:51:36Z FATAL error occurred while reading records: error occurred while waiting for connections: %!s(<nil>); failed closing writer[topics.test1]: failed to close writer: context canceled: prev error: panic recovered: assignment to entry in nil map: post cdc error: context canceled
apple@Apples-MacBook-Pro-2 olake % 
```
Steps to recreate in kafka ui simply add a message with a value null

drivers/kafka/internal/cdc.go
ImDoubD-datazip: we never know which type of data might the user puts in his kafka message, as you can put anything. Just for a safety check, this condition is present. let's not remove this for now.

drivers/kafka/internal/cdc.go
ImDoubD-datazip: what is `shouldProcessChange`, I don't understand anything from this, we are skipping the messages here, but marking the offset. shouldSkip is understandable here.

drivers/kafka/internal/cdc.go
ImDoubD-datazip: ok

drivers/kafka/internal/cdc.go
ImDoubD-datazip: removed

drivers/kafka/internal/cdc.go
ImDoubD-datazip: removed

saksham-datazip: i was asking for removing shouldSkip not the other way around ...

drivers/kafka/internal/cdc.go
saksham-datazip: shouldProcessChange means should we process the change or not ,which is written in this part of comment as well in streaamchanges ,but it is removed so no problem 
```
		// process the change
		err := processFn(ctx, abstract.CDCChange{
			Stream:    currentPartitionMeta.Stream,
			Timestamp: record.Message.Time,
			Kind:      "create",
			Data:      record.Data,
		})
		if err != nil {
			return false, err
		}
```

drivers/kafka/internal/cdc.go
```diff
@@ -192,14 +189,17 @@ func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kafka.Reader,
...
+			key  string
+			data map[string]interface{}
+		)
+
+		// parse message value and key
+		data, key, err = k.parseKafkaData(message)
+		if err != nil {
+			logger.Warnf("failed to parse message of topic: %s, partition: %d, offset %d, error: %s", message.Topic, message.Partition, message.Offset, err)
+		} else if data != nil {
+			// for cases of unparseable message, or null in message value, data map will be nil so nil check is required
```
saksham-datazip: The comment is a bit misleading. Unparseable messages are handled in the `err != nil` branch, so this block is only relevant when the message value is null.

drivers/kafka/internal/cdc.go
```diff
@@ -192,14 +189,17 @@ func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kafka.Reader,
...
 
-		var data map[string]interface{}
-		if message.Value != nil {
-			var key string
-			data, key, err = k.parseKafkaData(message)
-			if err != nil {
-				logger.Warnf("failed to parse message at offset %d: %s", message.Offset, err)
-				continue
-			}
+		var (
```
saksham-datazip, line 192: why are we defining this key value above when it is only needed in message.value block can't we revert this declaration back to way it was in staging ?

drivers/kafka/internal/cdc.go
ImDoubD-datazip: done change

saksham-datazip (approved):
LGTM

PR #925, opened 2026-04-23
fix(iceberg): create one positional delete file per data file ref

badalprasadsingh:
# Description

Compaction fails on tables with cdc upserts when multiple changes arrive for the same `_olake_id` in a single batch, due to the presence of positional delete file having multiple data file references in it.

A detailed explanation is provided [here](https://datazip.atlassian.net/wiki/x/B4BTGQ).

## Fix:
As this is a special case of OLake, where a positional delete only references the "new" data files. In case a positional delete references multiple data files, `lower_bound != upper_bound`,

Thus, Iceberg simply doesn't checks if the data file referenced in the positional delete is a new data file or old data file, and fails the compaction.

We fix this issue by making sure:

- Each positional delete file now references exactly one data file → `lower_bound == upper_bound` 
- Ensuring, complete S3 path is stored as string values, without any truncation

Iceberg's [DeleteGranularity Feature](https://iceberg.apache.org/javadoc/1.5.2/org/apache/iceberg/deletes/DeleteGranularity.html)

## Type of change

# How Has This Been Tested?

- [x] Multiple upserts in a single batch in a big table
- [x] Both partitioned and un-partitioned iceberg table
- [x] Backward Compatibility

## Documentation

### DIFF
--- a/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/IcebergUtil.java
+++ b/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/IcebergUtil.java
@@ -251,7 +251,8 @@ public class IcebergUtil {
           null,
           null,
           null)
-          .setAll(icebergTable.properties());
+          .setAll(icebergTable.properties())
+          .set("write.metadata.metrics.column.file_path", "full");
     } else {
       return new GenericAppenderFactory(
           icebergTable.schema(),
@@ -259,7 +260,8 @@ public class IcebergUtil {
           Ints.toArray(identifierFieldIds),
           TypeUtil.select(icebergTable.schema(), Sets.newHashSet(identifierFieldIds)),
           null)
-          .setAll(icebergTable.properties());
+          .setAll(icebergTable.properties())
+          .set("write.metadata.metrics.column.file_path", "full");
     }
   }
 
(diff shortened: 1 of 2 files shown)


destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/IcebergUtil.java
```diff
@@ -251,15 +251,17 @@ public static GenericAppenderFactory getTableAppender(Table icebergTable) {
...
     } else {
       return new GenericAppenderFactory(
           icebergTable.schema(),
           icebergTable.spec(),
           Ints.toArray(identifierFieldIds),
           TypeUtil.select(icebergTable.schema(), Sets.newHashSet(identifierFieldIds)),
           null)
-          .setAll(icebergTable.properties());
+          .setAll(icebergTable.properties())
+          .set("write.metadata.metrics.column.file_path", "full");
```
hash-data, line 264: this property will affect arrow write once verify

hash-data (approved):

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/IcebergUtil.java
badalprasadsingh: Tested, it is not affecting the path of arrow writer.

PR #930, opened 2026-04-27
feat: add schema filtering for discover in postgres

vikaxsh:
# Description

Add optional `schemas` config field to restrict discover to user-specified PostgreSQL schemas instead of fetching all accessible schemas. When omitted, existing behaviour is preserved (all non-system schemas are discovered).

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

> ⚠️ NOTE: Docs to be added for this PR 

## Related PR's (If Any):

### DIFF
--- a/drivers/postgres/internal/config.go
+++ b/drivers/postgres/internal/config.go
@@ -88,8 +88,9 @@ func (c *Config) Validate() error {
 	parsed.RawQuery = query.Encode()
 	c.Connection = parsed
 
-	for i, s := range c.Schemas {
-		if strings.TrimSpace(s) == "" {
+	for i := range c.Schemas {
+		c.Schemas[i] = strings.TrimSpace(c.Schemas[i])
+		if c.Schemas[i] == "" {
 			return fmt.Errorf("schemas[%d] must not be blank", i)
 		}
 	}
--- a/drivers/postgres/internal/config.go
+++ b/drivers/postgres/internal/config.go
@@ -88,9 +88,8 @@ func (c *Config) Validate() error {
 	parsed.RawQuery = query.Encode()
 	c.Connection = parsed
 
-	for i := range c.Schemas {
-		c.Schemas[i] = strings.TrimSpace(c.Schemas[i])
-		if c.Schemas[i] == "" {
+	for i, s := range c.Schemas {
+		if strings.TrimSpace(s) == "" {
 			return fmt.Errorf("schemas[%d] must not be blank", i)
 		}
 	}
--- a/drivers/postgres/internal/config.go
+++ b/drivers/postgres/internal/config.go
@@ -23,6 +23,7 @@ type Config struct {
 	MaxThreads       int               `json:"max_threads"`
 	RetryCount       int               `json:"retry_count"`
 	SSHConfig        *utils.SSHConfig  `json:"ssh_config"`
+	Schemas          []string          `json:"schemas,omitempty"`
 }
 
 // Capture Write Ahead Logs
@@ -86,6 +87,12 @@ func (c *Config) Validate() error {
 
 	parsed.RawQuery = query.Encode()
 	c.Connection = parsed
+
+	for i, s := range c.Schemas {
+		if strings.TrimSpace(s) == "" {
+			return fmt.Errorf("schemas[%d] must not be blank", i)
+		}
+	}
 
 	return nil
 }
--- a/drivers/postgres/internal/postgres.go
+++ b/drivers/postgres/internal/postgres.go
@@ -37,6 +37,17 @@ const (
 	// extends getPrivilegedTablesTmpl to restrict to user-specified schemas
 	getPrivilegedTablesFilteredTmpl = getPrivilegedTablesTmpl + `
 		AND nspname = ANY($1)`
+	// configured schema names that are missing from pg_namespace or lack USAGE for current_user
+	validateConfiguredSchemasTmpl = `
+		WITH requested AS (
+			SELECT DISTINCT btrim(unnest($1::text[])) AS schema_name
+		)
+		SELECT r.schema_name
+		FROM requested r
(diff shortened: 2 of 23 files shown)


drivers/postgres/internal/postgres.go
```diff
@@ -177,12 +182,24 @@ func (p *Postgres) CloseConnection() {
...
-	err := p.client.SelectContext(ctx, &tableNamesOutput, getPrivilegedTablesTmpl)
+
+	var (
+		tableNamesOutput []Table
+		err              error
+	)
+
+	if len(p.config.Schemas) > 0 {
+		logger.Infof("Schema filter applied, discovering only schemas: %v", p.config.Schemas)
+		err = p.client.SelectContext(ctx, &tableNamesOutput, getPrivilegedTablesFilteredTmpl, pq.Array(p.config.Schemas))
```
vishalm0509, line 193: does pgx not provide any alternative for `pq.Array()`, can you check once

drivers/postgres/internal/postgres.go
```diff
@@ -101,6 +117,9 @@ func (p *Postgres) Setup(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to ping database: %s", err)
 	}
+	if err := p.validateConfiguredSchemas(ctx, pgClient); err != nil {
+		return err
+	}
```
vaibhav-datazip: during sync if some schema isn't present now which has been added in config file, this will cause error. Shouldn't we just skip it as its not currently in sync 

example
If a user configures schemas: ["A", "B"], runs a successful discover, selects streams from A, and later the DBA removes the B schema, the sync will fail at Setup() even though the selected streams in streams.json are all from A which still exists.

drivers/postgres/internal/config.go
```diff
@@ -87,6 +88,12 @@ func (c *Config) Validate() error {
 	parsed.RawQuery = query.Encode()
 	c.Connection = parsed
 
+	for i, s := range c.Schemas {
+		if strings.TrimSpace(s) == "" {
+			return fmt.Errorf("schemas[%d] must not be blank", i)
+		}
+	}
```
vaibhav-datazip, line 95: why not trim space here for schema names as well ?

drivers/postgres/resources/spec.json
```diff
@@ -131,6 +131,14 @@
                 "mode"
             ]
         },
+        "schemas": {
+            "type": "array",
+            "title": "Schemas",
+            "description": "List of PostgreSQL schemas to discover and sync. If omitted or empty, all accessible non-system schemas are discovered.",
+            "items": {
+                "type": "string"
+            }
```
vaibhav-datazip, line 141: shouldnt we also add `"uniqueItems": true` here.
and default []


even `"items": { "type": "string", "minLength": 1 }`
`"minItems": 1  `

drivers/postgres/internal/postgres.go
vishalm0509: Now, the validation runs only for the **Check** command (test-connection)

drivers/postgres/internal/config.go
vishalm0509: I checked for jdbc_url params, we are not trimming value. It's configuration right.

drivers/postgres/resources/spec.json
vishalm0509: Added uniqueItems constraint

drivers/postgres/internal/postgres.go
```diff
@@ -31,7 +32,22 @@ const (
...
+		AND nspname != 'information_schema'` // Exclude information_schema
+
+	// extends getPrivilegedTablesTmpl to restrict to user-specified schemas
+	getPrivilegedTablesFilteredTmpl = getPrivilegedTablesTmpl + `
+		AND nspname = ANY($1)`
+	// configured schema names that are missing from pg_namespace or lack USAGE for current_user
+	validateConfiguredSchemasTmpl = `
+		WITH requested AS (
+			SELECT DISTINCT btrim(unnest($1::text[])) AS schema_name
+		)
```
vaibhav-datazip: why in validate we have trim , but not in fetching the table query ?

this will cause the validation to pass but 0 tables returned during query , either you can use trim space during validation

protocol/check.go
```diff
@@ -42,7 +42,11 @@ var checkCmd = &cobra.Command{
 			}
 
 			if configPath != "not-set" {
-				return connector.Setup(cmd.Context())
+				ctx := cmd.Context()
+				if err := connector.Setup(ctx); err != nil {
+					return err
+				}
+				return connector.ValidateConfiguredSchemas(ctx)
```
vaibhav-datazip: why not add it in discover.go as well ?

hash-data: why?

vishalm0509: Reverted changes

vishalm0509: Reverted all changes. No more validation required :)

drivers/postgres/internal/postgres.go
vishalm0509: No more validation

drivers/postgres/internal/config.go
```diff
@@ -23,6 +23,7 @@ type Config struct {
 	MaxThreads       int               `json:"max_threads"`
 	RetryCount       int               `json:"retry_count"`
 	SSHConfig        *utils.SSHConfig  `json:"ssh_config"`
+	Schemas          []string          `json:"schemas,omitempty"`
```
vaibhav-datazip, line 26: shouldn't we call this selected schemas or something similar to that, as that will make more sense

vaibhav-datazip (approved):

vishalm0509 (approved):
LGTM

PR #933, opened 2026-04-30
feat: add optimized chunking strategies for MSSQL

vaibhav-datazip:
# Description

- **IAM walk chunk planning**: Adds a fast physloc-boundary planner using sys.dm_db_database_page_allocations (LIMITED) to stream (file_id, page_id) and generate page-aligned %%physloc%% chunks without scanning/sorting the table; used when supported (SQL Server 2012+ + VIEW DATABASE STATE, not Azure SQL DB/MI).

- **Sampling fallback + shared abstraction**: When IAM walk is unavailable/fails, adds TABLESAMPLE SYSTEM + %%physloc%% sampling to estimate evenly-spaced boundaries using only SELECT permission; the sampling % calculation and clamp constants are shared across MSSQL + Oracle.

Dependencies / requirements:
- SQL Server **2012+**, **VIEW DATABASE STATE** permission, and not Azure SQL DB/MI for IAM-walk planning.
- Existing strategies remain as fallback when IAM-walk is unavailable.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Scenario A: MSSQL sync with IAM-walk enabled 
- [x] Scenario B: IAM-walk unsupported (capability/permission/platform) falls back cleanly
  - Remove/deny `VIEW DATABASE STATE` (or run on Azure SQL DB/MI) and re-run the same sync.
  - Verify logs show fallback to existing strategy and sync proceeds.

# Screenshots or Recordings

## Documentation

- [ ]  Documentation Link: [link to README, olake.io/docs, or olake-docs]

## Related PR's (If Any):


pkg/jdbc/jdbc.go
```diff
@@ -985,6 +1029,17 @@ func MSSQLTableRowStatsQuery() string {
...
+// MSSQLPhysLocSampleBoundaryQuery returns a query that uses TABLESAMPLE SYSTEM to
+// sample a percentage of data pages and return sorted %%physloc%% binary values.
+func MSSQLPhysLocSampleBoundaryQuery(stream types.StreamInterface, samplePercent float64) string {
+	quotedTable := QuoteTable(stream.Namespace(), stream.Name(), constants.MSSQL)
+	return fmt.Sprintf(`
+		SELECT %%%%physloc%%%%
+		FROM %s TABLESAMPLE SYSTEM (%.6f PERCENT) WITH (NOLOCK)
+		ORDER BY %%%%physloc%%%%
+	`, quotedTable, samplePercent)
+}
```
vishalm0509: Do we need with NO LOCK ? What if we include a row which got rollbacked ?

drivers/mssql/internal/backfill.go
```diff
@@ -20,6 +22,30 @@ import (
...
+
+// usableBytesPerPage is an upper bound for in-row payload per 8KB page
+// (IN_ROW_DATA max row size). Using the ceiling yields smaller chunks.
+const usableBytesPerPage = 8060
+
+// isPhysLocChunk reports whether the chunk's boundaries are %%physloc%%
+// hex literals (the shape produced by IAM walk and the iterative physloc
+// planner). Either Min or Max is sufficient to identify the format because
+// both planners always set at least one boundary.
+func isPhysLocChunk(chunk types.Chunk) bool {
```
vishalm0509: Can we move this util to the bottom ?

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +300,212 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+	logger.Debugf("TABLESAMPLE sampling %.4f%% of pages from [%s.%s] for chunk boundaries (approxRows=%d, chunks=%d)",
+		samplePercent, stream.Namespace(), stream.Name(), approxRowCount, numberOfChunks)
+
+	rows, err := m.client.QueryContext(ctx, jdbc.MSSQLPhysLocSampleBoundaryQuery(stream, samplePercent))
+	if err != nil {
+		return nil, fmt.Errorf("TABLESAMPLE %%%%physloc%%%% query failed: %s", err)
+	}
+	defer rows.Close()
+
+	var sampledLocs [][]byte
```
vishalm0509: sampledLocs ?

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +300,212 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+		max := utils.HexEncode(physLocBytes(pages[i]))
+		chunks.Insert(types.Chunk{Min: prev, Max: max})
+		prev = max
+	}
+	chunks.Insert(types.Chunk{Min: prev, Max: nil})
+
+	return nil
+}
+
+// probeIAMWalkCapability checks if IAM walk
```
vishalm0509: comment incomplete ?

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +300,212 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+func (m *MSSQL) probeIAMWalkCapability(ctx context.Context) bool {
+	var majorVersion, engineEdition int
+	err := m.client.QueryRowContext(ctx, jdbc.MSSQLIAMWalkServerPropertiesQuery()).Scan(&majorVersion, &engineEdition)
+	if err != nil {
+		logger.Debugf("IAM walk probe: failed to read server properties: %s", err)
+		return false
+	}
+	if majorVersion < 11 {
```
vishalm0509, line 400: constants ?

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +300,212 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+	// EngineEdition 5 = Azure SQL Database, 8 = Azure SQL Managed Instance.
+	// sys.dm_db_database_page_allocations is blocked on both.
+	if engineEdition == 5 || engineEdition == 8 {
+		logger.Debugf("IAM walk probe: EngineEdition %d (Azure SQL DB/MI) blocks the DMF", engineEdition)
+		return false
+	}
+
+	// Permission probe: TOP 0 evaluates the DMF without returning any rows.
+	// Failure here means the current login lacks VIEW DATABASE STATE.
+	rows, err := m.client.QueryContext(ctx, jdbc.MSSQLIAMWalkPermissionQuery())
```
vishalm0509, line 413: We have this query to check the permission

```go
// MSSQLViewDatabaseStatePermissionQuery checks for VIEW DATABASE STATE (all versions) or
// VIEW DATABASE PERFORMANCE STATE (SQL Server 2022+). Either permission grants access to
// sys.dm_cdc_log_scan_sessions.
func MSSQLViewDatabaseStatePermissionQuery() string {
	return `
		SELECT CAST(CASE
			WHEN HAS_PERMS_BY_NAME(NULL, 'DATABASE', 'VIEW DATABASE STATE') = 1 THEN 1
			WHEN HAS_PERMS_BY_NAME(NULL, 'DATABASE', 'VIEW DATABASE PERFORMANCE STATE') = 1 THEN 1
			ELSE 0
		END AS BIT)
	`
}
```

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +300,212 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+		return fmt.Errorf("failed to resolve object_id for IAM walk: %s", err)
+	}
+
+	rows, err := m.client.QueryContext(ctx, jdbc.MSSQLIAMWalkQuery(), objectID)
+	if err != nil {
+		return fmt.Errorf("failed to run IAM walk query: %s", err)
+	}
+	defer rows.Close()
+
+	pages := make([]uint64, 0, 1024)
```
vishalm0509, line 354: Why 1024 ?

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +300,212 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+// byte 7 into the lowest. slot_id is fixed at 0xFFFF so each boundary
+// represents "after the last possible row on the page": chunk filter
+// `%%physloc%% <= b` includes every row on b's page, and `> b` starts
+// cleanly on the next page.
+//
+// This lets us sort []uint64 with native uint64 comparison (instead of
+// allocating an 8-byte slice per page and sorting with bytes.Compare),
+// halving memory on large tables and skipping the per-page encode step
+// entirely; we encode only the few sampled boundaries.
+func packPhysLoc(fileID, pageID int32) uint64 {
```
vishalm0509:
```go
func physlocPageBoundarySortKey(fileID, pageID int32) uint64 {
	var b [8]byte
	binary.LittleEndian.PutUint32(b[0:4], uint32(pageID))
	binary.LittleEndian.PutUint16(b[4:6], uint16(fileID))
	binary.LittleEndian.PutUint16(b[6:8], 0xFFFF)
	return binary.BigEndian.Uint64(b[:])
}
```

Can you check this, claude suggested. If it's better we can use it

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +300,212 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+	}
+	chunks.Insert(types.Chunk{Min: prev, Max: nil})
+
+	return chunks, nil
+}
+
+// splitViaIAMWalk plans chunks for any heap or clustered table by reading
+// only the table's Index Allocation Map pages via
+// sys.dm_db_database_page_allocations.
+func (m *MSSQL) splitViaIAMWalk(ctx context.Context, stream types.StreamInterface, chunks *types.Set[types.Chunk]) error {
```
vishalm0509: > 1. The first chunk {nil, B1} is unbounded on the low side.
    Scan: WHERE %%physloc%% <= B1. This includes rows on any page below B1, including pages not in the IAM walk list (e.g. pages deallocated since stats were gathered, pages from forwarded-row pointers).

Can you verify this

drivers/mssql/internal/backfill.go
```diff
@@ -48,11 +74,14 @@ func (m *MSSQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
 
 		// Build query for the chunk
 		stmt := ""
-		if chunkColumn != "" {
+		switch {
+		case chunkColumn != "":
 			stmt = jdbc.MSSQLChunkScanQuery(stream, []string{chunkColumn}, chunk, filter)
-		} else if len(pkColumns) > 0 {
+		case isPhysLocChunk(chunk):
```
vishalm0509: A table with a BINARY(8) primary key produces PK boundaries that are "0x" + 16 hex chars = 18 characters — the exact same shape isPhysLocChunk uses to detect physloc boundaries. ChunkIterator would scan with WHERE %%physloc%% > X instead of WHERE pk > X. Wrong rows returned. Data loss / incorrect results. This is not theoretical — BINARY(8) PKs exist in real workloads (hash keys, external IDs).

vishalm0509 (commented):
1. Image

2. Please change PR title to that of MSSQL

pkg/jdbc/jdbc.go
vaibhav-datazip: we will get a physloc id only, so our chunk will be lets say something like
nil -> a
a -> b
b -> c
c -> d
d -> nil

we wont be missing any data or getting extra data.

drivers/mssql/internal/backfill.go
vaibhav-datazip: sampled physlocs -> sampledlocs

drivers/mssql/internal/backfill.go
vaibhav-datazip: completed

drivers/mssql/internal/backfill.go
vaibhav-datazip: every uint64 takes 8bytes, so 1 page id takes 8 bytes , 1024/8 -> 128 page ids can be stored and it will keep doubling if array increases.

so its a optimal starting point for small or medium tables

drivers/mssql/internal/backfill.go
vaibhav-datazip: both will produce same result without any error, the current one is slightly faster as it operates directly on bytes without using any function. the one you suggested is more readable.

drivers/mssql/internal/backfill.go
vaibhav-datazip: but we want all the pages the pages which belong to the table right ? so this wont be adding any unwanted data

drivers/mssql/internal/backfill.go
vaibhav-datazip: had asked Claude sonnet to do research on percentage of primary key datatypes, attaching the results here

Screenshot 2026-05-12 at 7 22 40 PM

drivers/mssql/internal/backfill.go
vishalm0509: physLocSamples ?? makes it easier while reading

drivers/mssql/internal/backfill.go
vishalm0509: It's a one time operation right, so should we optimize for negligible performance difference or readability ?

drivers/mssql/internal/backfill.go
vishalm0509: `normalizeBoundaryValue` if you check this function we are hexEncoding the boundary values. So in the ChunkIterator even if the bondaryValue is not physloc it will run

```go
		case isPhysLocChunk(chunk):
			stmt = jdbc.MSSQLPhysLocChunkScanQuery(stream, chunk, filter)
```

right ?

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +291,216 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 		})
 	}
 
-	if len(pkColumns) > 0 {
-		logger.Debugf("Stream %s: Using PK-based chunking with columns: %v", stream.ID(), pkColumns)
+	// physLocSampleThenFallback tries TABLESAMPLE physloc sampling and, if that
+	// produces no usable chunks, falls back to PK or iterative physloc scan.
+	physLocSampleThenFallback := func() error {
+		sampleChunks, sampleErr := m.splitViaPhysLocSample(ctx, stream, approxRowCount, numberOfChunks)
```
vishalm0509: There are some strategies implemented as inline (anonymous func) and some are methods. 

Can we standardize on one ?

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +291,216 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+		sampledLocs = append(sampledLocs, loc)
+	}
+	if err = rows.Err(); err != nil {
+		return nil, fmt.Errorf("failed to iterate sampled %%%%physloc%%%% rows: %s", err)
+	}
+
+	if int64(len(sampledLocs)) < numberOfChunks {
+		return nil, fmt.Errorf("TABLESAMPLE returned %d rows, need at least %d for %d chunks",
+			len(sampledLocs), numberOfChunks, numberOfChunks)
+	}
```
vishalm0509:
```suggestion
	if int64(len(sampledLocs)) < numberOfChunks {
		return nil, fmt.Errorf("TABLESAMPLE returned %d rows, need at least %d",
			len(sampledLocs), numberOfChunks)
	}
```

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +291,216 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+	// produces no usable chunks, falls back to PK or iterative physloc scan.
+	physLocSampleThenFallback := func() error {
+		sampleChunks, sampleErr := m.splitViaPhysLocSample(ctx, stream, approxRowCount, numberOfChunks)
+		if sampleErr == nil && sampleChunks.Len() > 0 {
+			logger.Infof("Stream %s: TABLESAMPLE physloc sampling produced %d chunks", stream.ID(), sampleChunks.Len())
+			chunks = sampleChunks
+			return nil
+		}
+		logger.Debugf("Stream %s: TABLESAMPLE physloc sampling failed %s", stream.ID(), sampleErr)
+		chunks = types.NewSet[types.Chunk]()
```
vishalm0509: Do we need this here ?

drivers/mssql/internal/backfill.go
vishalm0509: Did you check this ??

drivers/mssql/internal/backfill.go
vishalm0509: ??

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +291,216 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+	}
+	chunks.Insert(types.Chunk{Min: prev, Max: nil})
+
+	return nil
+}
+
+// probeIAMWalkCapability checks whether the connected SQL Server instance
+// supports IAM walk chunking via sys.dm_db_database_page_allocations.
+func (m *MSSQL) probeIAMWalkCapability(ctx context.Context) bool {
+	m.once.Do(func() {
```
vishalm0509: Why do we need caching here, it's a very small query right. We are not following this design in any other driver. 

Can we remove it ?

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +291,216 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+		// EngineEdition 5 = Azure SQL Database, 8 = Azure SQL Managed Instance.
+		// sys.dm_db_database_page_allocations is blocked on both.
+		if engineEdition == 5 || engineEdition == 8 {
+			logger.Debugf("IAM walk probe: EngineEdition %d (Azure SQL DB/MI) blocks the DMF", engineEdition)
+			return
+		}
+
+		// Permission probe: TOP 0 evaluates the DMF without returning any rows.
+		// Failure here means the current login lacks VIEW DATABASE STATE.
+		rows, err := m.client.QueryContext(ctx, jdbc.MSSQLIAMWalkPermissionQuery())
```
vishalm0509: here, we can use the already present query to check perm. There's a comment already above.

drivers/mssql/internal/backfill.go
```diff
@@ -270,17 +291,216 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+
+	chunks := types.NewSet[types.Chunk]()
+	step := float64(len(sampledLocs)) / float64(numberOfChunks)
+	var prev any = nil
+	for i := int64(0); i < numberOfChunks; i++ {
+		curr := utils.HexEncode(sampledLocs[int(float64(i)*step)])
+		chunks.Insert(types.Chunk{Min: prev, Max: curr})
+		prev = curr
+	}
+	chunks.Insert(types.Chunk{Min: prev, Max: nil})
```
vishalm0509: These are also present in both the sampling function in Oracle and MSSQL. Can this be abstracted as well ? Or is it intentional to not abstract it ?

drivers/mssql/internal/backfill.go
vaibhav-datazip: this is a db level check , so we just need to check this once  also this was based on your earlier comment 

https://github.com/datazip-inc/olake/pull/933#pullrequestreview-4269073173

drivers/mssql/internal/backfill.go
vaibhav-datazip: That call runs MSSQLIAMWalkPermissionQuery(), which executes
sys.dm_db_database_page_allocations(..., 'LIMITED') with TOP 0.

So we’re checking: can this login successfully run that DMF in this database?
If yes → we assume IAM-walk style queries are allowed (typically needs VIEW DATABASE STATE, and the engine/platform must allow the DMF).
If it errors → IAM walk is treated as unavailable.

drivers/mssql/internal/backfill.go
vaibhav-datazip: this chunks adding process is similar for all drivers still we haven't abstracted it our so following that same architecture.

drivers/mssql/internal/backfill.go
```diff
@@ -117,168 +131,351 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+
+// physLocBytes converts a packed uint64 boundary back to its 8-byte
+// %%physloc%% wire representation. Because packPhysLoc places physloc
+// byte 0 at the uint64's most significant 8 bits, writing the uint64 in
+// big-endian order reproduces the wire layout exactly.
+func physLocBytes(packed uint64) []byte {
+	b := make([]byte, 8)
+	binary.BigEndian.PutUint64(b, packed)
+	return b
+}
```
vishalm0509: Since these are utils we can take this to the bottom or utils ?

drivers/mssql/internal/backfill.go
```diff
@@ -117,168 +131,351 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+// probeIAMWalkCapability checks whether IAM walk can run on this server/login.
+// It is cheap (two round-trips) so we call it per stream instead of caching.
+func (m *MSSQL) probeIAMWalkCapability(ctx context.Context) bool {
+	var majorVersion, engineEdition int
+	err := m.client.QueryRowContext(ctx, jdbc.MSSQLIAMWalkServerPropertiesQuery()).Scan(&majorVersion, &engineEdition)
+	if err != nil {
+		logger.Debugf("IAM walk probe: failed to read server properties: %s", err)
+		return false
+	}
+	if majorVersion < 11 {
```
vishalm0509, line 400: Add a comment why `11`

drivers/mssql/internal/backfill.go
vaibhav-datazip: physlocbytes is only called for actual chunk boundaries we can discuss this

vishalm0509 (commented):
LGTM

drivers/mssql/internal/backfill.go
```diff
@@ -117,168 +130,334 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
 			}
+		}
 
-			columnType := ""
-			if len(pkCols) == 1 {
-				columnType, err = m.getColumnTypeMSSQL(ctx, stream, pkCols[0], tx)
-				if err != nil {
-					return fmt.Errorf("failed to get table column type: %s", err)
-				}
-			}
```
vikaxsh, line 144: can you please add all the comments again

pkg/jdbc/jdbc.go
```diff
@@ -660,6 +660,50 @@ func MSSQLPhysLocNextChunkEndQuery(stream types.StreamInterface, chunkSize int64
...
+// MSSQLIAMWalkQuery returns a streaming IAM-walk page list for a table:
+// (file_id, page_id) from sys.dm_db_database_page_allocations in LIMITED mode.
+// Params: @p1 = object_id. Requires VIEW DATABASE STATE; SQL Server 2012+;
+// blocked on Azure SQL DB/MI.
+func MSSQLIAMWalkQuery() string {
+	return `
+		SELECT
+			allocated_page_file_id AS file_id,
+			allocated_page_page_id AS page_id
+		FROM sys.dm_db_database_page_allocations(
```
vikaxsh, line 672: the following query is going to take schema stability lock and blocks ddl operation on the table

https://www.mssqltips.com/sqlservertip/6309/use-caution-with-sysdmdbdatabasepageallocations-in-sql-server/#:~:text=it%20does%20block%20DDL%20on%20the%20current%20target.

drivers/abstract/backfill.go
```diff
@@ -97,3 +97,13 @@ func (a *AbstractDriver) Backfill(mainCtx context.Context, backfilledStreams cha
 	utils.ConcurrentInGroupWithRetry(a.GlobalConnGroup, chunks, a.driver.MaxRetries(), chunkProcessor)
 	return nil
 }
+
+// ComputeSamplePercent returns the TABLESAMPLE / SAMPLE BLOCK percentage to use
+// for chunk boundary estimation. The result is clamped to
+// [constants.SamplePercentMin, constants.SamplePercentMax] so callers never
+// request a near-zero or near-full scan regardless of row-count estimates.
+func ComputeSamplePercent(approxRowCount, numberOfChunks int64) float64 {
```
vikaxsh: ComputeSamplePercent Should Be in utils, Not abstract,
The abstract package hosts the shared orchestration logic

drivers/mssql/internal/backfill.go
```diff
@@ -48,11 +57,14 @@ func (m *MSSQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
 
 		// Build query for the chunk
 		stmt := ""
-		if chunkColumn != "" {
+		switch {
+		case chunkColumn != "":
 			stmt = jdbc.MSSQLChunkScanQuery(stream, []string{chunkColumn}, chunk, filter)
-		} else if len(pkColumns) > 0 {
+		case isPhysLocChunk(chunk):
```
vikaxsh: Image
please once check

drivers/mssql/internal/backfill.go
```diff
@@ -117,168 +130,346 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+	if int64(len(physLocSamples)) < numberOfChunks {
+		return nil, fmt.Errorf("TABLESAMPLE returned %d rows, need at least %d",
+			len(physLocSamples), numberOfChunks)
+	}
+
+	chunks := types.NewSet[types.Chunk]()
+	step := float64(len(physLocSamples)) / float64(numberOfChunks)
+	var prev any = nil
+	for i := int64(0); i < numberOfChunks; i++ {
+		curr := physLocBoundary(utils.HexEncode(physLocSamples[int(float64(i)*step)]))
```
vikaxsh: Image

drivers/mssql/internal/backfill.go
vaibhav-datazip: this can lead to this error, but detecting physloc on the basis of tracking the physloc structure is also not possible (which was implemented earlier) the cleaner solution can be we concat a string something like olake_physloc_chunk and then during chunk iteration scrap that and read directly

drivers/mssql/internal/backfill.go
vaibhav-datazip: had this in mind earlier as well, but such condition was only possible if sampled rows or number of chunks are too high. which in our case wont be that high like in order of 15 - 16 . 

Added this so it looks handled in extreme cases

vikaxsh:
lgtm

drivers/mssql/internal/backfill.go
```diff
@@ -103,182 +109,318 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+				return fmt.Errorf("failed to get table column type: %s", err)
+			}
+		}
+
+		logger.Infof("Stream %s extremes - min: %v, max: %v", stream.ID(),
+			normalizeBoundaryValue(minVal, pkCols, columnType),
+			normalizeBoundaryValue(maxVal, pkCols, columnType),
+		)
+		chunks.Insert(types.Chunk{Min: nil, Max: normalizeBoundaryValue(minVal, pkCols, columnType)})
+
```
vikaxsh: add comment back here

drivers/mssql/internal/backfill.go
```diff
@@ -103,182 +109,318 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
-					break
-				}
-				if err != nil {
-					return fmt.Errorf("failed to get next chunk end: %s", err)
-				}
+// splitViaPrimaryKey divides the table into chunks by walking PK boundaries
+// under REPEATABLE READ (see jdbc.WithIsolation).
+func (m *MSSQL) splitViaPrimaryKey(ctx context.Context, stream types.StreamInterface, pkCols []string, chunkSize int64) (*types.Set[types.Chunk], error) {
+	chunks := types.NewSet[types.Chunk]()
+	err := jdbc.WithIsolation(ctx, m.client, false, func(tx *sql.Tx) error {
```
vikaxsh: why are we taking REPEATABLE READ lock here not read commited

drivers/mssql/internal/backfill.go
```diff
@@ -103,182 +109,318 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
-			// Get the minimum and maximum values for the primary key columns
-			minVal, maxVal, err := m.getTableExtremes(ctx, stream, pkCols, tx)
-			if err != nil {
-				return fmt.Errorf("failed to get table extremes: %s", err)
-			}
-			// Skip if table is empty
-			if minVal == nil {
-				return nil
-			}
+	if m.probeIAMWalkCapability(ctx) {
```
vikaxsh, line 134: why aren't we using iam walk as a primary statergy

vikaxsh (commented):
Could you please add performance benchmark results for each newly introduced chunking strategy, so we can compare the improvements achieved by each approach?

drivers/mssql/internal/backfill.go
vaibhav-datazip: Iam walk needs computation of physloc , similar to what the split via physloc strategy does. as Physloc is not stored by default in the table and needs to be calculated for every row. 

The reason to still keep iam walk and physloc split is because
- iam walk does chunking way faster than what physloc did , in 75 seconds
- physloc split might be required when there is a table with no primary key , and user has not given view database state access.

drivers/mssql/internal/backfill.go
vaibhav-datazip: we don't need to take it now , we can use read committed here as well , changing this as well

vaibhav-datazip:
Ran the tests for fhv_trips chunking 

split via primary key -> 14 mins

new strategies
sampling -> 40 seconds
iam walk -> 75 seconds

drivers/mssql/internal/backfill.go
vikaxsh: Since there is not much difference in chunk creation time between IAM-based chunking and sampling, we should prefer IAM-based chunking whenever the required permissions are available. Otherwise, we can fall back to sampling. IAM-based chunking provides better chunk quality and results in a more even distribution of data.

vaibhav-datazip: no, our goal is to improve overall sync time and not just chunking time, iam based chunking will give you boundary, but still the chunk iterator query will go through all the rows and compute their rowid whichever row falls in our chunk will then be decided for pick or not pick.

vikaxsh: If that's the case, it would be helpful if you could provide complete sync benchmarks for each strategy so we can make an informed decision. This is what I had requested earlier as well. Comparing only the chunk creation time may not be sufficient if the overall sync time is what ultimately matters.

vaibhav-datazip: the db cpu usage for parallel reads will be high in case of iam walk , 
here are the sync stats 

iam walk (prefetch 30k) -> 44629.42 rps (ran with 8 threads) -> 49530.00 seconds
with 30 threads the db went down  after 7720.00 seconds -> 38127.50 rps

sampling strategy (30 threads) -> 7214.00 seconds -> 302976.63 rps

vikaxsh: In the no-primary-key case, if the user configures a high number of threads, the sync may keep failing and could also put significantly higher load on the database. Is my understanding correct here?

vaibhav-datazip: yes, and this would've happened earlier as well, because the previous strategy also depends on this , we will have to add in docs `its advised to use less parallel threads in case when primary key is not present in your database`

vikaxsh (approved):

PR #934, opened 2026-04-30
feat(iceberg): skip equality deletes for cdc inserts

vikaxsh:
# Description

Equality deletes are only required during the backfill→CDC overlap window. Track this with a new `dedup_inserts` flag in the Iceberg `olake_2pc` table property: Java sets it to `true` on backfill commit; Go clears it to `false` after the first successful CDC commit. While the flag is `true`, CDC inserts emit `_op_type="i"` (equality delete + write); otherwise `_op_type="c"` (write only). Applies to both the arrow writer and the legacy gRPC writer.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/BaseDeltaTaskWriter.java
+++ b/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/BaseDeltaTaskWriter.java
@@ -54,8 +54,11 @@ abstract class BaseDeltaTaskWriter extends BaseTaskWriter<Record> {
     if (rowOperation == Operation.DELETE && !keepDeletes) {
       // deletes. doing hard delete. when keepDeletes = FALSE we dont keep deleted record
       writer.deleteKey(keyProjection.wrap(row));
+    } else if ("c".equals(String.valueOf(row.getField("_op_type")))) {
+      // Steady-state CDC insert: no prior committed row exists for this key, skip equality delete.
+      writer.write(row);
     } else {
-      // We are deleting key even for insert operations to avoid duplicate records for handling inserts happening while full-load
+      // "i" (first-CDC overlap) / UPDATE / READ: equality-delete the prior version, then write.
       writer.deleteKey(keyProjection.wrap(row));
       writer.write(row);
     }
--- a/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/BaseDeltaTaskWriter.java
+++ b/destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/BaseDeltaTaskWriter.java
@@ -54,13 +54,13 @@ abstract class BaseDeltaTaskWriter extends BaseTaskWriter<Record> {
     if (rowOperation == Operation.DELETE && !keepDeletes) {
       // deletes. doing hard delete. when keepDeletes = FALSE we dont keep deleted record
       writer.deleteKey(keyProjection.wrap(row));
-    } else if ("c".equals(String.valueOf(row.getField("_op_type")))) {
+    } else if (rowOperation == Operation.CREATE) {
       // Steady-state CDC insert: no prior committed row exists for this key, skip equality delete.
       writer.write(row);
     } else {
       // Phantom read possible: equality-delete before write to evict any prior committed version.
-      // Normalise "i" → "c" in the data file so downstream consumers see a consistent op type.
-      if ("i".equals(String.valueOf(row.getField("_op_type")))) {
+      // Normalise "i" → "c" so downstream sees a consistent op type.
+      if (rowOperation == Operation.INSERT) {
         row.setField("_op_type", "c");
       }
       writer.deleteKey(keyProjection.wrap(row));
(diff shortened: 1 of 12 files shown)


destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/BaseDeltaTaskWriter.java
```diff
@@ -54,8 +54,15 @@ public void write(Record row) throws IOException {
...
+      writer.write(row);
     } else {
-      // We are deleting key even for insert operations to avoid duplicate records for handling inserts happening while full-load
+      // Phantom read possible: equality-delete before write to evict any prior committed version.
+      // Normalise "i" → "c" so downstream sees a consistent op type.
+      if (rowOperation == Operation.INSERT) {
+        row.setField("_op_type", "c");
+      }
       writer.deleteKey(keyProjection.wrap(row));
       writer.write(row);
```
hash-data, line 65:
```suggestion
    } 
    else {
      // Phantom read possible: equality-delete before write to evict any prior committed version.
      // Normalise "i" → "c" so downstream sees a consistent op type.
      if (rowOperation == Operation.INSERT) {
        row.setField("_op_type", "c");
        writer.deleteKey(keyProjection.wrap(row));
      }
      writer.write(row);
    }
```

drivers/abstract/abstract.go
```diff
@@ -259,10 +259,12 @@ func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *er
 	var metadataState any
 	var closeErr error
 	if mtState != nil {
-		metadataState, closeErr = types.SetMetadataState(*mtState, threadID)
-		if closeErr != nil {
-			closeErr = fmt.Errorf("failed to set metadata state: %s", closeErr)
+		ms, setErr := types.SetMetadataState(*mtState, threadID)
+		if setErr != nil {
+			closeErr = fmt.Errorf("failed to set metadata state: %s", setErr)
```
hash-data, line 264: if there is an error we should cancel the context

drivers/abstract/cdc.go
```diff
@@ -122,19 +126,23 @@ func (a *AbstractDriver) streamChanges(mainCtx context.Context, pool *destinatio
...
 
-	defer handleWriterCleanup(cdcCtx, cdcCtxCancel, &err, writers, "", &finalMetadataState)
+	defer handleWriterCleanup(cdcCtx, cdcCtxCancel, &err, writers, "", &finalMetadataState, &clearDedup)
 
 	finalMetadataState, err = a.driver.StreamChanges(cdcCtx, streamIndex, metadataStates, func(ctx context.Context, change CDCChange) error {
 		writer := writers[change.Stream.ID()]
 		olakeColumns := map[string]any{
 			constants.OlakeID:        utils.GetKeysHash(change.Data, change.Stream.GetStream().SourceDefinedPrimaryKey.Array()...),
-			constants.OpType:         mapChangeKindToOperationType(change.Kind),
+			constants.OpType:         mapInsertOpType(change.Kind, dedupInserts[change.Stream.ID()]),
```
hash-data: it is not just insert operation type

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/BaseDeltaTaskWriter.java
vikaxsh: we need this for update as well

drivers/abstract/abstract.go
```diff
@@ -259,10 +259,14 @@ func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *er
...
+			cancel()
+			return
```
hash-data: is it good to return or let driver get closed? commit will not happen fine but server get closed

hash-data (dismissed):

destination/parquet/parquet.go
```diff
@@ -162,6 +162,11 @@ func (p *Parquet) Setup(_ context.Context, stream types.StreamInterface, schema
 func (p *Parquet) Write(_ context.Context, records []types.RawRecord) error {
 	// TODO: use batch writing feature of pq writer
 	for _, record := range records {
+		// Normalise "i" -> "c": Parquet has no equality-delete concept, downstream
```
hash-data: why we need here when we already did in flattening?

destination/parquet/parquet.go
```diff
@@ -162,16 +162,14 @@ func (p *Parquet) Setup(_ context.Context, stream types.StreamInterface, schema
...
 		if opType, ok := record.OlakeColumns[constants.OpType].(string); ok && opType == "i" {
 			record.OlakeColumns[constants.OpType] = "c"
-			// FlattenAndCleanData (normalized path) already merged OlakeColumns into Data,
-			// so fix the Data map as well when the field is present.
-			if record.Data != nil {
-				if _, exists := record.Data[constants.OpType]; exists {
-					record.Data[constants.OpType] = "c"
-				}
+			if v, ok := record.Data[constants.OpType].(string); ok && v == "i" {
+				record.Data[constants.OpType] = "c"
```
hash-data: do you need to check i type here ? as already olake columns says it is i ?

vikaxsh: We can remove it, but it’s better to keep it as an extra safeguard.

hash-data: unnecessary check it looks, better to keep checks simple

hash-data (approved):

PR #940, opened 2026-05-05
fix(mysql): fix CDC charset corruption for utf16/ucs2/latin1 columns

vikaxsh:
# Description

ENUM and string columns with non-UTF8 charsets (utf16, ucs2, latin1) were silently corrupted in CDC by blind []byte → string casts
Added collation-aware decoding in pkg/binlog/filter.go using TableMapEvent.CollationMap() / EnumSetCollationMap() (available MySQL 8.0.1+ with binlog_row_metadata=FULL)

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):


pkg/binlog/filter.go
```diff
@@ -214,7 +243,7 @@ func mysqlTypeName(t byte, unsigned bool) string {
 	case mysql.MYSQL_TYPE_LONG_BLOB:
 		return "LONGBLOB"
 	case mysql.MYSQL_TYPE_STRING:
-		return "STRING"
+		return "CHAR"
```
vaibhav-datazip: will this change require a state versioning check, as we are changing the mapping from string to char ?

pkg/binlog/filter.go
```diff
@@ -263,3 +292,57 @@ func (e *TableMapEvent) isNumericColumn(i int) bool {
...
+var mysqlStringDecoders = map[string]func([]byte) (string, error){
+	"utf8":    decodeRawString,
+	"utf8mb3": decodeRawString,
+	"utf8mb4": decodeRawString,
+	"ascii":   decodeRawString,
+	"latin1":  decodeLatin1,
+	"ucs2":    decodeUTF16BE, // UCS-2 is Big Endian, BMP-only subset of UTF-16
+	"utf16":   decodeUTF16BE,
+	"utf16le": decodeUTF16LE,
+}
```
vaibhav-datazip, line 333: There are some more common character sets 

Image

some of them are more used then our **current added once**

Image

pkg/binlog/filter.go
```diff
@@ -263,3 +292,57 @@ func (e *TableMapEvent) isNumericColumn(i int) bool {
...
+
+func decodeUTF16BE(b []byte) (string, error) {
+	out, err := unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM).NewDecoder().Bytes(b)
+	return string(out), err
+}
+
+func decodeUTF16LE(b []byte) (string, error) {
+	out, err := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewDecoder().Bytes(b)
+	return string(out), err
+}
```
vaibhav-datazip, line 352: these functions can be made cleaner if we implement it like this

```
var mysqlStringDecoders = map[string]func([]byte) (string, error){
	"utf8":    passthrough,
	"utf8mb3": passthrough,
	"utf8mb4": passthrough,
	"ascii":   passthrough,
	"latin1":  makeDecoder(charmap.ISO8859_1),
	"ucs2":    makeDecoder(utf16BEEncoding), // MySQL ucs2: Big Endian, BMP-only subset of UTF-16
	"utf16":   makeDecoder(utf16BEEncoding), // MySQL utf16: always Big Endian
	"utf16le": makeDecoder(utf16LEEncoding),
}

// passthrough returns the raw bytes as a Go string without any transcoding.
// Safe for charsets whose byte representation is already valid UTF-8 (utf8, utf8mb4, ascii).
func passthrough(b []byte) (string, error) {
	return string(b), nil
}

// makeDecoder returns a decoder function that transcodes bytes from enc to UTF-8.
func makeDecoder(enc encoding.Encoding) func([]byte) (string, error) {
	return func(b []byte) (string, error) {
		out, err := enc.NewDecoder().Bytes(b)
		return string(out), err
	}
}
```

pkg/binlog/filter.go
```diff
@@ -263,3 +292,57 @@ func (e *TableMapEvent) isNumericColumn(i int) bool {
...
+	return string(b), nil
+}
+
+func decodeLatin1(b []byte) (string, error) {
+	out, err := charmap.ISO8859_1.NewDecoder().Bytes(b)
+	return string(out), err
+}
+
+func decodeUTF16BE(b []byte) (string, error) {
+	out, err := unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM).NewDecoder().Bytes(b)
```
vaibhav-datazip, line 345: wont it be better to just the value of 

`unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM)

in a variable and directly use it (caching), instead of calling it everytime the function is called.

pkg/binlog/filter.go
vikaxsh: previously there were no mapping of string present in mysqlTypeToDataTypes function

pkg/binlog/filter.go
vikaxsh: will add in future if required in future

pkg/binlog/filter.go
vikaxsh: how is it more cleaner?

pkg/binlog/filter.go
vikaxsh: no need as expected performance improvement is not significant

pkg/binlog/filter.go
vaibhav-datazip: it reduces your 4 repetitive functions to just 2 , and in future also when you add more charsets new functions wont be required.

vikaxsh: They are not repetitive functions, each one represents different logic. In the future, if we need to support different charsets, each may require its own handling approach. This design also keeps the code cleaner and more readable and also expandable in the future

pkg/binlog/filter.go
vaibhav-datazip: yeah but it was being assigned so value that was `string` right ?

pkg/binlog/filter.go
```diff
@@ -118,24 +122,49 @@ func convertRowToMap(row []interface{}, tableMap *replication.TableMapEvent, col
 		return nil, fmt.Errorf("column count mismatch: expected %d, got %d", len(columns), len(row))
 	}
 
-	enumMap := tableMap.EnumStrValueMap()
+	enumRaw := tableMap.EnumStrValue                   // [][][]byte: one entry per ENUM column
+	enumCollationMap := tableMap.EnumSetCollationMap() // col idx -> collation ID for ENUM/SET
```
vaibhav-datazip: this was the table I created and tried syncing it 

```
CREATE TABLE olake_charset_set_cdc_test (
  id BIGINT NOT NULL AUTO_INCREMENT,
  loyalty_opt_in ENUM('YES','NO','AWAITING_CUSTOMER_INPUT')
    CHARACTER SET utf16 COLLATE utf16_bin DEFAULT NULL,
  tags_utf16 SET('sports','music','gaming','reading')
    CHARACTER SET utf16 COLLATE utf16_bin DEFAULT NULL,
  tags_utf8 SET('red','green','blue','yellow')
    CHARACTER SET utf8mb4 COLLATE utf8mb4_bin DEFAULT NULL,
  customer_name VARCHAR(100)
    CHARACTER SET utf16 COLLATE utf16_bin DEFAULT NULL,
  note_latin1 VARCHAR(100)
    CHARACTER SET latin1 COLLATE latin1_swedish_ci DEFAULT NULL,
  created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
  updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  PRIMARY KEY (id)
) ENGINE=InnoDB DEFAULT CHARSET=utf16 COLLATE=utf16_bin;
```

then inserted a row and synced using cdc

```
INSERT INTO olake_charset_set_cdc_test
  (loyalty_opt_in, tags_utf16,        tags_utf8,    customer_name, note_latin1)
VALUES
  ('YES',          'sports,reading',  'red,yellow', 'Diana',       'naïve');
```

the set values are acquiring integer values 

Screenshot 2026-05-07 at 7 37 10 PM

pkg/binlog/filter.go
```diff
@@ -123,9 +124,11 @@ func convertRowToMap(row []interface{}, tableMap *replication.TableMapEvent, col
 	}
 
 	enumRaw := tableMap.EnumStrValue                   // [][][]byte: one entry per ENUM column
+	setRaw := tableMap.SetStrValue                     // [][][]byte: one entry per SET column
 	enumCollationMap := tableMap.EnumSetCollationMap() // col idx -> collation ID for ENUM/SET
```
vaibhav-datazip: Shouldn't we also rename this variable to enumSetCollationMap

vaibhav-datazip (approved):

PR #946, opened 2026-05-08
feat: wire SIGINT/SIGTERM into root context for graceful shutdown

### LINKED ISSUE #926: improvement: wire SIGINT/SIGTERM into root context for graceful shutdown
## Problem

OLake does not wire OS signals (`SIGINT` / `SIGTERM`) into its root `context.Context`. When the operating system delivers a clean-shutdown signal — Kubernetes pod eviction, Karpenter node consolidation, `docker stop`, or Ctrl-C in a dev shell — the process exits immediately, without giving the CDC / backfill / destination-writer paths any chance to reach their existing `ctx.Done()` branches.

### Evidence

Searching the tree for any signal handling:

```
$ git grep -nE 'signal\.Notify|os\.Interrupt|syscall\.SIG(INT|TERM|HUP)' -- ':!LICENCE'
(no matches)
```

Confirmed on current `master` (`v0.6.5`).

The root command is built from cobra's default background context:

- `protocol/root.go` declares `RootCmd = &cobra.Command{...}` with no `SetContext` / `ExecuteContext` anywhere, so `cmd.Context()` inside every subcommand is `context.Background()` that never cancels on signals.
- `connector.go:18` — `protocol.CreateRootCommand(true, driver).Execute()` — uses plain `Execute()` (not `ExecuteContext`).
- `protocol/root.go:85` — `connector = abstract.NewAbstractDriver(RootCmd.Context(), ...)` — so `AbstractDriver.GlobalCtxGroup` / `GlobalConnGroup` are also rooted on a background context.

Meanwhile, the CDC / read paths already branch on `ctx.Done()` and would unwind cleanly if the context were cancellable:

- `pkg/waljs/pgoutput.go:47-108` — the pgoutput receive loop has `case <-ctx.Done(): return nil` as its top-level exit, and `ReceiveMessage` is called with a `context.WithTimeout(ctx, …)` child so it also honours parent cancellation.
- Backfill / incremental / CDC dispatchers in `drivers/abstract/{backfill,cdc,incremental}.go` all consume `context.Context` arguments derived from `mainCtx` and thread them through to each driver call.
- Concurrency primitives in `utils/cxgroup.go` (`CxGroup`) are constructed with a parent context that is currently always background, but would propagate cancellation automatically if that parent were signal-aware.

So the machinery to react to cancellation is already there — only the signal wiring is missing.

### Impact (Kubernetes / production)

We run OLake as a Kubernetes CronJob for Postgres → Iceberg CDC. Because `terminationGracePeriodSeconds` SIGTERMs pods on routine events — Karpenter node consolidation, image rollouts, cluster autoscaling, preemption of spot instances — every such event currently:

- kills the process mid-read without letting the pgoutput loop return through its `ctx.Done()` branch,
- gives no chance to send a final `AcknowledgeLSN` / standby-status update back to the Postgres replication slot,
- releases no DB / destination connections through their normal `defer` paths,
- eventually escalates to SIGKILL once the grace period elapses.

After pod restart this tends to mean reprocessing the last in-flight WAL segment and leaving the replication slot holding WAL longer than it needs to. Nothing catastrophic, but it's extra work and extra surface area on every consolidation event, and it's entirely avoidable.

## Proposed solution

Wrap the root context with `signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)` *before* `abstract.NewAbstractDriver` captures it, and defer `stop()` so the handler is deregistered cleanly. Roughly:

```go
// protocol/root.go
func CreateRootCommand(_ bool, driver any) *cobra.Command {
    RootCmd.AddCommand(commands...)

    ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
    RootCmd.SetContext(ctx)
    RootCmd.PersistentPostRunE = func(*cobra.Command, []string) error { stop(); return nil }

    connector = abstract.NewAbstractDriver(ctx, driver.(abstract.DriverInterface))
    return RootCmd
}
```

This is a single-location change: every existing `cmd.Context()` / `RootCmd.Context()` call-site (≈10 callers in `protocol/*.go` plus the `AbstractDriver.GlobalCtxGroup` / `GlobalConnGroup`) automatically inherits the signal-aware context. No other edits needed for the propagation to work.

Semantics of `signal.NotifyContext`:

- First SIGINT/SIGTERM cancels the context (callers see `ctx.Err() == context.Canceled`).
- Once fired, the handler is torn down, so a second signal falls through to the Go runtime default (terminate) — operators who lose patience and hit Ctrl-C twice still get the hard-exit behaviour they expect.

### Honest scoping note

Signal wiring alone is **not** full graceful shutdown — it is the prerequisite. After this change, `ctx.Done()` would fire throughout the sync, but several cleanup paths currently treat cancellation as "abort, don't commit" rather than "flush, then exit":

- `destination/iceberg/iceberg.go:205-210` — the Iceberg writer's `Close()` deliberately skips `writer.Close(ctx, finalMetadataState)` when `ctx.Done()` has fired and returns `ctx.Err()` instead.
- `drivers/abstract/cdc.go:107-108,131` — `PostCDC` and `handleWriterCleanup` are deferred but run with `cdcCtx`, which is derived from `mainCtx` and therefore already cancelled by the time the defers run.

A fully-graceful shutdown would probably want cleanup defers to use a fresh (or short-timeout) background context so commits / state flushes can still complete after cancellation. That is a larger design decision and feels like a second step on top of this issue rather than part of it — happy to scope that into a follow-up if the approach here is acceptable.

### Happy to raise a PR

I've searched open + closed issues and PRs on `datazip-inc/olake` for `SIGTERM` / `signal` / `graceful` / `shutdown` / `NotifyContext` — nothing prior. Happy to open a PR implementing the sketch above if maintainers agree on:

1. the location of the wiring (`CreateRootCommand` vs. `connector.RegisterDriver` via `ExecuteContext`), and
2. whether to include the destination-writer cleanup-context change in the same PR or keep it as a follow-up.

Jonathan-7:
# Description

Wires SIGINT/SIGTERM into the cobra root context so the CDC, backfill and destination-writer paths reach their existing `ctx.Done()` branches on Kubernetes pod eviction, `docker stop` or Ctrl-C, instead of being killed mid-read.

Today there is no signal handling on master — `git grep -nE 'signal\.Notify|os\.Interrupt|syscall\.SIG(INT|TERM|HUP)'` returns no matches. The cancellation machinery downstream is already in place:

- `pkg/waljs/pgoutput.go` — pgoutput receive loop already has `case <-ctx.Done(): return nil`.
- `drivers/abstract/{backfill,cdc,incremental}.go` — dispatchers already accept and propagate `context.Context`.
- `utils/cxgroup.go` — `CxGroup` propagates cancellation automatically once its parent is signal-aware.

Only the root-level signal wiring was missing.

`protocol.CreateRootCommand` now wraps the cobra root context with `signal.NotifyContext(ctx, SIGINT, SIGTERM)` before `abstract.NewAbstractDriver` captures it, and calls `RootCmd.SetContext(ctx)` so every subcommand inherits a signal-aware context. The wiring is split into a small testable helper, `signalAwareRootContext`. A small watcher goroutine calls `stop()` after the first cancellation so a subsequent SIGINT/SIGTERM falls through to the Go runtime default (terminate), matching the behavior an operator hitting Ctrl-C twice expects.

```go
func CreateRootCommand(_ bool, driver any) *cobra.Command {
    RootCmd.AddCommand(commands...)
    ctx := signalAwareRootContext(RootCmd.Context())
    RootCmd.SetContext(ctx)
    connector = abstract.NewAbstractDriver(ctx, driver.(abstract.DriverInterface))
    return RootCmd
}

func signalAwareRootContext(parent context.Context) context.Context {
    ctx, stop := signal.NotifyContext(parent, syscall.SIGINT, syscall.SIGTERM)
    go func() { <-ctx.Done(); stop() }() // restore default after first cancellation
    return ctx
}
```

**Scope** — signal wiring alone is not full graceful shutdown; it is the prerequisite. After this change, source/destination cancellation safety remains a per-driver/per-writer responsibility: each `PostCDC` and each destination `writer.Close` gates its final commit on `ctx.Done()`. Tested across drivers and destinations during review: Kafka + Parquet behaves consistently (offsets not committed, partial parquet files removed on cancel, retry reprocesses cleanly). For Postgres, the in-flight keepalive in `pkg/waljs/pgoutput.go` acknowledges the slot independently of the defer chain, so a SIGTERM landing between a keepalive and PostCDC can leave the slot ahead of `state.json`. This race exists with or without this PR's signal handling, worth its own follow-up ticket.

Fixes #926

## Type of change

# How Has This Been Tested?

- [x] Unit tests in `protocol/root_test.go` — `TestSignalAwareRootContextCancelsOnSignal` (table-driven, re-execs a helper child per signal case covering both SIGINT and SIGTERM) and `TestSignalAwareRootContextPreservesParentCancellation` (asserts parent-cancel propagation through the wrapper).
- [x] Manual: built the Postgres driver image locally and ran `sync` against a disposable Postgres with logical replication and a CDC-generating writer. `docker stop -t 30` on the running container produced a non-`143` exit and the process reached the deferred cleanup path rather than being killed mid-read.

`go build ./...`, `go vet ./...`, `gofmt`, and `golangci-lint` are clean.

# Screenshots or Recordings
N/A — no UI/output changes.

## Documentation

- [x] N/A (internal signal-handling improvement, no user-facing surface change)

## Related PR's (If Any):
None.

### DIFF
--- a/protocol/root.go
+++ b/protocol/root.go
@@ -1,9 +1,12 @@
 package protocol
 
 import (
+	"context"
 	"fmt"
 	"os"
+	"os/signal"
 	"path/filepath"
+	"syscall"
 
 	"github.com/datazip-inc/olake/constants"
 	"github.com/datazip-inc/olake/drivers/abstract"
@@ -80,11 +83,36 @@ var RootCmd = &cobra.Command{
 	},
 }
 
+// CreateRootCommand wires the cobra root for the given driver. It mutates
+// package-level state (RootCmd, connector) and installs a process-wide signal
+// handler, so it must be called at most once per process — the existing
+// connector.RegisterDriver entry point already enforces this.
 func CreateRootCommand(_ bool, driver any) *cobra.Command {
 	RootCmd.AddCommand(commands...)
-	connector = abstract.NewAbstractDriver(RootCmd.Context(), driver.(abstract.DriverInterface))
+
+	// Wire SIGINT/SIGTERM into the root context so CDC, backfill and
+	// destination-writer paths reach their existing ctx.Done() branches on
+	// pod eviction, docker stop, or Ctrl-C, instead of being killed mid-read.
+	ctx := signalAwareRootContext(RootCmd.Context())
+	RootCmd.SetContext(ctx)
+
+	connector = abstract.NewAbstractDriver(ctx, driver.(abstract.DriverInterface))
 
 	return RootCmd
+}
+
+func signalAwareRootContext(parent context.Context) context.Context {
+	ctx, stop := signal.NotifyContext(parent, syscall.SIGINT, syscall.SIGTERM)
+	// signal.NotifyContext keeps the signal handler installed until stop() is
+	// called. Releasing it after the first cancellation lets a subsequent
+	// SIGINT/SIGTERM fall through to the Go runtime default (terminate), which
+	// is the behavior an operator hitting Ctrl-C twice expects.
+	go func() {
+		<-ctx.Done()
+		stop()
+	}()
+
+	return ctx
 }
 
 func init() {
--- a/protocol/root.go
+++ b/protocol/root.go
@@ -101,6 +101,23 @@ func CreateRootCommand(_ bool, driver any) *cobra.Command {
 	return RootCmd
 }
 
+// signalAwareRootContext wraps parent so that the returned context cancels on
+// SIGINT / SIGTERM as well as on any parent cancellation. Used to wire pod
+// eviction, docker stop, and Ctrl-C through to the existing ctx.Done()
+// branches in CDC, backfill, and destination-writer paths.
+//
+// Source / destination consistency on cancel is still owned by each
+// driver.PostCDC and destination writer.Close implementation. This wrapper only
+// makes process signals visible through ctx.Done(); it does not make source
+// checkpoints and destination commits atomic. Any implementation that performs
+// a final commit after work has been written must continue to check ctx.Done()
+// before that commit and must treat a canceled context as a reason to avoid
+// advancing only one side of the source/destination boundary.
+//
+// The Kafka driver has a separate `TODO: Add 2PC support for Kafka` for a
+// future stricter contract. Other drivers may have similar source-specific
+// checkpointing constraints, so this helper should not be used as a substitute
+// for driver-level cancellation safety.
 func signalAwareRootContext(parent context.Context) context.Context {
 	ctx, stop := signal.NotifyContext(parent, syscall.SIGINT, syscall.SIGTERM)
 	// signal.NotifyContext keeps the signal handler installed until stop() is
--- a/protocol/root_test.go
+++ b/protocol/root_test.go
@@ -1,0 +1,59 @@
+package protocol
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"runtime"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func TestSignalAwareRootContextCancelsOnSIGTERM(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("process signal behavior differs on windows")
+	}
+
+	if os.Getenv("OLAKE_SIGNAL_CONTEXT_HELPER") == "1" {
+		ctx := signalAwareRootContext(context.Background())
+		proc, err := os.FindProcess(os.Getpid())
+		if err != nil {
+			t.Fatal(err)
+		}
+		if err := proc.Signal(syscall.SIGTERM); err != nil {
+			t.Fatal(err)
+		}
+
+		select {
+		case <-ctx.Done():
+			if ctx.Err() != context.Canceled {
+				t.Fatalf("expected canceled context, got %v", ctx.Err())
+			}
+			return
+		case <-time.After(time.Second):
+			t.Fatal("context was not canceled after SIGTERM")
+		}
+	}
+
+	cmd := exec.Command(os.Args[0], "-test.run=TestSignalAwareRootContextCancelsOnSIGTERM")
+	cmd.Env = append(os.Environ(), "OLAKE_SIGNAL_CONTEXT_HELPER=1")
+	if output, err := cmd.CombinedOutput(); err != nil {
+		t.Fatalf("signal helper failed: %v\n%s", err, output)
+	}
+}
+
+func TestSignalAwareRootContextPreservesParentCancellation(t *testing.T) {
+	parent, cancel := context.WithCancel(context.Background())
+	ctx := signalAwareRootContext(parent)
+	cancel()
+
+	select {
+	case <-ctx.Done():
+		if ctx.Err() != context.Canceled {
+			t.Fatalf("expected canceled context, got %v", ctx.Err())
+		}
+	case <-time.After(time.Second):
+		t.Fatal("context was not canceled after parent cancellation")
+	}
+}
--- a/protocol/root_test.go
+++ b/protocol/root_test.go
@@ -10,18 +10,59 @@ import (
 	"time"
 )
 
-func TestSignalAwareRootContextCancelsOnSIGTERM(t *testing.T) {
+// TestSignalAwareRootContextCancelsOnSignal verifies that signalAwareRootContext
+// cancels the returned context when the process receives SIGINT or SIGTERM.
+//
+// The test re-execs itself in a child process per signal case so that we can
+// safely deliver a real OS signal without affecting the parent test runner.
+// The OLAKE_SIGNAL_CONTEXT_HELPER env var switches the binary into "helper"
+// mode; OLAKE_TEST_SIGNAL selects which signal the helper sends to itself.
+func TestSignalAwareRootContextCancelsOnSignal(t *testing.T) {
 	if runtime.GOOS == "windows" {
 		t.Skip("process signal behavior differs on windows")
 	}
 
+	tests := []struct {
+		name   string
+		signal syscall.Signal
+		env    string
+	}{
+		{
+			name:   "SIGTERM cancellation",
+			signal: syscall.SIGTERM,
+			env:    "SIGTERM",
+		},
+		{
+			name:   "SIGINT cancellation",
+			signal: syscall.SIGINT,
+			env:    "SIGINT",
+		},
+	}
+
+	// Helper-mode branch: runs inside the re-execed child. Installs the
+	// signal handler under test, sends the chosen signal to ourselves, and
+	// asserts the context cancels with context.Canceled within a bounded
+	// timeout.
 	if os.Getenv("OLAKE_SIGNAL_CONTEXT_HELPER") == "1" {
 		ctx := signalAwareRootContext(context.Background())
-		proc, err := os.FindProcess(os.Getpid())
+
+		var signal syscall.Signal
+
+		switch os.Getenv("OLAKE_TEST_SIGNAL") {
+		case "SIGTERM":
+			signal = syscall.SIGTERM
+		case "SIGINT":
+			signal = syscall.SIGINT
+		default:
+			t.Fatal("unknown test signal")
+		}
+
+		currentProcess, err := os.FindProcess(os.Getpid())
 		if err != nil {
 			t.Fatal(err)
 		}
-		if err := proc.Signal(syscall.SIGTERM); err != nil {
+
+		if err := currentProcess.Signal(signal); err != nil {
 			t.Fatal(err)
 		}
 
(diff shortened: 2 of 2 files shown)


saksham-datazip:
Hey, please change the base branch of this PR from master to staging.
image

saksham-datazip:
Please update the PR description using the repository PR template instead of a custom format. Include the issue reference, checklist updates, testing details, and full change summary.

For reference, you can check existing PRs or follow:
https://olake.io/docs/community/issues-and-prs/

nayanj98:
@Jonathan-7 Can you also upload a short video of you showing your code fixes and demonstrating with your changes the olake sync happens successfully. It really helps for the reviewer to also understand how you have implemented your changes and most importantly to prevent heavily AI generated PR's.

Jonathan-7:
Hey @nayanj98 on the AI concern: I'm running OLake as a Kubernetes CronJob and graceful shutdown was something I needed for my own deployment. I built the Postgres driver image locally and tested the SIGTERM path against a real Postgres with CDC traffic. The `docker stop -t 30` recipe in the body is what I ran.

88 lines across two files. Recording a video is more work than just reading the diff. Happy to answer specific questions about the implementation here. If a video is a hard requirement close it, no hard feelings.

protocol/root.go
```diff
@@ -80,13 +83,38 @@ var RootCmd = &cobra.Command{
...
+// handler, so it must be called at most once per process — the existing
+// connector.RegisterDriver entry point already enforces this.
 func CreateRootCommand(_ bool, driver any) *cobra.Command {
 	RootCmd.AddCommand(commands...)
-	connector = abstract.NewAbstractDriver(RootCmd.Context(), driver.(abstract.DriverInterface))
+
+	// Wire SIGINT/SIGTERM into the root context so CDC, backfill and
+	// destination-writer paths reach their existing ctx.Done() branches on
+	// pod eviction, docker stop, or Ctrl-C, instead of being killed mid-read.
+	ctx := signalAwareRootContext(RootCmd.Context())
```
saksham-datazip, line 96: Due to this change, whenever Kafka receives a Ctrl+C, the deferred cleanup functions such as handleWriterCleanup and postCDC will be executed during shutdown.

However, inside PostCDC we currently have:

select {
case <-ctx.Done():
    return ctx.Err()
}

Because of this, once the shutdown signal cancels the context, PostCDC exits immediately without committing offsets to the consumer group while handleWriterCleanup would be executed successfully. As a result, in next sync it will read same data again causing data duplicacy.
For reference:
/Users/apple/Desktop/olake/drivers/kafka/internal/cdc.go

Steps to recreate :-
1->Run a sync in kafka 
2->Than, use ctrl+ c in between the sync
3->Than run again in same CG using state.json 
4->U will see duplicate data in some of the selected tables

protocol/root.go
```diff
@@ -80,13 +83,38 @@ var RootCmd = &cobra.Command{
...
+	RootCmd.SetContext(ctx)
+
```
saksham-datazip, line 98: I checked and confirmed this issue on Kafka in parquet format . Please do check for other drivers as well and both iceberg, Parquent format.(Check for its effect on downstream for once)

protocol/root_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+	"os"
+	"os/exec"
+	"runtime"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func TestSignalAwareRootContextCancelsOnSIGTERM(t *testing.T) {
```
saksham-datazip: Can u change a code to something like this as it tests both sigint and sigterm and also follow olake architecture of unit_tests.
```
func TestSignalAwareRootContextCancelsOnSignal(t *testing.T) {
	if runtime.GOOS == "windows" {
		t.Skip("process signal behavior differs on windows")
	}

	tests := []struct {
		name   string
		signal syscall.Signal
		env    string
	}{
		{
			name:   "SIGTERM cancellation",
			signal: syscall.SIGTERM,
			env:    "SIGTERM",
		},
		{
			name:   "SIGINT cancellation",
			signal: syscall.SIGINT,
			env:    "SIGINT",
		},
	}

	if os.Getenv("OLAKE_SIGNAL_CONTEXT_HELPER") == "1" {
		ctx := signalAwareRootContext(context.Background())

		var signal syscall.Signal

		switch os.Getenv("OLAKE_TEST_SIGNAL") {
		case "SIGTERM":
			signal = syscall.SIGTERM
		case "SIGINT":
			signal = syscall.SIGINT
		default:
			t.Fatal("unknown test signal")
		}

		currentProcess, err := os.FindProcess(os.Getpid())
		if err != nil {
			t.Fatal(err)
		}

		if err := currentProcess.Signal(signal); err != nil {
			t.Fatal(err)
		}

		select {
		case <-ctx.Done():
			if ctx.Err() != context.Canceled {
				t.Fatalf("expected canceled context, got %v", ctx.Err())
			}
		case <-time.After(time.Second):
			t.Fatalf("context was not canceled after %v", signal)
		}

		return
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			cmd := exec.Command(os.Args[0], "-test.run=TestSignalAwareRootContextCancelsOnSignal")
			cmd.Env = append(os.Environ(), "OLAKE_SIGNAL_CONTEXT_HELPER=1", "OLAKE_TEST_SIGNAL="+tt.env)

			if output, err := cmd.CombinedOutput(); err != nil {
				t.Fatalf("signal helper failed: %v\n%s", err, output)
			}
		})
	}
}

File: protocol/root_test.go [hunk 4]
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"context"
+	"os"
+	"os/exec"
+	"runtime"
+	"syscall"
+	"testing"
+	"time"
+)
+
```
saksham-datazip, line 12: Can u add a comment to this function ,of what it does?

protocol/root_test.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+	}
+
+	cmd := exec.Command(os.Args[0], "-test.run=TestSignalAwareRootContextCancelsOnSIGTERM")
+	cmd.Env = append(os.Environ(), "OLAKE_SIGNAL_CONTEXT_HELPER=1")
+	if output, err := cmd.CombinedOutput(); err != nil {
+		t.Fatalf("signal helper failed: %v\n%s", err, output)
+	}
+}
+
```
saksham-datazip, line 94: Add a comment here as well please

saksham-datazip:
@Jonathan-7 Could you please help resolve the review comments within the next 1–2 days? We internally need this PR; otherwise, we may have to address the comments on our end.

protocol/root.go
vikaxsh, line 96: we should move this to connector.go,  it is the process entry point. Every driver calls RegisterDriver exactly once. It already owns process-level concerns, Signal handling is equally a process-level concern, not a CLI-flag concern.

Jonathan-7:
Hey @saksham-datazip, took a closer look at the comments.

On the Kafka duplicate-data point, I couldn't reproduce it. Ran against a real Kafka 3.9.1 broker with about half a million records, killed the pod mid-flush, re-ran. Destination cleanup ran under cancellation, the partial parquet file got removed, PostCDC saw `context canceled` and didn't commit offsets. Retry reprocessed cleanly from the prior committed offset, no duplicates in the destination. Code-wise it lines up: Kafka's PostCDC short-circuits, parquet's `closePqFiles` deletes local files when `ctx.Err() != nil`.

One issue I did find on Postgres though: if SIGTERM lands right after a pgoutput keepalive, the slot advances server-side while `state.json` stays at the prior PostCDC checkpoint, and retry fails with `lsn mismatch`. The keepalive ack at `pkg/waljs/pgoutput.go:97-101` goes out on the wire independently of whether this PR's signal handling runs, so the race is the same with or without this change. Worth its own ticket.

In response I've pushed a table-driven test covering both SIGINT and SIGTERM as you suggested, comments on both test functions, and a corrected docstring on `signalAwareRootContext` that drops the 2PC claim, now scoped down to "wrapper propagates cancellation, source/destination consistency stays a per-driver concern".

For @vikaxsh on the `connector.go` relocation, reasonable suggestion, but leaning toward keeping it in `protocol/root.go` for this PR since the wrap happens right before the context is consumed there.

protocol/root_test.go
```diff
@@ -0,0 +1,112 @@
...
+			if output, err := cmd.CombinedOutput(); err != nil {
+				t.Fatalf("signal helper failed: %v\n%s", err, output)
+			}
+		})
+	}
+}
+
+// TestSignalAwareRootContextPreservesParentCancellation verifies that
+// cancelling the parent context still propagates through the signal-aware
```
hash-data:
```suggestion
// canceling the parent context still propagates through the signal-aware
```

vikaxsh (approved):

PR #949, opened 2026-05-12
feat: add support for mssql read replica

vishalm0509:
# Description

Adds optional jdbc_url_params to the MSSQL source config (e.g. read intent), detects Always On read secondaries after connect, and adjusts CDC so replica-safe paths are used (skip primary-only agent/msdb logic and skip capture-instance management on replicas). Spec and tests updated accordingly.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Scenario A
- Simulated a SQL Server **primary + readable secondary** using an Always On
Availability Group with `CLUSTER_TYPE = NONE` and ran OLake with secondary instance

- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/drivers/mssql/internal/config_test.go
+++ b/drivers/mssql/internal/config_test.go
@@ -181,6 +181,7 @@ func TestConfig_URI(t *testing.T) {
 				}
 			}(),
 			expectedContains: []string{
+				"mssql-host:1433",
 				"ApplicationIntent=ReadOnly",
 				"MultiSubnetFailover=true",
 			},
@@ -198,6 +199,7 @@ func TestConfig_URI(t *testing.T) {
 				},
 			},
 			expectedContains: []string{
+				"mssql-host:1433",
 				"database=realdb",
 			},
 			notExpected: []string{
@@ -218,6 +220,7 @@ func TestConfig_URI(t *testing.T) {
 				SSLConfiguration: &utils.SSLConfig{Mode: utils.SSLModeDisable},
 			},
 			expectedContains: []string{
+				"mssql-host:1433",
 				"encrypt=disable",
 			},
 		},
@@ -232,6 +235,7 @@ func TestConfig_URI(t *testing.T) {
 				SSLConfiguration: &utils.SSLConfig{Mode: utils.SSLModeRequire},
 			},
 			expectedContains: []string{
+				"mssql-host:1433",
 				"encrypt=true",
 				"TrustServerCertificate=true",
 			},
--- a/drivers/mssql/internal/config_test.go
+++ b/drivers/mssql/internal/config_test.go
@@ -150,7 +150,7 @@ func TestConfig_Validate(t *testing.T) {
 	}
 }
 
-func TestConfig_ConnectionString(t *testing.T) {
+func TestConfig_URI(t *testing.T) {
 	tests := []struct {
 		name             string
 		config           *Config
@@ -184,9 +184,24 @@ func TestConfig_ConnectionString(t *testing.T) {
 				"ApplicationIntent=ReadOnly",
 				"MultiSubnetFailover=true",
 			},
+		},
+		{
+			name: "config database overrides jdbc_url_params database",
+			config: &Config{
+				Host:     "mssql-host",
+				Port:     1433,
+				Database: "realdb",
+				Username: "sa",
+				Password: "Password!123",
+				JDBCURLParams: map[string]string{
+					"database": "wrongdb",
+				},
+			},
+			expectedContains: []string{
+				"database=realdb",
+			},
 			notExpected: []string{
-				"invalid-ssl-config:0",
-				"tls=skip-verify",
+				"database=wrongdb",
 			},
 		},
 		{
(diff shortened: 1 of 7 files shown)


drivers/mssql/internal/config_test.go
```diff
@@ -148,6 +150,94 @@ func TestConfig_Validate(t *testing.T) {
 	}
 }
 
+func TestConfig_ConnectionString(t *testing.T) {
+	tests := []struct {
+		name             string
+		config           *Config
+		expectError      bool
+		expectedContains []string
+		notExpected      []string
```
vaibhav-datazip: you don't need to add unit test if a function is not present in config.go

drivers/mssql/internal/mssql.go
```diff
@@ -125,6 +144,12 @@ func (m *MSSQL) buildConnectionString() string {
 	query := url.Values{}
 	query.Add("database", m.config.Database)
 
+	if len(m.config.JDBCURLParams) > 0 {
+		for k, v := range m.config.JDBCURLParams {
+			query.Add(k, v)
+		}
+	}
+
 	// Set encrypt parameter based on SSL configuration.
```
vaibhav-datazip: Why are we not building the connection uri or string  in config.go like we are doing in other drivers like mysql , oracle etc ? 

that will give remove this jdbc url params parsing from mssql.go file as well and keep the structure consistent

drivers/mssql/resources/spec.json
```diff
@@ -59,6 +59,14 @@
                 "mode"
             ]
         },
+        "jdbc_url_params": {
+            "type": "object",
+            "title": "JDBC URL Parameters",
+            "description": "Additional connection parameters passed to the go-mssqldb driver (e.g. {\"ApplicationIntent\": \"ReadOnly\"} to route to a readable Always On secondary replica)",
+            "additionalProperties": {
+                "type": "string"
```
vaibhav-datazip: also update ui schema

drivers/mssql/internal/mssql.go
```diff
@@ -31,6 +31,7 @@ type MSSQL struct {
 	lsnMap       sync.Map
 	streams      []types.StreamInterface
 	cdcSupported bool
+	readReplica  bool
```
vaibhav-datazip: readReplicaSupported ?

drivers/mssql/internal/mssql.go
```diff
@@ -125,6 +144,12 @@ func (m *MSSQL) buildConnectionString() string {
 	query := url.Values{}
 	query.Add("database", m.config.Database)
 
+	if len(m.config.JDBCURLParams) > 0 {
+		for k, v := range m.config.JDBCURLParams {
+			query.Add(k, v)
```
vaibhav-datazip: shouldn't we use query.set instead of query.add to avoid passing duplicate values , like encrypt can be passed from here or some other parameter

pkg/jdbc/jdbc.go
```diff
@@ -669,6 +669,18 @@ func MSSQLCDCSupportQuery() string {
...
 
+// MSSQLIsReadReplicaQuery returns SQL that yields whether the current connection
+// is to a read-only secondary replica.
+func MSSQLIsReadReplicaQuery() string {
+	return `SELECT CAST(CASE
+		WHEN EXISTS (
+			SELECT 1 FROM sys.dm_hadr_database_replica_states
+			WHERE is_local = 1 AND is_primary_replica = 0
+		) THEN 1
+		ELSE 0
```
vaibhav-datazip, line 680: this query might give us false results, eg.

If you are connected to Database A (which is a Primary/Read-Write database), but Database B on the exact same server instance is part of a different Availability Group and is acting as a Secondary/Read-Only replica, the query will find Database B.

The EXISTS clause will evaluate to true, and your query will return 1 (BIT True). It will incorrectly tell your application that it is connected to a read replica, even though Database A is fully writable


maybe you can try something like `WHERE is_local = 1 AND is_primary_replica = 0 AND database_id = DB_ID()`

drivers/mssql/internal/mssql.go
vishalm0509: Updated

drivers/mssql/resources/spec.json
vishalm0509: Done

drivers/mssql/internal/mssql.go
vishalm0509: I have named it `isReadReplica` , gtg ??

drivers/mssql/internal/mssql.go
vishalm0509: I have followed other driver's pattern. Should I change ?

drivers/mssql/internal/config_test.go
```diff
@@ -148,6 +150,124 @@ func TestConfig_Validate(t *testing.T) {
...
+			}
+			if err != nil {
+				t.Errorf("failed to validate config: %v", err)
+				return
+			}
+
+			connStr := tt.config.URI()
+
+			for _, expected := range tt.expectedContains {
+				if !strings.Contains(connStr, expected) {
```
vaibhav-datazip, line 262: in the unit tests here shouldn't we verify the whole string and not just the expected array things , also add port number changes , etc this will help us to assert that db username password and other things also added in uri correctly ?

vaibhav-datazip (approved):

vaibhav-datazip (approved):

PR #958, opened 2026-05-20
feat: kafka driver change

### LINKED ISSUE #794: kafka retries are not supported yet (due to rebalancing while creating new reader)
https://github.com/datazip-inc/olake/blob/f7e34a75f640b47400f372b05692ed454da02080/drivers/kafka/internal/kafka.go#L65-L69

---

###### This issue was generated by [todo-issue](https://github.com/DerJuulsn/todo-issue) based on a `TODO` comment in f7e34a75f640b47400f372b05692ed454da02080. It&#x27;s been assigned to @hash-data because they committed the code.

saksham-datazip:
# Description

Migrated the Kafka consumer implementation from Segment Kafka-Go to Franz-Go to leverage improved consumer group management, rebalance handling, and lifecycle APIs provided by Franz-Go.

As part of this migration:

- Added support for static membership using `instance.id` to improve retry and reconnect behavior. This helps avoid unnecessary rebalances during transient failures or consumer restarts when the same instance rejoins the group.

- Implemented rebalance detection using Franz-Go consumer group callbacks along with generation ID tracking stored in consumer metadata. During sync execution, the active generation is continuously validated against the latest assigned generation to detect stale consumers or lost partition ownership.

- Added graceful shutdown handling on successful rebalance detection. Instead of continuing to process records with outdated assignments, the sync now exits cleanly to prevent duplicate processing and stale partition consumption during consumer group transitions.

Fixes #794 

## Type of change

# How Has This Been Tested?

- [x] Tested retry scenarios by intentionally returning errors from `StreamChanges`. Verified that retries worked correctly and did not trigger unnecessary consumer group rebalances due to static membership support using `instance.id`.
- [x] Stopped the sync during active processing and triggered a consumer group rebalance. Verified that the sync detected the rebalance successfully and exited gracefully without processing duplicate data.

# Screenshots or Recordings
N/A

## Documentation

## Related PR's (If Any):


drivers/db2/go.mod
ImDoubD-datazip, line 1: write all the cases that we have to test against:
- discover of streams => check all datatypes, all kafka related columns
- sync without and with state => check on the final datatypes with destination, check for duplication of data or data is misses
- rebalance detection, and graceful shutdown of sync during rebalance
- retry on error during stream changes
- for all kafka versions, under normal circumstances (for example: no rebalance) sync should work properly.
- both JSON and Avro datatype matching.
- all authentication protocols of kafka to be tested.
- testing to be done in single partition in a topic, multiple partitions in topic, multiple topics with multiple partitions, topics with partitions having no-messages or some already committed messages.
- testing should be done in testing-server with less messages, 2 million messages, 20 million messages, 2 billion messages
- iceberg and parquet partitioning to be tested.

If any more, add the case that has to be tested.

drivers/db2/go.mod
```diff
@@ -102,14 +102,13 @@ require (
 	github.com/parquet-go/parquet-go v0.25.0 // indirect
 	github.com/paulmach/orb v0.12.0 // indirect
 	github.com/pelletier/go-toml/v2 v2.2.3 // indirect
-	github.com/pierrec/lz4/v4 v4.1.22 // indirect
+	github.com/pierrec/lz4/v4 v4.1.26 // indirect
```
ImDoubD-datazip, line 105: what is this required for?

drivers/kafka/internal/cdc.go
```diff
@@ -46,29 +47,51 @@ func (k *Kafka) PreCDC(ctx context.Context, streams []types.StreamInterface) err
...
 	// create a reader manager for kafka
 	k.readerManager = kafkapkg.NewReaderManager(kafkapkg.ReaderConfig{
-		BootstrapServers:            k.config.BootstrapServers,
 		MaxThreads:                  k.config.MaxThreads,
 		ConsumerGroupID:             k.consumerGroupID,
 		Dialer:                      k.dialer,
-		AdminClient:                 k.adminClient,
+		Admin:                       k.admin,
 		ThreadsEqualTotalPartitions: k.config.ThreadsEqualTotalPartitions,
 	})
```
ImDoubD-datazip, line 60: keep proper spacing.

drivers/kafka/internal/kafka.go
```diff
@@ -39,13 +39,15 @@ var InternalKafkaTopics = []string{"__amazon_msk_canary", "_schemas"}
...
-	adminClient          *kafka.Client
+	dialer               []kgo.Opt
+	client               *kgo.Client
 	state                *types.State
 	consumerGroupID      string
+	streams              []types.StreamInterface // set in PreCDC and used when restarting readers from StreamChanges
 	readerManager        *kafkapkg.ReaderManager
 	checkpointMessage    sync.Map // last message for each reader w.r.t. partition to be used for checkpointing
 	schemaRegistryClient *kafkapkg.SchemaRegistryClient
+	admin                *kadm.Client
```
ImDoubD-datazip: rename this to adminClient

drivers/mongodb/go.mod
```diff
@@ -116,6 +115,8 @@ require (
 	github.com/testcontainers/testcontainers-go v0.42.0 // indirect
 	github.com/tklauser/go-sysconf v0.3.16 // indirect
 	github.com/tklauser/numcpus v0.11.0 // indirect
+	github.com/twmb/franz-go v1.21.1 // indirect
+	github.com/twmb/franz-go/pkg/kmsg v1.13.1 // indirect
```
ImDoubD-datazip, line 119: can we somehow design the go mod such that kafka based dependencies do not come in other drivers go mods.

pkg/kafka/types.go
```diff
@@ -3,39 +3,42 @@ package kafka
...
 // ReaderConfig holds configuration for creating Kafka readers
 type ReaderConfig struct {
 	MaxThreads                  int
 	ThreadsEqualTotalPartitions bool
-	BootstrapServers            string
 	ConsumerGroupID             string
-	Dialer                      *kafka.Dialer
-	AdminClient                 *kafka.Client
+	Dialer                      []kgo.Opt
+	Admin                       *kadm.Client
```
ImDoubD-datazip: rename this to adminClient

drivers/kafka/internal/cdc.go
```diff
@@ -46,29 +47,51 @@ func (k *Kafka) PreCDC(ctx context.Context, streams []types.StreamInterface) err
...
+	// Return early if no non-empty partitions are assigned to the reader
+	pending := 0
+	for _, pk := range assigned {
+		if _, ok := k.readerManager.GetPartitionIndex(fmt.Sprintf("%s:%d", pk.Topic, pk.Partition)); ok {
+			pending++
+		}
+	}
+	if pending == 0 {
+		logger.Infof("reader %d has no partitions with pending data assigned; completing StreamChanges", readerID)
+		return nil, nil
```
ImDoubD-datazip: check if we can remove this, as we have discussed to bring custom balancer

drivers/kafka/internal/cdc.go
```diff
@@ -46,29 +47,51 @@ func (k *Kafka) PreCDC(ctx context.Context, streams []types.StreamInterface) err
...
+	// create new readers and warmup the consumer group
 	return k.readerManager.CreateReaders(ctx, streams, k.consumerGroupID)
 }
 
 func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates map[string]any, processFn abstract.CDCMsgFn) (any, error) {
-	// get reader
-	reader := k.readerManager.GetReader(readerID)
-	if reader == nil {
-		return nil, fmt.Errorf("reader not found for readerID %d", readerID)
+	// Restart reader to get a fresh client per StreamChanges attempt
```
ImDoubD-datazip: provide concrete comment on why this is done here.

drivers/kafka/internal/cdc.go
```diff
@@ -46,29 +47,51 @@ func (k *Kafka) PreCDC(ctx context.Context, streams []types.StreamInterface) err
...
+		return nil, fmt.Errorf("failed to restart reader %d: %v", readerID, err)
+	}
+
+	//Check if current reader has some non-empty partitions or not
+	assigned, err := k.getReaderAssignedPartitions(ctx, readerID)
+	if err != nil {
+		return nil, err
+	}
+	// Return early if no non-empty partitions are assigned to the reader
+	pending := 0
```
ImDoubD-datazip: what is this? please improve on this. this won't be taken granted for in future.

drivers/kafka/internal/cdc.go
```diff
@@ -159,8 +183,10 @@ func (k *Kafka) PostCDC(ctx context.Context, readerIdx int) error {
 			if reader == nil {
 				return fmt.Errorf("reader %s not found for commit", readerID)
 			}
+			_, generationID := reader.GroupMetadata()
```
ImDoubD-datazip, line 174: write clean code, provide spacing.

drivers/kafka/internal/cdc.go
```diff
@@ -100,6 +123,7 @@ func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates
 			}
 		}
 
+		// track processing state
```
ImDoubD-datazip: if adding comment, add meaningful comments. this is last message map.

drivers/kafka/internal/cdc.go
```diff
@@ -182,45 +208,66 @@ func (k *Kafka) PostCDC(ctx context.Context, readerIdx int) error {
 }
 
 // for processing messages from a Kafka reader.
-func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kafka.Reader, stopProcessFn func(record types.KafkaRecord) (bool, error)) error {
+func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kgo.Client, stopProcessFn func(record types.KafkaRecord) (bool, error)) error {
 	for {
-		message, err := reader.FetchMessage(ctx)
-		if err != nil {
-			return fmt.Errorf("error reading message in Kafka CDC sync: %s", err)
+		pollCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
ImDoubD-datazip: check if we can reduce this to 10 sec.

drivers/kafka/internal/cdc.go
```diff
@@ -182,45 +208,66 @@ func (k *Kafka) PostCDC(ctx context.Context, readerIdx int) error {
...
+		messages := reader.PollFetches(pollCtx)
+		pollCtxErr := pollCtx.Err()
+		cancel()
+		errs := messages.Errors()
```
ImDoubD-datazip: use proper naming and variables, directly access wherever possible.

drivers/db2/go.mod
saksham-datazip: Ran go mod tidy after adding the franz-go dependency, which updated the module indirect dependencie.

drivers/kafka/internal/kafka.go
saksham-datazip: Renamed

drivers/mongodb/go.mod
saksham-datazip: This is a pre-Existing case in codebase even segment driver was getting imported everywhere ;
And also mongo-db driver is getting imported in other driver's go.mod
```
go.mongodb.org/mongo-driver
```

And the problem for this is in types/kafka_types.go, 
this import exists 
```
import "github.com/twmb/franz-go/pkg/kgo" 
```
So any driver which imports types.go will import this driver as an indirect dependancy

pkg/kafka/types.go
saksham-datazip: Renamed

drivers/kafka/internal/cdc.go
saksham-datazip: Added a concrete comment

drivers/db2/go.mod
saksham-datazip: Here u can access all the test cases which u mentioned :-
https://datazip.atlassian.net/wiki/x/AoAJHg

drivers/kafka/internal/cdc.go
saksham-datazip: did gofmt -w . and than this happend

drivers/kafka/internal/cdc.go
saksham-datazip: ADDED CUSTOM BALANCER AND REMOVED THIS PART

drivers/kafka/internal/cdc.go
saksham-datazip: This is stored because cancel() will override the error if any

drivers/kafka/internal/cdc.go
```diff
@@ -50,25 +52,34 @@ func (k *Kafka) PreCDC(ctx context.Context, streams []types.StreamInterface) err
...
 
 func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates map[string]any, processFn abstract.CDCMsgFn) (any, error) {
-	// get reader
-	reader := k.readerManager.GetReader(readerID)
-	if reader == nil {
-		return nil, fmt.Errorf("reader not found for readerID %d", readerID)
+	// Restart the reader to create a fresh franz-go client for each StreamChanges attempt.
+	// franz-go keeps uncommitted offsets in memory, so restarting clears that state and
+	// ensures retries resume from the last committed offset. Since a static instance ID
+	// is used, this restart does not trigger a consumer group rebalance.
```
ImDoubD-datazip:
```
// Restart the reader to create a fresh franz-go client for each StreamChanges attempt.
// franz-go keeps uncommitted offsets in memory, so restarting clears that state and
// ensures retries resume from the last committed offset.
// Note: Since a static instance ID is used, this restart does not trigger a consumer group rebalance.
```

drivers/kafka/internal/cdc.go
```diff
@@ -182,45 +196,73 @@ func (k *Kafka) PostCDC(ctx context.Context, readerIdx int) error {
 }
 
 // for processing messages from a Kafka reader.
-func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kafka.Reader, stopProcessFn func(record types.KafkaRecord) (bool, error)) error {
+func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kgo.Client, stopProcessFn func(record types.KafkaRecord) (bool, error)) error {
```
ImDoubD-datazip, line 200:
```
// processKafkaMessages processes messages from a Kafka reader
// until stopProcessFn signals stop, a rebalance is detected, or the poll times out (reader caught up).
func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kgo.Client, stopProcessFn func(record types.KafkaRecord) (bool, error)) error {
	var iter *kgo.FetchesRecordIter

	for {
		// checked before every poll and every record so a rebalance signal is never delayed by full batch processing.
		if stop, err := k.readerManager.FetchExitState(); stop {
			return err
		}

		// iter being nil triggers first poll.
		// iter.Done() being true triggers next polls when the current batch is fully consumed.
		if iter == nil || iter.Done() {
			pollCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
			// poll for new messages
			fetches := reader.PollFetches(pollCtx)
			pollErr := pollCtx.Err()
			cancel()

			// no new messages for 10s means reader has caught up; exit cleanly without error.
			if errors.Is(pollErr, context.DeadlineExceeded) {
				logger.Warnf("poll context deadline exceeded; stopping reader")
				return nil
			}

			// any fetch error (including parent ctx cancellation) is non-retryable.
			// For more info, gp through: https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#Fetches.Errors
			if errs := fetches.Errors(); len(errs) > 0 {
				return fmt.Errorf("%w: error reading message in Kafka CDC sync: %s", constants.ErrNonRetryable, errs[0].Err)
			}

			// wrap batch into iterator
			iter = fetches.RecordIter()
			// check exit state again before processing the batch
			continue
		}

		msg := iter.Next()
		data, key, err := k.parseKafkaData(msg)
		if err != nil {
			logger.Warnf("failed to parse message of topic: %s, partition: %d, offset %d, error: %s", msg.Topic, msg.Partition, msg.Offset, err)
		} else if data != nil {
			ts, err := typeutils.ReformatDate(msg.Timestamp, true)
			if err != nil {
				return fmt.Errorf("failed to reformat date: %s", err)
			}
			data[Partition] = msg.Partition
			data[Offset] = msg.Offset
			data[Key] = key
			data[KafkaTimestamp] = ts
		}

		stop, err := stopProcessFn(types.KafkaRecord{Data: data, Message: msg})
		if err != nil {
			return err
		}
		if stop {
			return nil
		}
	}
}
```

drivers/kafka/internal/kafka.go
```diff
@@ -7,21 +7,21 @@ import (
 	"fmt"
 	"regexp"
 	"slices"
-	"strings"
 	"sync"
 	"time"
 
 	kafkapkg "github.com/datazip-inc/olake/pkg/kafka"
+	kplain "github.com/twmb/franz-go/pkg/sasl/plain"
+	kscram "github.com/twmb/franz-go/pkg/sasl/scram"
```
ImDoubD-datazip:
```
	kafkaplain "github.com/twmb/franz-go/pkg/sasl/plain"
	kafkascram "github.com/twmb/franz-go/pkg/sasl/scram"
```

drivers/kafka/internal/kafka.go
ImDoubD-datazip, line 37: var InternalKafkaTopics = []string{"__amazon_msk_canary", "_schemas", "__consumer_offsets"}

drivers/kafka/internal/kafka.go
```diff
@@ -119,30 +119,42 @@ func (k *Kafka) Setup(ctx context.Context) error {
...
+	k.config.RetryCount = utils.Ternary(k.config.RetryCount <= 0, 1, k.config.RetryCount+1).(int)
+
 	return nil
 }
 
 func (k *Kafka) Close() error {
-	k.adminClient = nil
-	k.dialer = nil
-	k.readerManager.Close()
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
ImDoubD-datazip: ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)

as in RemoveExistingConsumers, the timeout is already of 90 seconds. we can also remove from here, the timeout as it already exists.

drivers/kafka/internal/kafka.go
```diff
@@ -168,41 +179,60 @@ func (k *Kafka) ProduceSchema(ctx context.Context, streamName string) (*types.St
...
+		}
+
+		startOffset, exists := startOffsets.Lookup(streamName, partitionDetail.Partition)
+		if !exists {
+			return nil
+		}
+		endOffset, exists := endOffsets.Lookup(streamName, partitionDetail.Partition)
+		if !exists {
+			return nil
+		}
```
ImDoubD-datazip:
```
startOffset, startOffsetExists := startOffsets.Lookup(streamName, partitionDetail.Partition)
		endOffset, endOffsetExists := endOffsets.Lookup(streamName, partitionDetail.Partition)
		if !startOffsetExists || !endOffsetExists {
			return nil
		}
```

drivers/kafka/internal/kafka.go
```diff
@@ -168,41 +179,60 @@ func (k *Kafka) ProduceSchema(ctx context.Context, streamName string) (*types.St
...
-		if err := reader.SetOffset(partitionDetails.FirstOffset); err != nil {
-			logger.Warnf("failed to set offset for partition %d: %s, continuing to next partition", partitionDetails.Partition, err)
-			return nil
+		consumerOpts = append(
+			consumerOpts,
+			kgo.FetchMaxBytes(10e6),
+			kgo.ConsumePartitions(map[string]map[int32]kgo.Offset{
+				streamName: {
+					partitionDetail.Partition: kgo.NewOffset().At(startOffset.Offset),
+				},
```
ImDoubD-datazip: kgo.ConsumePartitions(map[string]map[int32]kgo.Offset{
				streamName: {partitionDetail.Partition: kgo.NewOffset().At(startOffset.Offset)},
			}),

drivers/kafka/internal/kafka.go
ImDoubD-datazip: no need for this, pass ctx directly, as processKafkaMessages already has ctx timeout.

drivers/kafka/internal/kafka.go
```diff
@@ -374,45 +391,36 @@ func (k *Kafka) checkPartitionCompletion(ctx context.Context, readerID int, comp
...
+	readerID, _ := k.readerManager.GetReaderIDAndClientID(readerIndex)
+	if readerID == "" {
+		return nil, fmt.Errorf("readerID not found for reader index %d", readerIndex)
 	}
 
-	resp, err := k.adminClient.DescribeGroups(ctx, &kafka.DescribeGroupsRequest{
-		Addr:     addr,
-		GroupIDs: []string{k.consumerGroupID},
-	})
+	response, err := k.adminClient.DescribeGroups(ctx, k.consumerGroupID)
```
ImDoubD-datazip: not response, change it to describeGroupResp

pkg/kafka/balancer.go
```diff
@@ -2,51 +2,78 @@ package kafka
...
-func (b *CustomGroupBalancer) AssignGroups(members []kafka.GroupMember, partitions []kafka.Partition) kafka.GroupMemberAssignments {
-	assignments := make(kafka.GroupMemberAssignments)
+// JoinGroupMetadata encodes consumer subscription metadata for group joining.
+func (b *CustomGroupBalancer) JoinGroupMetadata(topicInterests []string, _ map[string][]int32, generation int32) []byte {
+	memberMetadata := kmsg.NewConsumerMemberMetadata()
+	memberMetadata.Version = 3
+	memberMetadata.Topics = topicInterests
+	memberMetadata.Generation = generation
+	return memberMetadata.AppendTo(nil)
+}
```
ImDoubD-datazip:
```
func (b *CustomGroupBalancer) JoinGroupMetadata(topicInterests []string, _ map[string][]int32, _ int32) []byte {
	memberMetadata := kmsg.NewConsumerMemberMetadata()
	memberMetadata.Topics = topicInterests
	return memberMetadata.AppendTo(nil)
}
```

pkg/kafka/balancer.go
```diff
@@ -2,51 +2,78 @@ package kafka
...
 
-	// number of consumers to use
-	consumerIDCount := min(b.requiredConsumerIDs, len(members))
+// ParseSyncAssignment decodes topic partition assignments from SyncGroup response data.
+func (b *CustomGroupBalancer) ParseSyncAssignment(assignment []byte) (map[string][]int32, error) {
+	return kgo.ParseConsumerSyncAssignment(assignment)
+}
```
ImDoubD-datazip, line 29:
```
// MemberBalancer returns a GroupMemberBalancer for the given group members.
func (b *CustomGroupBalancer) MemberBalancer(members []kmsg.JoinGroupResponseMember) (kgo.GroupMemberBalancer, map[string]struct{}, error) {
	consumerBalancer, err := kgo.NewConsumerBalancer(b, members)
	return consumerBalancer, consumerBalancer.MemberTopics(), err
}
```

pkg/kafka/balancer.go
```diff
@@ -2,51 +2,78 @@ package kafka
...
-	// active partitions with data in partition index
-	activePartitions := make([]kafka.Partition, 0, len(partitions))
-	err := utils.ForEach(partitions, func(partition kafka.Partition) error {
-		if _, exists := b.partitionIndex[fmt.Sprintf("%s:%d", partition.Topic, partition.ID)]; exists {
-			activePartitions = append(activePartitions, partition)
-		}
-		return nil
-	})
+// MemberBalancer returns a GroupMemberBalancer for the given group members.
+func (b *CustomGroupBalancer) MemberBalancer(members []kmsg.JoinGroupResponseMember) (kgo.GroupMemberBalancer, map[string]struct{}, error) {
```
ImDoubD-datazip, line 32:
```
// Balance assigns active partitions to consumers using round-robin distribution.
func (b *CustomGroupBalancer) Balance(consumerBalancer *kgo.ConsumerBalancer, partitionsPerTopic map[string]int32) kgo.IntoSyncAssignment {
	// a new plan for partition assignment
	plan := consumerBalancer.NewPlan()
	// list of group members (consumers)
	members := consumerBalancer.Members()

	// number of consumers to use
	consumerCount := min(b.requiredConsumerIDs, len(members))

	// active partitions with data in partition index
	activePartitions := make([]types.PartitionKey, 0, len(b.partitionIndex))
	for topic, partitions := range partitionsPerTopic {
		for partition := range partitions {
			if _, ok := b.partitionIndex[fmt.Sprintf("%s:%d", topic, partition)]; ok {
				activePartitions = append(activePartitions, types.PartitionKey{Topic: topic, Partition: partition})
			}
		}
	}

	// partition assignment in round-robin manner across consumers
	for index, activePartition := range activePartitions {
		consumerIndex := index % consumerCount
		plan.AddPartition(&members[consumerIndex], activePartition.Topic, activePartition.Partition)
	}

	return plan
}
```

pkg/kafka/reader.go
```diff
@@ -40,46 +50,29 @@ func (r *ReaderManager) CreateReaders(ctx context.Context, streams []types.Strea
...
 	for readerIndex := range readersToCreate {
 		readerID := fmt.Sprintf("group_%s", utils.ULID())
-		clientID := fmt.Sprintf("olake-%s-%s", consumerGroupID, readerID)
-
-		// create a per-reader dialer with a unique clientID to identify assignments
-		dialerCopy := *r.config.Dialer
-		dialerCopy.ClientID = clientID
+		clientID := fmt.Sprintf("olake-%s-%s", r.config.ConsumerGroupID, readerID)
 
+		reader, err := r.CreateReader(readerID, clientID, readersToCreate, false)
```
ImDoubD-datazip:
```
		// create reader (rebalance callbacks disabled during initial reader creation and partition assignment)
		reader, err := r.CreateReader(readerID, clientID, readersToCreate, false)
```

pkg/kafka/reader.go
```diff
@@ -40,46 +50,29 @@ func (r *ReaderManager) CreateReaders(ctx context.Context, streams []types.Strea
...
+		if err != nil {
+			return fmt.Errorf("failed to create reader %d: %v", readerIndex, err)
+		}
 		r.readers = append(r.readers, &kafkaReader{
```
ImDoubD-datazip, line 64: // add reader to manager
		r.readers = append(r.readers, &kafkaReader{

pkg/kafka/reader.go
```diff
@@ -40,46 +50,29 @@ func (r *ReaderManager) CreateReaders(ctx context.Context, streams []types.Strea
...
-				Dialer: &dialerCopy,
-			}),
+			reader:   reader,
 		})
 	}
-	logger.Infof("created %d readers for %d total partitions, with consumer group %s", len(r.readers), totalPartitions, consumerGroupID)
-	return nil
+	logger.Infof("created %d readers for %d total partitions, with consumer group %s", len(r.readers), totalPartitions, r.config.ConsumerGroupID)
+	// wait for consumer group members to join and partitions to be assigned
+	return r.waitForConsumerGroupJoin()
```
ImDoubD-datazip:
```
	// wait for consumer group members to join and partitions to be assigned, with a 2-minute deadline.
	joinCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
	defer cancel()
	return r.waitForPartitionAssignment(joinCtx)
```

pkg/kafka/reader.go
ImDoubD-datazip:
```
func (r *ReaderManager) SetPartitions(ctx context.Context, stream types.StreamInterface) error {
	topic := stream.Name()
	topicDetail, err := r.GetTopicMetadata(ctx, topic)
	if err != nil {
		return err
	}

	// fetch start offset of all partitions in topic
	startOffsets, startOffsetErr := r.config.AdminClient.ListStartOffsets(ctx, topic)
	if startOffsetErr != nil {
		return fmt.Errorf("failed to list start offsets for topic %s: %s", topic, startOffsetErr)
	}

	// fetch last offset of all partitions in topic
	endOffsets, endOffsetErr := r.config.AdminClient.ListEndOffsets(ctx, topic)
	if endOffsetErr != nil {
		return fmt.Errorf("failed to list end offsets for topic %s: %s", topic, endOffsetErr)
	}

	// fetch already committed offset of partition
	committedTopicOffsets, committedOffsetsErr := r.FetchCommittedOffsets(ctx, topic)
	if err != nil {
		return fmt.Errorf("failed to fetch committed offsets for topic %s: %s", topic, committedOffsetsErr)
	}

	// build partition metadata
	for _, partitionDetail := range topicDetail.Partitions {
		startOffsetDetail, startOffsetExists := startOffsets.Lookup(topic, partitionDetail.Partition)
		endOffsetDetail, endOffsetExists := endOffsets.Lookup(topic, partitionDetail.Partition)
		if !startOffsetExists || !endOffsetExists {
			continue
		}

		// check if the partition has any messages at all, if not then skip
		if startOffsetDetail.Offset >= endOffsetDetail.Offset {
			logger.Infof("skipping empty partition %d for topic %s (first: %d, last: %d)", partitionDetail.Partition, topic, startOffsetDetail.Offset, endOffsetDetail.Offset)
			continue
		}

		committedOffset, hasCommittedOffset := committedTopicOffsets[partitionDetail.Partition]

		// if a committed offset is available and there are no new messages, skip
		if hasCommittedOffset && committedOffset >= endOffsetDetail.Offset {
			logger.Infof("skipping partition %d for topic %s, no new messages (committed: %d, last: %d)", partitionDetail.Partition, topic, committedOffset, endOffsetDetail.Offset)
			continue
		}

		r.partitionIndex[fmt.Sprintf("%s:%d", topic, partitionDetail.Partition)] = types.PartitionMetaData{
			Stream:      stream,
			PartitionID: partitionDetail.Partition,
			EndOffset:   endOffsetDetail.Offset,
		}
	}
	return nil
}

```


mostly naming changes

pkg/kafka/reader.go
```diff
@@ -108,104 +101,272 @@ func (r *ReaderManager) SetPartitions(ctx context.Context, stream types.StreamIn
...
+
+	for partition, offset := range topicOffsets {
+		committedTopicOffsets[partition] = offset.At
+	}
+
+	return committedTopicOffsets, nil
+}
+
+// RemoveExistingConsumers force removes all existing consumers from the consumer group and closes reader clients.
+func (r *ReaderManager) RemoveExistingConsumers(ctx context.Context, client *kgo.Client) error {
```
ImDoubD-datazip, line 217:
```
// RemoveExistingConsumers force removes all existing consumers from the consumer group and closes reader clients.
func (r *ReaderManager) RemoveExistingConsumers(ctx context.Context, client *kgo.Client) error {
	// coordinator may not be active immediately (due to broker startup or coordinator election); retry describe until ready.
	describeGroupCtx, cancel := context.WithTimeout(ctx, 90*time.Second)
	defer cancel()

	var describedGroups kadm.DescribedGroups
	for {
		var err error
		describedGroups, err = r.config.AdminClient.DescribeGroups(describeGroupCtx, r.config.ConsumerGroupID)
		if err == nil {
			break
		}
		select {
		case <-describeGroupCtx.Done():
			return fmt.Errorf("describe groups failed: %s", err)
		case <-time.After(2 * time.Second):
		}
	}

	describedGroup := describedGroups[r.config.ConsumerGroupID]
	if describedGroup.Err != nil {
		return fmt.Errorf("describe groups error: %s", describedGroup.Err)
	}

	if len(describedGroup.Members) > 0 {
		leaveGroupRequest := kmsg.NewPtrLeaveGroupRequest()
		leaveGroupRequest.Group = r.config.ConsumerGroupID
		for _, member := range describedGroup.Members {
			leaveGroupRequest.Members = append(leaveGroupRequest.Members, kmsg.LeaveGroupRequestMember{
				MemberID:   member.MemberID,
				InstanceID: member.InstanceID,
			})
		}

		leaveGroupResponse, err := leaveGroupRequest.RequestWith(describeGroupCtx, client)
		if err != nil {
			return fmt.Errorf("leave group request failed: %s", err)
		}
		if leaveGroupResponse.ErrorCode != 0 {
			return fmt.Errorf("leave group error code: %d", leaveGroupResponse.ErrorCode)
		}
	}

	for _, kafkaReader := range r.readers {
		kafkaReader.reader.Close()
	}

	return nil
}
```

pkg/kafka/reader.go
```diff
@@ -108,104 +101,272 @@ func (r *ReaderManager) SetPartitions(ctx context.Context, stream types.StreamIn
...
+
+	committedTopicOffsets := make(map[int32]int64)
+
+	topicOffsets, exists := offsets[topic]
+	if !exists {
+		return committedTopicOffsets, nil
+	}
+
+	for partition, offset := range topicOffsets {
+		committedTopicOffsets[partition] = offset.At
```
ImDoubD-datazip, line 210:
```
		// offset.At can be -ve if no committed offset in partition.
		if offset.At >= 0 {
			committedTopicOffsets[partition] = offset.At
		}
```

pkg/kafka/reader.go
```diff
@@ -20,7 +24,13 @@ func NewReaderManager(config ReaderConfig) *ReaderManager {
 }
 
 // CreateReaders creates Kafka readers based on the provided streams and configuration
-func (r *ReaderManager) CreateReaders(ctx context.Context, streams []types.StreamInterface, consumerGroupID string) error {
+func (r *ReaderManager) CreateReaders(ctx context.Context, streams []types.StreamInterface) error {
+	// populate topics from streams
+	r.topics = make([]string, 0, len(streams))
```
ImDoubD-datazip: remove this.

pkg/kafka/reader.go
```diff
@@ -108,104 +101,272 @@ func (r *ReaderManager) SetPartitions(ctx context.Context, stream types.StreamIn
...
-	fetchOffsetReq := &kafka.OffsetFetchRequest{
-		GroupID: r.config.ConsumerGroupID,
-		Topics:  map[string][]int{topic: partitionsToFetch},
-	}
+	readerID, clientID := r.GetReaderIDAndClientID(readerIndex)
+
+	currentReader.Close()
 
-	committedOffsetsResp, err := r.config.AdminClient.OffsetFetch(ctx, fetchOffsetReq)
+	reader, err := r.CreateReader(readerID, clientID, len(r.readers), true)
```
ImDoubD-datazip: make this `newReader` from reader

pkg/kafka/reader.go
```diff
@@ -108,104 +101,272 @@ func (r *ReaderManager) SetPartitions(ctx context.Context, stream types.StreamIn
...
 	if err != nil {
-		logger.Warnf("could not fetch committed offsets for group %s", r.config.ConsumerGroupID)
+		return nil, fmt.Errorf("%v: failed to recreate kafka reader %d after close: %v", constants.ErrNonRetryable, readerIndex, err)
```
ImDoubD-datazip: remove %v, make it %s. check all sites and change to %s for errors. for non retryable error, make it %w

pkg/kafka/reader.go
```diff
@@ -108,104 +101,272 @@ func (r *ReaderManager) SetPartitions(ctx context.Context, stream types.StreamIn
...
+			}),
+			kgo.OnPartitionsLost(func(_ context.Context, _ *kgo.Client, lost map[string][]int32) {
+				logger.Warnf("reader %s lost partitions: %+v", clientID, lost)
+				r.exitMode.Store(nonRetryableExit)
+			}),
+		)
 	}
-	return committedTopicOffsets
+
+	reader, err := kgo.NewClient(readerOpts...)
```
ImDoubD-datazip, line 331:
```
// detect consumer group rebalances during message processing.
	if enableRebalanceCallbacks {
		onRebalance := func(_ context.Context, client *kgo.Client, _ map[string][]int32) {
			if r.RebalanceDetected(client) {
				r.exitMode.Store(gracefulExit)
			}
		}
		readerOpts = append(readerOpts,
			kgo.OnPartitionsAssigned(onRebalance),
			kgo.OnPartitionsRevoked(onRebalance),
			// partition loss is unrecoverable; broker forcibly revoked assignment, treated as non-retryable error.
			kgo.OnPartitionsLost(func(_ context.Context, _ *kgo.Client, lost map[string][]int32) {
				logger.Warnf("reader %s lost partitions: %+v", clientID, lost)
				r.exitMode.Store(nonRetryableExit)
			}),
		)
	}

	// initial partition assignment happens here.
	reader, err := kgo.NewClient(readerOpts...)
```

pkg/kafka/reader.go
```diff
@@ -108,104 +101,272 @@ func (r *ReaderManager) SetPartitions(ctx context.Context, stream types.StreamIn
...
+	switch r.exitMode.Load() {
+	case normalProcessing:
+		return false, nil
+	case gracefulExit:
+		logger.Infof("stopping kafka CDC processing gracefully due to consumer group rebalance")
+		return true, nil
+	case nonRetryableExit:
+		return true, fmt.Errorf("%v: kafka sync aborted due to partition loss during consumer group rebalance", constants.ErrNonRetryable)
+	default:
+		return true, fmt.Errorf("%v: kafka sync aborted: unexpected exit mode", constants.ErrNonRetryable)
```
ImDoubD-datazip: change to %w, below also.

pkg/kafka/reader.go
```diff
@@ -108,104 +101,272 @@ func (r *ReaderManager) SetPartitions(ctx context.Context, stream types.StreamIn
...
 	}
+}
 
-	return nil
+// waitForConsumerGroupJoin blocks until Kafka completes partition assignment
+// for all readers in the consumer group.
+func (r *ReaderManager) waitForConsumerGroupJoin() error {
```
ImDoubD-datazip: take context.

pkg/kafka/reader.go
```diff
@@ -108,104 +101,272 @@ func (r *ReaderManager) SetPartitions(ctx context.Context, stream types.StreamIn
...
+	for {
+		var (
```
ImDoubD-datazip: add this:
```
		if ctx.Err() != nil {
			return fmt.Errorf("timed out waiting for consumer group to join: %s", ctx.Err())
		}
```

pkg/kafka/reader.go
```diff
@@ -108,104 +101,272 @@ func (r *ReaderManager) SetPartitions(ctx context.Context, stream types.StreamIn
...
+		for _, kafkaReader := range r.readers {
+			_, generationID := kafkaReader.reader.GroupMetadata()
+
+			if generationID < 0 || (expectedGenerationID >= 0 && expectedGenerationID != generationID) {
+				allReadersJoined = false
+				break
+			} else if expectedGenerationID < 0 {
+				expectedGenerationID = generationID
+			}
+		}
```
ImDoubD-datazip:
```
for _, kafkaReader := range r.readers {
			_, currentReaderGenerationID := kafkaReader.reader.GroupMetadata()
			// generation id -1 means not yet joined
			// mismatch means readers are on different generations, partition assignment not yet completed.
			if currentReaderGenerationID < 0 || (expectedGenerationID >= 0 && expectedGenerationID != currentReaderGenerationID) {
				allReadersJoined = false
				break
			}
			if expectedGenerationID < 0 {
				expectedGenerationID = currentReaderGenerationID
			}
		}
```

naming changes

pkg/kafka/reader.go
```diff
@@ -108,104 +101,272 @@ func (r *ReaderManager) SetPartitions(ctx context.Context, stream types.StreamIn
...
+
+		if allReadersJoined {
+			r.generationID.Store(expectedGenerationID)
+			// wait for 2 seconds to ensure the consumer group is stable
```
ImDoubD-datazip: change this to:
			// brief wait to let partition assignment fully propagate before fetching starts.

pkg/kafka/types.go
```diff
@@ -3,38 +3,48 @@ package kafka
...
+	"github.com/twmb/franz-go/pkg/kadm"
+	"github.com/twmb/franz-go/pkg/kgo"
+)
+
+// ReaderManager exit modes used to control CDC processing flow.
+const (
+	normalProcessing int32 = iota // normal CDC processing state
+	gracefulExit                  // stop processing during rebalance without triggering abstract-layer retries
+	nonRetryableExit              // stop processing due to unrecoverable consumer state
 )
```
ImDoubD-datazip, line 18: // ReaderManager exit modes used to control processing flow.
const (
	normalProcessing int32 = iota // normal processing state
	gracefulExit                  // stop processing on rebalance (without triggering retries)
	nonRetryableExit              // stop processing due to unrecoverable kafka errors
)

comments changes

pkg/kafka/types.go
```diff
@@ -3,38 +3,48 @@ package kafka
...
 	clientID string
-	reader   *kafka.Reader
+	reader   *kgo.Client
 }
 
 // ReaderManager manages Kafka readers and their metadata
 type ReaderManager struct {
 	config         ReaderConfig
 	readers        []*kafkaReader
+	topics         []string                           // consume topics, set in CreateReaders and used by CreateReader
```
ImDoubD-datazip: change comment to:
// topics to be consumed

pkg/kafka/types.go
```diff
@@ -3,38 +3,48 @@ package kafka
...
 	partitionIndex map[string]types.PartitionMetaData // get per-partition boundaries
+	exitMode       atomic.Int32                       // normalProcessing | gracefulExit | nonRetryableExit
+	generationID   atomic.Int32                       // Group generationId is used to detect rebalances
```
ImDoubD-datazip: change comment to: // consumer group generationId: used to detect rebalances

ImDoubD-datazip (commented):
Go through this branch: [`feat/kafka-driver-change-2`](https://github.com/datazip-inc/olake/tree/feat/kafka-driver-change-2) and check the changes.

### Also check for the performance of the segment driver and franz driver, in ts4, for a billion messages.

drivers/kafka/internal/kafka.go
saksham-datazip: No need to add __consumer_offsets as it is already internal topic supported and i did multiple discover and it was skipping __consumer_offsets and emptied this map than _schemas is listed in topics as well as __amazon_msk_canary

drivers/kafka/internal/kafka.go
saksham-datazip: Removed the 30s timeout in Close() because it capped the internal 90s DescribeGroups retries and, with Background(), left LeaveGroup unbounded. We now use one 90 cleanupCtx inside RemoveExistingConsumers for both describe and leave.

drivers/kafka/internal/kafka.go
saksham-datazip: Added logs as per our discussion 
```
startOffset, startOffsetExists := startOffsets.Lookup(streamName, partitionDetail.Partition)
if !startOffsetExists {
	return nil
}
endOffset, endOffsetExists := endOffsets.Lookup(streamName, partitionDetail.Partition)
if !endOffsetExists {
	return nil
}

File: pkg/kafka/reader.go
Comment (saksham-datazip) [reply pkg/kafka/reader.go]: Changed the naming as well

File: pkg/kafka/reader.go (hunk 27 above)
Comment (saksham-datazip) [reply pkg/kafka/reader.go:210]: This api returns all commited offset :- FetchOffsets
and than FetchCommittedOffsets returns a map with key as partition and value as offset and it might be negative so that case is handled it does not get skipped in any step in Setpartitions.
Thus this negative check is mostly redundant. If their is something else so just let me know

File: pkg/kafka/reader.go (hunk 25 above)
Comment (saksham-datazip) [reply pkg/kafka/reader.go]: Created the context inside waitForPartitionAssignment and passed ctx here

File: drivers/kafka/internal/cdc.go [hunk 40]
```diff
@@ -181,29 +195,54 @@ func (k *Kafka) PostCDC(ctx context.Context, readerIdx int) error {
...
 		}
 
-		var (
-			key  string
-			data map[string]interface{}
-		)
+		// iter being nil triggers first poll.
+		// iter.Done() being true triggers next polls when the current batch is fully consumed.
+		if iter == nil || iter.Done() {
+			pollCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
vikaxsh: shouldn't we make poll timeout user configurable?

drivers/kafka/internal/cdc.go
```diff
@@ -50,25 +52,34 @@ func (k *Kafka) PreCDC(ctx context.Context, streams []types.StreamInterface) err
...
 		MaxThreads:                  k.config.MaxThreads,
 		ConsumerGroupID:             k.consumerGroupID,
 		Dialer:                      k.dialer,
 		AdminClient:                 k.adminClient,
 		ThreadsEqualTotalPartitions: k.config.ThreadsEqualTotalPartitions,
 	})
-	return k.readerManager.CreateReaders(ctx, streams, k.consumerGroupID)
+
+	// remove stale consumers before creating new readers
+	if err := k.readerManager.RemoveExistingConsumers(ctx, k.client); err != nil {
```
vikaxsh, line 63: what if two jobs using same kafka source

pkg/kafka/reader.go
```diff
@@ -95,117 +90,295 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
+		kgo.ConsumerGroup(r.config.ConsumerGroupID),
+		kgo.ClientID(clientID),
+		kgo.InstanceID(readerID),
+		kgo.ConsumeTopics(r.topics...),
+		kgo.Balancers(&CustomGroupBalancer{
+			requiredConsumerIDs: requiredConsumers,
+			partitionIndex:      r.partitionIndex,
+		}),
+		kgo.FetchMinBytes(1),
+		kgo.FetchMaxBytes(10e6),
```
vikaxsh, line 305: how are we deciding these numbers?

drivers/kafka/internal/kafka.go
```diff
@@ -119,30 +118,39 @@ func (k *Kafka) Setup(ctx context.Context) error {
 		logger.Infof("initialized schema registry client for endpoint: %s", k.config.SchemaRegistry.Endpoint)
 	}
 
+	// check for default backoff count
+	k.config.RetryCount = utils.Ternary(k.config.RetryCount <= 0, 1, k.config.RetryCount+1).(int)
```
vikaxsh, line 123: why are we modifying configurable fields, if it is necessary we should add in validate

pkg/kafka/reader.go
```diff
@@ -95,117 +90,295 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
-	committedOffsetsResp, err := r.config.AdminClient.OffsetFetch(ctx, fetchOffsetReq)
+	for _, kafkaReader := range r.readers {
+		kafkaReader.reader.Close()
+	}
+
+	return nil
+}
+
+// RestartReader closes and recreates the reader using the same instanceID.
+func (r *ReaderManager) RestartReader(readerIndex int) (*kgo.Client, error) {
```
vikaxsh, line 270: don't we need to set exitMode again in between retries?

drivers/kafka/internal/kafka.go
```diff
@@ -119,30 +118,39 @@ func (k *Kafka) Setup(ctx context.Context) error {
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to list topics: %s", err)
 	}
 
 	var topicNames []string
-	for _, topic := range resp.Topics {
+	for topicName := range metadata {
 		// skip internal topics
-		if topic.Internal || slices.Contains(InternalKafkaTopics, topic.Name) {
+		if slices.Contains(InternalKafkaTopics, topicName) {
```
vikaxsh, line 151: don't we have topic.Internal equivalent in franz?

drivers/kafka/internal/cdc.go
```diff
@@ -181,29 +195,54 @@ func (k *Kafka) PostCDC(ctx context.Context, readerIdx int) error {
...
+			// no new messages for 10s means reader has caught up; exit cleanly without error.
+			if errors.Is(pollCtxErr, context.DeadlineExceeded) {
+				logger.Warnf("poll context deadline exceeded: %s", pollCtxErr)
+				return nil
+			}
+
+			// any fetch error (including parent ctx cancellation) is non-retryable.
+			// For more info, go through the documentation: https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#Fetches.Errors
+			if errs := fetches.Errors(); len(errs) > 0 {
+				return fmt.Errorf("%w: error reading message in Kafka CDC sync: %s", constants.ErrNonRetryable, errs[0].Err)
```
vikaxsh: why only returning first error

pkg/kafka/reader.go
```diff
@@ -95,117 +90,295 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
+		if !startOffsetExists {
+			logger.Infof("skipping partition %d for topic %s, start offset not found", partitionDetail.Partition, topic)
+			continue
+		}
+
+		endOffsetDetail, endOffsetExists := endOffsets.Lookup(topic, partitionDetail.Partition)
+		if !endOffsetExists {
+			logger.Infof("skipping partition %d for topic %s, end offset not found", partitionDetail.Partition, topic)
+			continue
+		}
```
vikaxsh: We have similar code duplication in `kafka.go` for producing the schema as well. Can we see if this logic can be consolidated into a single place?

pkg/kafka/reader.go
```diff
@@ -20,7 +24,12 @@ func NewReaderManager(config ReaderConfig) *ReaderManager {
...
 // CreateReaders creates Kafka readers based on the provided streams and configuration
-func (r *ReaderManager) CreateReaders(ctx context.Context, streams []types.StreamInterface, consumerGroupID string) error {
+func (r *ReaderManager) CreateReaders(ctx context.Context, streams []types.StreamInterface) error {
+	// populate topics from streams
+	for _, stream := range streams {
+		r.topics = append(r.topics, stream.Name())
+	}
+
 	r.partitionIndex = make(map[string]types.PartitionMetaData)
 	for _, stream := range streams {
```
vikaxsh: can we iterate streams only once here?

pkg/kafka/reader.go
```diff
@@ -95,117 +90,295 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
+
+		if allReadersJoined {
+			r.generationID.Store(expectedGenerationID)
+			// brief wait to let partition assignment fully propagate before fetching starts.
+			time.Sleep(2 * time.Second)
+			logger.Infof("consumer group %s stable: all readers assigned, generation id: %d", r.config.ConsumerGroupID, expectedGenerationID)
+			return nil
+		}
+
+		time.Sleep(500 * time.Millisecond)
```
vikaxsh:
```suggestion
		select {
case <-joinCtx.Done():
case <-time.After(500 * time.Millisecond):
}
```

pkg/kafka/reader.go
```diff
@@ -95,117 +90,295 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
+	if !exists {
+		return nil, fmt.Errorf("topic %s not found in metadata", topic)
 	}
+	return &topicDetail, nil
+}
 
-	return nil, fmt.Errorf("topic %s not found in metadata", topic)
+// FetchCommittedOffsets fetches committed offsets for a topic.
+func (r *ReaderManager) FetchCommittedOffsets(ctx context.Context, topic string) (map[int32]int64, error) {
+	offsets, err := r.config.AdminClient.FetchOffsets(ctx, r.config.ConsumerGroupID)
```
vikaxsh: don't we have segment equivalnet in franz to fetch topic level offset instead fetching all

pkg/kafka/balancer.go
```diff
@@ -3,50 +3,70 @@ package kafka
...
 
 	// active partitions with data in partition index
-	activePartitions := make([]kafka.Partition, 0, len(partitions))
-	err := utils.ForEach(partitions, func(partition kafka.Partition) error {
-		if _, exists := b.partitionIndex[fmt.Sprintf("%s:%d", partition.Topic, partition.ID)]; exists {
-			activePartitions = append(activePartitions, partition)
+	activePartitions := make([]types.PartitionKey, 0)
+	for topic, partitions := range partitionsPerTopic {
+		for partition := range partitions {
+			if _, ok := b.partitionIndex[fmt.Sprintf("%s:%d", topic, partition)]; ok {
```
vikaxsh: consider creating a helper for this 
fmt.Sprintf("%s:%d", topic, partition)

pkg/kafka/reader.go
saksham-datazip: These numbers were on staging so just used them here as well and upon experimentation  i added 61 mb of a message and it synced fine and even in docs it is mentioned that:-
// FetchMaxBytes sets the maximum amount of bytes a broker will try to send
// during a fetch, overriding the default 50MiB. Note that brokers may not obey
// this limit if it has records larger than this limit. Also note that this
// client sends a fetch to each broker concurrently, meaning the client will
// buffer up to <brokers * max bytes> worth of memory.
//
// This corresponds to the Java fetch.max.bytes setting.
//
// If bumping this, consider bumping BrokerMaxReadBytes.
//
// If what you are consuming is compressed, and compressed well, it is strongly
// recommended to set this option so that decompression does not eat all of
// your RAM.
func FetchMaxBytes(b int32) ConsumerOpt {
	return consumerOpt{func(cfg *cfg) { cfg.maxBytes = lazyI32(b) }}
}

drivers/kafka/internal/cdc.go
saksham-datazip: RemoveExistingConsumers removes consumers from consumer group and we are going to mention it in docs olake CG is just for olake no changes should be done from user side

pkg/kafka/reader.go
saksham-datazip: why we need to set exit mode in retries again ?it is already on normalProcessing mode and these mode are set upon rebalance callbacks when generation id changes

drivers/kafka/internal/kafka.go
saksham-datazip: it is internally and already skipped by franz-go itself
```
// ListTopics issues a metadata request and returns TopicDetails. Specific
// topics to describe can be passed as additional arguments. If no topics are
// specified, all topics are requested. Internal topics are not returned unless
// specifically requested. To see all topics including internal topics, use
// ListTopicsWithInternal.
//
// This returns an error if the request fails to be issued, or an *AuthError.
```
 for more reference see comments for ListTopics function called in line 150

drivers/kafka/internal/kafka.go
saksham-datazip: Because atleast once the sync should run and this is what happend in all other drivers

drivers/kafka/internal/cdc.go
saksham-datazip: This timeout is added because due to rebalance kafka might stuck in infinite loop so to resolve this we added a context timeout

pkg/kafka/reader.go
saksham-datazip: We used FetchOffsetsfortopic function they have one distinctions as before it was returning 0 for all offsets which were not commited and now it is returning -1 but it does not have functionality difference

pkg/kafka/reader.go
saksham-datazip: common out this part and ListStartOffsets code part as well and created functions for them in reader.go

drivers/kafka/internal/cdc.go
saksham-datazip: Returned all errors by using join and %w  as it is non-retryable error

pkg/kafka/reader.go
```diff
@@ -95,117 +89,316 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
+	if !exists {
+		return nil, fmt.Errorf("topic %s not found in metadata", topic)
+	}
+	return &topicDetail, nil
+}
+
+// FetchCommittedOffsets fetches committed offsets for a topic.
+func (r *ReaderManager) FetchCommittedOffsets(ctx context.Context, topic string) (map[int32]int64, error) {
+	offsets, err := r.config.AdminClient.FetchOffsetsForTopics(ctx, r.config.ConsumerGroupID, topic)
+	if err != nil {
```
vikaxsh, line 204: on staging we were logging the error with warn level, why are we now returning error here?

drivers/kafka/internal/cdc.go
vikaxsh, line 63: `RemoveExistingConsumers` might crash on the first run with newer Kafka versions. Could you please verify this? We may be attempting to remove consumer groups before they are created in kafka, which could result in a `GROUP_ID_NOT_FOUND` error.

drivers/kafka/internal/kafka.go
vikaxsh: add a todo

pkg/kafka/reader.go
saksham-datazip: So if response fails than this part of code will be skipped inside SetPartitions and their might be a possibility of infinite loop as well in staging if that response returned is nil.
```
		// if a committed offset is available and there are no new messages, skip
		if hasCommittedOffset && committedOffset >= idx.LastOffset {
			logger.Infof("skipping partition %d for topic %s, no new messages (committed: %d, last: %d)", idx.Partition, topic, committedOffset, idx.LastOffset)
			continue
		}
```

pkg/kafka/reader.go
```diff
@@ -95,117 +90,316 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
+}
 
-	return nil
+// waitForPartitionAssignment blocks until Kafka completes partition assignment
+// for all readers in the consumer group.
+func (r *ReaderManager) waitForPartitionAssignment(ctx context.Context) error {
+	joinCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
+	for {
+		if joinCtx.Err() != nil {
```
hash-data: err can be other reasons as well not only time out

drivers/kafka/internal/cdc.go
```diff
@@ -50,25 +52,34 @@ func (k *Kafka) PreCDC(ctx context.Context, streams []types.StreamInterface) err
...
 		ConsumerGroupID:             k.consumerGroupID,
 		Dialer:                      k.dialer,
 		AdminClient:                 k.adminClient,
 		ThreadsEqualTotalPartitions: k.config.ThreadsEqualTotalPartitions,
 	})
-	return k.readerManager.CreateReaders(ctx, streams, k.consumerGroupID)
+
+	// remove stale consumers before creating new readers
+	if err := k.readerManager.RemoveExistingConsumers(ctx, k.client); err != nil {
+		return fmt.Errorf("failed to remove existing consumers: %s", err)
```
hash-data, line 64: can we remove it

drivers/kafka/internal/cdc.go
```diff
@@ -181,29 +195,54 @@ func (k *Kafka) PostCDC(ctx context.Context, readerIdx int) error {
...
 
-		// parse message value and key
-		data, key, err = k.parseKafkaData(message)
+			// wrap batch into iterator
+			iter = fetches.RecordIter()
+			// check exit state again before processing the batch
+			continue
+		}
+
+		message := iter.Next()
```
hash-data: how many records will be there in a batch? 
and also there was a case where message can be large then 1 mb?

pkg/kafka/balancer.go
```diff
@@ -1,52 +1,70 @@
...
 // ProtocolName implements kafka.GroupBalancer interface
 func (b *CustomGroupBalancer) ProtocolName() string {
 	return "olake-kafka-round-robin"
 }
 
-// UserData implements kafka.GroupBalancer interface
-func (b *CustomGroupBalancer) UserData() ([]byte, error) {
-	return nil, nil
+// IsCooperative returns false to indicate that the balancer is not cooperative.
+func (b *CustomGroupBalancer) IsCooperative() bool {
```
hash-data, line 15: add comment why we are not using ?

pkg/kafka/balancer.go
```diff
@@ -1,52 +1,70 @@
...
+
+	// list of group members (consumers)
+	members := consumerBalancer.Members()
+	if len(members) == 0 {
+		return plan
+	}
 
 	// number of consumers to use
-	consumerIDCount := min(b.requiredConsumerIDs, len(members))
+	consumerCount := min(b.requiredConsumerIDs, len(members))
```
hash-data: why this statement?

pkg/kafka/reader.go
```diff
@@ -95,117 +90,316 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
 }
 
+// PartitionIndexKey returns the map key used for partition index lookups.
+func PartitionIndexKey(topic string, partition int32) string {
+	return fmt.Sprintf("%s:%d", topic, partition)
+}
+
+// ListTopicOffsets returns start and end offsets for all partitions in a topic.
+func (r *ReaderManager) ListTopicOffsets(ctx context.Context, topic string) (kadm.ListedOffsets, kadm.ListedOffsets, error) {
+	startOffsets, err := r.config.AdminClient.ListStartOffsets(ctx, topic)
```
hash-data, line 157: just to confirm is there any extra permission required w.r.t. the prev kafka driver?

pkg/kafka/reader.go
```diff
@@ -20,9 +25,13 @@ func NewReaderManager(config ReaderConfig) *ReaderManager {
...
+func (r *ReaderManager) CreateReaders(ctx context.Context, streams []types.StreamInterface) error {
 	r.partitionIndex = make(map[string]types.PartitionMetaData)
 	for _, stream := range streams {
+		// populate topics from streams
+		r.topics = append(r.topics, stream.Name())
+
+		// set partitions for the stream
 		if err := r.SetPartitions(ctx, stream); err != nil {
 			return fmt.Errorf("failed to set partitions for stream %s: %s", stream.ID(), err)
 		}
```
hash-data: just return partition and assign to map, instead of assigning inside function, better readability

pkg/kafka/reader.go
```diff
@@ -95,117 +90,316 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
+		kgo.FetchMaxBytes(10e6),
+		kgo.DisableAutoCommit(),
+		kgo.ConsumeResetOffset(kgo.NewOffset().AtStart()),
+	)
+
+	if enableRebalanceCallbacks {
+		// Exit gracefully when a rebalance is detected via assign/revoke callbacks.
+		onRebalance := func(_ context.Context, client *kgo.Client, _ map[string][]int32) {
+			if r.RebalanceDetected(client) {
+				r.exitMode.Store(gracefulExit)
```
hash-data, line 314: this will create problem in MO

pkg/kafka/reader.go
```diff
@@ -95,117 +90,316 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
+			expectedGenerationID int32 = -1
+		)
+		for _, kafkaReader := range r.readers {
+			_, currentReaderGenerationID := kafkaReader.reader.GroupMetadata()
+
+			// generation id -1 means not yet joined
+			// mismatch means readers are on different generations, partition assignment not yet completed
+			if currentReaderGenerationID < 0 || (expectedGenerationID >= 0 && expectedGenerationID != currentReaderGenerationID) {
+				allReadersJoined = false
+				break
```
hash-data: need to understand again

drivers/kafka/internal/cdc.go
```diff
@@ -79,10 +90,10 @@ func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates
 		}
 	}()
 
-	err := k.processKafkaMessages(ctx, reader, func(record types.KafkaRecord) (bool, error) {
+	err = k.processKafkaMessages(ctx, reader, func(record types.KafkaRecord) (bool, error) {
 		// get current partition metadata and key
 		currentPartitionKey := types.PartitionKey{Topic: record.Message.Topic, Partition: record.Message.Partition}
-		currentPartitionMeta, exists := k.readerManager.GetPartitionIndex(fmt.Sprintf("%s:%d", record.Message.Topic, record.Message.Partition))
+		currentPartitionMeta, exists := k.readerManager.GetPartitionIndex(kafkapkg.PartitionIndexKey(record.Message.Topic, record.Message.Partition))
```
hash-data: function name seem not conveying what actully it does

drivers/kafka/internal/cdc.go
```diff
@@ -181,29 +195,54 @@ func (k *Kafka) PostCDC(ctx context.Context, readerIdx int) error {
...
+// until stopProcessFn signals stop, a rebalance is detected, or the poll times out (reader caught up).
+func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kgo.Client, stopProcessFn func(record types.KafkaRecord) (bool, error)) error {
+	var iter *kgo.FetchesRecordIter
+
 	for {
-		message, err := reader.FetchMessage(ctx)
-		if err != nil {
-			return fmt.Errorf("error reading message in Kafka CDC sync: %s", err)
+		// checked before every poll and every record so a rebalance signal is never delayed by full batch processing.
+		if stopProcessing, err := k.readerManager.FetchExitState(); stopProcessing {
```
hash-data: we should use context here as well

drivers/kafka/internal/cdc.go
```diff
@@ -133,21 +144,21 @@ func (k *Kafka) PostCDC(ctx context.Context, readerIdx int) error {
 		}
 
 		// Type assert and validate messages
-		lastMessages, isValid := lastMessagesMeta.(map[types.PartitionKey]kafka.Message)
+		lastMessages, isValid := lastMessagesMeta.(map[types.PartitionKey]*kgo.Record)
```
hash-data, line 147: should not it fail?

hash-data (commented):
Nice Code

saksham-datazip:
> Nice Code

Thnks

drivers/kafka/internal/cdc.go
hash-data: check default timeout

pkg/kafka/reader.go
hash-data:
```

// waitForPartitionAssignment blocks until Kafka completes partition assignment
// for all readers in the consumer group.
func (r *ReaderManager) waitForPartitionAssignment(ctx context.Context) error {
	joinCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
	defer cancel()
	for {
		select {
			case <-joinCtx.Done():
				return fmt.Errorf("timed out waiting for partition assignment on consumer group %s: %s", r.config.ConsumerGroupID, joinCtx.Err())
			case <-time.After(2 * time.Second):
				var (
					allReadersJoined           = true
					expectedGenerationID int32 = -1
				)
				for _, kafkaReader := range r.readers {
					_, currentReaderGenerationID := kafkaReader.reader.GroupMetadata()
		
					// generation id -1 means not yet joined
					// mismatch means readers are on different generations, partition assignment not yet completed
					if currentReaderGenerationID < 0 || (expectedGenerationID >= 0 && expectedGenerationID != currentReaderGenerationID) {
						allReadersJoined = false
						break
					}
					if expectedGenerationID < 0 {
						expectedGenerationID = currentReaderGenerationID
					}
				}
		
				if allReadersJoined {
					r.generationID.Store(expectedGenerationID)
					// brief wait to let partition assignment fully propagate before fetching starts.
					time.Sleep(2 * time.Second)
					logger.Infof("consumer group %s stable: all readers assigned, generation id: %d", r.config.ConsumerGroupID, expectedGenerationID)
					return nil
				}
		}
	}
}
```

pkg/kafka/balancer.go
saksham-datazip: In Kafka, when `isCooperative` is set to `false`, any rebalance causes all consumers to first have their partitions revoked before receiving new assignments. However, in the Franz-go library, there are three rebalance callbacks. During experimentation, I observed that `OnPartitionsAssigned` is triggered for all consumers on every rebalance, regardless of whether any partitions are assigned to that particular consumer, even when cooperative mode is set to `true`. Because of this behavior, either approach works for our use case, as the callback is still invoked during rebalances.So due to this we dont need any comment as of odd behavior from franz-go side

pkg/kafka/reader.go
saksham-datazip: Had a discussion with you regarding it, so no need for it now because rebalance detetection callbacks are only valid in streamchanges not in preCDC.

pkg/kafka/balancer.go
saksham-datazip: Yes it is not required it was just in staging so didnt removed but removed this now

pkg/kafka/reader.go
saksham-datazip: Previously on first run it use to first check all Readers readerId and than wait for 500ms now i changes it andi it will first wait 500ms and than start to check all readers readerId but since in other part of code we are using this and no big changes so i changed it.

pkg/kafka/reader.go
saksham-datazip: I couldn't find any new permission as this api is similar to the api we were using in segment-io and internally they are calling the same thing kafka.It needs read access.
https://kafka.apache.org/43/security/authorization-and-acls/

pkg/kafka/reader.go
saksham-datazip: It was their in staging, so i tried to make minimal changes but yes i changed it

drivers/kafka/internal/cdc.go
saksham-datazip: changed

drivers/kafka/internal/cdc.go
saksham-datazip: Changed

drivers/kafka/internal/cdc.go
saksham-datazip: Changed

drivers/kafka/internal/cdc.go
saksham-datazip: This batch size behavior already existed in Segment, where we had a 10 MB limit, so I kept the same limit in franz-go as well.

What this means is that once the broker has accumulated data up to this limit, it will return the batch. If a single message itself is larger than 10 MB, Kafka will still return that message in a single batch, ignoring the configured fetch limit for that request. I verified this by testing with a 61 MB message.

Regarding how these settings work:

* **MinFetchSize**: Specifies the minimum amount of data the broker should try to accumulate before returning a fetch response.
* **MaxFetchSize**: Sets an upper limit on the amount of data returned in a fetch response. However, if the first available message exceeds this limit, Kafka will still return that message in a single batch.
* **FetchMaxWaitMs**: Specifies how long the broker should wait for data to become available. If the minimum fetch size is not reached within this timeout, the broker returns whatever data is available (or an empty response).

Regarding the default poll timeout we discussed earlier, that behavior is not supported in franz-go. Unlike the Java consumer, franz-go does not enforce a maximum time between polls because heartbeats are handled independently in the background.

Reference: https://github.com/twmb/franz-go/issues/140

drivers/kafka/internal/cdc.go
saksham-datazip: I was not able to consistently reproduce this issue on my machine under normal conditions. However, after increasing the session timeout to 2 minutes, commenting out the `removeExistingConsumer` logic, and adding additional debug logging for investigation, I was able to observe that extra consumers were accumulating in the consumer group during rebalancing.

One possible outcome of this situation is that all partitions get assigned to the pre-existing consumers, leaving the newly created consumer with no partitions. In that case, `PollFetches` eventually times out because there are no assigned partitions to fetch from, causing the sync to exit without processing any records. However, this is only one possible explanation and not a confirmed root cause. There may be other unexpected behaviors or outcomes resulting from the presence of these extra consumers as well.

To further validate the behavior, I re-enabled the `removeExistingConsumer` logic and added additional logging inside the function to track how many consumers were being removed. Following the same reproduction steps, the logs showed that the function was working as expected and successfully removing the stale consumers.
Here is the reference link for the same :-
https://datazip.atlassian.net/wiki/x/AQCsI

drivers/kafka/internal/cdc.go
hash-data: prev we were returning on empty list

drivers/kafka/internal/kafka.go
```diff
@@ -118,6 +118,7 @@ func (k *Kafka) Setup(ctx context.Context) error {
 		logger.Infof("initialized schema registry client for endpoint: %s", k.config.SchemaRegistry.Endpoint)
 	}
 
+	// TODO: Avoid modifying the configured/default retry count during initialization across all drivers.
```
hash-data, line 121: we can remove this, as this is something done intentionaly

drivers/kafka/internal/cdc.go
hash-data: had a discussion with you, decided to keep it

pkg/kafka/reader.go
```diff
@@ -95,117 +93,311 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
-	return nil
+// waitForPartitionAssignment blocks until Kafka completes partition assignment
+// for all readers in the consumer group.
+func (r *ReaderManager) waitForPartitionAssignment(ctx context.Context) error {
+	joinCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
+	for {
+		select {
+		case <-joinCtx.Done():
+			return joinCtx.Err()
```
hash-data, line 374: Now it will retry even if it is a timeout

saksham-datazip: No need for this as in preCDC no retry happens

drivers/kafka/internal/cdc.go
saksham-datazip: reverted

pkg/kafka/reader.go
```diff
@@ -95,117 +93,311 @@ func (r *ReaderManager) ShouldMatchPartitionCount() bool {
...
+		return false, nil
 	}
 
-	for kStr := range r.partitionIndex {
-		delete(r.partitionIndex, kStr)
+	switch r.exitMode.Load() {
+	case normalProcessing:
+		return false, nil
+	case gracefulExit:
+		logger.Infof("stopping kafka CDC processing gracefully due to consumer group rebalance")
```
ImDoubD-datazip: change this info to warn

ImDoubD-datazip (approved):
LGTM

PR #959, opened 2026-05-20
feat: implement 2PC support in parquet writer

### LINKED ISSUE #856: implement 2pc in parquet writer (difficulty: hard)
Currently 2 phase commit support is only added for destination iceberg , the same needs to be done for parquet destination as well

https://github.com/datazip-inc/olake/blob/2c441ebc628b29ae98b2e1c94433acefcecacdee/destination/parquet/parquet.go#L346-L350

---

###### This issue was generated by [todo-issue](https://github.com/DerJuulsn/todo-issue) based on a `TODO` comment in 2c441ebc628b29ae98b2e1c94433acefcecacdee. It&#x27;s been assigned to @vaibhav-datazip because they committed the code.

siiddhantt:
# Description

Fixes #856

Adds destination-side 2PC recovery for the S3-backed Parquet destination. Committed full-refresh chunks and CDC/incremental checkpoints are persisted in the destination so retries can recover when OLake fails before saving the external source state.

The S3 layout is:

```text
<namespace>/<table>/
  _olake_2pc/
    <sha256-thread-id>/        # full-refresh staging
      <partition>/<files>
      finish.json
    <partition>/<files>        # CDC/incremental staging
    finish.json
  metadata.json
  <partition>/<committed-files>
```

## Commit and recovery flow

For full refresh:

- each chunk uploads to an isolated staging directory derived from a deterministic SHA-256 hash of its logical thread ID
- `finish.json` stores the original thread ID and marks the chunk as ready for promotion
- after promotion, the original thread ID is added to `metadata.json`
- retries use `full_refresh_committed_ids` to skip committed chunks
- staging without `finish.json` is incomplete and is discarded

The bounded hash prevents source-derived thread IDs, which may contain long object paths, from exceeding object-store path-component limits. The original logical ID remains in `finish.json` and `metadata.json`, so retry matching is unchanged.

For CDC and incremental:

- files upload under the shared `_olake_2pc/` staging prefix
- `finish.json` stores the `types.MetadataState` supplied by abstract
- after promotion, the checkpoint is merged into table-level `metadata.json`
- setup rolls finished staging forward before returning durable metadata
- no-data syncs do not update destination metadata, matching Iceberg behavior

Writers create files in private local temporary directories. Setup and Close are serialized per stream, so one writer modifies the shared S3 staging prefix at a time. A later serialized writer resolves any finished staging left by an earlier writer before starting its own commit. Parallel Kafka checkpoints are merged before `metadata.json` is updated.

The commit boundary remains:

1. upload all data to hidden staging
2. write `finish.json`
3. promote staged data to final table paths
4. update `metadata.json`
5. delete staged data first and `finish.json` last

Failures before `finish.json` are discarded on retry. Failures after `finish.json` are rolled forward idempotently. All S3 list, read, write, copy, upload, and delete operations use the existing Parquet retry policy.

Local Parquet behavior remains unchanged and is marked for follow-up 2PC support.

## Type of change

- [ ] Breaking change

## How Has This Been Tested?

- [x] Repository test suite: `go test ./... -count=1`
- [x] Repeated Parquet tests: `go test ./destination/parquet -count=10`
- [x] Parquet race tests: `go test -race ./destination/parquet -count=1`
- [x] Parquet vet: `go vet ./destination/parquet`
- [x] Binary build: `go build -o /tmp/olake-pr959-final-check .`
- [x] Postgres + MinIO manual demo covering full refresh, incremental, and CDC
- [x] S3 source to Parquet destination regression using the long thread ID from the failing integration run
- [x] Full-refresh interruption and committed-chunk replay
- [x] Incremental and CDC stale-state recovery
- [x] Interrupted upload, promotion, metadata commit, and cleanup recovery
- [x] Incomplete staging cleanup and idempotent roll-forward
- [x] CDC/incremental no-data behavior
- [x] Parallel Kafka checkpoint merging
- [x] `DropStreams` removal of table data and 2PC metadata

# Screenshots or Recordings

WIP

## Documentation

- [ ] Documentation Link
- [x] N/A (destination-internal feature and test changes only)

## Related PR's (If Any):

- Reference: #698


siiddhantt:
Updated the PR with the `_completed.json`-inside-staging approach discussed in #856.

Current Parquet 2PC shape:

`<namespace>/<table>/_olake_2pc/<escaped-thread-id>.staging/_completed.json`

Summary of the refactor:
- kept `_completed.json` as the durable completed marker inside each staging folder
- full refresh uses `{}` markers and derives committed chunk/thread IDs from the staging folder name
- CDC/incremental/no-op progress stores optional `types.MetadataState` in `_completed.json`
- setup promotes completed staging and removes incomplete staging before retry

Also reran the focused/unit/build/lint checks and the Postgres-backed manual demo for full refresh, incremental, and CDC recovery.

siiddhantt:
@nayanj98 looping you in as @vaibhav-datazip suggested.

This PR is ready for review now and the description is updated with the final flow and test coverage.

nayanj98:
@siiddhantt Assigning @vaibhav-datazip to review your PR.

siiddhantt:
@vaibhav-datazip I found the CI failure. It looks test-harness related rather than a 2PC flow failure.

The integration cleanup helper deletes only final `.parquet` files between Parquet subtests, but now `_olake_2pc` is part of the table-local destination state. That leaves stale `_completed.json` markers behind, so the next subtest can skip already-committed chunks while the actual data files were manually removed by the test helper.

I’ll push a small fix to clear `_olake_2pc/` in the Parquet integration cleanup as well, matching how `DropStreams()` clears the full table prefix.

destination/parquet/parquet.go
```diff
@@ -260,7 +270,25 @@ func (p *Parquet) Check(_ context.Context) error {
 	return nil
 }
 
-func (p *Parquet) closePqFiles(ctx context.Context, _ any, closeOnError bool) error {
+func (p *Parquet) dataFiles() []parquetDataFile {
```
vaibhav-datazip: can you rename this function to a more appropriate name, also add a comment for this function

drivers/postgres/internal/cdc.go
```diff
@@ -143,15 +143,18 @@ func (p *Postgres) StreamChanges(ctx context.Context, _ int, metadataStates map[
...
-	// Skip it when remainingStreams is empty: all streams are already committed in
-	// Iceberg, so no WAL will be replayed and the slot/state relationship is irrelevant.
-	if len(remainingStreams) > 0 {
+	// Destination 2PC metadata can be ahead of the external source state when
+	// the destination commit succeeded but saving the source state failed. If all
+	// streams are already committed and the slot is exactly at that committed LSN,
+	// there is no WAL left to replay; PostCDC will persist the recovered state.
+	skipValidation := recoveryLSN != nil && len(remainingStreams) == 0 && *recoveryLSN == slot.LSN
+	if !skipValidation {
+		// validate global state (might got invalid during full load)
```
vaibhav-datazip: what's the need of this change ?

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,594 @@
...
+
+func (p *Parquet) local2PCPath() string {
+	return filepath.Join(p.config.Path, p.basePath, parquet2PCDir)
+}
+
+func (p *Parquet) localStagingPath(threadID string) string {
+	return filepath.Join(p.local2PCPath(), p.stagingDirName(threadID))
+}
+
+func (p *Parquet) stagingDataDir(basePath string) string {
```
vaibhav-datazip: can you change the variable name of the argument, also can you add short comments for the functions in this file

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,594 @@
...
+		if err != nil {
+			return nil, err
+		}
+		markers = append(markers, marker)
+	}
+	return markers, nil
+}
+
+func (p *Parquet) listS3CompletedMarkers(ctx context.Context) ([]parquet2PCCompletedMarker, error) {
+	prefix := p.s3ObjectPath(filepath.Join(p.basePath, parquet2PCDir)) + "/"
```
vaibhav-datazip: why are we saving staging files under _olake_2pc folder and not directly in the root folder

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,594 @@
...
+const (
+	parquet2PCDir           = "_olake_2pc"
+	parquet2PCCompletedFile = "_completed.json"
+	parquet2PCStagingSuffix = ".staging"
+)
+
+type parquet2PCCompletedMarker struct {
+	ThreadID      string
+	MetadataState *types.MetadataState
+	modTime       time.Time
```
vaibhav-datazip: do we need modTime ?

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,594 @@
...
+			}
+		}
+		if len(keys) == 0 {
+			return true
+		}
+		concurrency := min(len(keys), 8)
+		pageErr = utils.Concurrent(ctx, keys, concurrency, func(_ context.Context, key string, _ int) error {
+			return p.deleteS3Object(ctx, key)
+		})
+		return pageErr == nil
```
vaibhav-datazip: Retries should be added in this as well as other s3 related functions as well like we have in clearS3files

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,594 @@
...
+
+	markers := make([]parquet2PCCompletedMarker, 0)
+	for _, entry := range entries {
+		if !entry.IsDir() || !strings.HasSuffix(entry.Name(), parquet2PCStagingSuffix) {
+			continue
+		}
+		marker, err := p.readLocalCompletedMarker(entry.Name())
+		if os.IsNotExist(err) {
+			continue
+		}
```
vaibhav-datazip: why don't we create a map here only and not depend on list staging dirs function at all, we will mark the folders without completed marker as false otherwise true and data will be there

destination/parquet/parquet.go
siiddhantt: Done, renamed this to `stagedDataFiles()` and added a short comment. The helper returns the current thread’s staged parquet file identities used during close/upload/promotion, so the old `dataFiles()` name was too generic.

drivers/postgres/internal/cdc.go
siiddhantt: This is for the stale external state recovery case.

Scenario:
- Parquet commits destination progress at LSN B in `_completed.json`
- the Postgres slot is already at B
- OLake crashes before the external source state file is saved, so the next run still starts from old state A

Without this guard, Postgres fails on the A vs B LSN mismatch before destination metadata can repair the state. The skip is intentionally narrow: it only applies when destination metadata is ahead, all streams are already committed, and the replication slot is exactly at the destination-committed LSN. Other mismatches still fail as before.

destination/parquet/parquet_2pc.go
siiddhantt: Keeping staging under `_olake_2pc` is intentional. Staged files are not committed table data yet, so placing them under the table root can make raw Parquet readers see partial/uncommitted files. `_olake_2pc` keeps staging hidden but still table-local, and `DropStreams()` clears it naturally with the table prefix.

destination/parquet/parquet_2pc.go
siiddhantt: Yes, we need an ordering value because CDC/incremental/no-op syncs can leave multiple completed markers over time, and setup needs to restore the latest metadata state. I renamed `modTime` to `completedAt` and added a short comment to make that purpose clearer.

destination/parquet/parquet_2pc.go
siiddhantt: Agreed, added retries for the new S3 2PC operations using the same `RetryWithSkip` + `isRateLimitError` pattern used by existing Parquet S3 cleanup. This now covers staging upload, marker read/write, listing, copy, and delete paths.

destination/parquet/parquet_2pc.go
siiddhantt: Agreed, refactored this into a single staging-entry map. Each staging dir now maps to either a completed marker or nil, so setup can promote committed staging and delete incomplete staging without separately listing completed markers and staging dirs.

destination/parquet/parquet_2pc.go
siiddhantt: Done, renamed the argument to avoid confusion with `p.basePath`, and added short comments around the 2PC helpers.

vaibhav-datazip:
@siiddhantt once you are done with your changes after the review , please re-request the review from here so that I know the pr is ready again
Screenshot 2026-06-18 at 11 22 04 AM

siiddhantt:
@vaibhav-datazip sure thing, I've initiated a request thru that, and addressed the changes from earlier in the [e380d0b](https://github.com/datazip-inc/olake/pull/959/commits/e380d0b0d45ec9f7f453d49b3926eb78828dc264) commit

drivers/postgres/internal/cdc.go
vaibhav-datazip: similar bug was reported internally , PR #998 will be fixing it

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,608 @@
...
+	parquet2PCDir           = "_olake_2pc"
+	parquet2PCCompletedFile = "_completed.json"
+	parquet2PCStagingSuffix = ".staging"
+)
+
+type parquet2PCCompletedMarker struct {
+	ThreadID      string
+	MetadataState *types.MetadataState
+	// completedAt selects the latest CDC/incremental metadata marker.
+	completedAt time.Time
```
vaibhav-datazip: can't we just eliminate the completedAt field altogether.

here is the expected flow which we had discussed earlier 

sync completed -> all the data written -> promote the data to root -> now folder has just completed.json -> update the state and exit -> inside olake_2pc folder delete the folder with just one completed.json first and then proceed with rest of the sync

one idea which I have is during next sync when we see that state.json aligns with the data which is in metadata/ completed.json we just clear the whole file and folder associated with it and rest of the sync works normal , else we do the recovery path

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,608 @@
...
+		})
+		return err
+	})
+}
+
+// cleanupCommittedStaging removes local staging data after committed promotion.
+func (p *Parquet) cleanupCommittedStaging(threadID string) error {
+	if p.s3Client != nil {
+		return nil
+	}
```
vaibhav-datazip: can you add a comment here as well why its not needed for s3 case

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,608 @@
...
+	if os.IsNotExist(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	for idx := len(dirs) - 1; idx >= 0; idx-- {
+		dir := dirs[idx]
+		if err := os.Remove(dir); err != nil && !os.IsNotExist(err) && !os.IsExist(err) {
```
vaibhav-datazip: maybe os.IsExists part might be dead code , please check about that once as 

os.Remove never returns an EEXIST error  it only returns ENOENT (not exist), ENOTEMPTY (not empty), permission errors, etc. os.IsExist checks for EEXIST/EROFS, which os.Remove cannot emit. if its so you can remove it

destination/parquet/parquet_2pc_test.go
```diff
@@ -0,0 +1,315 @@
...
+	"testing"
+	"time"
+
+	"github.com/datazip-inc/olake/constants"
+	"github.com/datazip-inc/olake/destination"
+	"github.com/datazip-inc/olake/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoad2PCState(t *testing.T) {
```
vaibhav-datazip, line 28: can you add brief comments as well for the test cases in this file what they will do 

for reference you can check other unit test files like flatten_test, resolver_test etc.

drivers/postgres/internal/cdc.go
siiddhantt: Got it, I checked #998. It covers the broader multi-stream recovery case.
I’ll restore `cdc.go` to the staging version and leave the source-side recovery fix to #998 so #959 stays scoped to the Parquet destination.

destination/parquet/parquet_2pc_test.go
siiddhantt: Sure. I’ll add brief scenario comments to the table cases and top-level tests.

destination/parquet/parquet_2pc.go
siiddhantt: I checked this against Go’s implementation. `os.IsExist` is not technically dead here: `os.Remove` returns `ENOTEMPTY` for a non-empty directory, and Go maps `ENOTEMPTY` to `fs.ErrExist`, so `os.IsExist(err)` returns true.

However, after promotion these nested staging directories are expected to be empty. If one is still non-empty, silently ignoring it hides an incomplete cleanup. I’ll remove the exemption so the existing best-effort caller logs the condition.

destination/parquet/parquet_2pc.go
siiddhantt: I agree with the lifecycle you described: once the external state has durably caught up, the corresponding completed marker can be removed on a later sync.

I traced the current call path though. `state.json` is loaded and saved in `protocol/sync.go`, while `Parquet.Setup()` only receives the stream, schema, and `destination.Options`. Parquet can read `_completed.json`, but it cannot currently compare it with the external source state or prove that state persistence succeeded.

Until acknowledged markers can be deleted safely, multiple CDC/incremental metadata markers may remain. Removing `completedAt` would make selecting the recovery cursor/LSN depend on directory/map/S3 listing order, which is not commit ordering.

Supporting this cleanup flow would require either passing acknowledged source state into the destination or adding a post-state-save acknowledgement/cleanup hook. That is a wider interface change. For #959, I recommend retaining `completedAt` unless we want to expand the destination contract as part of this PR.

destination/parquet/parquet_2pc.go
siiddhantt: Agreed.  I’ll add a short comment here clarifying that.

destination/parquet/parquet_2pc.go
vaibhav-datazip: no , what I am meant is that, we don't expand post state save thing, instead in the next sync run can't we compare the fetched completed.json lsn values with state saved values and then delete them.

siiddhantt: Hi, I was working on the refactor and a quick update, before I push any changes and commit.

I tried the next-sync cleanup flow locally, as you suggested. The shape is working, but it touches a few more files than just Parquet because Parquet itself cannot compare LSN/cursor/binlog/resume-token semantics safely.

The local approach is:

- source/abstract layer compares the _completed.json metadata with the source state loaded at the start of the sync
- if saved source state has caught up, Parquet deletes that completed staging folder
- if the marker is still ahead, it stays and recovery uses it

One important thing I caught during the manual demo: the comparison has to use the source state loaded at sync start, not state mutated during the same run. Otherwise we can delete full-refresh markers too early before the external state file is durably saved.

One nuance: I have still kept completedAt internally for now, not as part of the commit/state protocol, but only as a fallback ordering helper. Since Parquet.Setup() currently returns a single types.MetadataState, if multiple CDC/incremental completed markers exist due to a previous failure or cleanup failure, Parquet still needs a deterministic way to select the recovery marker instead of relying on local/S3 listing order.

vaibhav-datazip: we can return the data similar to how we return it from iceberg , in that case there wont be a need of much change in the abstract layer right ?


<br>
<br>
<br>
<br>

> One important thing I caught during the manual demo: the comparison has to use the source state loaded at sync start, not state mutated during the same run. 

This is the exact thing I am suggesting , similar to how recovery is done currently for iceberg. we check the states before any sync related operation , if you implement the same for deletion there wont be a need to care about early deletes of chunks right ?

<br>
<br>
<br>
<br>

> I have still kept completedAt internally for now, not as part of the commit/state protocol, but only as a fallback ordering helper

you can think about the above solutions I have mentioned, I don't think this will be needed if those work. Because if cleanup failure occurs sync will stop and on next sync cleanup will take place

siiddhantt: Yes, I agree with the first part. Parquet already returns `_completed.json` metadata from `Setup()` in the same pre-sync recovery path where Iceberg returns `olake_2pc`, so I’ll keep recovery there and avoid any post-state-save hook.

Also agreed that cleanup/deletion must compare against the source state loaded at sync start. That fixes the early-delete issue from the manual demo, because we do not use state mutated during the same sync.

The remaining distinction from Iceberg is marker lifecycle. Iceberg stores one `olake_2pc` table property and overwrites it during table commit. Parquet stores per-thread `_completed.json` markers, so multiple CDC/incremental/no-op markers can exist until they are cleaned/compacted.

That is where `completedAt` is currently used: only as an internal fallback to choose the latest metadata marker if multiple completed markers are present. Without it, setup would depend on filesystem/S3 listing order.

Example:

`_olake_2pc/thread-a.staging/_completed.json` -> LSN/cursor 100  
`_olake_2pc/thread-b.staging/_completed.json` -> LSN/cursor 200

If source state is stale, recovery must pick 200. If both markers are still present and `completedAt` is removed, Parquet alone cannot know which one is latest unless cleanup/compaction has already guaranteed only one active metadata marker remains.

So I see two paths:

1. Narrow #959: keep `completedAt` internal, keep recovery destination-local, and leave acknowledged marker cleanup/compaction for a follow-up. This keeps the PR mostly in `destination/parquet` and is correct across full refresh, incremental, CDC, and no-op syncs.

2. Wider #959: remove `completedAt`, but add pre-sync acknowledged-marker cleanup/compaction. Full refresh cleanup is generic, but CDC/incremental comparison is source-specific: Postgres LSN, MySQL binlog, Mongo resume token, MSSQL LSN, or incremental cursor maps. So this needs an abstract/source-aware comparison path using the source state loaded before sync starts.

I’m fine with either direction or if there's another one to go for this.

vaibhav-datazip: > The remaining distinction from Iceberg is marker lifecycle. Iceberg stores one olake_2pc table property and overwrites it during table commit. Parquet stores per-thread _completed.json markers, so multiple CDC/incremental/no-op markers can exist until they are cleaned/compacted.


but we will be querying a particular thread folder's completed.json only right ? if that's the case , there should be just one completed file present for one thread, because rest of the files with same thread values will get deleted at start of the sync. be it cdc incremental or full_refresh so all paths will become generic right ?

siiddhantt: I think the small mismatch here is around “particular thread folder”.

For a single known thread folder, yes, there will only be one `_completed.json`, so no ordering issue there.

The reason I kept `completedAt` is that Parquet setup currently does not always know one previous thread folder to query directly. It discovers completed markers under `_olake_2pc` and rebuilds recovery state from them. 
That matters especially for CDC, because the CDC thread ID is currently generated with a ULID, so the next sync cannot derive the previous CDC thread folder from source state.

So the problem is not duplicate `_completed.json` files inside one thread folder. It is multiple completed thread folders existing over time, where setup has to choose the latest metadata marker without relying on local/S3 listing order.

vaibhav-datazip: so the thread id discrimination is the only problem in cdc sync right ?

in that case , we can change the way we are generating cdc thread id to make it deterministic , during initial development of 2pc for iceberg we were using, deterministic thread id for all sync modes but as it was unnecessary for cdc at that time that's why we had reverted it back to ulid.
you can also try that method to generate deterministic thread ids for cdc.

siiddhantt: Yes, agreed, CDC is the mode where the thread ID lookup is not deterministic today.

I’ll try the deterministic CDC thread ID approach. One thing I’ll keep careful is that the CDC thread ID should not be just the stream ID, otherwise every CDC run reuses the same staging folder and an old `_completed.json` could make a later failed attempt look committed.

I’ll derive it from the CDC source state loaded at the start of the sync, so a retry from the same stale state maps to the same marker folder, while a later sync after state advances gets a different logical folder.

If that works cleanly, I’ll remove `completedAt` and keep the PR narrower.

siiddhantt: Updated this in the latest commit.

I changed CDC thread IDs to be deterministic from the CDC source state loaded at sync start, before `PreCDC` mutates state. So a retry from the same stale state maps to the same `_completed.json` marker, while a later sync after state advances gets a new logical thread.

With that in place, Parquet no longer needs `completedAt`; setup now restores metadata only from the current thread’s completed marker instead of choosing by time/listing order.

drivers/abstract/cdc.go
```diff
@@ -119,7 +123,7 @@ func (a *AbstractDriver) streamChanges(mainCtx context.Context, pool *destinatio
 	dedupInserts := make(map[string]bool, len(streams))
 
 	for _, stream := range streams {
-		threadID := generateThreadID(stream.ID(), "")
+		threadID := generateThreadID(stream.ID(), cdcThreadHashes[stream.ID()])
```
vaibhav-datazip: As the generation of thread id changes there might be a need for backward compatibility test between staging and this branch

drivers/abstract/cdc.go
```diff
@@ -21,6 +22,9 @@ import (
 //   - Parallel: Process all streams simultaneously after all backfills complete
 //   - Concurrent: Start each stream's CDC immediately after its backfill completes (can overlap)
 func (a *AbstractDriver) RunChangeStream(mainCtx context.Context, pool *destination.WriterPool, streams ...types.StreamInterface) error {
+	// Hash the loaded source state before PreCDC mutates it, so CDC retries reuse the same destination thread.
+	cdcThreadHashes := a.cdcThreadHashes(streams)
+
```
vaibhav-datazip: Instead of hashing persisted state generically in `cdcThreadHashes`, shouldn't we introduce a driver-level helper (e.g. `GetCDCState` / `PersistedCDCCheckpoint`) that each driver implements and returns the relevant checkpoint from the loaded state file?

For example:
- Postgres → global LSN
- Mongo → resume token (`_data`)
- MSSQL → per-stream `lsn`
- MySQL → binlog position

and we can construct thread id by appending this 

I think that would be easier to read and debug (thread IDs would map directly to source position).

Worth considering if the added per-driver method is acceptable vs keeping this generic hash in abstract.

siiddhantt: Thanks @vaibhav-datazip , updated this now.

Instead of hashing the generic CDC state in the abstract layer, I changed it to use an optional `PersistedCDCCheckpoint` hook. Postgres/MySQL/Mongo/MSSQL return the readable persisted checkpoint they already use for CDC recovery, while Kafka does not implement the hook and keeps the existing generated thread IDs. 
That should avoid the Kafka/Iceberg shared-session issue while keeping deterministic CDC IDs only where the driver has a stable checkpoint.

Also added focused unit coverage for the abstract suffix behavior and each driver checkpoint shape.

drivers/abstract/cdc.go
```diff
@@ -119,7 +121,7 @@ func (a *AbstractDriver) streamChanges(mainCtx context.Context, pool *destinatio
 	dedupInserts := make(map[string]bool, len(streams))
 
 	for _, stream := range streams {
-		threadID := generateThreadID(stream.ID(), "")
+		threadID := generateThreadID(stream.ID(), cdcThreadSuffixes[stream.ID()])
```
vaibhav-datazip: if provider will be null then we wont get deterministic thread id right ? in that case we can encounter failure in recovery, for kafka this will happen

drivers/abstract/cdc.go
```diff
@@ -162,6 +164,31 @@ func (a *AbstractDriver) streamChanges(mainCtx context.Context, pool *destinatio
...
+			checkpoint = "initial"
+		}
+		threadSuffixes[stream.ID()] = checkpoint
+	}
+	return threadSuffixes
+}
+
 // mapInsertOpType returns the _op_type string for a CDC change.
 // Inserts emit "i" during the backfill overlap window (dedupInserts=true) and "c" otherwise.
 func mapChangeKindToOperationType(kind string, dedupInserts bool) string {
```
vaibhav-datazip: also can't we make this code more cleaner, like defining this in driver interface or returning the state directly from precdc

drivers/abstract/cdc.go
siiddhantt: Updated this.

Kafka now implements `PersistedCDCCheckpoint()` as well. It returns the persisted consumer group ID from state, then falls back to the configured consumer group ID. If neither exists, the abstract CDC flow uses the stable `initial` suffix, and parallel Kafka readers get `initial-reader[n]`, so retries do not fall back to random ULID thread IDs.

drivers/abstract/cdc.go
siiddhantt: Yeah, moved this into the driver interface path.

`PersistedCDCCheckpoint()` is now part of `DriverInterface`, so the abstract layer does not need the optional provider/type assertion anymore. The checkpoint is still read before `PreCDC()` because `PreCDC()` can initialize or mutate source state.

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,593 @@
...
+
+// metadataState normalizes close-time state into the MetadataState shape used for recovery.
+func (p *Parquet) metadataState(finalMetadataState any) (*types.MetadataState, error) {
+	var metadataState *types.MetadataState
+	if finalMetadataState == nil {
+		return metadataState, nil
+	}
+
+	switch state := finalMetadataState.(type) {
+	case *types.MetadataState:
```
hash-data: why we need to check pointer and normal state?

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,593 @@
...
+		if state == nil {
+			return metadataState, nil
+		}
+		stateCopy := *state
+		metadataState = &stateCopy
+	case types.MetadataState:
+		stateCopy := state
+		metadataState = &stateCopy
+	default:
+		stateFromPayload, err := types.SetMetadataState(finalMetadataState, p.options.ThreadID)
```
hash-data: we should not call this function from here, abstract only share what actually 2 pc state is

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,593 @@
...
+		if err != nil {
+			return nil, err
+		}
+		metadataState = stateFromPayload
+	}
+
+	if metadataState.ID == nil || fmt.Sprint(metadataState.ID) == "" {
+		metadataState.ID = p.options.ThreadID
+	}
```
hash-data: we dont need this kind of check

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,593 @@
...
+		if !knownIDs[id] {
+			state.FullRefreshCommittedIDs = append(state.FullRefreshCommittedIDs, id)
+			knownIDs[id] = true
+		}
+	}
+	return state, nil
+}
+
+// metadataState normalizes close-time state into the MetadataState shape used for recovery.
+func (p *Parquet) metadataState(finalMetadataState any) (*types.MetadataState, error) {
```
hash-data: i would say whatever abstract share just write that, and when sync start just take back. Like in iceberg we are doing

destination/parquet/parquet.go
```diff
@@ -301,62 +323,83 @@ func (p *Parquet) closePqFiles(ctx context.Context, _ any, closeOnError bool) er
...
 }
 
 func (p *Parquet) Close(ctx context.Context, finalMetadataState any) error {
-	// TODO: implement 2pc in parquet writer (difficulty: hard)
-	return p.closePqFiles(ctx, finalMetadataState, ctx.Err() != nil)
+	dataFiles := p.stagedDataFiles()
+	metadataState, err := p.metadataState(finalMetadataState)
+	if err != nil {
+		if closeErr := p.closePqFiles(ctx, true); closeErr != nil {
+			return closeErr
```
hash-data: we should also append the main error here

destination/parquet/parquet.go
```diff
@@ -301,62 +323,83 @@ func (p *Parquet) closePqFiles(ctx context.Context, _ any, closeOnError bool) er
...
-				Body:   file,
+		err := utils.Concurrent(ctx, dataFiles, concurrency, func(_ context.Context, info parquetDataFile, _ int) error {
+			err := utils.RetryWithSkip(ctx, 3, time.Minute, isRateLimitError, func(ctx context.Context) error {
+				file, err := os.Open(info.LocalPath)
+				if err != nil {
+					return fmt.Errorf("failed to open file %s: %s", info.LocalPath, err)
+				}
+				defer file.Close()
+
+				_, err = p.s3Uploader.UploadWithContext(ctx, &s3manager.UploadInput{
```
hash-data: comment for me, to check with old code

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,593 @@
...
+		return p.retryS3(ctx, func(ctx context.Context) error {
+			_, err := p.s3Client.PutObjectWithContext(ctx, &s3.PutObjectInput{
+				Bucket: aws.String(p.config.Bucket),
+				Key:    aws.String(p.s3ObjectPath(filepath.Join(p.basePath, parquet2PCDir, name))),
+				Body:   bytes.NewReader(data),
+			})
+			return err
+		})
+	}
+	return writeLocalFile(filepath.Join(p.local2PCPath(), name), data)
```
hash-data: why we are writing local file?

drivers/abstract/cdc.go
```diff
@@ -21,13 +21,15 @@ import (
 //   - Parallel: Process all streams simultaneously after all backfills complete
 //   - Concurrent: Start each stream's CDC immediately after its backfill completes (can overlap)
 func (a *AbstractDriver) RunChangeStream(mainCtx context.Context, pool *destination.WriterPool, streams ...types.StreamInterface) error {
+	isSequentialMode, isParallelMode, isConcurrentMode := a.driver.ChangeStreamConfig()
+	// Persisted checkpoints must be read before PreCDC mutates source state.
+	cdcCheckpoints := a.cdcCheckpoints(streams)
```
hash-data: we actually dont need it, it will make this complex instead of simplification. 
I would say let us follow this 
-> in full load there will be stream/olake_2pc/thread_id/files
-> in cdc stream/olake_2pc/files (as at a point in time the folder will be one only and it is unnecessary to maintain hierarchy)
-> in incremenatal also same

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,593 @@
...
+	})
+}
+
+// cleanupCommittedStaging removes local staging data after committed promotion.
+func (p *Parquet) cleanupCommittedStaging(threadID string) error {
+	if p.s3Client != nil {
+		// S3 has no empty directories; promotion deletes staged data objects and keeps _completed.json.
+		return nil
+	}
+	return p.cleanupLocalCommittedStaging(threadID)
```
hash-data: why we need this if we are renaming?

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,593 @@
...
+
+				stagingEntry := stagingEntries[parts[0]]
+				if stagingEntry.StagingDir == "" {
+					stagingEntry.StagingDir = parts[0]
+				}
+				if parts[1] == parquet2PCCompletedFile {
+					marker, err := p.readS3CompletedMarker(ctx, parts[0], *obj.Key)
+					if err != nil {
+						pageErr = err
+						return false
```
hash-data: in above comments i have asked a folder structure after that we not need this most probably

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,593 @@
...
+}
+
+type parquet2PCStagingEntry struct {
+	StagingDir string
+	Marker     *parquet2PCCompletedMarker
+}
+
+// load2PCState restores committed progress and reconciles staged attempts before setup returns.
+func (p *Parquet) load2PCState(ctx context.Context) (*types.MetadataState, error) {
+	stagingEntries, err := p.listStagingEntries(ctx)
```
hash-data:
```
Olake Parquet 2 pc flow simplification 
-> Current structure 
    -> stream/partition/data (staging branch)
-> Proposed structure 
    -> stream/
        -> olake_2pc/ 
            -> thread_id (for full load only) with empty finish.json
            -> partition/data_files (for cdc and incremental)
        -> partition/
            -> data files
        -> metadata.json (to have main 2 pc state, just like iceberg has)

-> Flow for full load + cdc 
    -> full load chunks will write data inside stream/olake_2pc/thread_id, after complete will write a constant file finish.json
    -> if data migration fails 
        -> next run will check thread id folder and its finish.json exist then do start migration, if not exist then check main metadata state, if folder not found then that chunk not even started
    -> if data migration success
        -> first update the main metadata.json with thread id
        -> after that remove the thread id folder 

-> Flow For CDC/Incremental 
    -> data will be written in stream/olake_2pc directly not other hierarchy, on complete it will write finish.json which will have finished state 
    -> if data migration fails
        -> next sync will check olake_2pc folder and if there finish.json exist start migration, and return that, if not exist then return main metadata.json state

```

hash-data: let me know views on it

hash-data (commented):
I would say great efforts. Some of the things need to be simplified. I have added my comments. Also try to simplify logic where we have to create staging dir etc.

hash-data:
One more case: when we are renaming the folder in localtest, we are not making sure that next time finish.json will exist or not. Also, how do we get to know if chunk got finished in that case?

destination/parquet/parquet_2pc.go
siiddhantt: Thanks @hash-data , I’m refactoring around the simplified flow you suggested.

`<stream>/metadata.json` will now be the only durable Parquet 2PC state. Each writer stages its files under `_olake_2pc/<thread-id>/`, writes `finish.json` once staging is durable, promotes the data, updates `metadata.json`, and then removes staging. 
During setup, finished staging is rolled forward, unfinished staging from the previous run is removed, and the durable metadata is returned to the abstract layer.

`finish.json` will now store the `MetadataState` supplied by abstract directly. I’m removing the previous completed-marker/`completedAt` model and reverting the deterministic CDC checkpoint changes.

I'm keeping transient staging isolated per writer for CDC/incremental as well. Kafka parallel mode can create multiple writers for the same stream concurrently, so a shared `_olake_2pc/finish.json` could let one writer promote or remove another writer’s unfinished files.

destination/parquet/parquet_2pc.go
siiddhantt: Addressed. The pointer/value type switch is gone. `finishState()` serializes what abstract supplies.

destination/parquet/parquet_2pc.go
siiddhantt: Got it. Metadata construction remains in abstract, so Parquet only persists and reads `types.MetadataState`

destination/parquet/parquet_2pc.go
siiddhantt: Removed the missing-ID check and Parquet-side ID patching. The abstract layer now remains responsible for constructing the metadata state.

destination/parquet/parquet_2pc.go
siiddhantt: Implemented this now. `finish.json` stores the state supplied by abstract, and `Setup()` returns the durable state from `metadata.json`.

destination/parquet/parquet.go
siiddhantt: Updated. If state serialization fails, `Close()` still closes the Parquet files and preserves both the original error and any close error.

destination/parquet/parquet_2pc.go
siiddhantt: This handles the local Parquet destination as well as S3. Local metadata is written through a temporary file and atomic rename, while the S3 path uses `PutObject`.

drivers/abstract/cdc.go
siiddhantt: Removed the deterministic CDC checkpoint and thread-ID changes.

I kept transient per-writer staging for CDC/incremental because the current Kafka parallel-reader path can create multiple writers for the same stream. A shared `finish.json` could let one writer promote or remove another writer’s in-progress files. The thread directory is only temporary isolation and `metadata.json` remains the durable state.

destination/parquet/parquet_2pc.go
siiddhantt: Simplified this. Local promotion now renames only the data files and leaves `finish.json` in staging until `metadata.json` is updated. The entire staging directory is removed afterward, so separate empty-directory cleanup is no longer needed.

destination/parquet/parquet_2pc.go
siiddhantt: Simplified this to one staging-entry scan. It is still needed to discover transient per-writer directories, but it now only distinguishes finished staging from unfinished staging.

siiddhantt:
> One more case: when we are renaming the folder in localtest, we are not making sure that next time finish.json will exist or not. Also, how do we get to know if chunk got finished in that case?

Local promotion now deliberately excludes `finish.json` from the rename. Only data files move to their final paths.

If the process stops after promotion but before updating `metadata.json`, `finish.json` remains in staging. The next setup sees it, completes the metadata commit, and then removes the staging directory. There is focused coverage for this failure boundary.

drivers/abstract/incremental.go
```diff
@@ -116,12 +116,22 @@ func (a *AbstractDriver) Incremental(mainCtx context.Context, pool *destination.
...
+				if err != nil {
+					return fmt.Errorf("failed to reformat metadata secondary cursor value: %s", err)
+				}
+				if isMetadataCursorAhead(metadataPrimaryCursorValue, metadataSecondaryCursorValue, maxPrimaryCursorValue, maxSecondaryCursorValue, secondaryCursor != "") {
+					logger.Infof("Stream[%s] cursor(s) mismatch, updating cursor(s) in state", stream.ID())
+					maxPrimaryCursorValue = metadataPrimaryCursorValue
+					if secondaryCursor != "" {
+						maxSecondaryCursorValue = metadataSecondaryCursorValue
+					}
+					return nil
```
vaibhav-datazip: why do we need this logic , aren't we saving the cursor fields in metadata file after reformat , so can we use then directly ?

also can there be a case when state cursor is ahead of metadata cursor ?

siiddhantt: hi @vaibhav-datazip, I checked this again for both Iceberg and Parquet. This decision needs to remain in the abstract incremental flow because the destination can only return its saved checkpoint; abstract is the layer that has both the destination checkpoint and the current source state and can decide whether recovery is actually needed.

A matching thread ID alone is not enough. It can mean either the destination is ahead because the external state save failed, or that the previous sync was a no-op and both cursors are equal. In the first case we recover from metadata; in the second we must continue reading, otherwise a later sync could skip newly added records. This is why we only return early when the metadata cursor is strictly ahead.

The metadata cursor is reformatted because it comes back through JSON and may no longer have the source column’s Go type. If source state is equal to or ahead of metadata, we continue normally and never move it backward. So keeping this in abstract gives Iceberg and Parquet the same recovery behavior, while `parquet_2pc.go` remains responsible only for persisting and returning metadata.

vaibhav-datazip: @siiddhantt 

> It can mean either the destination is ahead because the external state save failed, or that the previous sync was a no-op and both cursors are equal

how can a matching thread id mean that earlier sync was no op ? , because if no new parquet file was written we wont be committing the metadata, right ? so thread id of old sync wont be matching the new sync . you can go through [incremental section of this blog ](https://olake.io/blog/exactly-once-delivery-iceberg/#incremental). 


> The metadata cursor is reformatted because it comes back through JSON and may no longer have the source column’s Go type. 

even in the iceberg side we are fetching the data from `metadata.json` file , so why do we require it now ?

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,552 @@
...
+			for _, object := range page.Contents {
+				if object.Key == nil {
+					continue
+				}
+				relativePath := strings.TrimPrefix(*object.Key, prefix)
+				parts := strings.SplitN(relativePath, "/", 2)
+				if len(parts) != 2 {
+					continue
+				}
+
```
vaibhav-datazip: I don't think this will allow us to follow the structure which was discussed in this comment 

https://github.com/datazip-inc/olake/pull/959#discussion_r3551028543

as it should be, staging files of cdc and incremental should be inside olake_2pc directory directly and hence finish.json will be the only part (len = 1) . but here we are expecting 2 parts otherwise not marking it as finished

drivers/abstract/incremental.go
siiddhantt: You’re right about it. I checked the Iceberg commit path again and `commitThread()` returns before updating olake_2pc when no files were produced. Parquet currently differs by committing metadata on a no-file close, and that is what created the equal-cursor case I was reasoning about.
I’ll align Parquet with Iceberg: a CDC/incremental writer with no staged data files will not write finish.json or update metadata.json. Full-refresh empty chunks will still record their chunk ID. With that behavior, a matching incremental thread ID represents the recovery case directly, so I’ll revert the extra cursor comparison/reformat logic in abstract.

destination/parquet/parquet_2pc.go
siiddhantt: @vaibhav-datazip I checked this against Ankit’s layout after rebasing the latest staging changes, including Kafka 2PC. The current parser expects a writer-scoped directory because Kafka runs parallel streamChanges() calls, and each call can create a Parquet writer for the same stream. Setup and Close are serialized, but file writes happen before that lock, so a shared root finish.json could include files from another writer that is still active.
The direct layout is safe only if OLake guarantees one active CDC/incremental writer per destination table. Otherwise, we should retain writer-scoped staging or add coordination above Parquet. Can we confirm which behavior the Parquet 2PC implementation should support?

siiddhantt:
hi @hash-data and @vaibhav-datazip, I've pushed the refactor that we aligned on. S3 now uses per-thread staging for full refresh and shared staging for CDC/incremental, with `finish.json` for roll-forward recovery and table-level `metadata.json` for durable state. 
The local behavior remains unchanged and I've marked it as TODO. I've ran build and lint tests and also tried it out with a 100k-row Postgres + MinIO recovery demo. Do lmk if any changes are required.

utils/testutils/test_utils.go
```diff
@@ -378,6 +378,29 @@ func DeleteParquetFiles(t *testing.T, parquetDB, tableName string) error {
...
+		deletedCount++
+	}
+
+	metadataPath := parquetPath + "metadata.json"
+	if err := minioClient.RemoveObject(ctx, bucketName, metadataPath, minio.RemoveObjectOptions{}); err != nil {
+		return fmt.Errorf("failed to delete %s: %s", metadataPath, err)
+	}
+	deletedCount++
+
 	t.Logf("--- Cleanup Complete: Deleted %d files ---", deletedCount)
```
vaibhav-datazip: are you sure we should be deleting metadata.json every time , because this runs for each test case. you check from what all places this function gets called from 

so for one sync this will be deleting after full refresh, after cdc each operation like insert , update , delete etc.

vaibhav-datazip (commented):
please check there are some merge conflicts

utils/testutils/test_utils.go
siiddhantt: Good catch. I’ll keep the per-operation helper data-only and clear the full Parquet table/2PC state once at the start of each test suite instead. This way between insert/update/delete subtests, we delete only the Parquet data files and retain metadata.json

I’ll also resolve the new conflicts by integrating rolled files into the existing staging and commit flow.

destination/parquet/parquet.go
```diff
@@ -93,8 +96,25 @@ func (p *Parquet) initS3Writer() error {
...
+		return fmt.Errorf("failed to get relative parquet partition path: %s", err)
+	}
+
 	directoryPath := filepath.Join(p.config.Path, basePath)
+	if p.s3Client != nil {
+		if p.tempDir == "" {
+			if err := os.MkdirAll(p.config.Path, os.ModePerm); err != nil {
+				return fmt.Errorf("failed to create parquet temp path[%s]: %s", p.config.Path, err)
+			}
+			p.tempDir, err = os.MkdirTemp(p.config.Path, "olake-parquet-*")
```
hash-data: let us make this constant olake-parquet-*

hash-data:
hi @siiddhantt the current structure looks good, will review its correctness as well. Thanks for updating the structure

siiddhantt:
Thanks @hash-data, I’ll wait for the correctness review.

hash-data:
@siiddhantt  can you take pull of staging once?

siiddhantt:
@hash-data sure thing, I've resolved the coflicts and merged the latest staging branch.

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+			threadID, err := decodeThreadID(parts[0])
+			if err != nil {
+				return nil, err
+			}
+			entry = &parquet2PCStagingEntry{
+				prefix:   rootPrefix + parts[0] + "/",
+				threadID: threadID,
+			}
+			entries[parts[0]] = entry
+		}
```
hash-data: can we add an example here how the state look like at this step

hash-data: like what will be part 0 and part 1

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+	}
+
+	stagingEntries := make([]parquet2PCStagingEntry, 0, len(entries))
+	for _, entry := range entries {
+		stagingEntries = append(stagingEntries, *entry)
+	}
+	sort.Slice(stagingEntries, func(i, j int) bool {
+		return stagingEntries[i].prefix < stagingEntries[j].prefix
+	})
```
hash-data: by the way key iteration happened on sort order only, do we still need it ?

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+		return err
+	}
+
+	for _, entry := range entries {
+		if !entry.finished {
+			if err := p.deleteS3Prefix(ctx, entry.prefix); err != nil {
+				return err
+			}
+			continue
+		}
```
hash-data, line 68: add comment that this is when there is failure happened on prev sync

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+		if err != nil {
+			return err
+		}
+		finishedState, err := parseFinishState(data)
+		if err != nil {
+			return err
+		}
+		if finishedState != nil {
+			return fmt.Errorf("parquet 2pc metadata state is not supported in full-refresh staging")
+		}
```
hash-data: why we even checking it?

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+		if err := p.promoteS3Staging(ctx, entry.prefix); err != nil {
+			return err
+		}
+		if err := p.commitMetadata(ctx, entry.threadID, finishedState); err != nil {
```
hash-data: also are we just passing nil state here?

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+	if state == nil {
+		state = &types.MetadataState{}
+	}
+
+	if finishedState == nil {
+		if !slices.Contains(state.FullRefreshCommittedIDs, threadID) {
+			state.FullRefreshCommittedIDs = append(state.FullRefreshCommittedIDs, threadID)
+		}
+		dedupInserts := true
+		state.DedupInserts = &dedupInserts
```
hash-data: adding same nil state checks here?

hash-data: can you add comments?

destination/parquet/parquet_2pc.go
hash-data: also do we need to parse the finish state or directly we can pass nil?

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+func (p *Parquet) deleteS3Prefix(ctx context.Context, prefix string) error {
+	keys, err := p.listS3Keys(ctx, prefix)
+	if err != nil {
+		return err
+	}
+	if len(keys) == 0 {
+		return nil
+	}
+	return utils.Concurrent(ctx, keys, min(len(keys), 8), func(deleteCtx context.Context, key string, _ int) error {
+		return p.deleteS3Object(deleteCtx, key)
```
hash-data: what if finish.json get deleted first? and after that it fails

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+	prefix := p.stagingRootPrefix()
+	keys, err := p.listS3Keys(ctx, prefix)
+	if err != nil {
+		return err
+	}
+	if len(keys) == 0 {
+		return nil
+	}
+
+	finishKey := prefix + parquet2PCFinishFile
```
hash-data: let us create function for these, 
like: 
- cdcFinishFilePath
- BackfillFinishFilePath
- metadataFilePath (i think already present)

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+		state, err := mergeState(current.State, next.State)
+		if err != nil {
+			return err
+		}
+		current.State = state
+	}
+	for _, threadID := range next.FullRefreshCommittedIDs {
+		if !slices.Contains(current.FullRefreshCommittedIDs, threadID) {
+			current.FullRefreshCommittedIDs = append(current.FullRefreshCommittedIDs, threadID)
+		}
```
hash-data: we dont need to merge these right? merge only run for cdc level state?

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+	}
+
+	if finishedState == nil {
+		if !slices.Contains(state.FullRefreshCommittedIDs, threadID) {
+			state.FullRefreshCommittedIDs = append(state.FullRefreshCommittedIDs, threadID)
+		}
+		dedupInserts := true
+		state.DedupInserts = &dedupInserts
+	} else if err := mergeMetadataState(state, finishedState); err != nil {
+		return err
```
hash-data: do we need to merge state or directly write what finish.json have?

hash-data: what do you say?

destination/parquet/parquet.go
```diff
@@ -366,66 +368,137 @@ func (p *Parquet) closePqFiles(ctx context.Context, _ any, closeOnError bool) er
...
+
+	if err := p.closePqFiles(ctx.Err() != nil); err != nil {
+		return err
+	}
+	p.partitionedFiles = make(map[string][]*FileMetadata)
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
+	if err := p.recoverStaging(ctx); err != nil {
```
hash-data, line 487: any reason to call it from here, we need to make sure when this close is called that means there is no prev sync data present

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+	var awsErr awserr.Error
+	return errors.As(err, &awsErr) && (awsErr.Code() == s3.ErrCodeNoSuchKey || awsErr.Code() == "NotFound")
+}
+
+func (p *Parquet) retryS3(ctx context.Context, fn func(context.Context) error) error {
+	return utils.RetryWithSkip(ctx, 3, time.Minute, isRateLimitError, fn)
+}
+
+func (p *Parquet) stagingRootPrefix() string {
+	return p.s3ObjectPath(path.Join(p.basePath, parquet2PCDir)) + "/"
```
hash-data: where we are using s3 prefix? as sometime permission only get on path

destination/parquet/parquet.go
```diff
@@ -366,66 +368,137 @@ func (p *Parquet) closePqFiles(ctx context.Context, _ any, closeOnError bool) er
...
+			if err != nil {
+				return fmt.Errorf("failed to open file %s: %s", info.path, err)
+			}
+			defer file.Close()
 
-		err := utils.Concurrent(ctx, filesToUpload, concurrency, func(_ context.Context, info uploadInfo, _ int) error {
-			return p.uploadLocalFileToS3(info.s3KeyPath, info.filePath)
+			_, err = p.s3Uploader.UploadWithContext(retryCtx, &s3manager.UploadInput{
+				Bucket: aws.String(p.config.Bucket),
+				Key:    aws.String(stagingKey),
```
hash-data, line 414: s3prefix?

destination/parquet/parquet_2pc.go
```diff
@@ -0,0 +1,479 @@
...
+	if err := p.promoteS3Staging(ctx, prefix); err != nil {
+		return err
+	}
+	if err := p.commitMetadata(ctx, p.options.ThreadID, finishedState); err != nil {
+		return err
+	}
+	return p.deleteS3Prefix(ctx, prefix)
+}
+
+func finishState(finalMetadataState any) ([]byte, *types.MetadataState, error) {
```
hash-data: add comments for this function

hash-data (commented):
After checking the arch it looks really good, while reading the code it was very smooth. 

Some of my comments, we can discuss them, I think most of them are just minor suggestions. 

I am still thinking of some of edge cases will ping if find any. 

Again thanks for the pr

destination/parquet/parquet.go
siiddhantt: Done, I've extracted this to `parquetTempDirPattern`

destination/parquet/parquet_2pc.go
siiddhantt: Sure yeah, I've added an example explaining that parts[0] is the encoded thread directory and parts[1] is the staged file path

destination/parquet/parquet_2pc.go
siiddhantt: good catch, i've emoved the extra sort. The entries are now built directly from the sorted S3 keys.

destination/parquet/parquet_2pc.go
siiddhantt: unfinished staging belongs to a previous attempt that did not reach finish.json, ive added a comment clarifying it

destination/parquet/parquet_2pc.go
siiddhantt: added a short comment for it!

destination/parquet/parquet_2pc.go
siiddhantt: i've made it such that all 2PC keys now go through `s3ObjectKey()`, which prepends `config.Prefix`. This covers staging, finish, metadata, promotion, and cleanup keys.

destination/parquet/parquet.go
siiddhantt: this is needed for Kafka writer handoff. Multiple writers can finish setup before either closes; if the first close fails after writing finish.json, the next serialized close must roll that staging forward before reusing the shared prefix. The new test covers this exact case.

destination/parquet/parquet_2pc.go
siiddhantt: i think direct replacement would work for a single writer, but parallel Kafka readers can commit offsets for different partitions of the same stream. Replacing the state would drop the earlier reader’s offsets, so I kept a one-level merge only for the inner stream state and added a test for two readers.

destination/parquet/parquet_2pc.go
siiddhantt: yes, correct for full-refresh IDs. I removed them from the stream merge; they are now updated only through commitBackfillMetadata()

destination/parquet/parquet_2pc.go
siiddhantt: got it! i've added `sharedFinishObjectKey()`, `backfillFinishObjectKey()`, and `metadataObjectKey()`. I used “shared” because the same finish path is used by both CDC and incremental.

destination/parquet/parquet_2pc.go
siiddhantt: in that case it would lose the recovery marker, so cleanup now deletes all staged data first and finish.json last. I also added a test that fails data deletion and verifies the marker remains for retry

destination/parquet/parquet_2pc.go
siiddhantt: that nil branch existed because one function handled both backfill and stream commits. After splitting them, backfill does not pass a state and the stream commit cannot receive nil.

destination/parquet/parquet_2pc.go
siiddhantt: not anymore. I split the commit paths like this -> backfill calls `commitBackfillMetadata(threadID)`, while CDC/incremental pass a concrete MetadataState to `commitStreamMetadata()`

destination/parquet/parquet_2pc.go
siiddhantt: good point, full-refresh finish.json is only an existence marker. I removed the read and parse entirely so recovery now promotes the files and commits the thread ID directly.

destination/parquet/parquet.go
siiddhantt: the upload uses stagingObjectKey(), which reaches s3ObjectKey() through the staging prefix, so the configured S3 path is included. i've added a recovery test that verifies every object stays under that prefix.

vaibhav-datazip:
## Manually tested

all for postgres
- [x] CDC records written but finish.json not present yet and failure occurred
- [x] CDC sync fails during LSN acknowledgment
- [x] CDC sync fails before finish.json was deleted
- [x] CDC sync fails before writing metadata
- [x] CDC sync fails before promoting the files from staging
- [x] LSN moves ahead in logs
- [x] Happy scenario, 2 syncs running on full refresh + CDC
- [x] Backfill chunk killed after finish.json was written, then re-run


## Additional automated runs


Run for postgres (CDC, incremental, backfill, partitioned), mysql (CDC), mongodb (CDC), oracle (incremental  its only non-full-refresh mode):

- [x] Sync fails mid-upload, before finish.json
- [x] Sync fails after upload, before finish.json
- [x] Sync fails after finish.json, before promoting from staging
- [x] Sync fails after promote, before writing metadata
- [x] Sync fails after writing metadata, before deleting staging
- [x] Sync fails after LSN acknowledgment, before saving state — postgres CDC only


### Partitioned table, YYYY/MM folders — postgres

Same five windows again, on both staging layouts:

- [x] Backfill chunk killed at each of the five windows
- [x] CDC sync killed at each of the five windows
- [x] One record inserted per sync, killed at each window in turn on the same table
- [x] LSN moves ahead
- [x] Happy scenario, 2 syncs on full refresh + CDC


### Corrupt destination state — postgres
- [x] empty metadata.json planted
- [x] Foreign file planted in the staging root





- [ ] s3 prefix change
- [ ] Drop streams

destination/parquet/parquet.go
```diff
@@ -366,66 +370,143 @@ func (p *Parquet) closePqFiles(ctx context.Context, _ any, closeOnError bool) er
...
+	if !p.options.Backfill && len(dataFiles) == 0 {
+		if err := p.closePqFiles(true); err != nil {
+			return err
+		}
+		p.partitionedFiles = make(map[string][]*FileMetadata)
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		return nil
+	}
```
hash-data, line 462: can you explain this why we are doing it?

destination/parquet/parquet.go
hash-data: can we connect for this discussion ?

destination/parquet/parquet.go
siiddhantt: good catch @hash-data Those cleanup lines are leftovers from the earlier flow and are now redundant because `dataFiles` is already empty. I’ll remove them. 
Also the early return remains to avoid committing metadata for empty CDC/incremental runs.

siiddhantt:
> can we connect for this discussion ?

@hash-data  sure we can discuss this, do lmk if you'd want to schedule a meet at a convenient time.

also here although `Setup()` handles staging left by an earlier run, this call handles writer handoff within the same run. Parallel Kafka writers can both finish `Setup()` before either closes and even though their `Close()` calls are serialized, the first writer may still leave shared staging on failure, so the next writer must resolve it before uploading its own files.

vaibhav-datazip:
hi @siiddhantt , the integration test for s3 is failing while writing to pq destination, this might be due to minio limit of 256 bytes name . please check this once , and also in what all other cloud services such problem can occur

siiddhantt:
Hi @vaibhav-datazip, I reproduced the issue locally and you're right Base64 expanded the thread directory to 264 bytes, exceeding MinIO’s limit. I’ve pushed 63225fef, using a fixed 64-character SHA-256 directory while retaining the original thread ID in `finish.json`. This also avoids differing key limits across S3-compatible providers and the MinIO regression and recovery flow pass locally.

vaibhav-datazip:
manual
- [x] postgres: cdc: manual destination name: failed before finish.json written, in backfill 
- [x] postgres: cdc: manual destination name: failed after finish.json written
- [x] postgres: cdc: with partition regex, and custom destination
automated testing 

automated
- [x] kafka, mongodb, s3, mysql, postgers, with cdc incremental, and full refresh , failing before state update : tested on both minio and gcp bucket


### s3 source (the case that was failing in CI)

- [x] s3: minio: full refresh: multi-file chunk with long stream name — no `XMinioInvalidObjectName`, staging drained
- [x] s3: minio: full refresh: failed after finish.json
- [x] s3: minio: full refresh: failed after promotion, before metadata.json
- [x] s3: minio: incremental: happy + failed after finish.json + before metadata.json
- [x] s3: minio: incremental: recovery through a hashed staging dir — thread id read back from finish.json, chunk skipped
- [x] s3: gcp: full refresh: happy + failed after finish.json
- [x] s3: gcp: incremental: happy + failed after finish.json
- [x] s3: gcp: staging shape confirmed `_olake_2pc/<sha256>/finish.json` with the full thread id inside

### mongodb (all sync modes, both destinations)

- [x] mongodb: minio: full refresh: happy + failed after finish.json + before metadata.json
- [x] mongodb: minio: cdc: happy + failed after finish.json + before metadata.json
- [x] mongodb: minio: cdc: with real churn (insert/update/delete), failed after finish.json
- [x] mongodb: minio: strict cdc: with churn, failed after finish.json — changes only, no backfill leg
- [x] mongodb: minio: incremental: happy + failed after finish.json + before metadata.json + with churn
- [x] mongodb: gcp: full refresh / cdc / incremental: happy
- [x] mongodb: gcp: full refresh: failed after finish.json
- [x] mongodb: gcp: cdc: failed after finish.json
- [x] mongodb: gcp: strict cdc: with churn, failed after finish.json

### kafka (strict cdc)

- [x] kafka: minio: 6 partitions / 6 parallel writers: happy — all partition offsets merged into one metadata.json
- [x] kafka: minio: failed after some data files uploaded
- [x] kafka: minio: failed before finish.json
- [x] kafka: minio: failed after finish.json
- [x] kafka: minio: failed mid-promotion
- [x] kafka: minio: failed after promotion, before metadata.json
- [x] kafka: minio: failed after metadata.json, before staging cleanup

### postgres: partition regex with multiple chunks (minio)

- [x] postgres: minio: partition on year/month: mixed state — some chunks committed, one staged, rest pending → resume merged all three correctly
- [x] postgres: minio: staging mirrors the partition tree under the hash dir, promotion merges chunk subtrees into the same final dirs
- [x] postgres: minio: only one staged chunk dir exists at a time (per-stream Setup/Close serialization)
- [x] postgres: minio: partition on year only, all sync modes

vaibhav-datazip (approved):

hash-data (approved):

hash-data:
Thanks for the contribution @siiddhantt

siiddhantt:
Thanks @hash-data, @vaibhav-datazip, and @nayanj98. Appreciate all the feedback and testing that helped shape the final implementation.

PR #961, opened 2026-05-24
feat(s3): sample first and last files for schema inference

### LINKED ISSUE #739: : implement sampling of records from first and last files to get a more accurate schema
Collect sample records using smart JSON format detection

---

https://github.com/datazip-inc/olake/blob/298b31d3400e047b4e9c860c58f8f755001263f9/pkg/parser/json.go#L33-L38

---

###### This issue was generated by [todo-issue](https://github.com/DerJuulsn/todo-issue) based on a `TODO` comment in 298b31d3400e047b4e9c860c58f8f755001263f9. It&#x27;s been assigned to @hash-data because they committed the code.

Samyak-jain7:
# Description

Fixes #739.

This updates S3 JSON schema inference so discovery samples the oldest and newest files in a stream using `LastModified` instead of inferring from a single file or relying on lexicographic `FileKey` order.

The multi-sample path is JSON-only:

- the S3 driver selects first/last JSON sample files by `LastModified`
- each sampled file is opened as an independent reader
- the JSON parser parses each reader with the existing smart JSON format detection path, so JSONL, JSON arrays, and single-object JSON keep their own parsing behavior
- sampled JSON records are appended and resolved once with `typeutils.Resolve`

CSV and parquet schema inference stay on the existing single-reader path. While switching JSON inference to one `Resolve` call, this also fixes `Resolve` to use its existing common-type calculation at schema write time so mixed JSON field types in one inference call resolve correctly instead of panicking.

## Type of change

# How Has This Been Tested?

- [x] `go test -count=1 ./...`
- [x] `cd drivers/s3 && go test -count=1 ./...`
- [x] `go build ./...`
- [x] `cd drivers/s3 && go build ./...`
- [x] `go vet ./...`
- [x] `cd drivers/s3 && go vet ./...`
- [x] `go test -race -count=1 ./pkg/parser`
- [x] `cd drivers/s3 && go test -race -count=1 ./internal`
- [x] `gofmt -l drivers/s3/internal/s3.go drivers/s3/internal/s3_test.go pkg/parser/json.go pkg/parser/json_test.go utils/typeutils/resolver.go utils/typeutils/resolver_test.go`
- [x] `git diff --check upstream/staging...HEAD`
- [x] Link check: narrated video, raw terminal video, and validation release page all returned HTTP 200.
- [x] Local MinIO validation: uploaded two JSON files where the first file contains `first_only` and the last file contains `last_only`, ran OLake S3 `discover`, confirmed both sampled files were used and both fields appeared in the catalog with nullable schema types, then ran OLake S3 `sync` and confirmed it completed successfully with 2 records written to local Parquet output.

# Screenshots or Recordings

Narrated validation video: https://github.com/Samyak-jain7/olake/releases/download/olake-739-validation-20260524/olake-739-validation-narrated.mp4

Raw terminal validation video: https://github.com/Samyak-jain7/olake/releases/download/olake-739-validation-20260524/olake-739-terminal-sync-screenrecording.mov

Release page: https://github.com/Samyak-jain7/olake/releases/tag/olake-739-validation-20260524

## Documentation

## Related PR's (If Any):

None.


Samyak-jain7:
Hi @nayanj98, could you please review this PR when you get a chance? The CLA is signed, tests are listed in the PR body, and I added the narrated validation video as requested. Thanks!

nayanj98:
@Samyak-jain7 Great! I will checkout the video and also try to assign a reviewer to this PR as soon as possible.

nayanj98:
@Samyak-jain7 Can you please attach a simple and raw video of sync running and completing successfully. Thank you.

Samyak-jain7:
Hi @nayanj98, sure, here is a simple macOS Terminal screen recording of the sync run completing successfully:

https://github.com/Samyak-jain7/olake/releases/download/olake-739-validation-20260524/olake-739-terminal-sync-screenrecording.mov

It runs the S3 sync locally against MinIO, processes both files, writes 2 records, shows `Total records read: 2`, `Sync completed`, and exits with code 0. Thanks!

nayanj98:
@Samyak-jain7 Thanks a lot! Will get back to you on this soon.

nayanj98:
hey @Samyak-jain7 Assigning @vikaxsh to review your PR

Samyak-jain7:
Hi @vikaxsh, did you get a chance to take a look?

vikaxsh:
hi @Samyak-jain7
we can drop those merge funcs imo — let the json parser take multiple readers, parse records from each using the existing `parseJSONContent` (so JSONL / Array / single-object all work in their own paths), append into one big slice, and call `typeutils.Resolve(stream, allRecords...)` **once**. Resolve's existing "field seen before, missing in current record → nullable" rule then fires across files for free. driver side is just "open N readers, hand them all to parser."

we only need the multi-reader path for json — for parquet and csv we'll keep using a single reader as before. open to a cleaner shape if you see one but that's the gist.
- pick first/last by `LastModified` instead of lexicographic `FileKey`. key-based sort only matches "oldest/newest" if files are time-prefixed; for hash/uuid keys it's basically random. `LastModified` is already on `FileObject`.


wdyt?

Samyak-jain7:
Thanks, agreed with this shape. I pushed an update that:

- keeps CSV/parquet schema inference on the single-reader path
- for JSON, picks sample files by `LastModified`, opens the sample readers independently, parses each with the existing JSON format detection path, then resolves once across the collected records
- removes the custom sample-schema merge/nullability helpers

While switching JSON inference to one `Resolve` call, I also fixed `Resolve` to use its existing common-type calculation at schema write time so mixed JSON field types in the same call do not panic.

Tested:
- `go test ./pkg/parser -run 'TestJSONParser_InferSchema|TestJSONParser_InferSchemaFromReaders'`
- `go test ./drivers/s3/internal -run TestSchemaSampleFiles`
- `go test ./utils/typeutils -run TestResolve`
- `go test ./drivers/s3/... ./pkg/parser ./utils/typeutils`

Samyak-jain7:
Hi @vikaxsh, I rebased this on the latest `staging` and re-ran validation. Could you please take another look when you get a chance?

drivers/s3/internal/s3.go
```diff
@@ -291,20 +292,21 @@ func (s *S3) ProduceSchema(ctx context.Context, streamName string) (*types.Strea
...
-	firstFile := files[0]
-	logger.Infof("Inferring schema from file: %s (%d files in stream)", firstFile.FileKey, len(files))
-
 	var inferredStream *types.Stream
 	var err error
 
-	// Create appropriate parser and infer schema (format-specific logic from parser package)
 	switch s.config.FileFormat {
 	case FormatCSV:
+		firstFile := files[0]
```
saksham-datazip: we can simply avoid this extra variable and directly pass it into log and inferSchemaForCSV and inferSchemaForParquet functions

drivers/s3/internal/s3.go
```diff
@@ -291,20 +292,21 @@ func (s *S3) ProduceSchema(ctx context.Context, streamName string) (*types.Strea
...
+		logger.Infof("Inferring schema from file: %s (%d files in stream)", firstFile.FileKey, len(files))
 		inferredStream, err = s.inferSchemaForCSV(ctx, firstFile, stream)
 	case FormatJSON:
-		inferredStream, err = s.inferSchemaForJSON(ctx, firstFile, stream)
+		sampleFiles := schemaSampleFiles(files)
+		logger.Infof("Inferring JSON schema from %d sample file(s) out of %d files in stream", len(sampleFiles), len(files))
+		inferredStream, err = s.inferSchemaForJSON(ctx, sampleFiles, stream)
```
saksham-datazip: also can u make logs for all three modes consistent with each other as FormatJSON seems non consistent with other two ?

drivers/s3/internal/s3.go
```diff
@@ -291,20 +292,21 @@ func (s *S3) ProduceSchema(ctx context.Context, streamName string) (*types.Strea
...
 	stream := types.NewStream(streamName, "s3", &s.config.BucketName)
 
-	// Infer schema from the first file in the stream
-	firstFile := files[0]
-	logger.Infof("Inferring schema from file: %s (%d files in stream)", firstFile.FileKey, len(files))
-
 	var inferredStream *types.Stream
 	var err error
 
-	// Create appropriate parser and infer schema (format-specific logic from parser package)
```
saksham-datazip, line 301: why remove this comment ?

drivers/s3/internal/s3.go
```diff
@@ -322,6 +324,31 @@ func (s *S3) ProduceSchema(ctx context.Context, streamName string) (*types.Strea
 	return inferredStream, nil
 }
 
+func schemaSampleFiles(files []FileObject) []FileObject {
+	switch len(files) {
+	case 0:
+		return nil
+	case 1:
+		return files[:1]
+	}
```
saksham-datazip, line 329: when len(files) can be 0 at this point as we already had a check in ProduceSchema?
we can directly return it like 
```
	if len(files) == 1 {
		return files
	}

File: drivers/s3/internal/s3.go [hunk 5]
```diff
@@ -322,6 +324,31 @@ func (s *S3) ProduceSchema(ctx context.Context, streamName string) (*types.Strea
...
+	sort.Slice(sortedFiles, func(i, j int) bool {
+		if sortedFiles[i].LastModified == sortedFiles[j].LastModified {
+			return sortedFiles[i].FileKey < sortedFiles[j].FileKey
+		}
+		return sortedFiles[i].LastModified < sortedFiles[j].LastModified
+	})
+
+	firstFile := sortedFiles[0]
+	lastFile := sortedFiles[len(sortedFiles)-1]
+	if firstFile.FileKey == lastFile.FileKey {
```
saksham-datazip: When this can be true ?

drivers/s3/internal/s3.go
```diff
@@ -322,6 +324,31 @@ func (s *S3) ProduceSchema(ctx context.Context, streamName string) (*types.Strea
 	return inferredStream, nil
 }
 
+func schemaSampleFiles(files []FileObject) []FileObject {
```
saksham-datazip, line 326: Add a comment of this function does

drivers/s3/internal/s3.go
```diff
@@ -364,12 +391,31 @@ func (s *S3) inferSchemaForCSV(ctx context.Context, file FileObject, stream *typ
...
 
-// inferSchemaForJSON infers schema from a JSON file
-func (s *S3) inferSchemaForJSON(ctx context.Context, file FileObject, stream *types.Stream) (*types.Stream, error) {
-	return s.withFileReader(ctx, file.FileKey, func(reader io.Reader) (*types.Stream, error) {
-		jsonParser := parser.NewJSONParser(*s.config.GetJSONConfig(), stream)
-		return jsonParser.InferSchema(ctx, reader)
-	})
+// inferSchemaForJSON infers schema from JSON sample files.
+func (s *S3) inferSchemaForJSON(ctx context.Context, files []FileObject, stream *types.Stream) (*types.Stream, error) {
+	if len(files) == 0 {
```
saksham-datazip: when this can be true?

drivers/s3/internal/s3.go
```diff
@@ -364,12 +391,31 @@ func (s *S3) inferSchemaForCSV(ctx context.Context, file FileObject, stream *typ
...
+		return nil, fmt.Errorf("no JSON sample files provided")
+	}
+
+	readers := make([]io.Reader, 0, len(files))
```
saksham-datazip: Here, the logic is duplicated with function withFileReader, so can u make that function(withFileReader) accept multiple files and than read them ,so as to avoid code duplication and pass a multiple file code directly from ProduceSchema?And if you have any better approach than feel free to communicate.

pkg/parser/json.go
```diff
@@ -28,29 +33,72 @@ func NewJSONParser(config JSONConfig, stream *types.Stream) *JSONParser {
 
 // InferSchema reads the first few records of a JSON file to infer the schema
 // Supports JSONL (line-delimited), JSON Array, and single JSON object formats
-func (p *JSONParser) InferSchema(_ context.Context, reader io.Reader) (*types.Stream, error) {
+func (p *JSONParser) InferSchema(ctx context.Context, reader io.Reader) (*types.Stream, error) {
```
saksham-datazip: I think we can refractor these three new functions majorly into one just one while trying to keep the code as similar to staging as possible here their is no need for three functions can u confirm this once ?

saksham-datazip: If refractoring, please make sure to make the respective changes in unit tests as well.

utils/typeutils/resolver.go
```diff
@@ -22,7 +22,7 @@ func Resolve(stream *types.Stream, objects ...map[string]interface{}) error {
 	}
 
 	for column, field := range allfields {
-		stream.UpsertField(column, *field.dataType, field.isNullable(), false)
+		stream.UpsertField(column, field.getType(), field.isNullable(), false)
```
saksham-datazip, line 25: Can u help me confirm why you are making the changes here ? and also have u checked for affects on upstream from where ever it is being called from that their is no changes in behavior?

saksham-datazip (commented):
First Review

Samyak-jain7:
Pushed `e1fbe7ad` for the latest parser-interface review note and replied inline.

Validation rerun: `go test -count=1 ./pkg/parser`, `go test -count=1 ./drivers/s3/internal`, `go test -count=1 ./utils/typeutils`, root `go test -count=1 ./...`, S3 driver `go test -count=1 ./...`, root build/vet, S3 driver build/vet, gofmt/diff checks, and the validation links all returned 200.

drivers/s3/internal/s3.go
Samyak-jain7: Yep, cleaned this up. CSV/parquet now pass the first-file slice directly (`files[:1]`) instead of carrying a separate `firstFile` variable.

drivers/s3/internal/s3.go
Samyak-jain7: This changed a bit in the latest pass: the three formats still use the same log shape, but I dropped file keys from the log. It now reports sample counts for CSV, JSON, and parquet.

drivers/s3/internal/s3.go
Samyak-jain7: You are right, I should not have dropped that comment. Added it back above the format-specific parser switch.

drivers/s3/internal/s3.go
Samyak-jain7: Fair point. This helper is only called after `ProduceSchema` has checked that files exist, so I removed the empty case and kept the single-file return simple.

drivers/s3/internal/s3.go
Samyak-jain7: This was just over-defensive. Since we pick first/last from a non-empty sorted slice, that check does not add anything useful, so I removed it.

drivers/s3/internal/s3.go
Samyak-jain7: Added a short comment there explaining that we pick the first and last files by `LastModified` for JSON schema inference.

drivers/s3/internal/s3.go
Samyak-jain7: It should not be true from the current call path. I removed that redundant guard from `inferSchemaForJSON`.

drivers/s3/internal/s3.go
Samyak-jain7: Makes sense. I changed the reader helper to handle multiple files and reused it for JSON, so the open/close logic stays in one place.

pkg/parser/json.go
Samyak-jain7: I simplified this part. The extra helper is gone now; the multi-reader path still calls `parseJSONContent`, so JSONL / array / object detection stays the same.

Samyak-jain7: Adjusted the tests along with the refactor. The S3 helper tests no longer cover unreachable branches, and parser tests still cover mixed reader formats plus nullable fields.

utils/typeutils/resolver.go
Samyak-jain7: This is to avoid a real nil dereference. `Field.Merge` clears `dataType` when it sees mixed concrete types, so the final schema write needs `field.getType()` to compute the common type. I checked the current `Resolve` callers; single-type behavior is unchanged, and the mixed-type path is covered by tests.

drivers/s3/internal/s3.go
```diff
@@ -322,20 +323,52 @@ func (s *S3) ProduceSchema(ctx context.Context, streamName string) (*types.Strea
...
+	keys := make([]string, 0, len(files))
+	for _, file := range files {
+		keys = append(keys, file.FileKey)
+	}
+	return strings.Join(keys, ", ")
+}
+
+// withFileReader opens one or more CSV/JSON readers, closes them after the callback,
+// and passes the readers in the same order as the input files.
+func (s *S3) withFileReader(ctx context.Context, files []FileObject, callback func([]io.Reader) (*types.Stream, error)) (*types.Stream, error) {
```
saksham-datazip: Since this function is reading multiple files now shouldn't we name something like withFileReaders?

drivers/s3/internal/s3.go
```diff
@@ -291,21 +292,21 @@ func (s *S3) ProduceSchema(ctx context.Context, streamName string) (*types.Strea
...
-
 	var inferredStream *types.Stream
 	var err error
 
 	// Create appropriate parser and infer schema (format-specific logic from parser package)
 	switch s.config.FileFormat {
 	case FormatCSV:
-		inferredStream, err = s.inferSchemaForCSV(ctx, firstFile, stream)
+		logger.Infof("Inferring schema from file(s): %s (%d files in stream)", files[0].FileKey, len(files))
+		inferredStream, err = s.inferSchemaForCSV(ctx, files[0], stream)
```
saksham-datazip: we can make inferSchemaForCSV accept multiple files and pass something like []FileObject{files[0]} from here only so that these all three functions would look more generic?same for inferSchemaForParquet.

drivers/s3/internal/s3.go
```diff
@@ -322,20 +323,52 @@ func (s *S3) ProduceSchema(ctx context.Context, streamName string) (*types.Strea
...
-	}()
+		return sortedFiles[i].LastModified < sortedFiles[j].LastModified
+	})
+
+	firstFile := sortedFiles[0]
+	lastFile := sortedFiles[len(sortedFiles)-1]
+	return []FileObject{firstFile, lastFile}
+}
+
+func schemaSampleFileKeys(files []FileObject) string {
```
saksham-datazip: Can u think something about we are creating entire new function just for log isnt it a bit overkill ?and what do u think do we need FileKey in logs?

pkg/parser/json.go
```diff
@@ -28,42 +33,68 @@ func NewJSONParser(config JSONConfig, stream *types.Stream) *JSONParser {
 
 // InferSchema reads the first few records of a JSON file to infer the schema
 // Supports JSONL (line-delimited), JSON Array, and single JSON object formats
-func (p *JSONParser) InferSchema(_ context.Context, reader io.Reader) (*types.Stream, error) {
-	logger.Debug("Inferring JSON schema from sample data")
-	//TODO : implement sampling of records from first and last files to get a more accurate schema
-	// Collect sample records using smart JSON format detection
-	maxSamples := 100
+func (p *JSONParser) InferSchema(ctx context.Context, reader io.Reader) (*types.Stream, error) {
```
saksham-datazip: why these two functions exist they can be just merged into one?
InferSchemaFromReaders and InferSchema can just be one function?

saksham-datazip: Also look into this function on staging as ig their are multiple places in code which can be reverted and will be much clearer in that way.

pkg/parser/json.go
```diff
@@ -12,6 +12,11 @@ import (
 	"github.com/datazip-inc/olake/utils/typeutils"
 )
 
+const (
+	jsonSchemaMaxSamples           = 100
+	jsonSchemaMaxBytesForInference = 10 * 1024 * 1024
+)
+
```
saksham-datazip: Previously they were defined inside inferschema why do we need to shift here now ?

drivers/s3/internal/s3.go
Samyak-jain7: Renamed it to `withFileReaders` in `1f6bf7d9`.

drivers/s3/internal/s3.go
Samyak-jain7: Changed this. CSV and parquet now take `[]FileObject` too; `ProduceSchema` passes `files[:1]` for those formats since they still infer from only the first file.

drivers/s3/internal/s3.go
Samyak-jain7: I agree on this one. Removed the log-only helper and dropped file keys from that log. It now just logs the sample count consistently for CSV, JSON, and parquet.

pkg/parser/json.go
Samyak-jain7: I checked this against `pkg/parser.Parser`. I kept both methods because `InferSchema(ctx, reader)` is still the parser interface shape used by CSV/parquet. `InferSchemaFromReaders` is the JSON-specific entry point S3 needs when it samples first + last files. Merging them cleanly would mean broadening the parser interface for every parser, which feels like a bigger change than this PR needs.

Samyak-jain7: Rechecked against `staging` and reverted the drift that was not needed: constants are back inside inference, and the log helper is gone. I kept the multi-reader JSON path and single `Resolve` call because that is the actual first/last-file sampling behavior this PR adds.

pkg/parser/json.go
Samyak-jain7: No strong reason to keep them at package scope. Moved them back inside the inference function in `1f6bf7d9`.

drivers/s3/internal/s3.go
saksham-datazip: ohkk

drivers/s3/internal/s3.go
Samyak-jain7: This makes sense. I changed the parser interface to `InferSchema(ctx, readers ...io.Reader)` and removed the separate JSON `InferSchemaFromReaders` method. CSV/parquet still use the first reader, while JSON can now pass first + last readers through the same interface. Pushed in `e1fbe7ad`.

drivers/s3/internal/s3.go
saksham-datazip: Now ,you used this variable(sampleFiles) everywhere which is not even needed remove it and can u use the logs like this ?
```
logger.Infof("Inferring JSON schema for stream %s (%d file(s))", streamName, len(files))

For all three

Samyak-jain7: Done in `2ee25935`. Removed the `sampleFiles` locals from the switch and changed the CSV/JSON/parquet logs to the same stream-level format. JSON still passes `schemaSampleFiles(files)` into inference, so the first/last sampling behavior is unchanged.

saksham-datazip:
@Samyak-jain7 Please take your time and thoroughly verify the changes. This is the final review from my side, so I'd appreciate a final confirmation that everything looks good before we proceed.And let me now when you are done from your side

saksham-datazip:
@Samyak-jain7 Could you please confirm if you've completed a self-review of your PR?

nayanj98:
closing the PR as no response was received

PR #968, opened 2026-06-04
fix: thread retry after commit leads duplication of data

vaibhav-datazip:
# Description

Fixes a bug in the 2PC (Two-Phase Commit) state management where a thread failure during backfill sync could cause duplicate records to be written on retry.

### Root Cause

When a writer thread fails after Java has already committed data to Iceberg (Phase 1 complete) but before Go could persist the local state (Phase 2 incomplete), the retry mechanism would re-run the chunk from scratch. The bug had two parts:

1. **`destination/writers.go`** — `writerSchema` was caching `metadataState` the first time `NewWriter` was called for a stream. Every subsequent thread (including retries) re-used this stale snapshot of `olake_2pc` instead of fetching the current committed state from Java. This meant a retrying thread could not see that a previous attempt had already committed its chunk.

2. **`destination/iceberg/iceberg.go`** — In `Setup()`, when the Go-side schema was already cached (`globalSchema != nil`), the `else` branch was empty — it skipped calling Java entirely. This meant the `olake_2pc` property (the source of truth for which chunks are committed) was never fetched for threads 2, 3, ... including all retried threads.

### Fix

- Removed `metadataState *types.MetadataState` from the `writerSchema` cache in `writers.go`. Schema is still cached (expensive), but `olake_2pc` state is now always fetched fresh on every `NewWriter` call.
- In `iceberg.go` `Setup()`, the `else` branch now explicitly calls `GET_OR_CREATE_TABLE` to load the current table from the Iceberg catalog and read its `olake_2pc` property. The returned schema from Java is discarded (the cached Go schema is used instead); only the `olake_2pc` state is consumed.
- Extracted duplicated `olake_2pc` JSON parsing into a shared `parseTwoPCState` helper used by both branches.
- Added a temporary fault injection env var (`OLAKE_INJECT_2PC_FAULT=1`) in `legacy-writer/writer.go` to reliably reproduce the scenario where Java commits successfully but Go loses the response, triggering retries.

### How the fix works end-to-end

Chunk A — first attempt:

Go reads records → sends to Java (WRITE phase)
Java commits to Iceberg catalog, updates olake_2pc: { "full_refresh_committed_ids": ["thread-A-uuid"] }
Go receives error (network / fault injection) → retry queued
Chunk A — retry:

pool.NewWriter() → calls Java GET_OR_CREATE_TABLE → reads CURRENT olake_2pc
Check: thread ID "thread-A-uuid" found in FullRefreshCommittedIDs
→ "already committed, skipping" → no data written → no duplicate records ✓

## Type of change

# How Has This Been Tested?

A fault injection mechanism was added (`OLAKE_INJECT_2PC_FAULT=1`) that simulates a commit response loss, Java commits successfully, but Go always receives an error from `Close()`, forcing retries on every attempt.

**Test configuration:**
- Source: MySQL (`vbhv.users`, 10 rows, 2 chunks)
- Destination: Apache Iceberg (JDBC catalog, MinIO S3)
- Sync mode: CDC (backfill phase tested)
- Env var: `OLAKE_INJECT_2PC_FAULT=1`

- [x] **Duplicate write prevention**: Ran sync with fault injection enabled. Chunk `min[1]-max[nil]` wrote 10 records and committed on attempt 1. On retry attempts 1 and 2, log confirmed `"chunk min[1] and max[nil] already committed, skipping"` — no records re-written. Verified Iceberg table row count remained 10 (not 20 or 30).

- [x] **Fresh metadata fetch on retry**: Confirmed via Java logs on each retry attempt that `"Refreshing table metadata from new version: 00001-..."` was emitted before the skip check — proving the retry thread read the current `olake_2pc` state from the catalog, not a cached stale copy.

## Documentation

## Related PR's (If Any):
N/A


destination/iceberg/iceberg.go
```diff
@@ -166,6 +177,33 @@ func (i *Iceberg) Setup(ctx context.Context, stream types.StreamInterface, globa
...
 			return nil, nil, fmt.Errorf("failed to convert globalSchema of type[%T] to map[string]string", globalSchema)
 		}
+
+		// Schema is already cached from the first thread's GET_OR_CREATE_TABLE call.
+		// We still need to call GET_OR_CREATE_TABLE here because:
+		//   1. Every Java process starts fresh (icebergTable=nil); it must load the table
+		//      before it can read olake_2pc from table properties.
+		//   2. The schema we send is only used by Java as a fallback if the table does
+		//      not exist (orElseGet path in loadIcebergTable).
+		protoSchema := make([]*proto.IcebergPayload_SchemaField, 0, len(schema))
```
vikaxsh: why can't we remove globalSchema nil check entirely? 
GET_OR_CREATE_TABLE is called in both branches and we can delete else branch completely

vaibhav-datazip: In that case every thread would redundantly re-derive the schema and re-parse Java's response work that is already cached. Why not just use the cached schema if we already have it

vikaxsh: There isn't much computation we can simplify here; otherwise, we could try extracting the shared logic from the if-else blocks.

vaibhav-datazip: already have abstracted the common function of parsing two pc state , the schema parsing is not required in retry part

vikaxsh: GET_OR_CREATE_TABLE , 2pc state refresh are still same in both branches

destination/iceberg/iceberg.go
```diff
@@ -126,46 +142,44 @@ func (i *Iceberg) Setup(ctx context.Context, stream types.StreamInterface, globa
...
+		Metadata: &proto.IcebergPayload_Metadata{
+			Schema:          protoSchema,
+			DestTableName:   i.stream.GetDestinationTable(),
+			ThreadId:        i.server.serverID,
+			IdentifierField: &identifierField,
+		},
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get or create table: %s", err)
+	}
```
vaibhav-datazip, line 156: test this for cdc and incremental as well

destination/writers.go
```diff
@@ -176,14 +175,15 @@ func (w *WriterPool) NewWriter(ctx context.Context, stream types.StreamInterface
 		}
 
 		if streamArtifact.schema == nil {
-			// First thread for this stream: persist schema and the olake_2pc state
-			// so all subsequent threads (e.g. concurrent backfill chunks) can also
-			// check which chunks were already committed.
+			// First thread for this stream: cache the schema so subsequent threads
+			// skip the expensive GET_OR_CREATE_TABLE schema negotiation.
```
vikaxsh: comment is incorrect, subsequent threads do call GET_OR_CREATE_TABLE

destination/iceberg/iceberg.go
vaibhav-datazip: Tested no duplicate files are being created for cdc and incremental , tested this using  mysql driver

destination/iceberg/iceberg.go
```diff
@@ -126,46 +127,48 @@ func (i *Iceberg) Setup(ctx context.Context, stream types.StreamInterface, globa
...
-	} else {
-		// set global schema for current thread
-		var ok bool
-		schema, ok = globalSchema.(map[string]string)
-		if !ok {
-			return nil, nil, fmt.Errorf("failed to convert globalSchema of type[%T] to map[string]string", globalSchema)
+	if raw := ingestResponse.GetOlake_2PcState(); raw != "" {
+		var ms types.MetadataState
+		if err := json.Unmarshal([]byte(raw), &ms); err != nil {
+			return schema, nil, fmt.Errorf("failed to unmarshal 2pc metadata state: %s", err)
```
vikaxsh: we should return nil here because every other failure path in Setup returns (nil, nil, err)

destination/iceberg/iceberg.go
```diff
@@ -126,46 +127,48 @@ func (i *Iceberg) Setup(ctx context.Context, stream types.StreamInterface, globa
...
+			IdentifierField: &identifierField,
+		},
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get or create table: %s", err)
+	}
 
-		ingestResponse := response.(*proto.RecordIngestResponse)
+	ingestResponse := metaResp.(*proto.RecordIngestResponse)
+	if globalSchema == nil {
```
vikaxsh, line 159: Can we avoid checking for nil in two different places within the same function?

vaibhav-datazip: we will be parsing the schema only for the thread which don't have a schema cached yet so this check is required

destination/iceberg/iceberg.go
vaibhav-datazip: this was the code earlier so just placed it below, returning nil now

vikaxsh (approved):

PR #975, opened 2026-06-06
refactor(chunking): refactor MySQL helpers for unit testing

### LINKED ISSUE #928: Separate chunking-related logic from this function so the individual components can be unit tested independently.
https://github.com/datazip-inc/olake/blob/06aeb9260e7c348c8bd74ed3d93e1ba0023e2e71/drivers/mysql/internal/backfill.go#L64-L69

---

###### This issue was generated by [todo-issue](https://github.com/DerJuulsn/todo-issue) based on a `TODO` comment in 06aeb9260e7c348c8bd74ed3d93e1ba0023e2e71. It&#x27;s been assigned to @saksham-datazip because they committed the code.

vimla01:
# Description

Refactored the MySQL chunking flow in `GetOrSplitChunks` into smaller helper functions so the chunk sizing, chunk column selection, chunk bound selection, and split strategy routing can be unit tested independently.

This keeps the existing behavior the same, but makes the chunking code easier to test and reason about.

Fixes #928

## Type of change

# How Has This Been Tested?

- [x] Ran focused MySQL backfill unit tests:
  `go test ./drivers/mysql/internal -run 'Test(Chunk|LimitOffset|SplitEvenly|PrimaryKey)'`

# Screenshots or Recordings

[olake_928.webm](https://github.com/user-attachments/assets/973466d0-c67f-43a6-838f-132adeafcfba)

## Documentation

## Related PR's (If Any):

N/A


vimla01:
@nayanj98  could you please take a look at this PR?
I have attached a video showcasing the changes and a successful sync with Olake (iceberg & parquet both)

nayanj98:
@vimla01 Thanks a lot will assign a reviewer for this PR soon

nayanj98:
@vimla01 Assigning @vaibhav-datazip to review your PR

drivers/mysql/internal/backfill.go
```diff
@@ -61,336 +61,363 @@ func (m *MySQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
...
-		approxRowCount      int64
-		avgRowSize          any
-		approxTableSize     int64
-		columnCollationType string
-		dataMaxLength       sql.NullInt64
-	)
-
-	tableStatsQuery := jdbc.MySQLTableStatsQuery()
-	err := m.client.QueryRowContext(ctx, tableStatsQuery, stream.Name()).Scan(&approxRowCount, &avgRowSize, &approxTableSize)
+	stats, err := m.fetchTableStats(ctx, stream)
```
vaibhav-datazip: I don't think we need to abstract out every single logic into a function as these are too basic implementation , primarily we need to abstract out the chunking strategies into separate functions so that those logic be tested.

You can go through mssql driver code once, that has all the chunking strategies separated.

also, you can keep some of the logic in separate functions , after the changes you can add comment on pr description or separate comment, why you think the logic needs a separate function.

drivers/mysql/internal/backfill.go
```diff
@@ -61,336 +61,363 @@ func (m *MySQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
...
-
-		Final Step:
-		- If sufficient boundaries are not obtained, fallback to primary key-based chunking.
-		- Reduce boundary count to match expectedChunks if needed by evenly subsampling while preserving order.
-
-		Example:
-		minVal = "aa", maxVal = "az", expectedChunks = 4
-
-		Generated boundaries after refining boundaries using collation-aware DB queries:
-		["aa", "ai", "ar", "az"]
```
vaibhav-datazip, line 288: this comment is necessary as it explains the necessary, even the other deleted comments should be kept

drivers/mysql/internal/backfill.go
```diff
@@ -61,336 +61,363 @@ func (m *MySQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
...
-		})
+	if len(chunkBoundaries) < int(math.Ceil(float64(constants.MysqlChunkAcceptanceRatio*float64(expectedChunks)))) {
+		return nil, nil
+	}
+	return chunksFromBoundaries(condenseStrings(chunkBoundaries, expectedChunks)), nil
+}
 
-		logger.Infof("Chunking completed using splitEvenlyForString Method for stream %s", stream.ID())
-		return nil
+func expectedStringChunkCount(approxTableSize int64) int64 {
```
vaibhav-datazip: the newly added functions should have a comment on top of them explaining what they do

drivers/mysql/internal/backfill.go
```diff
@@ -478,6 +505,21 @@ func isStringSupportedPK(minVal any, maxVal any, dataMaxLength sql.NullInt64, da
 }
 
 type (
```
vaibhav-datazip: move this on top of all the functions

drivers/mysql/internal/backfill.go
```diff
@@ -61,336 +61,363 @@ func (m *MySQL) ChunkIterator(ctx context.Context, stream types.StreamInterface,
...
-	*/
-	splitEvenlyForInt := func(chunks *types.Set[types.Chunk]) error {
+func limitOffsetChunks(approxRowCount, chunkSize int64) *types.Set[types.Chunk] {
+	chunks := types.NewSet[types.Chunk]()
+	chunks.Insert(types.Chunk{
+		Min: nil,
+		Max: utils.ConvertToString(chunkSize),
+	})
+	lastChunk := chunkSize
+	for lastChunk < approxRowCount {
```
vaibhav-datazip: the chunking process earlier used to take place in an isolation level , will be better if you go through what was happening earlier and why that has been remove by you now

drivers/mysql/internal/backfill_test.go
```diff
@@ -0,0 +1,166 @@
+package driver
+
```
vaibhav-datazip, line 2: in this file unit tests for some functions are missing , will be better if you can think of edge test cases in those as well as already present functions and add them .

drivers/mysql/internal/backfill.go
vimla01: Updated this to keep the basic metadata and orchestration logic inside `GetOrSplitChunks`. Only the actual chunking strategies and reusable boundary calculations remain separate

drivers/mysql/internal/backfill.go
vimla01: Restored.

drivers/mysql/internal/backfill.go
vimla01: added comments above the introduced helper functions .

drivers/mysql/internal/backfill_test.go
vimla01: Added more focused unit tests for the extracted chunking helpers and existing utilities. I covered edge cases around numeric chunk ranges/overflow, limit-offset boundaries, composite primary key args, string boundary generation, charset encoding, padding, condensing, and strategy eligibility.

drivers/mysql/internal/backfill.go
vimla01: i checked the previous flow and restored the isolation wrapper for the limit-offset chunking path so the behavior remains the same as before

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,293 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 		if err != nil {
 			return nil, fmt.Errorf("failed to fetch column datatype and max length for column %s: %s", pkColumns[0], err)
 		}
-		// 1. Try Numeric Strategy
-		numericChunkBounds = isNumericAndEvenDistributed(minVal, maxVal, approxRowCount, chunkSize, dataType)
 
-		// 2. If not numeric, check for supported String strategy
+		// Prefer an arithmetic split for evenly distributed numeric keys.
+		numericChunkBounds = isNumericAndEvenDistributed(minVal, maxVal, approxRowCount, chunkSize, dataType)
```
vaibhav-datazip, line 151: can you restore this as well as the comments can be preserved and there dosent seem to be any structural/logical change

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,293 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
-			if minVal == nil {
-				return nil
-			}
-			chunks.Insert(types.Chunk{
-				Min: nil,
-				Max: utils.ConvertToString(minVal),
-			})
+	switch {
+	case numericChunkBounds != nil:
+		logger.Infof("Using splitEvenlyForInt Method for stream %s", stream.ID())
```
vaibhav-datazip: can you check other drivers as well for the log placement of chunking strategy being used, and debug being used instead of info , and make it consistent for strategy being used

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,293 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+		}
+		logger.Infof("Chunking completed using splitEvenlyForString Method for stream %s", stream.ID())
+		return chunks, nil
+	case len(pkColumns) > 0:
+		logger.Infof("Using splitViaPrimaryKey Method for stream %s", stream.ID())
+		return m.splitViaPrimaryKey(ctx, stream, minVal, maxVal, pkColumns, chunkSize)
+	default:
+		logger.Infof("Falling back to limit offset method for stream %s", stream.ID())
+		var chunks *types.Set[types.Chunk]
+		err := jdbc.WithIsolation(ctx, m.client, true, func(_ *sql.Tx) error {
```
vaibhav-datazip: shouldn't the isolation level be done inside the function

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,293 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
-		(-∞, "aa"), ["aa","ai"), ["ai","ar"), ["ar","az"), ["az", +∞)
-	*/
-	splitEvenlyForString := func(chunks *types.Set[types.Chunk]) error {
-		expectedChunks := int64(math.Ceil(float64(approxTableSize) / float64(constants.EffectiveParquetSize)))
-		expectedChunks = utils.Ternary(expectedChunks <= 0, int64(1), expectedChunks).(int64)
-
-		stringChunkStepSize := new(big.Int).Sub(stringChunkBounds.maxEncodedBigIntValue, stringChunkBounds.minEncodedBigIntValue)
-		stringChunkStepSize.Add(stringChunkStepSize, new(big.Int).Sub(big.NewInt(expectedChunks), big.NewInt(1)))
-		stringChunkStepSize.Div(stringChunkStepSize, big.NewInt(expectedChunks))
-
```
vaibhav-datazip, line 300: even the stringchunkstepsize is being used at one place only , can we use that only instead of defining a function for it

vaibhav-datazip:
@vimla01 once you are done with your changes after the review , please re-request the review from here so that I know the pr is ready again
Screenshot 2026-06-18 at 11 42 03 AM

drivers/mysql/internal/backfill.go
vimla01: Restored.

drivers/mysql/internal/backfill.go
vimla01: Checked the other drivers as well. Oracle already uses debug for chunking strategy selection, and I updated MySQL and MongoDB strategy-selection logs from Infof to Debugf for consistency.

drivers/mysql/internal/backfill.go
vimla01: Addressed. Removed the separate stringChunkStepSize helper and kept the step-size calculation inline at the single usage site.

drivers/mysql/internal/backfill.go
vaibhav-datazip: I don't think you have restored it everywhere like in limitOffsetChunks

drivers/mysql/internal/backfill.go
vaibhav-datazip: the primaryKeyChunkArgs has been abstracted, and in that comments are missing which were present earlier

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,284 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+	rangeSlice := []string{}
+	adjustedStepSize := new(big.Int).Set(stepSize)
+	adjustedStepSize.Add(adjustedStepSize, big.NewInt(stepShrinkFactor))
+	adjustedStepSize.Div(adjustedStepSize, big.NewInt(stepShrinkFactor+1))
+	currentBoundary := new(big.Int).Set(bounds.minEncodedBigIntValue)
+
+	for chunkIdx := int64(0); chunkIdx < expectedChunks*(stepShrinkFactor+1) && currentBoundary.Cmp(bounds.maxEncodedBigIntValue) < 0; chunkIdx++ {
+		rangeSlice = append(rangeSlice, decodeBigIntToCharsetString(currentBoundary))
+		currentBoundary.Add(currentBoundary, adjustedStepSize)
 	}
```
vaibhav-datazip: the test cases are not that necessary and are limited maybe we should put this also back to its original place

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,284 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
 
-	return chunks, err
+	return append(rangeSlice, bounds.MaxPadded)
+}
+
+// chunksFromBoundaries converts ordered boundary values into open-ended chunks.
+func chunksFromBoundaries(boundaries []string) *types.Set[types.Chunk] {
```
vaibhav-datazip: similarly for this one

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,284 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
-		(-∞, 0), [0,25), [25,50), [50,75), [75,100), [100, +∞)
-	*/
-	splitEvenlyForInt := func(chunks *types.Set[types.Chunk]) error {
-		chunks.Insert(types.Chunk{
-			Min: nil,
-			Max: utils.ConvertToString(numericChunkBounds.MinBoundary),
-		})
-		prev := numericChunkBounds.MinBoundary
-		for next := numericChunkBounds.MinBoundary + numericChunkBounds.ChunkStep; next <= numericChunkBounds.MaxBoundary; next += numericChunkBounds.ChunkStep {
-			// Guard against int64 overflow causing `next` to wrap and become ≤ prev
```
vaibhav-datazip, line 244: comment like these are still pending to be restored , please check again in all the changes of your pr to figure similar things 

you can use GitHub pr extension that might help

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,305 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
-		0 → 25 → 50 → 75 → 100
+// buildLimitOffsetChunks divides a table without a usable key into offset-based chunks.
+func buildLimitOffsetChunks(approxRowCount, chunkSize int64) *types.Set[types.Chunk] {
+	chunks := types.NewSet[types.Chunk]()
+
+	// Start with the first batch-sized range.
+	chunks.Insert(types.Chunk{
+		Min: nil,
+		Max: utils.ConvertToString(chunkSize),
+	})
```
vaibhav-datazip: see limit offset chunking itself is a function which should give us the chunks, creating a new function just to separate the chunks and isolation thing dosent make sense
even the test cases for limit offset chunk function wont be any better so we can combine them both again

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,305 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+		if len(rangeSlice) > len(chunkBoundaries) {
+			chunkBoundaries = slices.Clone(rangeSlice)
+		}
+		if len(rangeSlice) >= int(expectedChunks) {
+			break
+		}
+	}
+
+	if len(chunkBoundaries) < int(math.Ceil(float64(constants.MysqlChunkAcceptanceRatio*float64(expectedChunks)))) {
+		return nil, nil
```
vaibhav-datazip: why are not throwing error in this case , as done in int case

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,305 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
-		logger.Infof("Using splitEvenlyForString Method for stream %s", stream.ID())
-		err = splitEvenlyForString(chunks)
-	case len(pkColumns) > 0:
-		logger.Infof("Using splitViaPrimaryKey Method for stream %s", stream.ID())
-		err = splitViaPrimaryKey(stream, chunks)
-	default:
-		logger.Infof("Falling back to limit offset method for stream %s", stream.ID())
-		err = limitOffsetChunking(chunks)
+// expectedStringChunkCount estimates the target number of string-key chunks from table size.
+func expectedStringChunkCount(approxTableSize int64) int64 {
```
vaibhav-datazip: even this helper function is simple maths only used once you can put this back

vaibhav-datazip (commented):
each time you are done doing the changes , you should ask the reviewer for re review please do the same after making your changes this time as well

drivers/mysql/internal/backfill.go
vimla01: yes done

drivers/mysql/internal/backfill.go
vimla01: okay

drivers/mysql/internal/backfill.go
vimla01: okay  combined limit-offset chunk creation back into limitOffsetChunks

vimla01:
> each time you are done doing the changes , you should ask the reviewer for re review please do the same after making your changes this time as well

sry i missed requesting re-review on the previous one. i had intended to verify the changes the next day first and then request re-review, but got delayed. i'll make sure to do so for future updates.

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,290 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
-			chunks.Insert(types.Chunk{
-				Min: nil,
-				Max: utils.ConvertToString(minVal),
-			})
+	switch {
+	case numericChunkBounds != nil:
+		logger.Debugf("Using splitEvenlyForInt Method for stream %s", stream.ID())
+		chunks, err := splitEvenlyForInt(numericChunkBounds)
+		if err != nil {
+			logger.Warnf("int64 arithmetic overflow, falling back to splitViaPrimaryKey for stream %s", stream.ID())
```
vaibhav-datazip: see u are sending the error from the above function but its silently getting eaten , you should print that in this message 

check other places as well where ur changes are silently eating the changes

drivers/mysql/internal/backfill.go
vaibhav-datazip: had asked you earlier that you can explain the reason why certain functions are abstracted out [here](https://github.com/datazip-inc/olake/pull/975#discussion_r3408642494), 

if you don't think it was necessary ,can you put primarykeychunkargs function back as well,


A suggestion as well : when we say that some function might not be required to be abstracted out or some other change is suggested. you should check if there are more such changes which follow the similar pattern, and do the suggested change everywhere in the PR.

doing a self review of your pr before you push the changes for re review will help a lot, it will help your pr to get merged much sooner as well and with less comments

drivers/mysql/internal/backfill_test.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/datazip-inc/olake/types"
+)
+
+func TestSplitEvenlyForInt(t *testing.T) {
+	tests := []struct {
+		name      string
+		bounds    *NumericChunkBounds
+		expected  []types.Chunk
+		wantError bool
```
vaibhav-datazip: use the naming convention used in other files , check _test files present in the codebase like reformat_test, flatten_test etc.

drivers/mysql/internal/backfill.go
```diff
@@ -179,7 +179,31 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 		return m.splitViaPrimaryKey(ctx, stream, minVal, maxVal, pkColumns, chunkSize)
 	default:
 		logger.Debugf("Falling back to limit offset method for stream %s", stream.ID())
-		return m.limitOffsetChunks(ctx, stream, approxRowCount, chunkSize)
```
vaibhav-datazip: this is a chunking strategy please keep it separate only like you had done earlier , the comment was only for primaryKeyChunkArgs function, chunking strategies were supposed to be abstracted out so that unit tests can be added for it

drivers/mysql/internal/backfill_test.go
```diff
@@ -135,21 +103,21 @@ func TestIsNumericAndEvenDistributed(t *testing.T) {
 		approxRowCount int64
 		chunkSize      int64
 		dataType       string
-		wantBounds     bool
+		expected       bool
 	}{
-		{name: "supported bigint", minVal: 1, maxVal: 100, approxRowCount: 100, chunkSize: 10, dataType: "BIGINT", wantBounds: true},
+		{name: "supported bigint", minVal: 1, maxVal: 100, approxRowCount: 100, chunkSize: 10, dataType: "BIGINT", expected: true},
 		{name: "unsupported type", minVal: 1, maxVal: 100, approxRowCount: 100, chunkSize: 10, dataType: "varchar"},
```
vaibhav-datazip: also please can you add comments for the test cases you have added in this file , for reference you can check other unit test files like resolver_test, reformat_test , etc.

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,287 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
-				// Split the current value into parts
-				columns := strings.Split(utils.ConvertToString(currentVal), ",")
-
-				// Create args array with the correct number of arguments for the query
-				args := make([]interface{}, 0)
-				for columnIndex := 0; columnIndex < len(pkColumns); columnIndex++ {
-					// For each column combination in the WHERE clause, we need to add the necessary parts
-					for partIndex := 0; partIndex <= columnIndex && partIndex < len(columns); partIndex++ {
-						args = append(args, columns[partIndex])
-					}
```
vaibhav-datazip, line 171: can please add these comments back

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,287 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
-				}
-				var nextValRaw interface{}
-				err := tx.QueryRowContext(ctx, query, args...).Scan(&nextValRaw)
-				if err == sql.ErrNoRows || nextValRaw == nil {
-					break
-				} else if err != nil {
-					return fmt.Errorf("failed to get next chunk end: %s", err)
-				}
```
vaibhav-datazip, line 179: in the new changes you have created two if blocks , is there a reason for that of can we keep this if and else if structure only

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,287 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
-				args := make([]interface{}, 0)
-				for columnIndex := 0; columnIndex < len(pkColumns); columnIndex++ {
-					// For each column combination in the WHERE clause, we need to add the necessary parts
-					for partIndex := 0; partIndex <= columnIndex && partIndex < len(columns); partIndex++ {
-						args = append(args, columns[partIndex])
-					}
-				}
-				var nextValRaw interface{}
-				err := tx.QueryRowContext(ctx, query, args...).Scan(&nextValRaw)
-				if err == sql.ErrNoRows || nextValRaw == nil {
```
vaibhav-datazip, line 175: earlier nextval raw nil was also getting checked in this condition only , any reason you have made it a separate condition or can that be reverted back to this only ?

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,287 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+["aa", "ai", "ar", "az"]
+
+Chunks:
+(-inf, "aa"), ["aa","ai"), ["ai","ar"), ["ar","az"), ["az", +inf)
+*/
+func (m *MySQL) splitEvenlyForString(ctx context.Context, stream types.StreamInterface, bounds *StringChunkBounds, pkColumn, columnCollationType string, approxTableSize int64) (*types.Set[types.Chunk], error) {
+	expectedChunks := int64(math.Ceil(float64(approxTableSize) / float64(constants.EffectiveParquetSize)))
+	if expectedChunks <= 0 {
+		expectedChunks = 1
 	}
```
vaibhav-datazip: can you revert this back to the utils.ternary condition only

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,287 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+			currentBoundary.Add(currentBoundary, adjustedStepSize)
+		}
+
+		rangeSlice = append(rangeSlice, bounds.MaxPadded)
+
+		// Align boundaries with actual DB values using MySQL collation ordering
+		query, args := jdbc.MySQLDistinctAlignedPKValuesWithCollationQuery(stream, pkColumn, rangeSlice, columnCollationType, bounds.MinPadded, bounds.MaxPadded)
+		rows, err := m.client.QueryContext(ctx, query, args...)
+		if err != nil {
+			boundaryQueryErr = fmt.Errorf("failed to query distinct string chunk boundaries: %w", err)
```
vaibhav-datazip: in this code base we are using %s for errors instead of %w , so please revert %w to %s wherever you have done changes, you can check the other files of codebase as well to verify this , 

also can you use the older message for distinct boundary error

distinct boundary query failed for stream %s: %s

drivers/mysql/internal/backfill.go
vimla01: I had separated the checks while reviewing error handling, but the earlier if/else-if structure was better aligned with the existing flow. ill keep that

drivers/mysql/internal/backfill.go
vimla01: Reverted it back to checking `sql.ErrNoRows || nextValRaw == nil` in the same condition with the existing `else if err != nil` block.

drivers/mysql/internal/backfill.go
```diff
@@ -132,265 +145,285 @@ func (m *MySQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
...
+		rangeSlice = rangeSlice[:0]
+		for rows.Next() {
+			var val string
+			if err := rows.Scan(&val); err != nil {
+				rows.Close()
+				return nil, fmt.Errorf("failed to scan row: %s", err)
 			}
+			rangeSlice = append(rangeSlice, val)
 		}
+		rows.Close()
```
vaibhav-datazip: I don't think this will be required, once you can read comments on Close as well , it says rows.next will automatically close it once finished . 

so this rows.Close can be removed

drivers/mysql/internal/backfill_test.go
```diff
@@ -0,0 +1,257 @@
...
+import (
+	"database/sql"
+	"math"
+	"reflect"
+	"testing"
+
+	"github.com/datazip-inc/olake/types"
+)
+
+func TestSplitEvenlyForInt(t *testing.T) {
```
vaibhav-datazip, line 12: your unit tests are not running on the CI , you can check it in the passed workflows on your PR, it wont list any of the tests added there.

Some change might be required in unit-tests.yml file, can you check it ? also make it generic so that it works for all the files added in future as well

vimla01: Got it, I'll look into it.

.github/workflows/unit-tests.yml
```diff
@@ -19,4 +19,6 @@ jobs:
         run: go mod download
 
       - name: Run All Unit Tests
-        run: go test -v ./...
+        run: |
+          TEST_PACKAGES=$(git ls-files '*_test.go' | xargs -n1 dirname | sort -u | sed 's#^#./#' | xargs go list -f '{{ .ImportPath }}' | grep -v '/drivers/db2/internal$')
+          go test -v -count=1 -skip 'Integration|2PC|Performance' $TEST_PACKAGES
```
vaibhav-datazip: can you check if this will be the better approach , as it will help in running db2 unit tests as well

```suggestion
      - name: Install DB2 CLI Driver
        run: |
          go run -C drivers/db2 github.com/ibmdb/go_ibm_db/installer@v0.4.5 "$(pwd)"
          IBM_DB_HOME="$(pwd)/clidriver"
          echo "IBM_DB_HOME=$IBM_DB_HOME" >> $GITHUB_ENV
          echo "CGO_CFLAGS=-I$IBM_DB_HOME/include" >> $GITHUB_ENV
          echo "CGO_LDFLAGS=-L$IBM_DB_HOME/lib -Wl,-rpath,$IBM_DB_HOME/lib" >> $GITHUB_ENV
          echo "LD_LIBRARY_PATH=$IBM_DB_HOME/lib" >> $GITHUB_ENV

      - name: Run All Unit Tests
        run: |
          go list -m | awk '{print $1"/..."}' | xargs go test -v -count=1 -skip 'Integration|2PC|Performance'

```

vaibhav-datazip: you can check the db2 driver installation yml from integration test yml file as well

vimla01: yeah, i ve made the suggested changes. Thanks!!

vaibhav-datazip (approved):
LGTM

PR #978, opened 2026-06-09
feat: add configurable source vs destination column naming strategy

### LINKED ISSUE #938: Support for Case-Sensitive Column Names in Data Pipelines
### Problem

Currently, Olake converts all column names to lowercase during data loads from source to destination (e.g., in Parquet files). This behavior can cause issues for systems where case sensitivity is enabled.

For example, in our setup using Amazon Redshift with case sensitivity enabled, this transformation leads to inconsistencies between the source and destination schemas.

### Proposed Solution

Introduce an option in the Olake UI to preserve the original case of column names during data transfer.

To maintain backward compatibility, this option can be disabled by default and made configurable for users who require case-sensitive column handling.

vikaxsh:
# Description

Adds a `use_source_column_names` flag to `StreamMetadata` to control destination column naming.

By default, column names continue to be normalized using `utils.Reformat()` (e.g. `My-Column` → `my_column`). When `use_source_column_names: true`, original source column names are preserved.

To avoid scattered naming logic, a new `ResolveColumnName(key string) string` method has been introduced on `StreamInterface`. All schema generation, record flattening, column selection, and partition handling now use this method instead of calling `utils.Reformat()` directly.

The flag is also included in `GetStreamsDelta` comparison logic, allowing orchestrators to detect naming strategy changes and trigger a `clear-destination` run when required.

Fixes #938

## Type of Change

* [x] New feature
* [ ] Bug fix
* [ ] Breaking change
* [x] Documentation update

## Testing

* [ ] Verified default behavior (`use_source_column_names: false`) remains unchanged.
* [ ] Verified original source column names are preserved when `use_source_column_names: true`.
* [ ] Verified `GetStreamsDelta` detects naming strategy changes and triggers the expected re-sync flow.

## Documentation

## Related PRs

N/A


types/type_schema.go
vaibhav-datazip, line 202: I think there is a logical bug here, the corrected should be like this 

```suggestion
t.Properties.Range(func(_, value interface{}) bool {
if value.(*Property).DestinationColumnName != "" {
            found = true
            return false 
        }
        return true
        }
```

vaibhav-datazip: destination column name wont be consistent from what we want this will create inconsistency

types/catalog.go
vaibhav-datazip, line 252: for this new field you might need to update this comment as well

utils/typeutils/flatten.go
```diff
@@ -16,18 +16,13 @@ type Flattener interface {
...
 var keyCache sync.Map
 
-func NewFlattener() Flattener {
-	return &FlattenerImpl{
-		omitNilValues: true,
-	}
-}
-
-func getReformattedKey(key string) string {
+func getCachedReformattedKey(key string) string {
```
vaibhav-datazip: well the old name for this function suited better because it works even when keys are not cached

utils/typeutils/flatten.go
```diff
@@ -36,6 +31,20 @@ func getReformattedKey(key string) string {
...
+	return &FlattenerImpl{omitNilValues: true, keyFn: getCachedReformattedKey}
+}
+
+// NewFlattenerWith returns a flattener that resolves record keys using the
+// provided function. Pass stream.ResolveColumnName to honor the stream's
+// naming strategy; source-name mode returns keys unchanged, destination-name
+// mode applies utils.Reformat.
+func NewFlattenerWith(resolve func(string) string) Flattener {
+	return &FlattenerImpl{omitNilValues: true, keyFn: resolve}
+}
```
vaibhav-datazip, line 32: why not just rename this to NewFlattener instead of creating a new function

types/type_schema.go
vikaxsh: required for showing destination column names, if toggle is off

utils/typeutils/flatten.go
vikaxsh: Renamed to `resolveKey` since reformatting is no longer happening every time.

types/type_schema.go
vaibhav-datazip: One possible iteration order:
  visit "id"         → found = true
  visit "My Column"  → found = true
  visit "updated_at" → found = true
  visit "status"     → found = false   ← last one, this is what gets returned
  return false  ← WRONG, there ARE properties with DestinationColumnName set

so instead we should early return the function as soon as we find a destination column name non nil

utils/typeutils/filter_test.go
```diff
@@ -53,7 +53,7 @@ func TestFilterRecords_LegacyFilter(t *testing.T) {
 	}
 	schema := makeIcebergSchema(map[string]string{"id": "long", "name": "string"})
 
-	result, err := FilterRecords(ctx, records, filter, true, schema)
+	result, err := FilterRecords(ctx, records, filter, true, schema, getCachedReformattedKey)
```
vaibhav-datazip: you need to fix these tests

destination/iceberg/internal/utils.go
```diff
@@ -12,7 +12,7 @@ type ServerClient interface {
 // PartitionInfo represents an Iceberg partition column with its transform, preserving order.
 // Field is the original source column name (used for record.Data lookups before pre-shaping).
 // SchemaField is the reformatted destination column name (used for schema, Java partition spec,
```
vaibhav-datazip: schema field need not be reformatted, should be just resolved

types/type_schema.go
vaibhav-datazip: we should add a comment here as well to point out that this will be valid for write path only when use source column names will be false

vikaxsh: will improve this later if required, for now keeping the behaviour same with staging

types/type_schema.go
```diff
@@ -116,7 +116,7 @@ func (t *TypeSchema) AddTypes(column string, isOlakeColumn bool, types ...DataTy
 	if !found {
 		t.Properties.Store(column, &Property{
 			Type:                  NewSet(types...),
-			DestinationColumnName: utils.Reformat(column),
+			DestinationColumnName: utils.Reformat(column), // Used to render UI column names when source column names are disabled.
```
hash-data, line 119: ui rendering will always be same ?

utils/typeutils/fields.go
```diff
@@ -111,17 +110,12 @@ func (f Fields) ToProperties() map[string]*types.Property {
...
+func (f Fields) FromSchema(schema *types.TypeSchema, resolve func(string) string) {
 	schema.Properties.Range(func(key, value any) bool {
-		fieldName := key.(string)
-		property := value.(*types.Property)
-		// Use the destination column name if set, otherwise use field name for backward compatibility
-		// This ensures consistency with flattened record keys which are also normalized to lowercase
-		// Without this, both uppercase (from schema) and lowercase (from records) entries exist,
-		// causing type conflicts when creating the parquet schema in TypeSchema.ToParquet()
-		normalizedName := property.DestinationColumnName
-		normalizedName = utils.Ternary(normalizedName != "", normalizedName, fieldName).(string)
```
hash-data, line 123: this check does not exist in resolve? 
For example people who not upgraded and have dest column name same as field name and now also not turned the flag which is default behaviour after upgrade?

utils/typeutils/filter.go
```diff
@@ -33,7 +33,7 @@ func FilterRecords(ctx context.Context, records []types.RawRecord, filter types.
 	logger.Infof("filtering records with filter: %+v", filter)
 	conditions := make([]parsedCondition, len(filter.Conditions))
 	for i, cond := range filter.Conditions {
-		cond.Column = utils.Reformat(cond.Column)
+		cond.Column = resolve(cond.Column)
```
hash-data, line 36: dont we have to do this while unmarshalling filter?

utils/typeutils/flatten.go
```diff
@@ -48,30 +43,41 @@ func (f *FlattenerImpl) Flatten(data types.Record) (types.Record, error) {
 	return destination, nil
 }
 
+// resolveKey returns the output key for a column, using the instance-level cache
+// to avoid calling keyFn more than once per unique column name per batch.
+func (f *FlattenerImpl) resolveKey(key string) string {
+	if cached, ok := f.keyCache.Load(key); ok {
+		return cached.(string)
+	}
+	resolved := f.keyFn(key)
```
hash-data: keyFn not a good name, how about resolveColumnName?

destination/iceberg/iceberg.go
```diff
@@ -345,9 +345,12 @@ func (i *Iceberg) FlattenAndCleanData(ctx context.Context, records []types.RawRe
 		}
 
 		// parallel flatten data and detect schema difference
+		// One flattener per batch: the internal cache amortizes resolve calls
+		// across all records so each column name is resolved only once.
+		batchFlattener := typeutils.NewFlattener(i.stream.ResolveColumnName)
```
hash-data, line 350: we can create it stream level as well, just mentioning so that in refactoring we can take benefit of it.

types/catalog.go
```diff
@@ -52,6 +52,8 @@ type StreamMetadata struct {
 	StreamName     string `json:"stream_name"`
 	AppendMode     bool   `json:"append_mode,omitempty"`
 	Normalization  bool   `json:"normalization"`
+	// When enabled, source column names are preserved as-is; otherwise utils.Reformat() is applied to generate destination-safe lowercase column names.
+	UseSourceColumnNames bool `json:"use_source_column_names,omitempty"`
```
hash-data: remove omit empty

types/catalog.go
```diff
@@ -313,6 +315,7 @@ func GetStreamsDelta(oldStreams, newStreams *Catalog) *Catalog {
 				return (oldMetadata.Normalization != newMetadata.Normalization) ||
 					(oldMetadata.PartitionRegex != newMetadata.PartitionRegex) ||
 					(oldMetadata.Filter != newMetadata.Filter) ||
+					(oldMetadata.UseSourceColumnNames != newMetadata.UseSourceColumnNames) ||
```
hash-data, line 318: merge stream also checked?

utils/typeutils/filter.go
hash-data: also one thing, while user add a filter they add the column name as shown in ui, it will not according to source shown. 

So are we sure it will work?

types/type_schema.go
```diff
@@ -173,13 +173,14 @@ func (t *TypeSchema) ToIceberg(defaultColumns bool, stream StreamInterface, incl
 	t.Properties.Range(func(key, value interface{}) bool {
 		prop := value.(*Property)
 		colName := key.(string)
+		outName := stream.ResolveColumnName(colName)
 		// skip non-olake columns in defaultColumns mode unless explicitly included (e.g. partition columns)
-		if !isSelected(utils.Reformat(colName)) || (defaultColumns && !prop.OlakeColumn && !includeSet.Exists(colName)) {
+		if !isSelected(outName) || (defaultColumns && !prop.OlakeColumn && !includeSet.Exists(colName)) {
```
hash-data, line 178: selected columns are checked against reformatted names, but now if the flag is enabled this feature will not work

hash-data (commented):
couple of p0 to check

types/type_schema.go
vikaxsh: it would be like this
image

types/catalog.go
vikaxsh: Yes, UseSourceColumnNames is correctly retained

utils/typeutils/filter.go
vikaxsh: the input in stream is expected to contain source column names as-is

utils/typeutils/fields.go
vikaxsh: ok, adding back

types/type_schema.go
vikaxsh: handled in isSelected function

hash-data:
can you check the getType function which is using reformatted column name to load from properties?
image

vikaxsh:
> can you check the getType function which is using reformatted column name to load from properties? image

yeah it shouldn't cause any issue because it kept for keeping backward compatibility, let's say condition is false here and use source column in also default false,it fallback to reformat branch, now resolve column name function would return as it is

vaibhav-datazip (approved):

PR #979, opened 2026-06-09
feat: 2pc kafka

saksham-datazip:
# Description

Implemented support for Two-Phase Commit (2PC) in Kafka by persisting the consumer group id, partition, and offset information in the destination metadata. This enables reliable recovery and ensures Kafka consumption can resume from the last committed state.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Ran sync and checked metadata.
- [x] changed consumer_group_id in stats.json and confirmed that sync was failing
- [x] Verified partition recovery behavior by:
  1. Ran a sync with **4 partitions**.
  2. Skipped the offset commit for **partition 0**.
  3. Added more messages to **partition 0**.
  4. Added a **new partition** and produced messages to it.
  5. Ran the next sync and verified that the metadata evolved successfully.
  6. Confirmed that **partition 0** performed only the recovery sync and did **not** consume the newly added messages.
  7. Confirmed that messages in the newly added partition were consumed during the recovery sync.
  8. Ran another sync and verified that the pending messages from **partition 0**  were consumed successfully.
  
- [x] Ran empty sync 
- [x] Recreate topic and synced again and confirmed that when endOffset< metacommited offset sync fail successfully
- [x] Also tested on retention period when endOffset < metacommited offset sync successfully passess by skipping messages.

      
## Documentation

## Related PR's (If Any):


drivers/abstract/abstract.go
```diff
@@ -256,27 +256,46 @@ func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *er
...
+				var metadataState any
+				if mtStateValue != nil {
+					ms, setErr := types.SetMetadataState(mtStateValue, "")
+					if setErr != nil {
+						closeErr = fmt.Errorf("failed to set metadata state for stream[%s]: %s", streamID, setErr)
+						continue
+					}
+					types.SetDedupInserts(ms, dedupInserts)
+					metadataState = ms
+				}
```
vikaxsh: both the switch cases share common logic, consider placing it in one place

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -463,8 +463,8 @@ private void updateJsonState(Table table, UpdateProperties updateProperties, Str
               JsonNode payloadNode = mapper.readTree(payload);
               rootNode.put(STATE_FIELD_LATEST_THREAD_ID, threadId);
               if (payloadNode.isObject()) {
-                  // Merge payload directly into root node
-                  payloadNode.fields().forEachRemaining(entry -> rootNode.set(entry.getKey(), entry.getValue()));
+                  // Deep Merge payload directly into root node
+                  mergePayloadIntoRoot(rootNode, payloadNode);
```
vikaxsh, line 466: can we Restructure the state to remove the recursive merge

drivers/kafka/internal/cdc.go
```diff
@@ -363,3 +403,91 @@ func decodeAvroMessage(data []byte, codec *goavro.Codec) (interface{}, error) {
...
+			if rawMetadataStateValue == nil {
+				// if metadata state is not present, create a new one with empty offsets map.
+				streamMetadata[streamID] = &types.KafkaMetadataState{
+					Offsets: make(map[int32]int64),
+				}
+			} else {
+				// if metadata state is present, unmarshal it and cache it.
+				mtStateStr, ok := rawMetadataStateValue.(string)
+				if !ok {
+					return false, fmt.Errorf("failed to typecast metadata state of type[%T] to string", rawMetadataStateValue)
```
vikaxsh: add stream id in log

drivers/kafka/internal/cdc.go
```diff
@@ -363,3 +403,91 @@ func decodeAvroMessage(data []byte, codec *goavro.Codec) (interface{}, error) {
...
+					return false, fmt.Errorf("%w: stream[%s]: consumer_group_id mismatch (destination metadata=%q, current=%q), run clear destination and restart", constants.ErrNonRetryable, streamID, parsedMetadataStateValue.ConsumerGroupID, k.consumerGroupID)
+				}
+
+				// cache the parsed metadata state for this stream
+				streamMetadata[streamID] = &parsedMetadataStateValue
+			}
+		}
+
+		// kafkaCommittedOffset = broker next offset, metaCommittedOffset = destination next offset.
+		// SetPartitions already ensures kafkaCommittedOffset < EndOffset; franz-go resets to log start when kafkaCommittedOffset < startOffset.
```
vikaxsh: what is SetPartitions here

drivers/kafka/internal/cdc.go
```diff
@@ -363,3 +403,91 @@ func decodeAvroMessage(data []byte, codec *goavro.Codec) (interface{}, error) {
...
+func (k *Kafka) syncCommittedOffsetsWithMetadata(ctx context.Context, readerID int, reader *kgo.Client, metadataStates map[string]any, assignedPartitions []types.PartitionKey) (bool, error) {
+	streamMetadata := make(map[string]*types.KafkaMetadataState)
+	var recordsToCommit []*kgo.Record
+
+	// iterate over all assigned partitions for this reader
+	for _, assignedPartition := range assignedPartitions {
+		currentPartitionID := assignedPartition.Partition
+		currentTopic := assignedPartition.Topic
+
+		partitionMeta, ok := k.readerManager.GetPartitionMeta(kafkapkg.PartitionMetadataKey(currentTopic, currentPartitionID))
```
vikaxsh, line 418: GetPartitionMeta skips empty partitions while assignedPartitions still contains them, shouldn't we skip instead of returning error here

drivers/kafka/internal/cdc.go
saksham-datazip: Added

drivers/kafka/internal/cdc.go
saksham-datazip: Changed

drivers/kafka/internal/cdc.go
saksham-datazip: Empty partition will never be assigned to Reader due to PartitionsForStream function

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -486,6 +486,35 @@ private void updateJsonState(Table table, UpdateProperties updateProperties, Str
       }
   }
 
+  // Some drivers (e.g. Kafka) can have multiple writers updating metadata for the same stream.
+  // Perform a one-level merge to preserve fields written by other writers.
+  private void mergePayloadIntoRoot(ObjectNode rootNode, JsonNode payloadNode) {
+      payloadNode.fields().forEachRemaining(entry -> {
+          String incomingStateKey = entry.getKey();
+          ObjectNode incomingStateValue = parseJSONObject(entry.getValue());
```
vikaxsh, line 493: please once check can we avoid object parsing using Jackson's built-in setAll method

drivers/kafka/internal/cdc.go
```diff
@@ -69,6 +67,23 @@ func (k *Kafka) PreCDC(ctx context.Context, streams []types.StreamInterface) err
...
+
+	// Recover broker offsets from destination metadata if needed.
+	isRecoveryPerformed, err := k.syncCommittedOffsetsWithMetadata(ctx, readerID, k.readerManager.GetReader(readerID), metadataStates, assignedPartitions)
+	if err != nil {
+		return nil, fmt.Errorf("reader[%d]: sync committed offsets with metadata failed: %s", readerID, err)
+	}
+
+	// A successful recovery stops processing for this reader so the next run starts from the recovered offsets.
+	if isRecoveryPerformed {
+		return nil, nil
```
vikaxsh, line 85: consider adding a log here

vikaxsh (commented):
please add  2pc integration tests got Kafka as well

drivers/kafka/internal/cdc.go
saksham-datazip: added

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
saksham-datazip: I used the setAll method for now. If you have another approach in mind, could you please leave a dummy function or a placeholder implementation?

.github/workflows/kafka-rebalance-tests.yml
```diff
@@ -0,0 +1,98 @@
+name: Kafka Rebalance Tests
```
vikaxsh, line 1: why do we need to create different workflow?

utils/testutils/test_utils.go
```diff
@@ -900,7 +903,52 @@ func (cfg *IntegrationTest) testIceberg2PCCDCRecovery(
 		},
 	}
 
-	for _, tc := range twoPCCDCTestCases {
+	kafkaTwoPCCDCTestCases := []syncTestCase{
```
vikaxsh: if keys are same we can dynamically  insert values based on driver instead of creating different struct

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -122,46 +136,171 @@ func ExecuteQueryJSON(ctx context.Context, t *testing.T, streams []string, opera
...
+	case "stop_rebalance":
+		stopRebalanceTrigger()
+		t.Logf("stopped rebalance trigger consumer")
+
 	default:
 		t.Fatalf("unsupported operation: %s", operation)
 	}
 }
 
+func waitForSyncProgress(ctx context.Context, t *testing.T, statsPath string) {
```
vikaxsh: why do we need stats file pooling?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -122,46 +136,171 @@ func ExecuteQueryJSON(ctx context.Context, t *testing.T, streams []string, opera
...
+			for {
+				err := client.ProduceSync(ctx, records...).FirstErr()
+				if err == nil {
+					break
+				}
+				if ctx.Err() != nil {
+					require.NoError(t, err, "failed writing kafka rebalance bulk batch (batch_size=%d): %v", batchCount, err)
+				}
+				time.Sleep(200 * time.Millisecond)
+			}
```
vikaxsh: reimplements same logic as writeMessagesWithRetry

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -8,21 +8,30 @@ import (
...
 )
 
 const (
-	partitionCount        = 5
-	avroSchemaRegistryURL = "http://127.0.0.1:8081"
+	partitionCount             = 5
+	rebalanceBulkMessageCount  = 100_000
+	rebalanceBulkPartition     = int32(0)
+	rebalanceBulkBatchSize     = 500
+	kafkaJSONIntegrationBroker = "127.0.0.1:29092"
```
vikaxsh, line 33: what about avoro?

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -122,46 +136,171 @@ func ExecuteQueryJSON(ctx context.Context, t *testing.T, streams []string, opera
...
+		// simulate 2PC failure after destination commit: consumer offset on partition 0 lags at 1
+		commitConsumerGroupOffset(ctx, t, client, KafkaJsonConsumerGroupID, streams[0], 0, 1)
+		writeMessagesWithRetry(ctx, t, client, &kgo.Record{Key: jsonKey, Value: jsonValue, Partition: 0})
+		// add a new partition with one message to simulate evolution of schema map in destination metadata
+		addKafkaPartitions(ctx, t, client, streams[0], 1)
+		writeMessagesWithRetry(ctx, t, client, &kgo.Record{Key: jsonKey, Value: jsonValue, Partition: 5})
+		t.Logf("Rolled back partition %d, added partition %d with 1 message, and added 1 message on partition %d for topic '%s'", 0, 5, 0, streams[0])
+
+	case "insert_rebalance":
+		for written := 0; written < rebalanceBulkMessageCount; written += rebalanceBulkBatchSize {
```
vikaxsh: wrap this logic in a function

.github/workflows/kafka-rebalance-tests.yml
saksham-datazip: We need a saperate workflow because kafka rebalance is a new edge case, not a part of 2pc or normal integration tests.

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: The reason is that we want the rebalance to happen only while processing `streamChanges`. Even with around 100k messages, the `streamChanges` code path for a single partition executes very quickly, so we needed a reliable way to determine when it had actually started reading messages.

Since `stats.json` is updated periodically during the sync (approximately every 2 seconds), we're using it as an indicator to confirm that message processing has begun before triggering the rebalance.

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: We dont need to implement avro as 2pc is format independent and also for rebalance testing , it is destination independent as well as format independent.

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: reimplemented

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Wrapped

utils/testutils/test_utils.go
saksham-datazip: done changes removed extra struct

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -462,8 +462,8 @@ private void updateJsonState(Table table, UpdateProperties updateProperties, Str
               JsonNode payloadNode = mapper.readTree(payload);
               rootNode.put(STATE_FIELD_LATEST_THREAD_ID, threadId);
               if (payloadNode.isObject()) {
-                  // Merge payload directly into root node
-                  payloadNode.fields().forEachRemaining(entry -> rootNode.set(entry.getKey(), entry.getValue()));
+                  // One-level merge payload into root node
+                  mergePayloadIntoRoot(rootNode, payloadNode);
```
vaibhav-datazip, line 466: test the recovery for backward compatibility for all sync modes

drivers/kafka/internal/cdc.go
```diff
@@ -364,3 +403,94 @@ func decodeAvroMessage(data []byte, codec *goavro.Codec) (interface{}, error) {
...
+		}
+
+		// destination metadata offset must not exceed the current partition end offset.
+		// this can happen when a topic is deleted and recreated with fewer messages.
+		if metaCommittedOffset > partitionMeta.EndOffset {
+			return false, fmt.Errorf("%w: stream[%s] topic %s partition %d metadata offset: %d exceeds partition end offset: %d, run clear destination and restart", constants.ErrNonRetryable, streamID, currentTopic, currentPartitionID, metaCommittedOffset, partitionMeta.EndOffset)
+		}
+
+		// kafkaCommittedOffset must not exceed metaCommittedOffset.
+		if kafkaCommittedOffset >= 0 && kafkaCommittedOffset > metaCommittedOffset {
```
vaibhav-datazip, line 477: `kafkaCommittedOffset > metaCommittedOffset` when can this condition hold ?

saksham-datazip: This can be true when user uses the olake specific consumer_group and commit the offset and apart from that i tried to find if their is any other way kafka_commited_offset increases automatically like in postgres case but couldn't find anything.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
saksham-datazip: I tested the backward compatibility using the following approach:
1-> Ran a full refresh on both the staging and feat/kafka-2pc branches and verified that all metadata fields were present as expected and compared them.

2-> Ran a CDC sync on both branches and confirmed that the id field was empty in both cases, while all other fields matched the expected output.
kafka_2pc branch :-
```
    "olake_2pc" : "{\"full_refresh_committed_ids\":[\"public.testing_nothing_min['(0,0)']-max['(4294967295,0)']\"],\"dedup_inserts\":false,\"id\":\"\",\"state\":\"{\\\"lsn\\\":\\\"365/F55ED730\\\"}\"}",
```

staging branch :-
```
 "olake_2pc" : "{\"full_refresh_committed_ids\":[\"public.testing_nothing_min['(0,0)']-max['(4294967295,0)']\"],\"dedup_inserts\":false,\"id\":\"\",\"state\":\"{\\\"lsn\\\":\\\"365/F56207E8\\\"}\"}",

```


3->For the backward compatibility test, I started a completely new sync by running a full refresh. I then inserted a new record and ran a CDC sync with state, but intentionally prevented postCDC from acknowledging the source or updating stats.json by commenting out that part of the code.

After that, I inserted another new record and ran the sync again on the kafka-2pc branch. Since the previous sync had not been acknowledged, the first run performed only the recovery sync and did not read the newly inserted record. On the subsequent sync, after recovery was complete, the new record was successfully processed as expected.

pkg/kafka/types.go
```diff
@@ -47,6 +47,11 @@ type CustomGroupBalancer struct {
 	partitionMeta map[string]types.PartitionMetaData
 }
 
+// NewCustomGroupBalancer returns a balancer that assigns only partitions present in partitionMeta.
+func NewCustomGroupBalancer(partitionMeta map[string]types.PartitionMetaData) *CustomGroupBalancer {
+	return &CustomGroupBalancer{partitionMeta: partitionMeta}
+}
+
```
vaibhav-datazip, line 54: when we have this constructor now , shouldn't we use it in CreateReader in reader.go as well

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -122,46 +136,168 @@ func ExecuteQueryJSON(ctx context.Context, t *testing.T, streams []string, opera
...
+// stopRebalanceTrigger stops the rebalance trigger consumer.
+func stopRebalanceTrigger() {
+	if rebalanceTriggerCancel != nil {
+		rebalanceTriggerCancel()
+		rebalanceTriggerCancel = nil
+	}
+	// Wait for the goroutine to fully exit and the explicit LeaveGroup to be sent.
+	// Without this, the static member lingers in the broker (kgo.InstanceID skips LeaveGroup
+	// on Close) and may still hold a partition assignment when the next sync starts.
+	if rebalanceTriggerDone != nil {
```
vaibhav-datazip, line 298: this may trigger a race condition , as rebalanceTriggerDone is set inside a goroutine . and we are not waiting for it to finish here

saksham-datazip: changed

pkg/kafka/types.go
saksham-datazip: Yes changed that as well

drivers/kafka/internal/kafka_test.go
```diff
@@ -7,6 +7,70 @@ import (
 	"github.com/datazip-inc/olake/utils/testutils"
 )
 
+func kafkaJsonBaseConfig() *testutils.IntegrationTest {
+	testConfig := testutils.GetTestConfig(string(constants.Kafka), "json")
+	kafkaJsonStatsPath = testConfig.HostStatsPath
```
vikaxsh: why do we need this we know stats file always would be ./testdata/stats.json

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -122,46 +136,172 @@ func ExecuteQueryJSON(ctx context.Context, t *testing.T, streams []string, opera
...
+}
+
+// waitForSyncProgress waits for sync progress to start.
+func waitForSyncProgress(ctx context.Context, t *testing.T, statsPath string) {
+	t.Helper()
+
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	for {
```
vikaxsh: ticker/select is boilerplate you don't need.  can use require.Eventually

drivers/kafka/internal/kafka_test_utils.go
```diff
@@ -63,14 +72,19 @@ const (
 )
 
 var (
+	// rebalance trigger
+	rebalanceTriggerCancel context.CancelFunc
+	rebalanceTriggerDone   chan struct{} // closed when the trigger goroutine has fully exited
```
vikaxsh, line 78: consider making it a struct , in concurrent run might cause issue because of package-level globals

drivers/kafka/internal/kafka_test.go
saksham-datazip: We should not hardcode the value but you are correct our current way is not too direct so i changed it to directly fetch hoststatspath from testutils.

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: Changed

drivers/kafka/internal/kafka_test_utils.go
saksham-datazip: As much as i could understand ,this will not happen as of now and making it struct will not gave us any value.

vikaxsh (approved):

vaibhav-datazip (approved):

PR #985, opened 2026-06-11
fix: auto select cursor field for default incremental sync mode (#976)

### LINKED ISSUE #976: Feat: Auto select a cursor value incase of default Incremental sync mode
### Problem

Currently when a source doesn't support CDC sync the default sync mode is set as incremental in streams.json . And the cursor field which is used to track new records is set to blank in that case, so that user can select the best suited field for it. But if the user forgets to add any cursor field in it, this will give an error as shown below

```
"message":"error occurred while reading records: failed to run incremental sync: backfill failed: failed to fetch max cursor values: failed to scan primary cursor value: ERROR: zero-length delimited identifier at or near \"\"\"\" (SQLSTATE 42601)"}
```

This can lead to problem in olake ui as well. When someone dosen't click the stream even once while creating a job the cursor field won't get populated from ui side, as the streams.json sends it empty.

### Solution

Select the available cursor fields 0th value and set it as cursor field while generating the streams.json

Saquib1973:
# Description

Automatically assigns the first available cursor field when the default sync mode is  `INCREMENTAL` during the discovery phase. Additionally, updates `mergeCatalogs` to prevent an old catalog's empty cursor field from overwriting the newly auto-selected default, while still preserving explicit user choices.

Fixes #976

## Type of change

# How Has This Been Tested?

Tested locally using a PostgreSQL database configured without CDC support to force the `INCREMENTAL` fallback.

- [x] Ran the `discover` command on PostgreSQL to verify `streams.json` correctly populates the `cursor_field` with the 0th available field instead of an empty string.
- [x] Ran the `sync` command on PostgreSQL to confirm the pipeline executes successfully without the `zero-length delimited identifier` SQL error.
- [x] Tested the merge stream catalog functionality to ensure that newly generated default cursor fields are not overwritten by empty strings from legacy catalogs.

# Screenshots or Recordings

https://github.com/user-attachments/assets/be6d847f-7972-43ef-88d9-d1ea70c0f6f4

## Documentation

## Related PR's (If Any):

None


Saquib1973:
@nayanj98 my pr is ready for review

nayanj98:
Assigning @vaibhav-datazip to review your PR

drivers/abstract/abstract.go
```diff
@@ -113,6 +113,9 @@ func (a *AbstractDriver) Discover(ctx context.Context, maxDiscoverThreads int, i
 			convStream.SyncMode = types.CDC
 		} else if convStream.SupportedSyncModes.Exists(types.INCREMENTAL) {
 			convStream.SyncMode = types.INCREMENTAL
+			if convStream.CursorField == "" && convStream.AvailableCursorFields.Len() > 0 {
```
vaibhav-datazip: isn't cursorField = "" guard redundant ?
also can you add a crisp and clear comment here why this has been added and how it helps

types/catalog.go
```diff
@@ -145,7 +145,9 @@ func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
 		if exists {
 			// preserve metadata from old
 			newStream.Stream.SyncMode = oldStream.Stream.SyncMode
-			newStream.Stream.CursorField = oldStream.Stream.CursorField
+			if oldStream.Stream.CursorField != "" {
+				newStream.Stream.CursorField = oldStream.Stream.CursorField
+			}
```
vaibhav-datazip, line 152: add a test case for this in catalog_test as well

drivers/abstract/abstract.go
```diff
@@ -113,6 +113,9 @@ func (a *AbstractDriver) Discover(ctx context.Context, maxDiscoverThreads int, i
...
+				convStream.CursorField = convStream.AvailableCursorFields.Array()[0]
+			}
```
vaibhav-datazip, line 119: also currently when we run incremental sync with empty cursor field it throws the error which was earlier added in the issue, wont it be better that we throw clear error telling that cursor field is empty instead when cursor field is empty

drivers/abstract/abstract.go
Saquib1973: According to the issue, user can select cursorField during discover
- If cursorField already filled skip
- If the cursorField is empty and there is a availableCursorFields, set cursrorField to the first availableCursorField

drivers/abstract/abstract.go
Saquib1973: added a check for cursorField in incremental mode : https://github.com/datazip-inc/olake/pull/985/changes/8ad877fa9b5a1d9fb461671ad3630cad75dd1c54

image

drivers/abstract/abstract.go
vaibhav-datazip: no I am saying can this ever happen that we get cursorField as non empty in discover function here

Saquib1973: In Discover(), every column is added to AvailableCursorFields using WithCursorField() while generating productSchema, but cursorField is never set.

types/catalog.go
Saquib1973: added a test case : https://github.com/Saquib1973/olake/commit/7eb4c6273866cf22fabd0ca22dd6e1aff4b79931

drivers/abstract/abstract.go
vaibhav-datazip: then why do we need the check of cursorField==""

types/catalog_test.go
```diff
@@ -699,6 +699,74 @@ func TestCatalogMergeCatalogs(t *testing.T) {
 				},
 			},
 		},
+		// when old stream has empty CursorField, new-catalogs CursorField should be used instead of being overwritten
+		{
```
vaibhav-datazip, line 703: can you add one more test if not present (once check the test cases), when cursor field is given earlier and new is changed

Saquib1973: there already exist a test case for this -> single stream merge

drivers/abstract/abstract.go
Saquib1973: removed the redundant check : https://github.com/Saquib1973/olake/commit/d98470fe97ccd77d9a36b23f75f8e90b93ad3a95

vaibhav-datazip:
@Saquib1973 please check the integration tests once and fix them

Saquib1973:
> @Saquib1973 please check the integration tests once and fix them

can you please start the integration test for this commit : https://github.com/datazip-inc/olake/pull/985/changes/c227a88efa0aaf0380ede579d2d8fbe67e698628

utils/testutils/test_utils.go
```diff
@@ -1155,6 +1156,66 @@ func (cfg *IntegrationTest) Test2PCIntegration(t *testing.T) {
...
+
+	normActual, err := strip(actual, true) // Validate and strip actual catalog
+	if err != nil {
+		return nil, nil, err
+	}
+
+	normExpected, err := strip(expected, false) // Strip expected catalog for 1:1 comparison
+	if err != nil {
+		return nil, nil, err
+	}
```
vaibhav-datazip: what's the need of removing it from expected ?it isn't present there at all right ?

Saquib1973: yes that is not required as , expected does not have any cursor_field. I have removed strip for expected stream : https://github.com/datazip-inc/olake/pull/985/changes/af6a16742740cd3da22c4afd48a0284c91f4aa2e

vaibhav-datazip:
LGTM @vikaxsh will be doing a 2nd review

utils/testutils/test_utils.go
```diff
@@ -1189,8 +1245,15 @@ func (cfg *IntegrationTest) TestIntegration(t *testing.T) {
 			if err != nil {
 				return fmt.Errorf("failed to read actual streams JSON: %s", err)
 			}
-			if !utils.NormalizedEqual(string(streamsJSON), string(testStreamsJSON)) {
-				return fmt.Errorf("streams.json does not match expected test_streams.json\nExpected:\n%s\nGot:\n%s", string(streamsJSON), string(testStreamsJSON))
+
+			// Validate and strip the non-deterministic auto-selected cursor_field
+			// from incremental streams before comparing the catalogs.
```
vikaxsh, line 1250: we can make it deterministic my sorting available cursor fields array and update  each test stream file with that value

drivers/abstract/incremental.go
```diff
@@ -16,6 +16,13 @@ import (
 )
 
 func (a *AbstractDriver) Incremental(mainCtx context.Context, pool *destination.WriterPool, streams ...types.StreamInterface) error {
+	for _, stream := range streams {
```
vikaxsh, line 19: Why do we need this check here? We're already validating the cursor field in the `Validate` function in `types/stream_configured.go`.

PR #987, opened 2026-06-12
fix: handle-toast-updates

### LINKED ISSUE #984: Postgres CDC to S3 Parquet sets large jsonb / LOB column to null
## Summary

  When using Olake with PostgreSQL as the source and S3 Parquet as the destination, some `jsonb` columns are being written as `null` during CDC sync.

  This looks similar to the LOB/null column issue seen with AWS DMS.

  ## Environment

  - Source: PostgreSQL
  - Destination: S3 Parquet
  - Sync mode: CDC
  - Affected column type: `jsonb`
  - Example affected column: `invoice_details`

  ## Issue

  During CDC, the `jsonb` column becomes `null` in the destination Parquet files, even though the source PostgreSQL table contains non-null values.

  The affected `jsonb` values are relatively large. Most values in this column are around 3 KB to 15 KB.

  ## Expected behavior

Image

Image

  The `jsonb` column should be replicated correctly to S3 Parquet during CDC, preserving the original value from PostgreSQL.

  ## Actual behavior

  The `jsonb` column is written as `null` in the destination during CDC for affected rows.

  ## Impact

  This causes data loss in the destination because important JSON payloads are missing after CDC replication.

  Using append-only mode with query-side deduplication may work as a temporary workaround, but CDC mode should ideally preserve the full `jsonb` value.

  ## Questions

  1. Is there currently a size limit for `jsonb` / LOB columns in PostgreSQL CDC?
  2. Could this be related to PostgreSQL TOASTed values?
  3. Is there any configuration to force Olake to fetch full values for large `jsonb` columns during CDC?
  4. Is support for large `jsonb` / LOB columns in CDC planned?

  ## Additional context

  The issue happens with PostgreSQL CDC to S3 Parquet. The affected column is `jsonb`, and most values are larger than 2 KB, so this may be related to PostgreSQL TOAST /
  large object handling.

saksham-datazip:
# Description
Previously, unchanged TOAST columns in update events could be emitted as null values when pgoutput marked the column as unchanged and did not send the actual column data.
This behavior still exists for other replica identity modes.
The fix preserves the existing value from the old tuple for REPLICA IDENTITY FULL tables, preventing unchanged TOAST columns from becoming null after updates.

Fixes #984

## Type of change

# How Has This Been Tested?
- [x] Tested INSERT, UPDATE, DELETE, and TRUNCATE events on tables with REPLICA IDENTITY FULL and compared results with staging. Verified the fix correctly preserves unchanged TOAST columns.
- [x] Tested INSERT, UPDATE, DELETE, and TRUNCATE events on tables with REPLICA IDENTITY DEFAULT and compared results with staging. Behavior matched staging.
- [x] Tested INSERT, UPDATE, DELETE, and TRUNCATE events on tables with REPLICA IDENTITY USING INDEX and compared results with staging. Behavior matched staging.
- [x] Tested INSERT and TRUNCATE events on tables with REPLICA IDENTITY NOTHING and compared results with staging. Behavior matched staging.
# Screenshots or Recordings
N/A

## Documentation

## Related PR's (If Any):

### DIFF
--- a/pkg/waljs/pgoutput.go
+++ b/pkg/waljs/pgoutput.go
@@ -116,6 +116,9 @@ func (p *pgoutputReplicator) processPgoutputWAL(ctx context.Context, walData []b
 
 	switch msg := logicalMsg.(type) {
 	case *pglogrepl.RelationMessage:
+		if _, relationVisited := p.relationIDToMsgMap[msg.RelationID]; !relationVisited && msg.ReplicaIdentity != 'f' {
+			logger.Warnf("table[%s.%s] replica identity is not FULL, unchanged TOAST column values may be lost during CDC UPDATE events; set REPLICA IDENTITY FULL to avoid data loss", msg.Namespace, msg.RelationName)
+		}
 		p.relationIDToMsgMap[msg.RelationID] = msg
 		return nil
 	case *pglogrepl.BeginMessage:
@@ -136,7 +139,7 @@ func (p *pgoutputReplicator) processPgoutputWAL(ctx context.Context, walData []b
 	}
 }
 
-func (p *pgoutputReplicator) tupleValuesToMap(rel *pglogrepl.RelationMessage, tuple *pglogrepl.TupleData, oldTuple *pglogrepl.TupleData) (map[string]any, error) {
+func (p *pgoutputReplicator) tupleValuesToMap(rel *pglogrepl.RelationMessage, tuple, oldTuple *pglogrepl.TupleData) (map[string]any, error) {
 	data := make(map[string]any)
 	if tuple == nil {
 		return data, nil
--- a/pkg/waljs/pgoutput.go
+++ b/pkg/waljs/pgoutput.go
@@ -136,7 +136,7 @@ func (p *pgoutputReplicator) processPgoutputWAL(ctx context.Context, walData []b
 	}
 }
 
-func (p *pgoutputReplicator) tupleValuesToMap(rel *pglogrepl.RelationMessage, tuple *pglogrepl.TupleData) (map[string]any, error) {
+func (p *pgoutputReplicator) tupleValuesToMap(rel *pglogrepl.RelationMessage, tuple *pglogrepl.TupleData, oldTuple *pglogrepl.TupleData) (map[string]any, error) {
 	data := make(map[string]any)
 	if tuple == nil {
 		return data, nil
@@ -148,6 +148,9 @@ func (p *pgoutputReplicator) tupleValuesToMap(rel *pglogrepl.RelationMessage, tu
 		}
 		colName := rel.Columns[idx].Name
 		colType := rel.Columns[idx].DataType
+		if col.DataType == pglogrepl.TupleDataTypeToast && oldTuple != nil && idx < len(oldTuple.Columns) {
+			col = oldTuple.Columns[idx]
+		}
 		if col.Data == nil {
 			data[colName] = nil
 			continue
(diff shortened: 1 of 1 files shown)


pkg/waljs/pgoutput.go
```diff
@@ -148,6 +148,9 @@ func (p *pgoutputReplicator) tupleValuesToMap(rel *pglogrepl.RelationMessage, tu
 		}
 		colName := rel.Columns[idx].Name
 		colType := rel.Columns[idx].DataType
+		if col.DataType == pglogrepl.TupleDataTypeToast && oldTuple != nil && idx < len(oldTuple.Columns) {
```
vikaxsh, line 157: please add a comment here

pkg/waljs/pgoutput.go
```diff
@@ -136,7 +136,7 @@ func (p *pgoutputReplicator) processPgoutputWAL(ctx context.Context, walData []b
 	}
 }
 
-func (p *pgoutputReplicator) tupleValuesToMap(rel *pglogrepl.RelationMessage, tuple *pglogrepl.TupleData) (map[string]any, error) {
+func (p *pgoutputReplicator) tupleValuesToMap(rel *pglogrepl.RelationMessage, tuple *pglogrepl.TupleData, oldTuple *pglogrepl.TupleData) (map[string]any, error) {
```
vikaxsh:
```suggestion
func (p *pgoutputReplicator) tupleValuesToMap(rel *pglogrepl.RelationMessage, tuple , oldTuple *pglogrepl.TupleData) (map[string]any, error) {
```

pkg/waljs/pgoutput.go
```diff
@@ -148,6 +148,9 @@ func (p *pgoutputReplicator) tupleValuesToMap(rel *pglogrepl.RelationMessage, tu
 		}
 		colName := rel.Columns[idx].Name
 		colType := rel.Columns[idx].DataType
+		if col.DataType == pglogrepl.TupleDataTypeToast && oldTuple != nil && idx < len(oldTuple.Columns) {
+			col = oldTuple.Columns[idx]
+		}
```
vikaxsh, line 159: @hash-data Should we log a warning when a NULL TOAST column cannot be captured with the default replica identity?

hash-data: there will be a lot of warnings, if we can log at once will be fine.

hash-data: While setting up CDC, we can print what the replica identity is being set to?

pkg/waljs/pgoutput.go
saksham-datazip: Changed

pkg/waljs/pgoutput.go
saksham-datazip: ADDED

vikaxsh (approved):

JGustavo0:
> Previously, unchanged TOAST columns in update events could be emitted as null values when pgoutput marked the column as unchanged and did not send the actual column data.
> This behavior still exists for other replica identity modes.

Hum...debezium already has a flag for this. I think the path forward is to propagate the flag downstream to help with the merge. Otherwise, we won't be able to differentiate between an update to `null` and an unchanged `TOAST` column.
Am I missing something or not understanding this correctly?

saksham-datazip:
> Hum...debezium already has a flag for this. I think the path forward is to propagate the flag downstream to help with the merge. Otherwise, we won't be able to differentiate between an update to `null` and an unchanged `TOAST` column. Am I missing something or not understanding this correctly?

@JGustavo0 ,We will get back to you soon regarding this.

saksham-datazip:
@JGustavo0 Good catch! You're absolutely right. We'll add support for this soon. Thanks for highlighting it.

saksham-datazip:
@JGustavo0 Thanks for pinpointing the issue! Starting from **v0.8.0**, we've replaced `null` with a placeholder for unchanged TOAST column values during update events.

PR #990, opened 2026-06-15
feat: add bulk read path for backfill and incremental sync DB2

vishalm0509:
# Description

Adds a fast bulk-read path for the DB2 driver using a patched `go_ibm_db` driver. Instead of reading one row at a time via `database/sql` Scan, OLake now fetches rows in batches and converts them with pre-compiled per-column converters.
**Requires a companion PR to `go_ibm_db`** (block fetch, `ReadBatch`, `QueryWithArgs`, null mask support).

**Why**
The standard `MapScan` path does one CGO `SQLFetch` per row and reflection-based conversion per cell. On large tables this becomes the bottleneck. The ReadBatch path:
- Fetches **200 rows per ODBC call** (block fetch via `SQL_ATTR_ROW_ARRAY_SIZE`)
- Writes directly into **typed Go slices**, skipping `database/sql.Scan`
- Overlaps **I/O and CPU** with a producer–consumer pipeline
- Resolves column types **once at setup**, not on every cell

**Key pieces:**
| Component | Role |
|-----------|------|
| `readBatchConcurrent` | Entry point for backfill + incremental reads |
| `buildColBuffers` | Allocates typed `*[]T` slices per column scan type |
| `colBatch` + `freeCh` | Reuses buffer sets across fetches (no per-batch alloc in steady state) |
| `buildResolvedConverters` | Pre-compiles one converter closure per column |
| Null mask (`nulls[][]bool`) | Distinguishes SQL NULL from zero-values (e.g. `NULL` vs `0.00` decimal) |
| `QueryWithArgs` | Supports incremental queries with bound parameters (`?` placeholders) |

**Read path selection:**
- **Default:** ReadBatch (full refresh backfill + incremental)
- **Opt-out:** `DB2_READ_MODE=mapscan` → existing `MapScanConcurrent` path (set as env)

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] **Full refresh** — ReadBatch backfill on test tables with mixed data types (`NUMERIC_TEST`, `OLAKE_RB_TEST`)
- [x] **Incremental** — backfill + delta read on `OLAKE_INCR_TEST` (parameterized queries via `QueryWithArgs`)
- [x] **MapScan parity** — same streams synced with `DB2_READ_MODE=mapscan`; output matches ReadBatch path
- [x] **Destination schema** — column types verified in Athena/Iceberg against DB2 source types
- [x] **Data correctness** — source vs destination row comparison on synced tables

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

--- 

⚠️ Breaking Changes Made 
1. `timestamp`, `time`, `date` are synced as UTC.
2. `decfloat` is stored in the destination as String (earlier float64)
> 	// DECFLOAT (16/34) exceeds float64 range and precision; the ODBC driver returns
	// values as decimal text, so preserve them as string to avoid ParseFloat overflow.


drivers/db2/internal/db2.go
```diff
@@ -253,18 +256,82 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
 		return nil, typeutils.ErrNullValue
 	}
 
+	// ColumnTypeDatabaseTypeName returns the DB2 type name in uppercase as-is,
+	// so this comparison must be uppercase "TIME".
```
saksham-datazip: we can use .toUppercase function to keep it generic?

drivers/db2/internal/incremental.go
```diff
@@ -42,7 +51,6 @@ func (d *DB2) StreamIncrementalChanges(ctx context.Context, stream types.StreamI
 		if err := jdbc.MapScan(rows, record, d.dataTypeConverter); err != nil {
 			return fmt.Errorf("failed to scan record: %s", err)
 		}
-
```
saksham-datazip, line 45: why this space is removed?

saksham-datazip, line 45: why remove this space?

drivers/db2/internal/db2.go
```diff
@@ -253,18 +256,82 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
...
 	return typeutils.ReformatValue(olakeType, value)
 }
+
+// isFixedLengthCharType returns true for DB2 types that pad values to the
+// declared width with trailing spaces. Only these types should have trailing
+// spaces trimmed; variable-length types (VARCHAR, CLOB, etc.) store spaces
+// as meaningful data and must not be trimmed.
+func isFixedLengthCharType(dbTypeName string) bool {
+	switch strings.ToLower(dbTypeName) {
+	case "char", "character", "chararr", "chararray", "graphic":
```
saksham-datazip: From where we get that these are only dynamic length datatypes?

saksham-datazip: If columnType will be of Capital charachters than shouldnt we use these constant string in uppercase itself ?

drivers/db2/internal/db2.go
```diff
@@ -253,18 +256,82 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
...
-			return strings.TrimSpace(string(v)), nil
+	if isFixedLengthCharType(columnType) && olakeType == types.String {
+		switch v := value.(type) {
+		case string:
+			return typeutils.ReformatValue(olakeType, strings.TrimRight(v, " "))
+		case []byte:
+			return typeutils.ReformatValue(olakeType, strings.TrimRight(string(v), " "))
 		}
 	}
+	return d.reformatResolved(value, olakeType)
```
vaibhav-datazip:
```suggestion
	return typeutils.ReformatValue(olakeType, value)
```

drivers/db2/internal/db2.go
```diff
@@ -253,18 +256,82 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
...
+}
 
+// TODO(BEFORE MERGE): make the comments concise after review
+// reformatResolved performs the value conversion once the olake DataType is
```
saksham-datazip: if merging this function is in todo is fine than its ok else can't we make it inline ?

drivers/db2/internal/db2.go
```diff
@@ -20,6 +21,8 @@ import (
 	"golang.org/x/crypto/ssh"
 )
 
+const db2DefaultFetchSize = 200
```
saksham-datazip: From where we took this value ?And shouldnt we define it in constants.go ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+
+import (
+	"context"
+	"database/sql/driver"
+	"fmt"
+	"io"
+	"reflect"
+	"time"
+
+	goibmdb "github.com/ibmdb/go_ibm_db"
```
saksham-datazip: shouldnt this goibmdb name also convey that this driver is modified?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+
+	"github.com/datazip-inc/olake/drivers/abstract"
+	"github.com/datazip-inc/olake/utils"
+)
+
+// batchChannelDepth controls how many filled rowsets can be queued between the
+// producer and consumer. 4 lets the producer run ~800 rows ahead.
+const batchChannelDepth = 4
```
saksham-datazip: Should we define it in constants.go or all constants from all files in this db2 driver at same place?And how did we select this value ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+// row, convertAssign reflection per cell), this file uses two features of the
+// patched driver:
+//
+//  1. Block fetch (SQL_ATTR_ROW_ARRAY_SIZE): SQLFetch returns db2DefaultFetchSize
+//     rows per CGO call instead of one.
+//
+//  2. ReadBatch(*[]T, ...): reads the entire current rowset directly into
+//     typed Go slices, bypassing database/sql.Scan and convertAssign entirely.
+//
+// Producer-consumer concurrency
```
saksham-datazip, line 17: Can we add a comment on what producer and consumer does?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
+package driver
```
saksham-datazip, line 1: Are we sure that this is the right place for this file, it shouldn't be in package folder as in other driver we dont have extra files in internal folder?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+// colBatch bundles a reusable set of column buffers with the matching
+// []interface{} argument slice handed to ReadBatch. Recycling these across
+// fetches (via a free-list channel) removes the per-batch buildColBuffers
+// allocations that the alloc profile showed at ~25% of all heap allocations.
+//
+// nulls holds one []bool per column (each sized fetchSize). The driver's
+// ReadBatchWithNulls fills nulls[c][i] = true when column c, row i is SQL NULL.
+// This is what lets the fast path preserve the NULL-vs-zero-value distinction
+// for every type — the typed destination slices alone cannot express NULL.
+type colBatch struct {
```
saksham-datazip, line 61: Can we define all types struct altogether meaning 2 struct are defined here one abova and also don't we need one new file defining types.go like the way kafka has ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+// batchMsg is the unit sent from producer to consumer: one filled colBatch plus
+// the row count for that fetch.
+type batchMsg struct {
+	n     int
+	batch *colBatch
+}
+
+// readBatchConcurrent executes query (with optional args) via the patched driver,
+// fetches rows in bulk with ReadBatch, converts them to OLake records, and calls onMessage.
+func (d *DB2) readBatchConcurrent(ctx context.Context, query string, args []any, onMessage abstract.BackfillMsgFn) error {
```
saksham-datazip, line 70: This function readBatchConcurrent is the main function other function buildColBuffers is just a wrapper around it so we should keep readBatchConcurrent above buildColBuffers for ease not the other way around?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+		c, ok := driverConn.(*goibmdb.Conn)
+		if !ok {
+			return fmt.Errorf("readBatchConcurrent: not a *go_ibm_db.Conn (%T)", driverConn)
+		}
+
+		// set the fetch size for the connection to the default value.
+		c.SetFetchSize(db2DefaultFetchSize)
+
+		var rows *goibmdb.Rows
+		// len(args) = 0: backfill for full_refresh or cdc sync
```
saksham-datazip: CDC not supported so this is wrong

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+		// len(args) > 0: incremental sync
+		if len(args) == 0 {
+			dr, err := c.Query(query, nil)
+			if err != nil {
+				return fmt.Errorf("readBatchConcurrent: query: %w", err)
```
saksham-datazip: Change %w to %s or %v

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+		}
+
+		// set the fetch size for the connection to the default value.
+		c.SetFetchSize(db2DefaultFetchSize)
+
+		var rows *goibmdb.Rows
+		// len(args) = 0: backfill for full_refresh or cdc sync
+		// len(args) > 0: incremental sync
+		if len(args) == 0 {
+			dr, err := c.Query(query, nil)
```
saksham-datazip: change this variable name dr to something meaningful

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+			}
+		}
+		defer rows.Close()
+
+		// Collect column metadata once.
+		colNames := rows.Columns()
+		nCols := len(colNames)
+		scanTypes := make([]reflect.Type, nCols)
+		colTypeNames := make([]string, nCols)
+		for i := 0; i < nCols; i++ {
```
saksham-datazip: we can use range here for this loop?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+		// Pre-compile per-column converters once, with the olake DataType
+		// resolved a single time per column (not per cell). See
+		// buildResolvedConverters.
+		convFuncs := d.buildResolvedConverters(colTypeNames)
+
+		// Read the fetch size back from the rows object — this is the value the
+		// driver actually applied to SQL_ATTR_ROW_ARRAY_SIZE (confirming that
+		// SetFetchSize took effect) and is used to pre-allocate buffer sets.
+		fetchSize := rows.FetchSize()
+		if fetchSize <= 0 {
```
saksham-datazip: We already set this above in this function so why do it again ?

drivers/db2/internal/db2.go
```diff
@@ -253,18 +256,82 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
...
+
+// buildResolvedConverters compiles one converter closure per column with the
+// olake DataType (and TIME flag) resolved exactly once — not per cell. This is
+// the completed form of the optimisation buildReadBatchConverters only started:
+// the column type string is parsed a single time at setup instead of on every
+// one of the (rows × columns) cells.
+func (d *DB2) buildResolvedConverters(colTypeNames []string) []func(interface{}) (interface{}, error) {
+	funcs := make([]func(interface{}) (interface{}, error), len(colTypeNames))
+	for i, typeName := range colTypeNames {
+		tn := typeName // captured per column; never changes
```
saksham-datazip: do we need this extra variable

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+// buildColBuffers allocates a fresh set of typed column slices for one ReadBatch
+// call. The element type for each column is determined from its ColumnTypeScanType.
+func buildColBuffers(scanTypes []reflect.Type, fetchSize int) []colBuffer {
+	timeType := reflect.TypeOf(time.Time{})
+	bufs := make([]colBuffer, len(scanTypes))
+
+	for i, t := range scanTypes {
+		switch {
+		case t.Kind() == reflect.Bool:
+			s := make([]bool, fetchSize)
```
vaibhav-datazip: can we use better names instead of single words like c , s etc

drivers/db2/internal/backfill.go
```diff
@@ -49,6 +50,16 @@ func (d *DB2) ChunkIterator(ctx context.Context, stream types.StreamInterface, c
...
 
+	logger.Debugf("Starting backfill for %s with chunk %v using query: %s", stream.ID(), chunk, stmt)
+
+	// TODO: decide how to handle read_mode: env or flag
+	readMode := os.Getenv("DB2_READ_MODE")
+	if readMode != "mapscan" {
+		logger.Infof("[DB2] backfill read path: readbatch (DB2_READ_MODE=%q, fetch_size=%d)", readMode, db2DefaultFetchSize)
+		return d.readBatchConcurrent(ctx, stmt, args, OnMessage)
+	}
+	logger.Infof("[DB2] backfill read path: mapscan (DB2_READ_MODE=%q, fetch_size=%d)", readMode, db2DefaultFetchSize)
```
vaibhav-datazip: why can't we use read batch concurrent only incase the driver has data ready in its buffer ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+			for {
+				var cb *colBatch
+				select {
+				case <-ctx.Done():
+					return nil
+				case cb = <-freeCh:
+				}
+
+				n, err := rows.ReadBatch(cb.args, cb.nulls)
+				if err == io.EOF {
```
vaibhav-datazip: it will be better if you use error.is

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+			defer close(batchCh)
+			for {
+				var cb *colBatch
+				select {
+				case <-ctx.Done():
+					return nil
+				case cb = <-freeCh:
+				}
+
+				n, err := rows.ReadBatch(cb.args, cb.nulls)
```
vaibhav-datazip: can you once check if in ODBC drivers there can be case where n > 0 as well as err can be EOF , 

in that case we can miss some data as of now

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+		// Free-list of reusable buffer sets. Sized to cover every colBatch that
+		// can be in flight simultaneously: one being filled by the producer, up
+		// to batchChannelDepth queued in batchCh, and one being drained by the
+		// consumer. The consumer returns each set to freeCh once it is done, so
+		// no per-batch column-buffer allocation happens in steady state.
+		const inFlight = batchChannelDepth + 2
+		freeCh := make(chan *colBatch, inFlight)
+		for k := 0; k < inFlight; k++ {
+			bufs := buildColBuffers(scanTypes, fetchSize)
+			args := make([]interface{}, nCols)
```
vaibhav-datazip: shouldn't we rename it to something different as args already exists in this function

drivers/db2/internal/backfill.go
```diff
@@ -49,6 +50,16 @@ func (d *DB2) ChunkIterator(ctx context.Context, stream types.StreamInterface, c
 		stmt = jdbc.DB2RidChunkScanQuery(stream, chunk, filter)
 	}
 
+	logger.Debugf("Starting backfill for %s with chunk %v using query: %s", stream.ID(), chunk, stmt)
+
+	// TODO: decide how to handle read_mode: env or flag
+	readMode := os.Getenv("DB2_READ_MODE")
+	if readMode != "mapscan" {
+		logger.Infof("[DB2] backfill read path: readbatch (DB2_READ_MODE=%q, fetch_size=%d)", readMode, db2DefaultFetchSize)
+		return d.readBatchConcurrent(ctx, stmt, args, OnMessage)
```
vaibhav-datazip: also why no transaction is being used in this read ?

drivers/db2/internal/db2.go
```diff
@@ -253,18 +256,82 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
...
 
+// TODO(BEFORE MERGE): make the comments concise after review
+// reformatResolved performs the value conversion once the olake DataType is
+// already known. The ReadBatch fast path resolves the DataType a single time
+// per column (see buildResolvedConverters) and calls this directly, avoiding
+// the per-cell strings.ToLower/Split inside ExtractAndMapColumnType — which the
+// CPU profile showed accounting for ~26% of total sync CPU.
+func (d *DB2) reformatResolved(value interface{}, olakeType types.DataType) (interface{}, error) {
 	return typeutils.ReformatValue(olakeType, value)
 }
```
vaibhav-datazip: what's the point of this function can't we directly call the typeutils function

drivers/db2/internal/db2.go
```diff
@@ -253,18 +256,82 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
...
+// olake DataType (and TIME flag) resolved exactly once — not per cell. This is
+// the completed form of the optimisation buildReadBatchConverters only started:
+// the column type string is parsed a single time at setup instead of on every
+// one of the (rows × columns) cells.
+func (d *DB2) buildResolvedConverters(colTypeNames []string) []func(interface{}) (interface{}, error) {
+	funcs := make([]func(interface{}) (interface{}, error), len(colTypeNames))
+	for i, typeName := range colTypeNames {
+		tn := typeName // captured per column; never changes
+		olakeType := typeutils.ExtractAndMapColumnType(tn, db2TypeToDataTypes)
+		funcs[i] = func(raw interface{}) (interface{}, error) {
```
saksham-datazip: Can we have a better name for this funcs variable ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+	// driver's ensureCapXxx only reallocates when len < n, and we pre-allocate
+	// to fetchSize which equals the driver's rowset size, the underlying array
+	// is never replaced and the captured slice header stays valid.
+	getValue func(i int) interface{}
+}
+
+// buildColBuffers allocates a fresh set of typed column slices for one ReadBatch
+// call. The element type for each column is determined from its ColumnTypeScanType.
+func buildColBuffers(scanTypes []reflect.Type, fetchSize int) []colBuffer {
+	timeType := reflect.TypeOf(time.Time{})
```
saksham-datazip: Can we inline this ?No need for this extra variable.

drivers/db2/internal/readbatch.go
saksham-datazip: can we make this variable name better?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+					return nil
+				case cb = <-freeCh:
+				}
+
+				n, err := rows.ReadBatch(cb.args, cb.nulls)
+				if err == io.EOF {
+					return nil
+				}
+				if err != nil {
+					return fmt.Errorf("ReadBatch: %w", err)
```
saksham-datazip: Make it %s or %w .

saksham-datazip: Also look for other places where u  have used %w and fix them

drivers/db2/internal/db2.go
```diff
@@ -253,18 +256,82 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
...
+				if err != nil && !errors.Is(err, typeutils.ErrNullValue) {
+					return nil, err
+				}
+				return v, nil
+			}
+			if isFixedLengthCharType(tn) && olakeType == types.String {
+				switch v := raw.(type) {
+				case string:
+					return typeutils.ReformatValue(olakeType, strings.TrimRight(v, " "))
+				case []byte:
```
saksham-datazip: In readbatch , we need this string handling after patch so can you look into patched driver code if it is possible that only it returns []byte only ?No need to change its functionality right .

drivers/db2/internal/db2.go
```diff
@@ -253,18 +256,82 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
...
+					return typeutils.ReformatValue(olakeType, strings.TrimRight(string(v), " "))
+				}
+			}
+			v, err := d.reformatResolved(raw, olakeType)
```
saksham-datazip: No need for this function we can revert it right ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,290 @@
...
+// fetches (via a free-list channel) removes the per-batch buildColBuffers
+// allocations that the alloc profile showed at ~25% of all heap allocations.
+//
+// nulls holds one []bool per column (each sized fetchSize). The driver's
+// ReadBatchWithNulls fills nulls[c][i] = true when column c, row i is SQL NULL.
+// This is what lets the fast path preserve the NULL-vs-zero-value distinction
+// for every type — the typed destination slices alone cannot express NULL.
+type colBatch struct {
+	bufs  []colBuffer
+	args  []interface{}
```
saksham-datazip: this args variable name exist twice here one for querry argument and one in this type interface and also when we call this inside readbatch function in driver we have named this parameter as dst so can we keep this consistent.

drivers/db2/internal/db2.go
vishalm0509: Done

drivers/db2/internal/readbatch.go
vishalm0509: Done

drivers/db2/internal/backfill.go
vishalm0509: Let's discuss

drivers/db2/internal/readbatch.go
vishalm0509: Done

drivers/db2/internal/readbatch.go
vishalm0509: Yes, there can be such cases as we are using multi row fetch: 
https://ftpdocs.broadcom.com/cadocs/0/CA%20Chorus%2003%200%2000-ENU/Bookshelf_Files/HTML/Performance%20Handbook%20for%20DB2/Multi_Row_FETCH.html

Made changes

drivers/db2/internal/readbatch.go
vishalm0509: Done

drivers/db2/internal/db2.go
vishalm0509: Right, removed

drivers/db2/internal/backfill.go
vishalm0509: Even for mapScan Concurrent we are not using any Isolation level, it uses the default one (Cursor Stability). Will look into this...

drivers/db2/internal/db2.go
vishalm0509: We receive the value in uppercase only right. I am not able to understand the comment

drivers/db2/internal/incremental.go
vishalm0509: Reverted

drivers/db2/internal/db2.go
vishalm0509: These are the character types, other types like VARCHAR have varying length.

vishalm0509: for multiple types, we are using lowercase and for single type we are using as it is (uppercase). that's the practice already followed in the DB2 driver code in OLake

drivers/db2/internal/db2.go
vishalm0509: removed

drivers/db2/internal/db2.go
vishalm0509: we have conducted an experiment and there are no much difference in speed after setting the value to `200`. 

If there's more memory footprint we can reduce the size via env as well. 

Not adding it to the `constants.go` as it is specific to this file

drivers/db2/internal/readbatch.go
vishalm0509: Yes, we will update it. Not yet updated it as driver code not yet merged

drivers/db2/internal/readbatch.go
vishalm0509: Already present.

drivers/db2/internal/readbatch.go
vishalm0509: Yes, this is an internal function. This is the right place.

drivers/db2/internal/readbatch.go
vishalm0509: I have coupled all the structs together. types file is not required in this case.

drivers/db2/internal/readbatch.go
vishalm0509: Yes, makes sense.

drivers/db2/internal/readbatch.go
vishalm0509: just the comment, I have removed it. In case of `full_refresh` we don't have args. So the case is still correct.

drivers/db2/internal/readbatch.go
vishalm0509: done

drivers/db2/internal/db2.go
vishalm0509: Earlier we were handling both the cases. Now only the check for fixedLenCharacter was added.

vishalm0509: I have removed that as well, since we are not handling that case in other drivers.

drivers/db2/internal/readbatch.go
vishalm0509: We set it in the driver, above. Now we are getting the set value from the driver and using it to create the buffer.

drivers/db2/internal/db2.go
vishalm0509: Not required. Removed

drivers/db2/internal/readbatch.go
vishalm0509: I have added a comment why `+ 2` is added. 4 is just a number, tried with 2, 3, 4. 

Need to discuss and check if we can increase it more without OOM risk.

drivers/db2/internal/incremental.go
saksham-datazip, line 45: Please revert this

drivers/db2/internal/db2.go
```diff
@@ -258,13 +261,33 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
...
+			if strings.EqualFold(typeName, "TIME") {
+				v, err := typeutils.ReformatTimeValue(raw)
+				if err != nil && !errors.Is(err, typeutils.ErrNullValue) {
+					return nil, err
+				}
+				return v, nil
+			}
 
-	// in db2, string based types come in byte format
-	if olakeType == types.String {
```
saksham-datazip, line 263: Can you confirm why we are removing this check from here ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,330 @@
...
+		} else {
+			driverArgs := make([]driver.Value, len(args))
+			for i, a := range args {
+				driverArgs[i] = a
+			}
+			rows, err = conn.QueryWithArgs(query, driverArgs)
+			if err != nil {
+				return fmt.Errorf("readBatchConcurrent: query with args: %s", err)
+			}
+		}
```
saksham-datazip: I think ,we can directly use querrywithArgs instead of calling querry and thus revert the changes we did in driver code as well for this function.

drivers/db2/internal/db2.go
saksham-datazip: We are using this variable db2DefaultFetchSize mainly in readbatch.go and apart from that we are just using it for logs so can we keep both variables together db2DefaultFetchSize and batchChannelDepth.
Just for reference we can use something like this inside db2.go:-
```
const db2DefaultFetchSize = 200
// batchChannelDepth controls how many filled rowsets can be queued between the
// producer and consumer. 4 lets the producer run ~800 rows ahead.
const batchChannelDepth = 4
```

drivers/db2/internal/db2.go
```diff
@@ -258,13 +261,33 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
...
+// buildResolvedConverters compiles one converter closure per column with the
+// olake DataType (and TIME flag) resolved exactly once — not per cell. This is
+// the completed form of the optimisation buildReadBatchConverters only started:
+// the column type string is parsed a single time at setup instead of on every
+// one of the (rows × columns) cells.
+func (d *DB2) buildResolvedConverters(colTypeNames []string) []func(interface{}) (interface{}, error) {
+	convFuncs := make([]func(interface{}) (interface{}, error), len(colTypeNames))
+	for i, typeName := range colTypeNames {
+		olakeType := typeutils.ExtractAndMapColumnType(typeName, db2TypeToDataTypes)
+		convFuncs[i] = func(raw interface{}) (interface{}, error) {
```
saksham-datazip: This function is only called from readbatch.go only but defined inside db2.go so can we define this function inside readbatch.go only?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,330 @@
...
+			}
+			return nil
+		}
+
+		return utils.ConcurrentF(ctx, consumer, producer)
+	})
+}
+
+// makeColBuffer allocates a typed column slice and wires dest/getValue for ReadBatch.
+func makeColBuffer[T any](fetchSize int) colBuffer {
```
saksham-datazip: Can we change the ordering of makeColBuffer and newColBatch as newColBatch is calling makeColBuffer only ??

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,330 @@
...
+	dests []interface{}
+	nulls [][]bool
+}
+
+// batchMsg is the unit sent from producer to consumer: one filled colBatch plus
+// the row count for that fetch.
+type batchMsg struct {
+	n     int
+	batch *colBatch
+}
```
saksham-datazip, line 66: Can you confirm it once please that if we can merge these two structs colBatch and colBuffer ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,330 @@
...
+// fetches (via a free-list channel) removes the per-batch buildColBuffers
+// allocations that the alloc profile showed at ~25% of all heap allocations.
+//
+// nulls holds one []bool per column (each sized fetchSize). The driver's
+// ReadBatchWithNulls fills nulls[c][i] = true when column c, row i is SQL NULL.
+// This is what lets the fast path preserve the NULL-vs-zero-value distinction
+// for every type — the typed destination slices alone cannot express NULL.
+type colBatch struct {
+	bufs  []colBuffer
+	dests []interface{}
```
saksham-datazip: these two 	bufs  and dests are representing the same thing so can you look into this for once that if we can change the input in patched driver or some other approach we can follow to remove this redundancy ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,330 @@
...
+			if err != nil {
+				return fmt.Errorf("readBatchConcurrent: query with args: %s", err)
+			}
+		}
+		defer rows.Close()
+
+		// Collect column metadata once.
+		colNames := rows.Columns()
+		nCols := len(colNames)
+		scanTypes := make([]reflect.Type, nCols)
```
saksham-datazip, line 106: Can we add a comment of what scanTypes is ?

drivers/db2/internal/db2.go
```diff
@@ -258,13 +261,33 @@ func (d *DB2) dataTypeConverter(value interface{}, columnType string) (interface
...
+			if strings.EqualFold(typeName, "TIME") {
+				v, err := typeutils.ReformatTimeValue(raw)
+				if err != nil && !errors.Is(err, typeutils.ErrNullValue) {
+					return nil, err
```
saksham-datazip: in case of ReformatTimeValue when error can be not nil ?also when error can be ErrNullValue?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,330 @@
...
+		// the patched driver resets SQL_ATTR_ROW_ARRAY_SIZE to 1. In that case
+		// rows.FetchSize() returns 1 and we skip the producer-consumer goroutine
+		// pair (two channel hops per row) and run a simple inline loop instead.
+		fetchSize := rows.FetchSize()
+
+		// processBatch converts one filled colBatch (n rows) to OLake records.
+		// Shared by both the inline (FetchSize=1) and concurrent paths below.
+		processBatch := func(cb *colBatch, n int) error {
+			for rowIdx := 0; rowIdx < n; rowIdx++ {
+				record := make(map[string]interface{}, nCols)
```
saksham-datazip: Can we use range here ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,330 @@
...
+		// driver actually applied to SQL_ATTR_ROW_ARRAY_SIZE. When the result
+		// set contains non-bindable columns (CLOB/DBCLOB/XML), BindColumns in
+		// the patched driver resets SQL_ATTR_ROW_ARRAY_SIZE to 1. In that case
+		// rows.FetchSize() returns 1 and we skip the producer-consumer goroutine
+		// pair (two channel hops per row) and run a simple inline loop instead.
+		fetchSize := rows.FetchSize()
+
+		// processBatch converts one filled colBatch (n rows) to OLake records.
+		// Shared by both the inline (FetchSize=1) and concurrent paths below.
+		processBatch := func(cb *colBatch, n int) error {
```
saksham-datazip: Can you please check if we can change cb variable name to something more meaningful?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,330 @@
...
+		// ── Concurrent path (FetchSize>1, all columns bindable) ─────────────
+		// Buffered channel: producer stages batchChannelDepth rowsets ahead.
+		batchCh := make(chan batchMsg, batchChannelDepth)
+
+		// Free-list of reusable buffer sets. Sized to cover every colBatch that
+		// can be in flight simultaneously: one being filled by the producer, up
+		// to batchChannelDepth queued in batchCh, and one being drained by the
+		// consumer. The consumer returns each set to freeCh once it is done, so
+		// no per-batch column-buffer allocation happens in steady state.
+		const inFlight = batchChannelDepth + 2
```
saksham-datazip: This is another const so as per our discussion we should group them together ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,330 @@
...
+					}
+				} else {
+					freeCh <- cb
+				}
+				switch {
+				case errors.Is(readErr, io.EOF):
+					return nil
+				case readErr != nil:
+					return fmt.Errorf("ReadBatch: %s", readErr)
+				}
```
saksham-datazip: Can we put this rest part of code inside case cb = <-freeCh: ?for producer?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,330 @@
...
+			defer close(batchCh)
+			for {
+				var cb *colBatch
+				select {
+				case <-ctx.Done():
+					return nil
+				case cb = <-freeCh:
+				}
+
+				n, readErr := rows.ReadBatch(cb.dests, cb.nulls)
```
saksham-datazip: shouldnt we do error handling first for this ?

drivers/db2/internal/incremental.go
vishalm0509: Removed mapScanConcurrent

drivers/db2/internal/db2.go
vishalm0509: Removed datatype converter

drivers/db2/internal/db2.go
vishalm0509: Moved.

drivers/db2/internal/readbatch.go
vishalm0509: Okay, a top-down approach, got it. Done

drivers/db2/internal/readbatch.go
vishalm0509: There is a scope to merge them. Done.

drivers/db2/internal/readbatch.go
vishalm0509: Done

drivers/db2/internal/readbatch.go
vishalm0509: Done

drivers/db2/internal/db2.go
vishalm0509: It's declaration states that it can return an error, but actually never returns any. Leaving the function untouched,  have updated the caller to return result directly.

drivers/db2/internal/readbatch.go
vishalm0509: Done

drivers/db2/internal/readbatch.go
vishalm0509: "batch" ?

I have updated to that

drivers/db2/internal/readbatch.go
vishalm0509: Done

drivers/db2/internal/readbatch.go
vishalm0509: We can.

drivers/db2/internal/readbatch.go
vishalm0509: There can be certain rows even after we get an error (EOF) so we handle those first if present and then check the errors.

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,308 @@
...
+
+		// Inline path (FetchSize=1, non-bindable columns present)
+		// When the driver downgraded to FetchSize=1 due to CLOB/DBCLOB/XML
+		// columns, producer-consumer goroutine overhead (two channel round-trips
+		// per row) costs more than it saves. Run a single-goroutine loop instead.
+		if fetchSize == 1 {
+			batch := newColBatch(scanTypes, 1)
+			for {
+				if err := ctx.Err(); err != nil {
+					return nil
```
vaibhav-datazip: shouldn't we return ctx err here

drivers/db2/internal/backfill.go
```diff
@@ -49,21 +49,10 @@ func (d *DB2) ChunkIterator(ctx context.Context, stream types.StreamInterface, c
...
-	defer tx.Rollback()
-
 	logger.Debugf("Starting backfill for %s with chunk %v using query: %s", stream.ID(), chunk, stmt)
 
-	setter := jdbc.NewReader(ctx, stmt, func(ctx context.Context, query string, queryArgs ...any) (*sql.Rows, error) {
-		return tx.QueryContext(ctx, query, args...)
-	})
-
-	return jdbc.MapScanConcurrent(setter, d.dataTypeConverter, OnMessage)
+	logger.Infof("[DB2] backfill read path: readbatch (fetch_size=%d)", db2DefaultFetchSize)
```
vaibhav-datazip: can we add this in debug log only ? , as this will be called for each chunk and just create noise

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,308 @@
...
+				return typeutils.ReformatTimeValue(raw)
+			}
+
+			v, err := typeutils.ReformatValue(olakeType, raw)
+			if err != nil && !errors.Is(err, typeutils.ErrNullValue) {
+				return nil, err
+			}
+			return v, nil
+		}
+	}
```
vaibhav-datazip, line 301: can you once check that the string data which we are getting is correct and dosent have extra spaces as well

as earlier , when the driver returned the data in bytes array we used to convert it to string and then trimmed spaces.

the reason being char(10) -> "check" would've had check followed by 5 spaces -> something like this "check     "

drivers/db2/internal/backfill.go
vishalm0509: That log is not needed I feel. Removed

drivers/db2/internal/readbatch.go
vishalm0509: Yes. Done

drivers/db2/internal/readbatch.go
vishalm0509: Yes, now we do not get string in bytes as we not using `database/sql`. WE get it as string only

vaibhav-datazip: do we get extra spaces or are those handled as well at the driver side ?

drivers/db2/internal/readbatch.go
vishalm0509: Done

drivers/db2/internal/readbatch.go
vishalm0509: We get the extra space, and we are removing it here in OLake side.

drivers/db2/internal/backfill.go
saksham-datazip: Here their are two logs simulaneously also regarding second log it gaves an impression that readbatch path is one of the path which seems kind of misleading so can you please look into this ?Also if changing make the same changes in incremental.go

drivers/db2/internal/readbatch.go
saksham-datazip: Can you check again once i dont think they are yet merged .Also with this changes we might be able to reduce the complication inside producer function so check that once as well.

drivers/db2/internal/db2.go
saksham-datazip: Function removed itself as of now

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,317 @@
...
+			// recover so a panic in ReadBatch returns an error instead of crashing the process.
+			defer func() {
+				if r := recover(); r != nil {
+					err = fmt.Errorf("readBatchConcurrent producer: panic: %v", r)
+				}
+			}()
+			defer close(batchCh)
+			for {
+				var batch *colBatch
+				select {
```
saksham-datazip: Can we directly inline delare this variable?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,317 @@
...
+// turns a raw ReadBatch value into the OLake representation (e.g. int64,
+// timestamp, "HH:MM:SS" for TIME). DB2 type -> OLake type mapping.
+func (d *DB2) buildResolvedConverters(colTypeNames []string) []func(interface{}) (interface{}, error) {
+	convFuncs := make([]func(interface{}) (interface{}, error), len(colTypeNames))
+	for i, typeName := range colTypeNames {
+		olakeType := typeutils.ExtractAndMapColumnType(typeName, db2TypeToDataTypes)
+		convFuncs[i] = func(raw interface{}) (interface{}, error) {
+			if strings.EqualFold(typeName, "TIME") {
+				return typeutils.ReformatTimeValue(raw)
+			}
```
saksham-datazip, line 287: Need to discuss

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,317 @@
...
+					n, readErr := rows.ReadBatch(batch.dests, batch.nulls)
+					if n > 0 {
+						select {
+						case <-ctx.Done():
+							return ctx.Err()
+						case batchCh <- batchMsg{n: n, batch: batch}:
+						}
+					} else {
+						freeCh <- batch
+					}
```
saksham-datazip: Can you once check if we even avoid this if n>0 check even if dont use this check consumer will treat no-op and directly do this operation freeCh <- batch?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,317 @@
...
+					switch {
+					case errors.Is(readErr, io.EOF):
+						return nil
+					case readErr != nil:
+						return fmt.Errorf("ReadBatchConcurrent producer: %s", readErr)
```
saksham-datazip: Should we directly use if clause instead of switch case here for better readability ?

drivers/db2/internal/readbatch.go
```diff
@@ -0,0 +1,317 @@
...
+		// per row) costs more than it saves. Run a single-goroutine loop instead.
+		if fetchSize == 1 {
+			logger.Infof("[DB2] readBatchConcurrent: block fetch downgraded to fetch_size=1 (non-bindable columns such as CLOB/DBCLOB/XML); using inline read path")
+			batch := newColBatch(scanTypes, 1)
+			for {
+				if err := ctx.Err(); err != nil {
+					return ctx.Err()
+				}
+				n, readErr := rows.ReadBatch(batch.dests, batch.nulls)
+				if n > 0 {
```
saksham-datazip: Can you once check if we can directly pass n by pointer ?instead of returning it ?

drivers/db2/internal/backfill.go
vishalm0509: This log is not present. Already removed

drivers/db2/internal/readbatch.go
vishalm0509: No, I think think the current approach is better

drivers/db2/internal/readbatch.go
vishalm0509: Not doing it

drivers/db2/internal/backfill.go
saksham-datazip: Now it is removed

vaibhav-datazip (approved):

saksham-datazip (approved):
LGTM

PR #991, opened 2026-06-17
feat: add support for capture instance management on read replica

### LINKED ISSUE #966: MSSQL Capture Instance Management with Read Replica connection
### Problem

_What problem this feature is going to solve. Why this problem is important_

For the MSSQL connector which can connect to the read replica to replicate, capture instance management is disabled. This means that the automated create/delete of capture instances on the primary cannot work, like it would if connecting to the primary.

### Solution

Proposed solution is to be able to connect to the primary for capture instance management, but read replica for replication when capture instance management is enabled regardless of the JDBC connection parameters to use the read replica.

vishalm0509:
# Description

## What
Adds optional `primary_config` to the MSSQL driver so capture instance management `(sp_cdc_enable_table / sp_cdc_disable_table)` can run against the Always On primary while ingestion stays on the read-only secondary.

## Why 
Read replicas cannot execute CDC write operations. When `manage_capture_instances` is enabled on a secondary, those calls were skipped entirely. This lets users connect to a readable replica for CDC reads while still managing capture instances on the primary.

## How
- Introduced `PrimaryConfig` (host, port, username, password) on the main config; database, SSL, and JDBC params are inherited from the main connection.
- Refactored connection setup into `setupDBConnection`; when on a replica with `manage_capture_instances + primary_config`, opens a second connection to the primary.
- Reuses the existing `ssh_config` tunnel for both connections (one bastion, two DB dials).
- Routes `sp_cdc_*` writes through primaryClient in manageCaptureInstances; DDL reads remain on the secondary.
- Updated `spec.json` and UI schema to expose `primary_config` when `manage_capture_instances` is enabled.

## Assumptions
- The primary and secondary nodes are accessible through the same ssh tunnel
- The primary and secondary nodes will have same ssl modes
- For primary node in ci mgmnt jdbc params are not required.

Fixes #966

## Type of change

# How Has This Been Tested?
- [x] Happy scenario , only host node provided (which happens to be primary). so capture instance is getting  created from main host and everything is working fine for fr+cdc and subsequent cdc sync.
- [x] only host node provided and it happens to be secondary , no primary node provided. sync runs for fr+cdc and cdc as well, but when alter table happens no ci management is done.
- [x] secondary is provided as main host and capture instance management is toggled on , and primary node is also provided there , in this case as well sync runs perfectly for fr+cdc, and subsequent cdc sync , also if alter table happens the ci is getting managed.
- [x] if both main host and ci mgmnt host are given , and we give the primary node in main host and the secondary in ci mgmnt host then too sync is running fine and capture instance is getting managed.
- [x] Tested it for ssh tunnel for both primary and secondary.

# Screenshots or Recordings

image

image

## Documentation

## Related PR's (If Any):

### DIFF
--- a/drivers/mssql/internal/cdc.go
+++ b/drivers/mssql/internal/cdc.go
@@ -240,7 +240,7 @@ func (m *MSSQL) manageCaptureInstances(ctx context.Context, streamIDs []string, 
 	mgmtClient := m.client
 	if m.isReadReplica {
 		if m.primaryClient == nil {
-			logger.Warn("manage_capture_instances enabled on replica but no primary_config provided; skipping")
+			logger.Warn("manage_capture_instances enabled on replica but no primary_config provided. capture instance management is skipped")
 			return nil
 		}
 		mgmtClient = m.primaryClient
--- a/drivers/mssql/internal/cdc.go
+++ b/drivers/mssql/internal/cdc.go
@@ -235,11 +235,15 @@ func (m *MSSQL) manageCaptureInstances(ctx context.Context, streamIDs []string, 
 		return nil
 	}
 
-	// Read replicas are read-only; sp_cdc_enable_table / sp_cdc_disable_table require
-	// write access to the primary.
+	// Read replicas are read-only; sp_cdc_enable_table / sp_cdc_disable_table require write access to the primary.
+	// When primary_config is provided, use that connection.
+	mgmtClient := m.client
 	if m.isReadReplica {
-		logger.Debug("manage_capture_instances is enabled but the connection targets a read-only replica; capture instance management is skipped")
-		return nil
+		if m.primaryClient == nil {
+			logger.Warn("manage_capture_instances enabled on replica but no primary_config provided; skipping")
+			return nil
+		}
+		mgmtClient = m.primaryClient
 	}
 
 	// Fetch DDL history for all streams in bulk
@@ -287,7 +291,7 @@ func (m *MSSQL) manageCaptureInstances(ctx context.Context, streamIDs []string, 
 		for idx, capture := range instances {
 			if idx != activeIdx && (currentCursorLSN == "" || capture.startLSN <= currentCursorLSN) {
 				query := jdbc.MSSQLCDCDisableCaptureInstanceQuery()
-				_, err := m.client.ExecContext(ctx, query, capture.schema, capture.table, capture.instanceName)
+				_, err := mgmtClient.ExecContext(ctx, query, capture.schema, capture.table, capture.instanceName)
 				if err != nil {
 					return fmt.Errorf("failed to delete obsolete capture instance %s for %s: %w", capture.instanceName, streamID, err)
 				}
@@ -311,7 +315,7 @@ func (m *MSSQL) manageCaptureInstances(ctx context.Context, streamIDs []string, 
 				newInstanceName := fmt.Sprintf("olake_%s_%d", streamPart, time.Now().Unix())
 
 				createCaptureInstanceQuery := jdbc.MSSQLCDCCreateCaptureInstanceQuery()
-				_, err := m.client.ExecContext(ctx, createCaptureInstanceQuery, latestCapture.schema, latestCapture.table, newInstanceName)
+				_, err := mgmtClient.ExecContext(ctx, createCaptureInstanceQuery, latestCapture.schema, latestCapture.table, newInstanceName)
 				if err != nil {
 					return fmt.Errorf("failed to create new capture instance for schema drift on %s: %w", streamID, err)
 				}
--- a/drivers/mssql/internal/cdc.go
+++ b/drivers/mssql/internal/cdc.go
@@ -238,12 +238,11 @@ func (m *MSSQL) manageCaptureInstances(ctx context.Context, streamIDs []string, 
 	// Read replicas are read-only; sp_cdc_enable_table / sp_cdc_disable_table require write access to the primary.
 	// When primary_config is provided, use that connection.
 	mgmtClient := m.client
-	if m.isReadReplica {
-		if m.primaryClient == nil {
-			logger.Warn("manage_capture_instances enabled on replica but no primary_config provided. capture instance management is skipped")
-			return nil
-		}
+	if m.primaryClient != nil {
 		mgmtClient = m.primaryClient
+	} else if m.isReadReplica {
+		logger.Warn("manage_capture_instances enabled on read replica but no primary_config provided. capture instance management is skipped")
+		return nil
 	}
 
 	// Fetch DDL history for all streams in bulk
--- a/drivers/mssql/internal/config.go
+++ b/drivers/mssql/internal/config.go
@@ -22,25 +22,30 @@ type Config struct {
 	SSLConfiguration       *utils.SSLConfig  `json:"ssl"`
 	ManageCaptureInstances bool              `json:"manage_capture_instances"`
 	SSHConfig              *utils.SSHConfig  `json:"ssh_config"`
+	PrimaryConfig          *PrimaryConfig    `json:"primary_config,omitempty"`
+}
+
+// PrimaryConfig holds connection details for the primary replica when the main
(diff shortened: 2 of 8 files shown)


drivers/mssql/internal/cdc.go
```diff
@@ -235,11 +235,15 @@ func (m *MSSQL) manageCaptureInstances(ctx context.Context, streamIDs []string,
...
-	// Read replicas are read-only; sp_cdc_enable_table / sp_cdc_disable_table require
-	// write access to the primary.
+	// Read replicas are read-only; sp_cdc_enable_table / sp_cdc_disable_table require write access to the primary.
+	// When primary_config is provided, use that connection.
+	mgmtClient := m.client
 	if m.isReadReplica {
-		logger.Debug("manage_capture_instances is enabled but the connection targets a read-only replica; capture instance management is skipped")
-		return nil
+		if m.primaryClient == nil {
+			logger.Warn("manage_capture_instances enabled on replica but no primary_config provided; skipping")
```
vaibhav-datazip: shouldn't we add the same line in logs as earlier 
```
capture instance management is skipped
```

drivers/mssql/internal/config.go
```diff
@@ -65,14 +70,52 @@ func (c *Config) Validate() error {
...
+	if c.PrimaryConfig != nil {
+		if err := c.PrimaryConfig.Validate(); err != nil {
+			return err
+		}
+	}
+
 	return utils.Validate(c)
 }
 
+func validateSQLConnection(host string, port int, username, password string, isPrimary bool) error {
```
vaibhav-datazip: isPrimaryNode, wouldn't that be better naming ?

drivers/mssql/internal/config.go
vaibhav-datazip, line 146: have you checked if we can connect with same ssl move to the primary node as well

drivers/mssql/internal/mssql.go
```diff
@@ -152,6 +136,56 @@ func (m *MSSQL) Close() error {
...
+func setupDBConnection(
+	ctx context.Context,
+	uri string,
+	host string,
+	existingSSH *ssh.Client,
+	sshCfg *utils.SSHConfig,
+	maxConns int,
+) (*sqlx.DB, *ssh.Client, error) {
+	sshTunnel := existingSSH
+	if sshTunnel == nil && sshCfg != nil && sshCfg.Host != "" {
```
vaibhav-datazip: well instead of returning the whole connection , can we just not get the custom dialer from this function and then use it to create connections to both secondary and primary node ?

also , can this happen in some case that primary node is only accessible via ssh tunnel whereas our secondary node is accessible without tunnel ?

drivers/mssql/internal/config.go
```diff
@@ -65,14 +70,52 @@ func (c *Config) Validate() error {
...
+	return c.buildURI(c.Host, c.Port, c.Username, c.Password)
+}
+
+// primaryURI returns the sqlserver:// connection string for the primary replica,
+// inheriting database, SSL, and JDBC params from the main config.
+func (c *Config) primaryURI() string {
+	if c.PrimaryConfig == nil {
+		return ""
+	}
+	return c.buildURI(c.PrimaryConfig.Host, c.PrimaryConfig.Port, c.PrimaryConfig.Username, c.PrimaryConfig.Password)
```
vaibhav-datazip: this is copying all the jdbc url params as well in primary connection , what if user has passed `pplicationIntent=ReadOnly` , you can also check how common it is to pass this parameter for read replica connections

drivers/mssql/internal/mssql.go
```diff
@@ -116,7 +80,21 @@ func (m *MSSQL) Setup(ctx context.Context) error {
...
+				m.config.primaryURI(),
+				m.config.PrimaryConfig.Host,
+				m.sshClient,
+				m.config.SSHConfig,
+				1,
+			)
+			if err != nil {
+				return fmt.Errorf("failed to connect to primary for capture instance management: %w", err)
+			}
+			logger.Info("conected to primary MSSQL connection for capture instance management")
```
vaibhav-datazip: connected spelling

drivers/mssql/internal/mssql.go
```diff
@@ -116,7 +80,21 @@ func (m *MSSQL) Setup(ctx context.Context) error {
 
 	m.isReadReplica = m.detectReadReplica(ctx)
 	if m.isReadReplica {
```
vaibhav-datazip, line 90: On Azure SQL Managed Instance, this DMV may return no rows even on secondaries. If detection returns false (assume primary), m.primaryClient is never created, and manageCaptureInstances falls through to use m.client.

we should primary client whenever its provided

drivers/mssql/internal/mssql.go
```diff
@@ -152,6 +136,56 @@ func (m *MSSQL) Close() error {
...
+		}
+		connector.Dialer = &mssqlSSHDialer{sshClient: sshTunnel, host: host}
+		client = sqlx.NewDb(sql.OpenDB(connector), "sqlserver")
+	} else {
+		db, err := sql.Open("sqlserver", uri)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to open MSSQL connection: %s", err)
+		}
+		client = sqlx.NewDb(db, "sqlserver")
+	}
```
vaibhav-datazip: maybe the logic in this function can be simplified , because the way mssql handles dial function is somewhat different please check once

drivers/mssql/internal/mssql.go
```diff
@@ -116,7 +80,21 @@ func (m *MSSQL) Setup(ctx context.Context) error {
...
+			m.primaryClient, _, err = setupDBConnection(
+				ctx,
+				m.config.primaryURI(),
+				m.config.PrimaryConfig.Host,
+				m.sshClient,
+				m.config.SSHConfig,
+				1,
+			)
+			if err != nil {
+				return fmt.Errorf("failed to connect to primary for capture instance management: %w", err)
```
vaibhav-datazip: change it to %s

drivers/mssql/internal/config.go
```diff
@@ -65,14 +70,52 @@ func (c *Config) Validate() error {
...
+		}
+	}
+
 	return utils.Validate(c)
 }
 
+func validateSQLConnection(host string, port int, username, password string, isPrimary bool) error {
+	prefix := utils.Ternary(isPrimary, "primary_config:", "").(string)
+	if host == "" {
+		return fmt.Errorf("%sempty host name", prefix)
```
vaibhav-datazip, line 85: %s empty

drivers/mssql/internal/cdc.go
vishalm0509: Makes sense. Done

drivers/mssql/internal/config.go
vishalm0509: Okay

drivers/mssql/internal/mssql.go
vishalm0509: Done

drivers/mssql/internal/config.go
vishalm0509: I have confirmed with DevOps regd. this. We will use the same SSL Connection.

drivers/mssql/internal/mssql.go
vishalm0509: I have made changes

drivers/mssql/internal/config.go
vishalm0509: We are now using a separate JSBC_URL_PARAMS for primary database

drivers/mssql/internal/mssql.go
vishalm0509: I have updated the query that'll work in all the instances.

drivers/mssql/internal/mssql.go
vishalm0509: Simplified it.

vaibhav-datazip (approved):

PR #1011, opened 2026-06-30
feat: track source bytes read in stats

vaibhav-datazip:
# Description

Adds `BytesRead` : the total source bytes read and processed and committed in destinaton, to Olake's `stats.json` output across all drivers. Previously only row count (`ReadCount`) was tracked; now both rows and bytes are reported, enabling user to track the bytes which were read and processed by olake.

Also fixes a bug where `ReadCount` (rows) was overcounted when a chunk failed and was retried, the failed attempt's rows were never subtracted, so retried chunks inflated the final row count in `stats.json`.

The PR also adds CPU usage calculation ratio stat.

### What changed

**Per-thread rollback — `destination/writers.go`**

Each `WriterThread` tracks its own contribution to the pool-wide `ReadCount` and `BytesCommitted` via three atomic counters (`recordsPushed`, `recordsFiltered`, `bytesPushed`). `Push(ctx, record, sourceBytes int64)` is the single write path: both the row count and bytes are added live as records are read. If the chunk/run fails, `rollbackStats()` subtracts exactly this thread's numbers from the pool stats. Since each retry creates a fresh `WriterThread` with zeroed counters, retried attempts cannot inflate the final totals.

**Stats output**

`stats.json` now includes `"Bytes Committed"` alongside `"Synced Records"` and `"Speed"`, updated every 2 seconds. The final summary line (`Total records read / Total bytes committed`) now prints  on successful sync

## Type of change

- [x] New feature (non-breaking change which adds functionality) — `BytesCommitted` metric
- [x] Bug fix (non-breaking change which fixes an issue) — `ReadCount` overcounting on retry

# How Has This Been Tested?

Tested against live MySQL (CDC) and Postgres (backfill) → Iceberg destination. Byte accuracy verified against each DB's own storage size functions (mysql, postgres, mongodb, mssql, db2).

- [x] Postgres full refresh cdc and incremental came accurate , without performance loss tested on fhv trips multiple times
- [x] Mysql full refresh cdc and incremental came accurate , without performance loss tested on fhv trips multiple times
- [x] Mssql full refresh cdc and incremental came accurate
- [x] DB2 full refresh cdc and incremental came accurate  
- [x] Injected flush error mid-CDC sync: 2 rows read live (59 B added), flush failed → `rollbackStats()` fired → **`Total records read: 0 | Total bytes committed: 0 B`** 

# Screenshots or Recordings

N/A

## Documentation

## Related PR's (If Any):

N/A


pkg/waljs/pgoutput.go
```diff
@@ -245,6 +327,7 @@ func (p *pgoutputReplicator) emitDelete(ctx context.Context, m *pglogrepl.Delete
 		Kind:         "delete",
 		Data:         values,
 		ExtraColumns: map[string]any{CDCLSN: p.socket.ClientXLogPos.String()},
+		Bytes:        rowBytes,
```
mihir-datazip: I see Bytes being set in `abstract.CDCChange` instances here and `pkg/binlog/filter.go` but we skip it in places like `pkg/waljs/filter.go` any particular reason for it?

mihir-datazip: Anyways it will be better to use a NewCDCChange() instead of making the structure manually everywhere

destination/writers.go
```diff
@@ -265,12 +285,24 @@ func (wt *WriterThread) flush(ctx context.Context, buf []types.RawRecord) (err e
...
+
 func (wt *WriterThread) Close(ctx context.Context, finalMetadataState any) (err error) {
 	select {
 	case <-ctx.Done():
-		err := wt.writer.Close(ctx, finalMetadataState)
-		if err != nil {
-			return fmt.Errorf("failed to close writer: %s", err)
+		// Aborted before commit: writer.Close releases resources without committing,
+		// so nothing this thread pushed reaches the destination — roll its stats back.
+		wt.rollbackStats()
```
mihir-datazip, line 289: Shouldn't we do a wt.group.Block() here?

drivers/abstract/cdc.go
mihir-datazip, line 108: Shouldn't we handle rollback here as well?

drivers/db2/internal/readbatch.go
```diff
@@ -138,8 +141,9 @@ func (d *DB2) readBatchConcurrent(ctx context.Context, query string, args []any,
 						return fmt.Errorf("column %s: %s", colName, err)
 					}
 					record[colName] = conv
+					rowBytes += db2ColumnBytes(conv, colTypeNames[colIdx])
```
mihir-datazip: DB2 alone computes bytes on the post-conversion value, while every other driver measures raw pre-conversion scan values. Why so?

protocol/sync.go
```diff
@@ -146,10 +146,14 @@ var syncCmd = &cobra.Command{
 		}
 
 		state.LogWithLock()
-		// TODO: record count also contain records which arrived in retry attempts, need to remove them
+		// ReadCount/RecordsFiltered are rolled back per-thread on failed or retried
```
mihir-datazip, line 158: We removed this TODO, did we add the logic to remove records with retry attempts?

protocol/sync.go
mihir-datazip, line 154: The PR description claims the final summary "now prints even on sync failure," but RunE still returns at the connector.Read error check before ever reaching the Total records read | Total bytes committed log at line 154, and no defer prints it.
Example: (A sync that commits 90 of 100 chunks then exhausts retries emits no byte/row summary at all)

mihir-datazip (dismissed):

drivers/abstract/cdc.go
vaibhav-datazip: if something fails in post cdc, the data is already committed in that case we wont be needing to revert the number of records, because in next sync those wont sync again because of 2pc. this is currently in iceberg and the parquet writer 2pc pr is in progress

drivers/db2/internal/readbatch.go
vaibhav-datazip: changed it to raw

protocol/sync.go
vaibhav-datazip: we have added the revert logic for number of records hence this todo has been solved

protocol/sync.go
vaibhav-datazip: corrected the PR description

utils/logger/logger.go
```diff
@@ -159,11 +161,15 @@ func StatsLogger(ctx context.Context, statsFunc func() (int64, int64, int64)) {
 				stats := map[string]interface{}{
 					"Writer Threads":           runningThreads,
 					"Synced Records":           syncedRecords,
+					"Bytes Committed":          bytesCommitted,
 					"Memory":                   fmt.Sprintf("%d mb", memStats.HeapInuse/(1024*1024)),
```
hash-data: can we also check the memory, which is not accurate

drivers/abstract/abstract.go
```diff
@@ -18,6 +18,18 @@ type CDCChange struct {
 	Kind         string
 	Data         map[string]any
 	ExtraColumns map[string]any // Driver-specific CDC metadata (e.g., LSN, binlog position, resume token)
+	Bytes        int64
+}
+
+func NewCDCChange(stream types.StreamInterface, timestamp time.Time, kind string, data, extraColumns map[string]any, sourceBytes int64) CDCChange {
```
hash-data, line 25: let us rename it to NewCDCRecord

pkg/binlog/filter.go
```diff
@@ -371,3 +368,62 @@ func decodeBytesToString(b []byte, collationID uint64) (string, error) {
 	}
 	return decoder(b)
 }
+
+// mysqlCDCRowBytes returns the approximate InnoDB on-disk byte sum for a
+// binlog row, using the raw decoded Go values and their MySQL type names.
```
hash-data: create a separate file

drivers/postgres/internal/bytes.go
```diff
@@ -0,0 +1,95 @@
...
+//
+// NUMERIC / DECIMAL use PostgreSQL's base-10000 digit encoding. All other
+// variable-width types (VARCHAR, TEXT, BYTEA, JSON, JSONB, arrays …) use the actual
+// length of the value. NULL values return 0.
+func pgStorageBytes(rawVal any, colType *sql.ColumnType) int64 {
+	if rawVal == nil {
+		return 0
+	}
+
+	switch strings.ToUpper(colType.DatabaseTypeName()) {
```
hash-data, line 32: as database type and column going to be same for a batch cant we cache something here

drivers/s3/internal/bytes.go
```diff
@@ -0,0 +1,57 @@
...
+	case int32, uint32, float32:
+		return 4
+	case int, int64, uint, uint64, float64:
+		return 8
+	case time.Time:
+		return 8
+	case map[string]any:
+		var total int64
+		for _, e := range x {
+			total += valueBytes(e)
```
hash-data, line 47: recursive call often lead to performance decrease

pkg/jdbc/reader.go
```diff
@@ -100,26 +104,38 @@ func MapScan(rows *sql.Rows, dest map[string]any, converter func(value interface
...
+
+	rawVals := make([]any, len(columns))
+	for i := range scanValues {
+		rawVals[i] = *(scanValues[i].(*any)) // Dereference pointer before storing
+	}
+
+	var rowBytes int64
+	if len(rowBytesFn) > 0 && rowBytesFn[0] != nil {
+		rowBytes = rowBytesFn[0](rawVals, colTypes)
 	}
```
hash-data: the thing is if we have column name and type same why we are checking each time for a row? why we cant cache it and use?

drivers/abstract/abstract.go
vaibhav-datazip: but its a constructor for cdc change struct , so should I change the struct name as well

pkg/jdbc/reader.go
```diff
@@ -100,26 +105,41 @@ func MapScan(rows *sql.Rows, dest map[string]any, converter func(value interface
...
 
 	if err := rows.Scan(scanValues...); err != nil {
-		return err
+		return 0, err
+	}
+
+	var colSizer func(*sql.ColumnType, any) int64
+	if len(colBytesFn) > 0 && colBytesFn[0] != nil {
+		colSizer = colBytesFn[0]
 	}
```
hash-data: why we need this check?

pkg/jdbc/reader.go
```diff
@@ -100,26 +105,41 @@ func MapScan(rows *sql.Rows, dest map[string]any, converter func(value interface
...
 
+	var rowBytes int64
 	for i, col := range columns {
 		rawData := *(scanValues[i].(*any)) // Dereference pointer before storing
+		if colSizer != nil {
+			rowBytes += colSizer(colTypes[i], rawData)
```
hash-data:
```suggestion
			rowBytes += calculateSize(colTypes[i], rawData)
```

hash-data: or sizeOf

pkg/jdbc/reader.go
```diff
@@ -128,6 +148,11 @@ func MapScanConcurrent(setter *Reader[*sql.Rows], converter func(value interface
 		scanDests []any // reused pointers for rows.Scan
 	)
 
+	var bytesFn func([]*sql.ColumnType) func(i int, v any) int64
+	if len(rowBytesFn) > 0 && rowBytesFn[0] != nil {
+		bytesFn = rowBytesFn[0]
+	}
```
hash-data: ??

utils/logger/logger.go
```diff
@@ -159,11 +159,17 @@ func StatsLogger(ctx context.Context, statsFunc func() (int64, int64, int64)) {
...
+					"Bytes Read":               bytesRead,
 					"Speed":                    fmt.Sprintf("%.2f rps", speed),
 					"Seconds Elapsed":          fmt.Sprintf("%.2f", timeElapsed),
 					"Estimated Remaining Time": estimatedSeconds,
 				}
+				if memInfo, err := mem.VirtualMemory(); err == nil {
+					stats["Memory"] = fmt.Sprintf("%d mb", memInfo.Used/(1024*1024))
+				}
+				if cpuPercent, err := cpu.Percent(0, false); err == nil && len(cpuPercent) > 0 {
+					stats["CPU Utilization"] = math.Round(cpuPercent[0]*100) / 10000
```
hash-data, line 172: only 0th we need?

drivers/postgres/internal/bytes.go
```diff
@@ -0,0 +1,143 @@
...
+//	OID                        → 4 bytes
+//	MONEY                      → 8 bytes
+//
+// NUMERIC / DECIMAL use PostgreSQL's base-10000 digit encoding. All other
+// variable-width types (VARCHAR, TEXT, BYTEA, JSON, JSONB, arrays …) use the actual
+// length of the value.
+func pgClassifyCol(typeName string) (uint8, int64) {
+	switch strings.ToUpper(typeName) {
+	case "INT2", "SMALLINT", "SMALLSERIAL":
+		return pgColFixed, 2
```
hash-data: prev logic was fine why we cant return the functions and store them while iterating over map?

pkg/jdbc/reader.go
vaibhav-datazip: was using variadic here, as some functions used to pass it others didn't but had removed that it became unnecessary. 
removed the variadic function so this condition is not necessary

pkg/jdbc/reader.go
vaibhav-datazip: used sizeOf here also the bytesFn has been renamed to bytesCalculatorFunction

pkg/jdbc/reader.go
vaibhav-datazip: same here variadic function was used now removed

drivers/postgres/internal/bytes.go
vaibhav-datazip: well I think this one is better approach , we already know what kind of column is here in this row we call row sizer, the array already has those values and we can access them in O1 , while the once needing calculations will be called through cases in our switch case

utils/logger/logger.go
vaibhav-datazip: yes, because we are passing perCPU as false, so we get the aggregate value for all the cpus. the returned value is at 0th index of array of float

pkg/jdbc/reader.go
```diff
@@ -167,10 +182,26 @@ func MapScanConcurrent(setter *Reader[*sql.Rows], converter func(value interface
 
 	// Consumer: convert + emit records.
 	consumer := func(ctx context.Context) error {
+		var sizeOf map[string]func(v any) int64
 		for vals := range valuesCh {
+			// colTypes is safely visible here: the producer sets it before the first
+			// channel send. Build the per-column sizers once for the result set, keyed
+			// by column name (unique for a single-table scan).
+			if sizeOf == nil {
+				sizeOf = make(map[string]func(v any) int64, len(columns))
```
hash-data, line 191: any reason to have this inside the loop?

utils/logger/logger.go
```diff
@@ -463,3 +470,19 @@ func SetupAndStartProcess(processName string, cmd *exec.Cmd) error {
...
+	const unit = 1024
+	if b < unit {
+		return fmt.Sprintf("%d B", b)
+	}
+	div, exp := int64(unit), 0
+	for n := b / unit; n >= unit; n /= unit {
+		div *= unit
+		exp++
+	}
+	return fmt.Sprintf("%.2f %cB", float64(b)/float64(div), "KMGTP"[exp])
```
hash-data, line 487: can you share an example in pr for this

pkg/jdbc/reader.go
vaibhav-datazip: yes, the producer scans through the coltypes and pushes it into the channel, once its in channel we can read it

utils/logger/logger.go
vaibhav-datazip: `Total records read: 15 | Total bytes read: 360 B`

hash-data (approved):

PR #1012, opened 2026-07-01
feat: rolling writer in s3 destination

mihir-datazip:
# Description
This change is regarding https://datazip.atlassian.net/browse/OL-2905

The parquet destination previously buffered an entire partition in the writer's memory and uploaded one file per partition at the end of the sync. Memory grew linearly with partition size and produced a single oversized object. This branch streams each partition to a local file and rolls it into multiple size-bounded files, uploading each as it's sealed.

Fixes # (issue)

## Type of change

# How Has This Been Tested?
- Unit tests to verify rolling, without rolling and partitioned rolling
- Integration tests to verify rolling on a limited size(1MB)
- For benchmarking multiple file uploads, I added a temporary time.Sleep() of 10sec in each file upload and compared master and the new code.
#### Old Writer
image

#### New Writer
image

# Screenshots or Recordings
Old writer(without rolling) -> 2GB single parquet
image

New Writer(with 512 MB rolling) -> 4 files with ~512MB size per file
image

UI testing
image

## Documentation

- [x] Documentation Link: https://github.com/datazip-inc/olake-docs/

## Related PR's (If Any):
https://github.com/datazip-inc/olake-docs/pull/450


destination/parquet/parquet.go
```diff
@@ -129,6 +147,17 @@ func (p *Parquet) Setup(_ context.Context, stream types.StreamInterface, schema
 	p.basePath = filepath.Join(p.stream.GetDestinationDatabase(nil), p.stream.GetDestinationTable())
 	p.schema = make(typeutils.Fields)
 
+	if p.maxFileBytes <= 0 {
+		maxFileSizeMB := float64(defaultMaxFileSizeMB)
+		if p.config.MaxFileSizeMB > 0 {
+			maxFileSizeMB = p.config.MaxFileSizeMB
+		}
+		p.maxFileBytes = int64(maxFileSizeMB * 1024 * 1024)
+	}
```
vikaxsh: we should add <0 validation check in config validate fn

destination/parquet/parquet.go
```diff
@@ -129,6 +147,17 @@ func (p *Parquet) Setup(_ context.Context, stream types.StreamInterface, schema
...
+	if p.checkIntervalForRoll <= 0 {
```
vikaxsh: since defaultRollCheckInterval is not user input, when are we expecting <= 0

destination/parquet/resources/spec.json
```diff
@@ -42,6 +42,13 @@
           "type": "string",
           "title": "S3 Endpoint",
           "description": "Specifies the endpoint URL for S3 compatible services (e.g., MinIO)"
+        },
+        "max_file_size_mb": {
```
vikaxsh, line 46: we need to add this in ParquetUISchema as wll

vikaxsh: also once test with ui as well

destination/parquet/parquet.go
```diff
@@ -173,42 +202,75 @@ func (p *Parquet) Write(_ context.Context, records []types.RawRecord) error {
...
 			maps.Copy(recordsMap, record.OlakeColumns)
 
-			_, err = partitionFile.writer.(*pqgo.GenericWriter[any]).Write([]any{recordsMap})
+			_, err = partitionFile.writer.Write([]any{recordsMap})
 		}
 		if err != nil {
 			return fmt.Errorf("failed to write in parquet file: %s", err)
 		}
+
+		if p.checkForRoll(i, len(records)) {
```
vikaxsh, line 225: we can check size once per writer batch for all partitions, and    remove this overhead

drivers/postgres/internal/postgres_test.go
```diff
@@ -53,6 +53,17 @@ func TestPostgres2PC(t *testing.T) {
 	postgresBaseConfig().Test2PCIntegration(t)
 }
 
+func TestPostgresParquetRollingIntegration(t *testing.T) {
```
vikaxsh: Instead of spinning up a separate, dedicated test suite just to verify file rolling in the integration environment, we can embed this check directly into VerifyParquetSync inside utils/testutils/test_utils.go along with existing runs only

vikaxsh: thats how we can also make it consistent with all the drivers

destination/parquet/parquet.go
mihir-datazip: In tests where we might need different check interval...
As this is a config, we should handle it regardless

destination/parquet/parquet.go
mihir-datazip: We don't know if the single batch itself should roll to multiple files...
I remember discussing this with @hash-data to decouple the batch size to the memory constraints of the files

drivers/postgres/internal/postgres_test.go
mihir-datazip: The issue there is that our current data for Sync test is tiny (~1-3KB) and the parquet-go writer itself has a 32KB write buffer so should we increase the seed data for existing tests in order to test rolling in sync?

mihir-datazip: Done

destination/parquet/parquet.go
mihir-datazip: Added

vikaxsh: we can remove this <0 check from here

vikaxsh: we can remove ==0 check as well

drivers/postgres/internal/testdata/parquet_destination_rolling.json
```diff
@@ -0,0 +1,11 @@
+{
```
vikaxsh: we can add "max_file_size_mb" in existing file only

destination/parquet/parquet.go
mihir-datazip: How do we set default values then? in case of empty field, we wish to use 512MB still

utils/testutils/parquet_rolling.go
```diff
@@ -0,0 +1,164 @@
...
+
+// testParquetRolling exercises the parquet writer's memory/size-based file rolling against a
+// real S3 (MinIO) destination. It runs as a sub-test of the common Sync flow (reusing that
+// flow's container c): it seeds a single bulk table, syncs it with a tiny max_file_size_mb, and
+// asserts the writer split the output into multiple size-bounded files without losing rows.
+//
+// The seeded data (~8MB) is far smaller than the backfill chunk size (EffectiveParquetSize,
+// 2GB), so the whole table is one chunk handled by one writer — meaning multiple bounded
+// files can only come from rolling, not from chunk fan-out.
+func (cfg *IntegrationTest) testParquetRolling(ctx context.Context, t *testing.T, c testcontainers.Container) error {
```
vikaxsh: we can use "currentTestTable", and avoid discover and frequent edit streams, it already contains col_text

utils/testutils/parquet_rolling.go
```diff
@@ -0,0 +1,164 @@
...
+	t.Helper()
+	ctx := context.Background()
+	const bucket = "warehouse"
+	prefix := fmt.Sprintf("%s/%s/", cfg.DestinationDB, table)
+
+	mc, err := minio.New("localhost:9000", &minio.Options{
+		Creds:  credentials.NewStaticV4("admin", "password", ""),
+		Secure: false,
+	})
+	require.NoError(t, err, "failed to create MinIO client")
```
vikaxsh: we can extract out this common logic, already present in DeleteParquetFiles fn

destination/parquet/parquet.go
vikaxsh: this code is enough
 ```
 maxFileSizeMB := float64(defaultMaxFileSizeMB)
		if p.config.MaxFileSizeMB > 0 {
			maxFileSizeMB = p.config.MaxFileSizeMB
		}
		p.maxFileBytes = int64(maxFileSizeMB * 1024 * 1024) 
```
		
		
it will set to default

Comment (mihir-datazip) [reply destination/parquet/parquet.go]: Done

File: drivers/postgres/internal/testdata/parquet_destination_rolling.json (hunk 6 above)
Comment (mihir-datazip) [reply drivers/postgres/internal/testdata/parquet_destination_rolling.json]: Done

File: utils/testutils/parquet_rolling.go (hunk 7 above)
Comment (mihir-datazip) [reply utils/testutils/parquet_rolling.go]: Done

File: utils/testutils/parquet_rolling.go (hunk 8 above)
Comment (mihir-datazip) [reply utils/testutils/parquet_rolling.go]: Unified

File: utils/testutils/parquet_rolling.go [hunk 9]
```diff
@@ -0,0 +1,126 @@
...
+		return fmt.Errorf("failed to delete parquet files before rolling sync: %s", err)
+	}
+	defer func() {
+		if err := DeleteParquetFiles(t, cfg.DestinationDB, testTable); err != nil {
+			t.Logf("cleanup: failed to delete parquet files: %v", err)
+		}
+	}()
+
+	// Full load (no --state) with a small buffer so rolled files hug the threshold.
+	syncCmd := syncCommand(*cfg.TestConfig, false, "parquet", "--destination-buffer-size", "1000")
```
vikaxsh: why do we need to configure --destination-buffer-size

mihir-datazip: Not really needed. We can remove it

vikaxsh:
lgtm

vikaxsh (approved):

Issue #1000, opened 2026-06-23
feat: save 2pc metadata state for every sync


vaibhav-datazip:
### Problem

Currently the metadata.json is generated only for the syncs in which there is some modification in the data of that stream. Because of this there is added handling for metadata vs state.json lsn. 

The reason being even without no new records the lsn or the resume tokens for various databases can move ahead because of keep alive checks or other misc. things

Also cases like one out of many streams commits data so lsn updates in metadata and acknowledgment is also done but state save fails, now lets say the person unselects that particular stream so the lsn validation will fail. 

### Solution

Update and generate new metadata file for each sync even when no new records are added.

ishanyawhocares:
hello, can I please be assigned this issue?

@vaibhav-datazip @nayanj98

PR #1013, opened 2026-07-01
Feat/issue 1000 save metadata every sync

ishanyawhocares:
If StreamChanges returned nil metadata state but no error, fall back to the last-known metadata state that was read at the start of the sync (metadataStates map). This ensures handleWriterCleanup always calls writer.Close with a non-nil payload, so the metadata file is always written.

fixes issue #1000


vaibhav-datazip:
Can you add a video of testing this and show the metadata file values in the sync when 0 records were ingested.

nayanj98:
@ishanyawhocares Have assigned @vaibhav-datazip to review your PR

vaibhav-datazip:
hi @ishanyawhocares , we will be closing this pr soon if there is no response from your side

ishanyawhocares:
Hello @vaibhav-datazip, I will upload the video explanation to this by today.

vaibhav-datazip:
@ishanyawhocares , we are closing this pr as of now, as no testing video was provided and this doesn't seem to solve the issue mentioned.

PR #1015, opened 2026-07-02
fix: discover fails for streams with "." in schema or table name

### LINKED ISSUE #1010: Discovery fails if there is MySQL table with "." in table name
## Environment

- **Olake version**:  0.7.7
- **OS**: `Ubuntu 22.04`

## Description
I have MySQL database and there is table with "."  in table name. When I try to run discovery process for this database OLake fails because it cant properly construct table name.

## Steps to reproduce

1. Create table that has "." in name. For example 
database name: "main" 
table name: "user1.test_table"
2. run Olake discovery

## Expected behavior

Olake should provide list of tables from "main" database.

## Actual behavior
Olake fails with error

FATAL error occurred while waiting for connection group: 
failed with non retryable error: 
failed to produce schema for stream main.user1.test_table: 
failed to process table[main.user1.test_table]: 
invalid stream name format: main.user1.test_table

Probably error is  here  https://github.com/datazip-inc/olake/blob/0948d81fe2d994ddfef28c3500a8410956d859bb/drivers/mysql/internal/mysql.go#L207

in lines 207-208

parts := strings.Split(streamName, ".")    
if len(parts) != 2 {

should be changed on 

parts := strings.SplitN(streamName, ".", 2)    
if len(parts) < 2 {

vaibhav-datazip:
# Description

Discover failed whenever a source table or schema name contained a `.` (e.g. `user1.test_table`). `GetStreamNames` joined schema and table into a single `"namespace.name"` string, and `ProduceSchema` split that string back on `.` , so an embedded dot broke the parse (`invalid stream name format: ...`) in mysql/oracle/mssql/db2, postgres. Because discover runs all streams in one non-retryable concurrency group, a single dotted table aborted discovery for the entire database.

Fixed by replacing the string round-trip with a `types.StreamID{Namespace, Name}` struct carried through the `DriverInterface.GetStreamNames`/`ProduceSchema` contract, across all 8 drivers (mysql, postgres, oracle, mssql, db2, mongodb, s3, kafka). The schema/table boundary is never re-parsed, so a `.` anywhere in the namespace or name is handled correctly.

Fixes #1010

## Type of change

# How Has This Been Tested?

- [x] MySQL: table with `.` in name (`user1.test_table`): discover succeeds
- [x] MySQL: table with `.` in name: full refresh sync, verified rows land correctly in destination
- [x] MySQL: table with `.` in name: CDC insert, update, delete verified end-to-end
- [x] MySQL: table with `.` in name: incremental sync insert verified
- [x] MySQL: database/schema with `.` in name (`sales.eu`): discover + full refresh sync, verified rows queryable
- [x] MySQL: clear destination   with `.` in table name 
- [x] MongoDB: Full refresh + CDC sync verified with tables with `.` in name

# Screenshots or Recordings

## Documentation


drivers/kafka/internal/kafka.go
```diff
@@ -138,28 +138,29 @@ func (k *Kafka) Close() error {
...
 	return topicNames, nil
 }
 
 // TODO: for avro, we use decode messages to get stream properties similar to JSON, we should directly use the avro schema to get stream properties
-func (k *Kafka) ProduceSchema(ctx context.Context, streamName string) (*types.Stream, error) {
+func (k *Kafka) ProduceSchema(ctx context.Context, streamID types.StreamID) (*types.Stream, error) {
+	streamName, streamNamespace := streamID.Name, streamID.Namespace
 	logger.Infof("producing schema for topic [%s]", streamName)
-	stream := types.NewStream(streamName, "topics", nil)
+	stream := types.NewStream(streamID.Name, streamNamespace, nil)
```
saksham-datazip: We already stored streamName than why we are fetching it again can't we use streamID.Name directly ?
Also if changing than please look into other drivers as well like in mongoDb.

vikaxsh:
it would break here
image

image

vikaxsh:
stream id is still using dot
image

types/stream.go
```diff
@@ -9,6 +9,17 @@ import (
...
 )
 
+// StreamID is the fully qualified identity of a source stream, carried as separate fields for namespace and table name
+type StreamID struct {
+	Namespace string // schema / database / owner
+	Name      string // table / collection / topic
+}
+
+// String returns the canonical "namespace.name" identifier, matching Stream.ID(). It is used for logging only.
+func (s StreamID) String() string {
```
vikaxsh, line 19: in logs as well it could be confusing to understood table name if it contains dot

vaibhav-datazip: user should be expecting the log to be like <namespace>.<table name> other wise we can output the log with a space

vaibhav-datazip:
> it would break here image
> 
> image

the names in paths are already normalized so it shouldn't break

vaibhav-datazip:
> stream id is still using dot image

we are never using it for splitting operations

types/stream.go
vikaxsh: there would be confusion in understanding what is table name if we print like  "public.my.table"

vikaxsh:
> > it would break here image
> > image
> 
> the names in paths are already normalized so it shouldn't break

We should keep it consistent. If we're avoiding dot-separated keys and using a struct key, let's use that everywhere so it's not confusing later.

vikaxsh (approved):

PR #1018, opened 2026-07-03
chore(s3-driver): abstract out aws to generic s3 object store

mihir-datazip:
# Description

Abstracts the S3 source driver's direct dependency on the AWS SDK behind a new storage-agnostic `pkg/objstorage` package, so drivers depend on a small `Store` interface instead of a vendor SDK. This is groundwork for generalizing file-based drivers to other object stores (GCS, Azure Blob) and makes the driver unit-testable without a live bucket.

Ticket: OL-2945

**What changed:**

- **New `pkg/objstorage` package:**
  - `Store` interface — minimal read-oriented API over a single bucket (`Check`, `List`, `Open`, `OpenRange`), safe for concurrent use. Write ops can be added when destinations migrate onto it.
  - `NewS3Store` — implementation backed by `aws-sdk-go-v2`, supporting both AWS S3 and S3-compatible endpoints (MinIO, GCS interop, R2) with static credentials or the default credential chain. All AWS-specific logic (path-style addressing, pagination, provider error shapes) now lives here.
  - `ErrNotFound` sentinel — implementations wrap missing-key errors so callers match with `errors.Is` instead of string-matching provider errors (`"NoSuchKey"`/`"NotFound"`).
  - `ReaderAt` — generic range-request-backed `io.ReaderAt` for Parquet streaming; replaces the driver-local `S3RangeReader` (`range_reader.go` deleted).
  - `MemStore` — in-memory `Store` with error injection, for unit tests.
- **S3 driver migration:** `drivers/s3` now holds an `objstorage.Store` instead of `*s3.Client` and has zero AWS SDK imports (moved to indirect in `drivers/s3/go.mod`). Discovery, schema inference, and file processing go through the interface; behavior is unchanged.
- **Bug fix (gzip reader leak):** closing a gzip-decompressed reader previously closed only the gzip layer — `gzip.Reader.Close` does not close the underlying object body, leaking the HTTP connection. New `utils.NewCompositeReadCloser` closes both.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) — gzip reader left the underlying S3 body unclosed
- [x] Code refactor (non-breaking; no user-facing behavior or config change)

# How Has This Been Tested?

All new code paths are covered by unit tests (no live bucket needed — driver tests run against `MemStore`):

- [x] `go test ./pkg/objstorage/...` — MemStore round-trip/list/range/error-injection/concurrency, ReaderAt offset/EOF/error/concurrency behavior, S3Store endpoint-vs-AWS config modes, not-found wrapping
- [x] `cd drivers/s3 && go test ./internal/...` — stream discovery, file-pattern filtering, list-error propagation, cursor-field schema; `processFile` end-to-end for CSV, gzip CSV, JSONL, Parquet (streaming + in-memory fallback), missing-key skip, open-error propagation
- [x] `go test ./utils/...` — composite ReadCloser close ordering and first-error propagation

# Screenshots or Recordings

N/A

## Documentation

- [x] N/A (refactor and test changes only; no user-facing config or behavior change)

## Related PR's (If Any):


pkg/objstorage/memstore.go
```diff
@@ -0,0 +1,133 @@
+package objstorage
```
vikaxsh: since memstore is strictly a mock for testing, we can move it to sub package, either here or in utils

pkg/objstorage/reader_at.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/datazip-inc/olake/utils/logger"
+)
+
+// ReaderAt adapts ranged Store reads into io.ReaderAt (as required by the
+// parquet parser via parser.NewParquetReaderWrapper). It is stateless per
+// call - every ReadAt issues one OpenRange request - and therefore safe for
+// concurrent use. The context is captured at construction because io.ReaderAt
+// cannot thread a context through ReadAt; canceling it aborts subsequent reads.
+type ReaderAt struct {
+	ctx   context.Context
```
vikaxsh, line 18: context should be part of fn argumets

pkg/objstorage/memstore.go
vikaxsh: IMO, instead of building a full in-memory store, we could define a narrow s3api interface covering just the SDK calls s3Store uses (HeadBucket, GetObject, ListObjectsV2), and mock that directly. This gives us more control in tests

pkg/objstorage/s3.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+			if err := fn(info); err != nil {
+				return err
+			}
+		}
+
+		if !aws.ToBool(result.IsTruncated) {
+			logger.Debugf("Completed S3 listing: processed %d pages", pageCount)
+			return nil
+		}
```
vikaxsh, line 133: can add ctx cancellation check

pkg/objstorage/objstorage.go
```diff
@@ -0,0 +1,45 @@
...
+	// object. Pagination is handled internally; keys arrive in provider
+	// listing order (lexicographic for S3). An error returned by fn aborts
+	// the walk and is returned as-is.
+	List(ctx context.Context, prefix string, fn func(ObjectInfo) error) error
+	// Open returns the full object body. The caller must close the reader.
+	Open(ctx context.Context, key string) (io.ReadCloser, error)
+	// OpenRange returns object bytes [offset, offset+length) with length > 0.
+	// A range extending past the object end is truncated. The caller must
+	// close the reader.
+	OpenRange(ctx context.Context, key string, offset, length int64) (io.ReadCloser, error)
```
vikaxsh, line 51: Let's  move OpenRange into a separate interface. Since it's only used for Parquet parsing, the core Store interface shouldn't require it.

mihir-datazip: Done

pkg/objstorage/reader_at.go
mihir-datazip: ReaderAt.ReadAt must satisfy io.ReaderAt, whose signature is fixed as ReadAt(p []byte, off int64) — it cannot carry a ctx

pkg/objstorage/s3.go
mihir-datazip: Done

pkg/objstorage/memstore.go
mihir-datazip: Done

vikaxsh:
Also, could you take a look at this library and see if it would be useful for this PR?
https://github.com/google/go-cloud

mihir-datazip:
> Also, could you take a look at this library and see if it would be useful for this PR? https://github.com/google/go-cloud

@vikaxsh lets not add it for now, keeping our codebase light weight is also important. We can definitely add it once the requirement comes.

pkg/objstorage/s3.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		return nil, wrapNotFound(err)
+	}
+	return result.Body, nil
+}
+
+// OpenRange returns object bytes [offset, offset+length) using an S3 ranged GET.
+func (s *s3Store) OpenRange(ctx context.Context, key string, offset, length int64) (io.ReadCloser, error) {
+	// S3 range format: "bytes=start-end" (inclusive)
+	rangeHeader := fmt.Sprintf("bytes=%d-%d", offset, offset+length-1)
```
vikaxsh, line 156: do we need to add length check here?

vikaxsh (commented):
Mostly lgtm, before merge need to test (CSV, Json, Paruqet) includes gzipped files, with fullload+incremental mode

mihir-datazip:
> Mostly lgtm, before merge need to test (CSV, Json, Paruqet) includes gzipped files, with fullload+incremental mode

We can merge this after https://github.com/datazip-inc/olake/pull/1019 so we can be confident

pkg/objstorage/s3.go
mihir-datazip: Added

drivers/s3/internal/s3.go
```diff
@@ -148,72 +113,50 @@ func (s *S3) GetStreamNames(ctx context.Context) ([]types.StreamID, error) {
...
 		}
 
-		// Check if there are more results
-		if !aws.ToBool(result.IsTruncated) {
-			logger.Infof("Completed S3 discovery: processed %d pages, discovered %d files", pageCount, totalDiscovered)
-			break
+		fileObj := FileObject{
+			FileKey:      obj.Key,
+			Size:         obj.Size,
+			LastModified: obj.LastModified.UTC().Format("2006-01-02T15:04:05Z"),
```
vikaxsh, line 144: why utc?

vikaxsh (commented):
LGTM. Before merging, let's make sure we test the following:
1. Parquet files with `streaming_enabled=true` and `streaming_enabled=false`.
2. `.csv.gz` files for both full load and incremental syncs.

drivers/s3/internal/s3.go
mihir-datazip: We do expect cross-timezone cases — that's what the .UTC() is for. Every string in the cursor domain (discovered files and persisted state) is produced by this one line and nothing ever parses them back, so normalizing here guarantees all comparisons happen in one zone no matter the host TZ or what location the provider's SDK returns. Without it, e.g. a sync on an IST host with a local-located time.Time would persist a cursor 5.5h ahead, and the next sync would silently skip every file uploaded in that window. For real S3 the SDK already returns UTC so it's a no-op — but the generic store interface doesn't guarantee that, so we normalize instead of assuming.

mihir-datazip:
> LGTM. Before merging, let's make sure we test the following:
> 
> 1. Parquet files with `streaming_enabled=true` and `streaming_enabled=false`.
> 2. `.csv.gz` files for both full load and incremental syncs.

1. Added Integration Tests for streaming_enabled=true and false.
2. We already had covered the .gz formats in the ITs

.github/workflows/build-and-release-driver.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set up JDK 17
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
```
vikaxsh, line 40: why?

.github/workflows/integration-tests.yml
```diff
@@ -51,6 +56,24 @@ jobs:
...
+            tests/go.work
+            Makefile
+            drivers/*/driver.mk
+            .golangci.yml
+            .github/workflows/integration-tests.yml
+        run: |
+          changed=$(.github/scripts/changed-in-paths.sh <<<"$GO_CHECK_PATHS")
+          echo "changed=$changed" >> "$GITHUB_OUTPUT"
+          echo "Go-check inputs changed: $changed"
+
```
vikaxsh, line 76: This step exists only to force go-cache to re-run when its cache is warm, because make test.lint and gosec are bolted onto that job. That means test-module linting is gated on a cache state, which is going to be confusing the first time someone asks "why didn't lint run on my PR."

Two separate issues:

1. Three of the seven patterns are dead. tests/**/*.go, tests/**/go.mod, and tests/go.work are exactly the hashFiles() inputs of the go-caches key. Change any of them and go-cached != 'true' already fires the first clause of the if at :163. Only Makefile, drivers/*/driver.mk, .golangci.yml, and this workflow file actually add anything.

2. Better: delete the step. Move test.lint into golang-ci.yml next to olake.lint, and the tests-module gosec into security-ci.yaml next to the root gosec that just grew -exclude-dir=tests. That removes this step, the go-checks-changed output, GO_CHECK_PATHS, and changed-in-paths.sh's only caller — and leaves go-cache as a pure cache warm gated on go-cached != 'true' alone.

mihir-datazip: You need to understand the use case on why they are different.

1. Most of the time people will add changes to codebase only and tests/ is a safety net so we can assume less number of changes in the test cases itself.
2. Due to this we also need to check lint separately for tests. Combining them as said in point 2 will trigger it almost in each PR which is not needed.
3. The cache and lint are inter related in tests. We have faster lint in case of gocache already warm so the point that `This step exists only to force go-cache to re-run` is missing the faster lint we get by combining the cache and lint steps.
4. `Only Makefile, drivers/*/driver.mk, .golangci.yml, and this workflow file actually add anything` -> and that is exactly what we need, in case of updating linter rules or adding new steps in go related checks in tests

.github/workflows/build-and-release-driver.yml
mihir-datazip: setup-java@v4 was added in https://github.com/datazip-inc/olake/pull/1038
This causes 2 maven caches in v3 as well as in v4 which takes 0.4GB extra space in our git cache which has total limit of 10GB.

vikaxsh (approved):

PR #1020, opened 2026-07-06
fix: s3 driver parquet parser data types

mihir-datazip:
# Description

The S3 driver's parquet parser produced destination schemas and values inconsistent with what the database drivers emit for equivalent data. This PR fixes the conversion of the parquet source datatypes in scope so the S3 parser lines up with the DB drivers:

- **Timestamp (INT64, millis/micros unit)** — values were converted through `time.Unix(0, raw*1000)`, which materializes nanoseconds in an int64. Any instant outside ~1678–2262 overflowed and wrapped to garbage (e.g. `9999-12-31` in micros became year **1816**, `1600-01-01` became **2184**). Now converted via `time.UnixMilli`/`time.UnixMicro`, which are overflow-safe.
- **Timestamp / Date sub-second truncation** — values were formatted with `time.Format(time.RFC3339)`, which has no fractional seconds, so `.123`/`.123456` components were silently dropped. Temporal values are now returned as `time.Time` (like DB drivers), so full precision reaches the destination. Date (INT32 days) is likewise returned as `time.Time` instead of an RFC3339 string.
- **INT32 / FLOAT widening** — `int32` values were widened to `int64` and `float32` to `float64`, causing the destination to promote columns to `long`/`double` on the first batch (and storing float noise like `3.140000104904175` instead of `3.14`). Values now keep their native width, matching the inferred `Int32`/`Float32` schema.
- **Time (INT32/INT64, millis/micros/nanos unit)** — was truncated to whole seconds-of-day (`13:45:30.123` → `49530`), silently discarding sub-second precision. Now the raw stored value is returned as `Int64` with precision conserved (`13:45:30.123` in millis → `49530123`), so no information is lost. Destination type stays `Int64`/`long`, but the stored value changes from the old truncated seconds — streams synced with the old behavior carry truncated values.
- **Unsigned 32-bit integers (INT32 physical)** — read as a signed `int32`, values above 2³¹−1 wrapped negative. Now widened through `uint32 → int64`, and the schema maps unsigned-32 columns to `Int64` (matches pg/mysql). uint8/uint16 stay `Int32`; uint64 stays `Int64`.
- **UUID (16-byte)** — previously took the generic byte-array path and, being valid UTF-8, surfaced as control-character garbage. Now formatted as a canonical `8-4-4-4-12` hex string, matching how DB drivers emit uuid.
- **INT96 legacy timestamp** — `val.String()` emitted the raw 96-bit integer, which disagreed with the inferred `Timestamp` schema and collapsed the column to `String`. Now converted to `time.Time` from the Julian day (high 32 bits) + nanoseconds-of-day (low 64 bits).
- **Decimal (INT32/INT64/BYTE_ARRAY/FIXED_LEN_BYTE_ARRAY)** — conversion was already correct (unscaled + scale → float64, incl. two's-complement negatives); added missing FIXED_LEN_BYTE_ARRAY test coverage.

Note: this changes the destination **type** for parquet unsigned-32 columns (`Int32`→`Int64`) and INT96 columns (`String`→`Timestamp`), and changes stored **values** for TIME columns; streams synced with the old behavior should be re-discovered.

Fixes OL-2946

## Type of change

# How Has This Been Tested?

- [x] Unit tests (`go test ./pkg/parser/`):
  - `TestMapParquetTypeToOlake_LogicalTypes` / `_PhysicalTypes` — build real parquet schemas and assert the mapping for date, timestamp (all units), time (→ `Int64`), decimal, uuid (→ `String`), and signed/unsigned integers (unsigned-32 → `Int64`).
  - Timestamp/date value tests assert exact `time.Time` results (previously compared with 1-second tolerance, which masked the truncation), plus regression cases for `9999-12-31` in **both** micros and millis, and pre-epoch (negative) values.
  - TIME tests: INT32 and INT64 storage, millis/micros/nanos units, whole-second values, and the `isAdjustedToUTC=false` variant — all asserting the raw `int64` value with precision conserved.
  - INT96 test: epoch, a 2024 instant with nanosecond precision, and `9999-12-31`, all as `time.Time`.
  - UUID test: canonical `8-4-4-4-12` hex formatting of a 16-byte value.
  - Unsigned-integer test: uint8/uint16 stay `int32`; uint32 widens to `int64` (incl. the 2³¹ and 2³²−1 wrap cases); uint64 stays `int64`.
  - Decimal tests extended with FIXED_LEN_BYTE_ARRAY(16), including negative two's complement.
  - `TestParquetParser_Integration` — writes a real parquet file across the full matrix (int32/int64/uint32/uint64/float/double/decimal/date/timestamp ms-µs-ns/time ms-µs-ns) and verifies `InferSchema` + `StreamRecords` end to end, including exact preservation of `9007199254740993` (2⁵³+1) and `2147483648` (2³¹, unsigned-32 not wrapping negative).
- [x] End-to-end A/B sync: a seed parquet file covering the full datatype matrix (decimal backed by INT32/INT64/FIXED, DATE, TIMESTAMP in ms/µs/ns with extreme values 9999-12-31 / 1600-01-01 / 2262-04-12, TIME in ms/µs) was uploaded to S3 and synced with the old and the fixed binary. Outputs were diffed programmatically cell by cell — the fixed output matches the source (and the Postgres driver's output for shared columns) exactly; see table below.

# Screenshots or Recordings
Seed file: https://drive.google.com/file/d/1zsPdpwBzjeWxqeYxkaytgPklBpXM5l1b/view?usp=sharing
Old Parquet: https://drive.google.com/file/d/1HdJJWGpXES_8LSWy67YmRVLuy4bGS2sT/view?usp=share_link
New Parquet: https://drive.google.com/file/d/1icgMj8ff3B4cFvHrFdHOyeuMYwiECO9z/view?usp=share_link

Before/after from the A/B sync of the same source file:

| column | old parser output | fixed parser output |
|---|---|---|
| INT32 columns | INT64 (promoted) | INT32, values exact |
| FLOAT column (`3.14`) | DOUBLE `3.140000104904175` | FLOAT `3.14` |
| TIMESTAMP millis `.123` | `.000` (fraction lost) | `.123` |
| TIMESTAMP micros `.123456` | `.000000` | `.123456` |
| TIMESTAMP nanos `.123456789` | whole seconds | `.123456` (destination stores micros) |
| TIMESTAMP micros `9999-12-31` / `1600-01-01` / `2262-04-12` | years **1816 / 2184 / 1677** (int64 nanosecond overflow) | exact |
| TIME millis `13:45:30.123` | INT64 `49530` (truncated) | INT64 `49530123` (precision kept) |
| TIME micros `13:45:30.123456` | INT64 `49530` | INT64 `49530123456` |

## Documentation


pkg/parser/parquet.go
```diff
@@ -192,10 +192,10 @@ func mapParquetTypeToOlake(pqType pq.Type) types.DataType {
 			return types.Timestamp
 		}
 
-		// Time with precision (stored as INT32 or INT64)
-		// We convert to seconds, so map to Int64
+		// Time-of-day with precision (stored as INT32 or INT64), formatted as
+		// "15:04:05[.fraction]" to match how MySQL/Postgres emit TIME columns
 		if logicalType.Time != nil {
-			return types.Int64
+			return types.String
```
vikaxsh: why are we converting to string?

vikaxsh: won't be backward compatible

pkg/parser/parquet.go
```diff
@@ -258,33 +258,38 @@ func parquetValueToInterfaceWithType(val pq.Value, fieldType pq.Type) interface{
 
 	logicalType := fieldType.LogicalType()
 
-	// Handle temporal types with logical type annotations
+	// Handle temporal types with logical type annotations.
+	// Temporal values are returned as time.Time (not formatted strings) so
```
vikaxsh: need to edit this comment, temporal is not used anywhere here

vikaxsh (commented):
we need to test this pr with all possible datatyes, we can take reference from here https://github.com/apache/parquet-testing

in my testing for one datatype found

destination(datatype: VARCHAR)
Image

<br/>

source (datatype: TIME)
Image

pkg/parser/parquet.go
mihir-datazip: It is changed to String because entire point of OL-2946 is that the s3 parquet parser should produce the same olake/destination type as the DB drivers for an equivalent column. 
The old s3 behavior — types.Int64 = seconds-of-day — was both inconsistent with those drivers and lossy: the old value converter did seconds = rawValue / 1e9, so 13:45:30.123456 collapsed to the integer 49530, dropping the fraction and any human meaning.

TIME is the only column in this PR whose destination type changes — bigint → string. Iceberg (and the parquet destination's schema-merge) cannot evolve long → string — it's not an allowed transition. So any existing destination table that already materialized a TIME column as bigint will hard-fail on the next sync with a type conflict.

What we should do here?

pkg/parser/parquet.go
mihir-datazip: Done

pkg/parser/parquet.go
mihir-datazip: I reverted to the timestamp being int64 as discussed

pkg/parser/csv.go
```diff
@@ -209,6 +201,17 @@ func (p *CSVParser) StreamRecords(ctx context.Context, reader io.Reader, callbac
 	return nil
 }
 
+// fieldType resolves a header's type from the stream schema. A header the schema does not
+// carry is a column that appeared after discover: hand its cells through as strings, the
+// way the JSON parser hands through unknown fields, so the destination can evolve.
+func (p *CSVParser) fieldType(header string) types.DataType {
+	fieldType, err := p.stream.Schema.GetType(header)
+	if err != nil {
```
vikaxsh, line 209: why are we avoiding error?

vikaxsh: we cannot convert column datatype to string if it missed during sampling

pkg/parser/parquet.go
```diff
@@ -96,78 +120,291 @@ func (p *ParquetParser) StreamRecords(ctx context.Context, reader io.Reader, cal
...
+			continue
+		}
+		record[field.Name()] = parquetValueToInterfaceWithType(value, field.Type())
+	}
+	return record, nil
+}
+
+// leafValue finds the value a non-repeated leaf column contributed to a row. Row values are
+// ordered by column but the index is not the column number once groups are present.
+func leafValue(row pq.Row, column int) (pq.Value, bool) {
```
vikaxsh: called once per field per row would be O(fields x row_len) per row, ->  O(n²) , can we optimize this

constants/state_version.go
```diff
@@ -38,7 +38,7 @@ package constants
 //     * Now these byte slices are parsed and converted into int64
 
 const (
-	LatestStateVersion = 6
+	LatestStateVersion = 7
```
vikaxsh: aren't we using this in the pr?

pkg/parser/parquet.go
```diff
@@ -337,8 +599,10 @@ func parquetValueToInterfaceWithType(val pq.Value, fieldType pq.Type) interface{
 		}
 		return base64.StdEncoding.EncodeToString(byteData)
 	case pq.Int96:
-		// Int96 is typically used for timestamps in legacy Parquet files
-		return val.String()
+		// Int96 is a legacy timestamp, and the schema maps it to Timestamp, so return a
+		// time.Time. String() emitted the raw 96 bit integer in decimal, which disagreed
+		// with the schema and collapsed the column to a string.
+		return int96ToTime(val.Int96())
```
vikaxsh: won't be backward compatible

pkg/parser/parquet.go
```diff
@@ -12,9 +12,20 @@ import (
...
 )
 
+const (
+	// julianDayOfUnixEpoch is the Julian day number for 1970-01-01, the origin Int96
+	// timestamps are offset from.
+	julianDayOfUnixEpoch = 2440588
+
+	// rowReadBatchSize is how many rows are read from a row group at a time, bounding how
+	// much of a row group is held in memory while still amortizing reads over the network.
+	rowReadBatchSize = 256
```
vikaxsh, line 28: reducing it might impact rps, please once test

pkg/parser/parquet.go
```diff
@@ -12,9 +12,20 @@ import (
 	"github.com/datazip-inc/olake/types"
 	"github.com/datazip-inc/olake/utils/logger"
 	pq "github.com/parquet-go/parquet-go"
+	"github.com/parquet-go/parquet-go/deprecated"
 	"github.com/shopspring/decimal"
 )
```
vikaxsh, line 20: This Parquet file could be structured more cleanly
like
-  reorder funcs top-down: entry points -> row streaming ->     row decode -> reconstruction ->chema mapping -> value conversion -> io adapter.
- Bundle schema/fields/leafOffsets/hasGroupField into one rowDecoder struct with decode(row) method — replaces buildRecord + param threading.
- Split convertReconstructed's 4 inline branches into convertLeaf/convertListNode/convertMapNode/convertStructNode, dispatched from a slim top func.
- how can we avoid nested if's in mapParquetTypeToOlake/parquetValueToInterfaceWithType

vikaxsh (commented):
please merge integration test pr into this pr

pkg/parser/csv.go
mihir-datazip: Fixed

pkg/parser/parquet.go
mihir-datazip: Done

constants/state_version.go
mihir-datazip: Added checks

pkg/parser/parquet.go
mihir-datazip: Added check for backwards compatibility

pkg/parser/parquet.go
mihir-datazip: Done

pkg/parser/parquet.go
mihir-datazip: Ran banchmark using the following:
```
package parser

import (
	"bytes"
	"context"
	"fmt"
	"runtime"
	"testing"
	"time"

	pq "github.com/parquet-go/parquet-go"
)

// Benchmarks comparing the two row group readers at realistic scale: a wide flat schema
// (20 columns, 2 nullable) spread over several row groups, and a nested schema that only
// the row-at-a-time reader can handle. Batch sizes climb past a row group's row count, so
// "wholegroup" reads a whole row group in one ReadRows call.
//
//	go test ./pkg/parser/ -run=^$ -bench='BenchmarkBigFlat|BenchmarkBigNested' -benchtime=5x -count=3 -benchmem
//	go test ./pkg/parser/ -run=^$ -bench=BenchmarkPeakHeap -benchtime=1x -count=3

type benchFlatRow struct {
	ID       int64     `parquet:"id"`
	UserID   int64     `parquet:"user_id"`
	Seq      int32     `parquet:"seq"`
	Bucket   int32     `parquet:"bucket"`
	Amount   float64   `parquet:"amount"`
	Rate     float64   `parquet:"rate"`
	Score    float32   `parquet:"score"`
	Name     string    `parquet:"name"`
	Category string    `parquet:"category"`
	Country  string    `parquet:"country"`
	Payload  string    `parquet:"payload"`
	Flag     bool      `parquet:"flag"`
	Active   bool      `parquet:"active"`
	TS       time.Time `parquet:"ts,timestamp(microsecond)"`
	Created  time.Time `parquet:"created,timestamp(millisecond)"`
	Opt      *int64    `parquet:"opt,optional"`
	OptStr   *string   `parquet:"opt_str,optional"`
	Extra1   int64     `parquet:"extra1"`
	Extra2   int64     `parquet:"extra2"`
	Extra3   string    `parquet:"extra3"`
}

type benchInner struct {
	A int64  `parquet:"a"`
	B string `parquet:"b"`
}

type benchNestedRow struct {
	ID    int64             `parquet:"id"`
	Name  string            `parquet:"name"`
	TS    time.Time         `parquet:"ts,timestamp(microsecond)"`
	Tags  []string          `parquet:"tags"`
	Attrs map[string]string `parquet:"attrs"`
	Inner benchInner        `parquet:"inner"`
}

func makeFlatFile(b *testing.B, numRows, rowsPerGroup int, codec pq.WriterOption) []byte {
	b.Helper()
	rows := make([]benchFlatRow, numRows)
	base := time.Date(2024, 1, 1, 0, 0, 0, 0, time.UTC)
	countries := []string{"IN", "US", "DE", "JP", "BR"}
	for i := range rows {
		v := int64(i)
		s := fmt.Sprintf("opt-%d", i%97)
		rows[i] = benchFlatRow{
			ID: v, UserID: v * 7, Seq: int32(i % 100000), Bucket: int32(i % 512),
			Amount: float64(i) * 1.5, Rate: float64(i%1000) / 3.0, Score: float32(i%50) * 0.5,
			Name: fmt.Sprintf("record-%d", i%1000), Category: "category-value",
			Country: countries[i%len(countries)], Payload: "payload-payload-payload-payload",
			Flag: i%2 == 0, Active: i%3 == 0,
			TS:      base.Add(time.Duration(i) * time.Second),
			Created: base.Add(time.Duration(i) * time.Millisecond),
			Extra1:  v * 3, Extra2: v * 11, Extra3: "extra-extra-extra",
		}
		if i%4 != 0 {
			rows[i].Opt = &v
			rows[i].OptStr = &s
		}
	}

	var buf bytes.Buffer
	opts := []pq.WriterOption{pq.MaxRowsPerRowGroup(int64(rowsPerGroup))}
	if codec != nil {
		opts = append(opts, codec)
	}
	w := pq.NewGenericWriter[benchFlatRow](&buf, opts...)
	if _, err := w.Write(rows); err != nil {
		b.Fatal(err)
	}
	if err := w.Close(); err != nil {
		b.Fatal(err)
	}
	return buf.Bytes()
}

func makeNestedFile(b *testing.B, numRows, rowsPerGroup int) []byte {
	b.Helper()
	rows := make([]benchNestedRow, numRows)
	base := time.Date(2024, 1, 1, 0, 0, 0, 0, time.UTC)
	for i := range rows {
		tags := make([]string, i%4)
		for j := range tags {
			tags[j] = fmt.Sprintf("tag-%d", j)
		}
		rows[i] = benchNestedRow{
			ID: int64(i), Name: fmt.Sprintf("record-%d", i%1000),
			TS:   base.Add(time.Duration(i) * time.Second),
			Tags: tags,
			Attrs: map[string]string{
				"k1": "v1",
				"k2": fmt.Sprintf("v-%d", i%50),
			},
			Inner: benchInner{A: int64(i) * 2, B: "inner-value"},
		}
	}

	var buf bytes.Buffer
	w := pq.NewGenericWriter[benchNestedRow](&buf, pq.MaxRowsPerRowGroup(int64(rowsPerGroup)))
	if _, err := w.Write(rows); err != nil {
		b.Fatal(err)
	}
	if err := w.Close(); err != nil {
		b.Fatal(err)
	}
	return buf.Bytes()
}

func benchDiscard(_ context.Context, _ map[string]any) error { return nil }

// runColumnar drives the flat column-major path over every row group of the file.
func runColumnar(b *testing.B, parser *ParquetParser, fileBytes []byte) {
	pqFile, err := pq.OpenFile(bytes.NewReader(fileBytes), int64(len(fileBytes)))
	if err != nil {
		b.Fatal(err)
	}
	decoder := newRowDecoder(pqFile.Schema())
	count := 0
	for _, rg := range pqFile.RowGroups() {
		if err := parser.streamRowGroupColumns(context.Background(), rg, decoder, benchDiscard, &count); err != nil {
			b.Fatal(err)
		}
	}
}

// runRows drives the row-at-a-time path at a given batch size. batchSize <= 0 means
// "size the buffer to the whole row group", the closest equivalent to an unbounded read.
func runRows(b *testing.B, parser *ParquetParser, fileBytes []byte, batchSize int) {
	pqFile, err := pq.OpenFile(bytes.NewReader(fileBytes), int64(len(fileBytes)))
	if err != nil {
		b.Fatal(err)
	}
	decoder := newRowDecoder(pqFile.Schema())
	count := 0
	for _, rg := range pqFile.RowGroups() {
		size := batchSize
		if size <= 0 {
			size = int(rg.NumRows())
		}
		if err := parser.streamRowGroupRows(context.Background(), rg, decoder, size, benchDiscard, &count); err != nil {
			b.Fatal(err)
		}
	}
}

// peakHeap samples HeapAlloc while fn runs and reports the maximum. Used in a separate
// pass from the timing runs so the stop-the-world sampling does not skew throughput.
func peakHeap(b *testing.B, fn func()) {
	stop := make(chan struct{})
	done := make(chan uint64, 1)
	go func() {
		var peak uint64
		var ms runtime.MemStats
		ticker := time.NewTicker(2 * time.Millisecond)
		defer ticker.Stop()
		for {
			select {
			case <-stop:
				done <- peak
				return
			case <-ticker.C:
				runtime.ReadMemStats(&ms)
				if ms.HeapAlloc > peak {
					peak = ms.HeapAlloc
				}
			}
		}
	}()
	fn()
	close(stop)
	b.ReportMetric(float64(<-done)/(1<<20), "peakHeapMB")
}

// BenchmarkBigFlat compares the two readers on a wide flat schema (20 columns, 2 of them
// nullable) split across several row groups, uncompressed and snappy.
func BenchmarkBigFlat(b *testing.B) {
	const numRows = 500000
	const rowsPerGroup = 100000

	for _, c := range []struct {
		name  string
		codec pq.WriterOption
	}{
		{"uncompressed", nil},
		{"snappy", pq.Compression(&pq.Snappy)},
	} {
		fileBytes := makeFlatFile(b, numRows, rowsPerGroup, c.codec)
		parser := NewParquetParser(ParquetConfig{}, types.NewStream("bench", "bench", nil))
		b.Logf("%s: %d rows, %d bytes, %.1f MB", c.name, numRows, len(fileBytes), float64(len(fileBytes))/(1<<20))

		b.Run(c.name+"/columnar", func(b *testing.B) {
			b.ReportAllocs()
			for n := 0; n < b.N; n++ {
				runColumnar(b, parser, fileBytes)
			}
			b.ReportMetric(float64(numRows)*float64(b.N)/b.Elapsed().Seconds(), "rows/s")
		})

		for _, batch := range []int{256, 1024, 4096, -1} {
			label := fmt.Sprintf("%s/batch=%d", c.name, batch)
			if batch < 0 {
				label = c.name + "/batch=wholegroup"
			}
			b.Run(label, func(b *testing.B) {
				b.ReportAllocs()
				for n := 0; n < b.N; n++ {
					runRows(b, parser, fileBytes, batch)
				}
				b.ReportMetric(float64(numRows)*float64(b.N)/b.Elapsed().Seconds(), "rows/s")
			})
		}
	}
}

// BenchmarkBigNested exercises the row path only: lists, maps and a nested struct cannot
// be read by the column-major path today.
func BenchmarkBigNested(b *testing.B) {
	const numRows = 200000
	const rowsPerGroup = 100000

	fileBytes := makeNestedFile(b, numRows, rowsPerGroup)
	parser := NewParquetParser(ParquetConfig{}, types.NewStream("bench", "bench", nil))
	b.Logf("nested: %d rows, %d bytes, %.1f MB", numRows, len(fileBytes), float64(len(fileBytes))/(1<<20))

	for _, batch := range []int{256, 1024, 4096, -1} {
		label := fmt.Sprintf("batch=%d", batch)
		if batch < 0 {
			label = "batch=wholegroup"
		}
		b.Run(label, func(b *testing.B) {
			b.ReportAllocs()
			for n := 0; n < b.N; n++ {
				runRows(b, parser, fileBytes, batch)
			}
			b.ReportMetric(float64(numRows)*float64(b.N)/b.Elapsed().Seconds(), "rows/s")
		})
	}
}

// BenchmarkPeakHeap reports peak heap for one pass of each path. Run with -benchtime=1x.
func BenchmarkPeakHeap(b *testing.B) {
	const numRows = 500000
	const rowsPerGroup = 100000

	fileBytes := makeFlatFile(b, numRows, rowsPerGroup, nil)
	parser := NewParquetParser(ParquetConfig{}, types.NewStream("bench", "bench", nil))
	b.Logf("file held on heap: %.1f MB (included in every peak below)", float64(len(fileBytes))/(1<<20))

	b.Run("columnar", func(b *testing.B) {
		peakHeap(b, func() {
			for n := 0; n < b.N; n++ {
				runColumnar(b, parser, fileBytes)
			}
		})
	})
	for _, batch := range []int{256, -1} {
		label := fmt.Sprintf("batch=%d", batch)
		if batch < 0 {
			label = "batch=wholegroup"
		}
		b.Run(label, func(b *testing.B) {
			peakHeap(b, func() {
				for n := 0; n < b.N; n++ {
					runRows(b, parser, fileBytes, batch)
				}
			})
		})
	}
}

```

## Outcome:
```
500k rows × 20 columns (int64/int32/float64/float32/string/bool/timestamp + 2 nullable), 5 row groups of 100k, uncompressed and snappy. Nested case: 200k rows with list + map + nested struct. Apple M4, decode-only from an in-memory reader. Median of 3 × 5 iterations.

Flat schema — uncompressed

┌──────────────────────────────┬────────┬────────┬─────────┬───────────┐
│             path             │ ns/op  │ rows/s │  B/op   │ allocs/op │
├──────────────────────────────┼────────┼────────┼─────────┼───────────┤
│ columnar (prod path today)   │ 456 ms │ 1,096k │ 2.00 GB │ 17.75M    │
├──────────────────────────────┼────────┼────────┼─────────┼───────────┤
│ batch=256 (current constant) │ 607 ms │ 824k   │ 1.47 GB │ 18.26M    │
├──────────────────────────────┼────────┼────────┼─────────┼───────────┤
│ batch=1024                   │ 631 ms │ 793k   │ 1.48 GB │ 18.31M    │
├──────────────────────────────┼────────┼────────┼─────────┼───────────┤
│ batch=4096                   │ 643 ms │ 777k   │ 1.58 GB │ 18.61M    │
├──────────────────────────────┼────────┼────────┼─────────┼───────────┤
│ batch=wholegroup (100k)      │ 774 ms │ 646k   │ 2.29 GB │ 21.24M    │
└──────────────────────────────┴────────┴────────┴─────────┴───────────┘

Snappy tracks it exactly: columnar 1,010k · 256 → 799k · 1024 → 795k · 4096 → 766k · wholegroup → 606k rows/s.

Nested schema (list + map + struct, row path only)

┌────────────┬────────┬────────┐
│   batch    │ ns/op  │ rows/s │
├────────────┼────────┼────────┤
│ 256        │ 353 ms │ 566k   │
├────────────┼────────┼────────┤
│ 1024       │ 351 ms │ 570k   │
├────────────┼────────┼────────┤
│ 4096       │ 354 ms │ 563k   │
├────────────┼────────┼────────┤
│ wholegroup │ 387 ms │ 516k   │
└────────────┴────────┴────────┘

Peak heap (500k flat rows, one pass)

columnar 371 MB · batch=256 267 MB · batch=wholegroup 448 MB
```

tests/s3/s3_test.go
```diff
@@ -0,0 +1,38 @@
+package s3
```
vikaxsh, line 1: add s3 in tests/go.work,  S3 absent from SkipCDCDrivers

tests/s3/s3_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/datazip-inc/olake/tests/testutils"
+)
+
+func TestS3Integration(t *testing.T) {
+	t.Parallel()
+
+	t.Run("Variants", func(t *testing.T) {
+		for _, variant := range S3TestVariants {
+			t.Run(variant.Name, func(t *testing.T) {
+				t.Parallel()
```
vikaxsh: check this comment from kafka can we get same error here?

Image

tests/s3/s3_test.go
```diff
@@ -0,0 +1,38 @@
+package s3
+
+import (
+	"testing"
+
+	"github.com/datazip-inc/olake/lib/constants"
```
vikaxsh: use tests/testutils/constants

tests/s3/s3_test.go
```diff
@@ -0,0 +1,38 @@
...
+					// has not seen (see S3TestVariant.BuildEvolvedFile), so the update sync
+					// must land it in the destination as a string column.
+					UpdatedDestinationDataTypeSchema: variant.UpdatedDestinationSchema,
+					ExecuteQuery:                     ExecuteQueryFactory(variant),
+					DestinationDB:                    S3DestinationDB,
+					CursorField:                      S3CursorField,
+					PartitionRegex:                   S3PartitionRegex,
+					FilterConfig:                     S3FilterConfig,
+				}
+				cfg.TestIntegration(t)
```
vikaxsh, line 36: column selection tests are missing

tests/testutils/test_utils.go
```diff
@@ -250,7 +250,7 @@ func updateSelectedStreamsCommand(config TestConfig, namespace, partitionRegex,
 	}
 	streamConditions := make([]string, len(stream))
 	for i, s := range stream {
-		s = Ternary(slices.Contains(constants.SkipCDCDrivers, constants.DriverType(config.Driver)), strings.ToUpper(s), s).(string)
+		s = Ternary(slices.Contains(constants.UppercaseStreamDrivers, constants.DriverType(config.Driver)), strings.ToUpper(s), s).(string)
```
vikaxsh, line 253: what are UppercaseStreamDrivers?

tests/s3/s3_util_test.go
```diff
@@ -0,0 +1,1073 @@
...
+const (
+	// writerLegacy is the legacy Iceberg writer, which truncates timestamptz to millis.
+	writerLegacy s3DestinationWriter = "legacy"
+	// writerArrow is the Arrow Iceberg writer, which keeps micros. The Parquet
+	// destination also answers to this value: it is not observable from the config (its
+	// block just leaves arrow_writes where the Arrow block set it) and it keeps micros
+	// exactly like the Arrow writer (types.ToNewParquet pins every timestamp column to
+	// parquet.Microsecond), so it never needs telling apart.
+	writerArrow s3DestinationWriter = "arrow"
+)
```
vikaxsh, line 594: why do we require this?

tests/s3/s3_util_test.go
```diff
@@ -0,0 +1,1073 @@
...
+	"github.com/parquet-go/parquet-go/deprecated"
+	"github.com/stretchr/testify/require"
+)
+
+// The S3 integration test reuses the MinIO instance from the Iceberg destination stack
+// (destination/iceberg/local-test/docker-compose.yml) as the source object store. Every
+// format variant shares one bucket, isolated by per-variant path prefixes, so the parallel
+// subtests never touch each other's keys.
+const (
+	s3TestBucket = "olake-s3-test"
```
vikaxsh: Why aren't we reading this from the source config instead of hardcoding it?

pkg/parser/parquet.go
```diff
@@ -96,136 +112,495 @@ func (p *ParquetParser) StreamRecords(ctx context.Context, reader io.Reader, cal
...
+	for {
+		page, err := pages.ReadPage()
+		if err == io.EOF {
+			return values, nil
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to read page: %s", err)
+		}
+
+		buf := make([]pq.Value, page.NumValues())
```
vikaxsh: Avoid per-page allocations and copies by reading page values directly into the final buffer

vikaxsh (commented):
Please check the failing workflows as well

tests/s3/s3_test.go
mihir-datazip: Good catch, it was skipped after lib removal merge...

tests/s3/s3_test.go
mihir-datazip: Yes we should add it temporarily before the revamp lands

tests/s3/s3_test.go
mihir-datazip: Done

tests/s3/s3_test.go
mihir-datazip: Added column selection as well

tests/testutils/test_utils.go
mihir-datazip: So some drives change the stream names to uppercase. We can do a case insensitive match but that might break for other drivers with case sensitive names, so I added a list of drivers where we change to uppercase

tests/s3/s3_util_test.go
mihir-datazip: We need test coverage of all the cases. The difference in the timestamp precision in legacy and arrow writers should be asserted explicitly

tests/s3/s3_util_test.go
mihir-datazip: Added

pkg/parser/parquet.go
mihir-datazip: Done

vikaxsh (approved):

PR #1025, opened 2026-07-09
fix: cdc insert failure during gcp

vaibhav-datazip:
# Description

CDC syncs to Iceberg on **GCS via the s3-compat flavor (HMAC keys)** fail at commit with:

```
software.amazon.awssdk.services.s3.model.NoSuchKeyException: The specified key does not exist. (Service: S3, Status Code: 404, Request ID: null)
  at org.apache.iceberg.aws.s3.S3FileIO.deleteFile(S3FileIO.java:162)
  at org.apache.iceberg.io.BaseTaskWriter$BaseRollingWriter.closeCurrent(BaseTaskWriter.java:348)
```

**Root cause**:
Iceberg opens the equality-delete writer, but Parquet only materializes the object on GCS when the writer receives rows.
 On close, Iceberg deletes the "empty file" , which was never uploaded. AWS returns 204 for delete-of-missing-key (so this is invisible on AWS/MinIO); **GCS's S3-interop returns 404 NoSuchKey**, which escapes Iceberg's `catch (UncheckedIOException)` and kills the sync.
The failure starts only from the **second** CDC sync after full refresh (the `dedup_inserts` overlap window makes sync 1 write eq-deletes for inserts), matching the reported "first few CDC syncs worked" timeline.

**Fix:** restore AWS delete semantics at the S3-client boundary, using Iceberg's S3-scoped hook.

- New `OlakeS3ClientFactory` (implements Iceberg's `S3FileIOAwsClientFactory`): builds the standard S3 client via `AwsClientFactories.from(properties)`, wrapped so that `deleteObject` treats `NoSuchKey` as success. All other operations and errors (`NoSuchBucket`, `AccessDenied`, timeouts) propagate unchanged.
- `java_client.go` registers it via `s3.client-factory-impl` whenever a custom `s3_endpoint` is configured. This property is consumed only by `S3FileIO`, other things continues to work unchanged alongside this fix.
- Native-AWS configs (no custom endpoint) register nothing and behave identically to before.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Reproduced and validated end-to-end on a local Lakekeeper

- [x] **Repro before fix** : 
full refresh → CDC sync 1 with inserts  (dedup window) → **insert-only CDC sync 2 `NoSuchKeyException`**, stack identical to the customer incident, state not advanced, pod exit 1.
- [x] Tested same after changes , sync completes without errors
- [x] tested with jdbc minio setup, and local parquet sync  
- [x] test on lakekeeper (with and without partitions)
- [x]  jdbc minio setup as destination
- [x]  glue without s3 endpoint
- [x]  glue with  custom endpoint of gcp bucket also partitions
- [x] arrow_write: true , syncing to lakekeeper gcp setup (with and without partition)
- [x] arrow_writes : true, syncing to glue for backward compatability test  

# Screenshots or Recordings

## Documentation

- [ ] Documentation Link

## Related PR's (If Any):

### DIFF
--- a/destination/iceberg/java_client.go
+++ b/destination/iceberg/java_client.go
@@ -21,6 +21,9 @@ import (
 // defaultServerPort is the port the single shared JVM listens on.
 const defaultServerPort = 50051
 
+// Java class implementing Iceberg's AwsClientFactory
+const olakeAwsClientFactoryClass = "io.debezium.server.iceberg.OlakeAwsClientFactory"
+
 type serverInstance struct {
 	port            int
 	cmd             *exec.Cmd
@@ -55,7 +58,7 @@ func getServerConfigJSON(config *Config, port int, arrowWriterEnabled bool) ([]b
 		serverConfig["catalog-impl"] = "org.apache.iceberg.aws.glue.GlueCatalog"
 		// if custom glue endpoint creds are passed
 		if config.UseGlueAdditionalConfig {
-			addMapKeyIfNotEmpty("client.factory", "io.debezium.server.iceberg.OlakeAwsClientFactory")
+			addMapKeyIfNotEmpty("client.factory", olakeAwsClientFactoryClass)
 			addMapKeyIfNotEmpty("glue.access-key-id", config.GlueAccessKey)
 			addMapKeyIfNotEmpty("glue.secret-access-key", config.GlueSecretKey)
 			addMapKeyIfNotEmpty("glue.endpoint", config.GlueEndpoint)
@@ -103,6 +106,9 @@ func getServerConfigJSON(config *Config, port int, arrowWriterEnabled bool) ([]b
 
 	if config.S3Endpoint != "" {
 		serverConfig["s3.endpoint"] = config.S3Endpoint
+		// GCS's S3-interop returns 404 when deleting a missing key; AWS returns 204.
+		// Iceberg expects the AWS behavior, so use our factory, whose S3 client treats NoSuchKey-on-delete as success.
+		serverConfig["client.factory"] = olakeAwsClientFactoryClass
 	}
 	serverConfig["io-impl"] = "org.apache.iceberg.io.ResolvingFileIO"
 	serverConfig["s3.ssl-enabled"] = utils.Ternary(config.S3UseSSL, "true", "false").(string)
--- a/destination/iceberg/java_client.go
+++ b/destination/iceberg/java_client.go
@@ -18,11 +18,12 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
-// defaultServerPort is the port the single shared JVM listens on.
-const defaultServerPort = 50051
-
-// Java class implementing Iceberg's S3FileIOAwsClientFactory
-const olakeS3ClientFactoryClass = "io.debezium.server.iceberg.OlakeS3ClientFactory"
+const (
+	// defaultServerPort is the port the single shared JVM listens on.
+	defaultServerPort = 50051
+	// Java class implementing Iceberg's S3FileIOAwsClientFactory
+	olakeS3ClientFactoryClass = "io.debezium.server.iceberg.OlakeS3ClientFactory"
+)
 
 type serverInstance struct {
 	port            int
@@ -106,10 +107,11 @@ func getServerConfigJSON(config *Config, port int, arrowWriterEnabled bool) ([]b
 
 	if config.S3Endpoint != "" {
 		serverConfig["s3.endpoint"] = config.S3Endpoint
-		// GCS's S3-interop returns 404 when deleting a missing key; AWS returns 204.
-		// Iceberg expects the AWS behavior, so use our S3-only factory, whose client treats NoSuchKey-on-delete as success.
-		serverConfig["s3.client-factory-impl"] = olakeS3ClientFactoryClass
 	}
+	// Some s3-compatible stores (GCS's S3-interop) return 404 when deleting a
+	// missing key where AWS returns 204, and Iceberg expects the AWS behavior.
+	// The wrapper is a no-op on stores with AWS semantics.
+	serverConfig["s3.client-factory-impl"] = olakeS3ClientFactoryClass
 	serverConfig["io-impl"] = "org.apache.iceberg.io.ResolvingFileIO"
 	serverConfig["s3.ssl-enabled"] = utils.Ternary(config.S3UseSSL, "true", "false").(string)
 	// Marshal the config to JSON
--- a/destination/iceberg/java_client.go
+++ b/destination/iceberg/java_client.go
@@ -21,8 +21,8 @@ import (
 // defaultServerPort is the port the single shared JVM listens on.
 const defaultServerPort = 50051
 
-// Java class implementing Iceberg's AwsClientFactory
-const olakeAwsClientFactoryClass = "io.debezium.server.iceberg.OlakeAwsClientFactory"
+// Java class implementing Iceberg's S3FileIOAwsClientFactory
+const olakeS3ClientFactoryClass = "io.debezium.server.iceberg.OlakeS3ClientFactory"
 
 type serverInstance struct {
 	port            int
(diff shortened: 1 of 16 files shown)


destination/iceberg/java_client.go
```diff
@@ -21,6 +21,9 @@ import (
 // defaultServerPort is the port the single shared JVM listens on.
 const defaultServerPort = 50051
 
+// Java class implementing Iceberg's S3FileIOAwsClientFactory
+const olakeS3ClientFactoryClass = "io.debezium.server.iceberg.OlakeS3ClientFactory"
+
```
saksham-datazip: Can we write it like this ?
const (
	// defaultServerPort is the port the single shared JVM listens on.
	defaultServerPort = 50051
	// Java class implementing Iceberg's S3FileIOAwsClientFactory
	olakeS3ClientFactoryClass = "io.debezium.server.iceberg.OlakeS3ClientFactory"
)

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeS3ClientFactory.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.iceberg.aws.s3.S3FileIOAwsClientFactory;
+
+import software.amazon.awssdk.services.s3.DelegatingS3Client;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.DeleteObjectRequest;
+import software.amazon.awssdk.services.s3.model.DeleteObjectResponse;
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+
+/**
```
saksham-datazip: this line should be /*

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeS3ClientFactory.java
```diff
@@ -0,0 +1,66 @@
...
+    private static class DeleteTolerantS3Client extends DelegatingS3Client {
+        DeleteTolerantS3Client(S3Client delegate) {
+            super(delegate);
+        }
+
+        @Override
+        public DeleteObjectResponse deleteObject(DeleteObjectRequest request) {
+            try {
+                return super.deleteObject(request);
+            } catch (S3Exception e) {
```
saksham-datazip, line 50: Can you once confirm if we should merge these two functions deleteObject and isNoSuchKey?Also can you once confirm when e.awsErrorDetails().errorCode() in isNoSuchKey function can be null?

destination/iceberg/java_client.go
```diff
@@ -103,6 +106,9 @@ func getServerConfigJSON(config *Config, port int, arrowWriterEnabled bool) ([]b
 
 	if config.S3Endpoint != "" {
 		serverConfig["s3.endpoint"] = config.S3Endpoint
+		// GCS's S3-interop returns 404 when deleting a missing key; AWS returns 204.
+		// Iceberg expects the AWS behavior, so use our S3-only factory, whose client treats NoSuchKey-on-delete as success.
+		serverConfig["s3.client-factory-impl"] = olakeS3ClientFactoryClass
 	}
```
badalprasadsingh, line 110: So basically, anyone using `GCS` as the object store must provide the `S3Endpoint` as the parameter for this fix to work?

I understand right now, there is no other way to detect a `GCS` object store. But can this be fix be made more natural?

Please confirm this once: if `GCS` object store using `HMAC` keys is unreachable on _not_ passing `S3 Endpoint`. If it is unreachable, then we have to add this in doc as well, and then the fix would be valid as well.

Otherwise: confirm once it the person is even passing `S3Endpoint` on using Lakekeeper as destination.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeS3ClientFactory.java
```diff
@@ -0,0 +1,66 @@
...
+                if (isNoSuchKey(e)) {
+                    return DeleteObjectResponse.builder().build();
+                }
```
badalprasadsingh, line 53: This is `deleteObject`. Might fail on `deleteObjects`, as these are two different calls. 

Case: deleting multiple files due to lack of data.

Not applicable in our case, as second `equality delete file` or `position delete file` would just not open as there was no data for the first one never had any data. 

Have we tested it for partitioning? If not, please do it.

badalprasadsingh (changes requested):
Reviewed @vaibhav-datazip . Not tested.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeS3ClientFactory.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * It builds the standard S3 client and wraps it to treat NoSuchKey-on-delete as
+ * success: AWS reports deletes of missing keys as successful, while GCS's
+ * S3-interop returns 404 NoSuchKey, and Iceberg expects the AWS behaviour (e.g.
+ * its cleanup of empty writer files that were never uploaded). Other errors,
+ * including NoSuchBucket, still propagate.
+ */
+public class OlakeS3ClientFactory implements S3FileIOAwsClientFactory {
+
+    private AwsClientFactory delegate;
```
badalprasadsingh, line 29: Please make it `transient` as it is in `OLakeAwsClientFactory.java`.

destination/iceberg/java_client.go
vaibhav-datazip: gcs is reachable without endpoint as well, in case of s3/ parquet writer documentation we have added this

Screenshot 2026-07-09 at 9 58 44 PM

maybe we can add this , or we can register s3.client-factory-impl unconditionally

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeS3ClientFactory.java
vaibhav-datazip: partitioning is tested it works fine , and delete objects is not getting called from anywhere so wont be required for now

destination/iceberg/java_client.go
vaibhav-datazip: added this unconditionally

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeS3ClientFactory.java
vaibhav-datazip: it looks cleaner this way

 can be null on non parseable errors,  e.g. an empty/HTML 404 body from a proxy

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeS3ClientFactory.java
vaibhav-datazip: what will be the benefit of this  being made transient ?

destination/iceberg/java_client.go
badalprasadsingh: As its now un-conditionally, we need to test it once with Glue. As it falls in its path as well.

Previously, it was not catched as we generally don't pass `S3Endpoint` with Glue.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeS3ClientFactory.java
badalprasadsingh: Its a standard `nit` practise for OLake's standlone JVM. Its fine for now, non-blocking as the adoption fix should go first.

Context: `transient` tells Java to not serialize a field.

destination/iceberg/java_client.go
vaibhav-datazip: tested on glue as well, with no separate as s3 endpoint as well as with gcp as storage working fine

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeS3ClientFactory.java
badalprasadsingh: Cool

badalprasadsingh (approved):
LGTM. Not Tested.

saksham-datazip (approved):

PR #1026, opened 2026-07-09
feat(iceberg): upgrade Apache Iceberg from 1.7.2 to 1.10.2

vikaxsh:
# Description
This PR upgrades the Apache Iceberg version  **1.7.2** to the latest stable  **1.10.2** release.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/destination/iceberg/java_client.go
+++ b/destination/iceberg/java_client.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"strconv"
 	"strings"
 	"syscall"
@@ -122,17 +121,6 @@ func getServerConfigJSON(config *Config, port int, arrowWriterEnabled bool, gcpC
 	return json.Marshal(serverConfig)
 }
 
-// sweepStaleGCPCredsTempFiles clears creds temp files orphaned by a killed prior run (SIGKILL or reclaimPort); safe since only one shared JVM/temp file exists at a time.
-func sweepStaleGCPCredsTempFiles() {
-	matches, err := filepath.Glob(filepath.Join(os.TempDir(), "olake-gcp-creds-*.json"))
-	if err != nil {
-		return
-	}
-	for _, m := range matches {
-		os.Remove(m)
-	}
-}
-
 // writeGCPCredsTempFile materializes inline gcp_credentials_json content to a
 // temp file, since GoogleAuthManager only reads credentials from a file path.
 // Returns "" if credsJSON is empty.
@@ -150,10 +138,6 @@ func writeGCPCredsTempFile(credsJSON string) (string, error) {
 		return "", fmt.Errorf("failed to write gcp_credentials_json to temp file: %w", err)
 	}
 	f.Close()
-	if err := os.Chmod(f.Name(), 0o600); err != nil {
-		os.Remove(f.Name())
-		return "", fmt.Errorf("failed to set permissions on gcp credentials temp file: %w", err)
-	}
 	return f.Name(), nil
 }
 
@@ -170,8 +154,6 @@ func startServer(config *Config) (*serverInstance, error) {
 
 	// Forcefully kill any existing process on this port before starting
 	reclaimPort(port)
-	// Clean up any creds temp file orphaned by a killed prior run.
-	sweepStaleGCPCredsTempFiles()
 
 	// GoogleAuthManager only reads credentials from a file path (gcp.auth.credentials-path);
 	// materialize inline JSON content to a temp file so the property has something to point at.
--- a/destination/iceberg/java_client.go
+++ b/destination/iceberg/java_client.go
@@ -26,19 +26,20 @@ const (
 )
 
 type serverInstance struct {
-	port            int
-	cmd             *exec.Cmd
-	client          proto.RecordIngestServiceClient
-	arrowClient     proto.ArrowIngestServiceClient
-	conn            *grpc.ClientConn
-	defaultServerID string
+	port               int
+	cmd                *exec.Cmd
+	client             proto.RecordIngestServiceClient
+	arrowClient        proto.ArrowIngestServiceClient
+	conn               *grpc.ClientConn
+	defaultServerID    string
+	gcpCredentialsTemp string // temp file holding gcp_credentials_json content, if used; cleaned up on Shutdown
 }
 
 // getServerConfigJSON builds the catalog/storage-level config the JVM consumes
 // at startup. Per-stream concepts (namespace, upsert, identifier-fields,
 // partition spec) are deliberately *not* included here — they ride on every
 // per-request payload instead. See StreamMetaCtx.
-func getServerConfigJSON(config *Config, port int, arrowWriterEnabled bool) ([]byte, error) {
+func getServerConfigJSON(config *Config, port int, arrowWriterEnabled bool, gcpCredsTemp string) ([]byte, error) {
 	serverConfig := map[string]interface{}{
 		"port":                 fmt.Sprintf("%d", port),
 		"warehouse":            config.IcebergS3Path,
@@ -88,6 +89,8 @@ func getServerConfigJSON(config *Config, port int, arrowWriterEnabled bool) ([]b
 		addMapKeyIfNotEmpty("rest.auth.type", config.RestAuthType)
 		addMapKeyIfNotEmpty("credential", config.RestCredential)
 		addMapKeyIfNotEmpty("scope", config.RestScope)
+		addMapKeyIfNotEmpty("gcp.auth.credentials-path", gcpCredsTemp)
+		addMapKeyIfNotEmpty("gcp.auth.scopes", config.GCPAuthScopes)
 	default:
 		return nil, fmt.Errorf("unsupported catalog type: %s", config.CatalogType)
 	}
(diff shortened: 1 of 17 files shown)


destination/iceberg/java_client.go
```diff
@@ -118,6 +121,30 @@ func getServerConfigJSON(config *Config, port int, arrowWriterEnabled bool) ([]b
...
+	if err != nil {
+		return "", fmt.Errorf("failed to create temp file for gcp_credentials_json: %w", err)
+	}
+	if _, err := f.WriteString(credsJSON); err != nil {
+		f.Close()
+		os.Remove(f.Name())
+		return "", fmt.Errorf("failed to write gcp_credentials_json to temp file: %w", err)
+	}
+	f.Close()
+	if err := os.Chmod(f.Name(), 0o600); err != nil {
```
mihir-datazip: Do we need to do Chmod separately? We already create the File with 0o66 in `os.CreateTemp()`

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeAwsClientFactory.java
```diff
@@ -43,6 +44,11 @@ public S3Client s3() {
         return delegate.s3();
     }
 
+    @Override
+    public S3AsyncClient s3Async() {
+        return delegate.s3Async();
+    }
```
badalprasadsingh, line 50: We fixed a recent issue in this [PR](https://github.com/datazip-inc/olake/pull/1025/changes), where `GCS` used to give a `404 NoSuchKey` error, on deleting a file which does not exist. AWS on the other hand gives success.

For that, we added `DeleteTolerantS3Client`. We need to wrap it and all the other calls of `s3Async` with a similar function.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeS3ClientFactory.java
```diff
@@ -38,6 +39,11 @@ public S3Client s3() {
         return new DeleteTolerantS3Client(delegate.s3());
     }
 
+    @Override
+    public S3AsyncClient s3Async() {
+        return delegate.s3Async();
+    }
```
badalprasadsingh, line 45: Same comment for wrapping up with `DeleteTolerantS3Client` here as well.

destination/iceberg/resources/spec.json
```diff
@@ -204,6 +204,17 @@
                   "title": "REST Token",
                   "description": "Bearer token for token-based authentication"
                 },
+                "gcp_credentials_json": {
+                  "type": "string",
+                  "format": "password",
+                  "title": "GCP Credentials JSON",
+                  "description": "Inline GCP service account key JSON, used when REST Auth Type is org.apache.iceberg.gcp.auth.GoogleAuthManager (e.g. BigLake). Falls back to Application Default Credentials if left empty"
```
badalprasadsingh, line 211: nit: why to add this in description `org.apache.iceberg.gcp.auth.GoogleAuthManager`? Keep it simple.

destination/iceberg/resources/spec.json
```diff
@@ -204,6 +204,17 @@
...
+                },
+                "gcp_auth_scopes": {
+                  "type": "string",
+                  "title": "GCP Auth Scopes",
+                  "description": "Comma-separated OAuth2 scopes for GoogleAuthManager. Defaults to https://www.googleapis.com/auth/cloud-platform if left empty"
```
badalprasadsingh, line 216: It is said we fall back to `https://www.googleapis.com/auth/cloud-platform`, does it happens in side the JVM engine? As we have no such logic here.

Please confirm this.

destination/iceberg/java_client.go
```diff
@@ -173,8 +210,13 @@ func startServer(config *Config) (*serverInstance, error) {
 	appendEnv("AWS_REGION", config.Region)
 	appendEnv("AWS_SESSION_TOKEN", config.SessionToken)
 	appendEnv("AWS_PROFILE", config.ProfileName)
+	// GCSFileIO auths separately from GoogleAuthManager; only via ADC env var. No-op if empty.
+	appendEnv("GOOGLE_APPLICATION_CREDENTIALS", gcpCredsTemp)
```
badalprasadsingh, line 210: For the `GCP` auth mechnaism, we already took `gcp.auth.credentials-path` and `gcp.auth.scopes` fields above. Why do we have to take it as well from environment?

In case it is for the same purpose apply a nil check and prioritize one: better prioritize what you take in `destination.json` and in case nil fall back to the env variable.

destination/iceberg/java_client.go
```diff
@@ -241,6 +287,9 @@ func (s *serverInstance) Shutdown(ctx context.Context) {
 			_ = s.cmd.Process.Kill()
 		}
 	}
+	if s.gcpCredentialsTemp != "" {
+		os.Remove(s.gcpCredentialsTemp)
```
badalprasadsingh, line 287: It is good that we are cleaning up the `s.gcpCredentialsTemp` but have we tested it for `SIGKILL`, basically `Ctrl+C` while the sync is running, in case not, then it will still remain int `temp`.

1. Fix it for `SIGKILL` case.
2. Also, please confirm if `reclaimPort()` function also makes sure, this credentials file do not remain in `temp` folder.

badalprasadsingh (changes requested):
Please test the following. In case any of its tested, please ignore.

`P0`:
1. **Delete Granularity (+ Backward Compatibility)** : Update a single row of data multiple times, and run cdc sync (including equality deletes & position deletes) using OLake version >= `0.7.0`. Then, do a similar sync (update a single row of data multiple times and run cdc sync with this version, and query the table using spark
2. **Delete Granularity (Compaction)** : For an ongoing compaction, run cdc syncs in between where you have updated a single row of data multiple times, and test if any compaction fails or succeeds
3. **Partitioning Test with Deletes (+ Backward Compatibility)** : For all the transforms across the 6 data types that we handle, run the sync using this version of OLake (Iceberg `1.10.2`) on a table already having partitioned data from previous OLake version (Iceberg `1.7.2`)
4. **CDC Failure due to Delete in GCS** : [Context PR](https://github.com/datazip-inc/olake/pull/1025), please test the exact scenario because of which it failed. Please ask for the test case from @vaibhav-datazip 
5. **Polaris Catalog Test** : Please once test the flow with Polaris. (Ignore if already tested)

`P1`:

6. **Clear Destination** : test `purge` behavior, test if it is changed in `iceberg-1.10.2`, specially for GCS. Please test it in `BigLake` Catalog.  
7. **Trino & Athena Test (Backward Compatibility)** with Query Engine `Trino`. 
8. **Spark Compaction Test (+Backward Compatibility)** : We allow users to do Spark compaction on our iceberg tables. Should work on both pre-existing iceberg tables with new syncs coming from `iceberg-1.10.2`, and fresh iceberg tables created by `iceberg-1.10.2`. 

`P2`:

9. **Sort Order** : Right now, we false fully say that our data is sorted in parquet files in `metadata.json`. This did not cause issue to query engines in `1.7.2`. Can you please check we have face no issue for the same configuration in iceberg `1.10.2`, Check line `.withSortOrder(IcebergUtil.getIdentifierFieldsAsSortOrder(schema))` in `IcebergUtil.java`.

destination/iceberg/resources/spec.json
vikaxsh: please find here https://github.com/apache/iceberg/pull/13212/changes
image

destination/iceberg/resources/spec.json
vikaxsh: it would be better if we can confirm with product

destination/iceberg/java_client.go
vikaxsh: Not redundant, different consumers. gcp.auth.credentials-path/gcp.auth.scopes only configure GoogleAuthManager, which handles catalog REST calls. GCSFileIO , the thing that actually reads/writes GCS objects , is a completely separate code path with no equivalent property

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeAwsClientFactory.java
vikaxsh: we can avoid as of today neither delete method uses s3Async()

destination/iceberg/java_client.go
```diff
@@ -118,6 +122,41 @@ func getServerConfigJSON(config *Config, port int, arrowWriterEnabled bool) ([]b
 	return json.Marshal(serverConfig)
 }
 
+// sweepStaleGCPCredsTempFiles clears creds temp files orphaned by a killed prior run (SIGKILL or reclaimPort); safe since only one shared JVM/temp file exists at a time.
+func sweepStaleGCPCredsTempFiles() {
+	matches, err := filepath.Glob(filepath.Join(os.TempDir(), "olake-gcp-creds-*.json"))
```
badalprasadsingh: If two OLake Jobs run on the same machine, won't it delete it the file used by the second OLake job?

destination/iceberg/resources/spec.json
```diff
@@ -204,6 +204,17 @@
...
+                },
```
badalprasadsingh: Non-blocking (confirm with Product once) : In our REST Catalog forum on OLake UI, we already have a lot of fields. Check if we should gate these two fields:
1. `config.RestAuthType` == `org.apache.iceberg.gcp.auth.GoogleAuthManager` or,
2. Add a boolean field and then only display these fields on the UI

destination/iceberg/java_client.go
badalprasadsingh: As this function acts global

badalprasadsingh (changes requested):
Pending Tests:
`P0`

1. Using Snowflake as query engine - required for Interoperability
2. Schema Evolution Backward Compatibility

_might add few more tests here_

destination/iceberg/resources/spec.json
badalprasadsingh: Please confirm it with Product as we need to change the description.

destination/iceberg/java_client.go
vikaxsh: each container and pods gets its own /tmp dir and we can remove this cleanup as well

destination/iceberg/resources/spec.json
vikaxsh: description approved from product side

destination/iceberg/resources/spec.json
vikaxsh: confirmed with product , org.apache.iceberg.gcp.auth.GoogleAuthManager is fine

badalprasadsingh (approved):

PR #1029, opened 2026-07-11
FIX: enforce CDC initial wait time defaults for MySQL and Postgres

### LINKED ISSUE #492: fix: fail cdc sync if initial_wait_time provided is zero

kaizakin:
# Description

MYSQL
- validate the initial wait time and fail the setup if it's 0 or negative
- perform query based checks to identify whether the current user has permission to read binlog

POSTGRES
- distinguish `initial_wait_time` "not provided" from "explicitly set to 0"
   - if not provided set the default value `1200`
   - if set to 0 fail it immediately
- perform query based checks to validate `wal_level` `max_replication_slots` `max_wal_senders` 
- check whether the current user has the replication or superuser privilege required for CDC  

Fixes #492

## Type of change 

/king bug 

# How Has This Been Tested?
running sync locally

# Screenshots or Recordings
Olake CLI and Olake UI (MySQL)

https://github.com/user-attachments/assets/c19e680a-242d-401e-8f4a-3abbb6cd4a5e

Olake CLI and Olake UI (Postgres)

https://github.com/user-attachments/assets/97bdd787-4ff7-4c08-ba4b-ff6f5408f2a4

## Documentation

## Related PR's (If Any):
NONE


kaizakin:
Requesting review :- @vishalm0509 @vaibhav-datazip @nayanj98

nayanj98:
@kaizakin Sorry for the delay assigning @vaibhav-datazip to review your PR

drivers/mysql/internal/mysql.go
```diff
@@ -135,31 +135,51 @@ func (m *MySQL) Setup(ctx context.Context) error {
...
-			// default set 10 sec
-			cdc.InitialWaitTime = 10
+		if cdc.InitialWaitTime <= 0 {
+			// Fail the setup if initial_wait_time is 0 or negative
+			return fmt.Errorf("failed to setup CDC: Initial wait time must be greater than 0")
 		}
+
+		// Permission check for CDC
+		var grants []string
+		if err := client.SelectContext(ctx, &grants, "SHOW GRANTS FOR CURRENT_USER()"); err != nil {
```
vaibhav-datazip, line 154: I don't think these many checks are even required, aren't we checking during discover that cdc is supported or not ?

and based on that only we get the supported sync modes, the issue raised related to initial wait time, we are changing it to default , also can't we set initial wait time to some default value if its set to 0 or less

kaizakin: @vaibhav-datazip I have added these permission checks based on your comments in this PR https://github.com/datazip-inc/olake/pull/563#discussion_r2429056783

kaizakin: > also can't we set initial wait time to some default value if its set to 0 or less

makes sense, but i didn't add because the issue clearly asks to fail the sync when `initial_wait_time` is set to 0.

although on the postgres side if the `initial_wait_time` is left empty i have added logic to set it to `1200` (default value)
if it's provided but not within the allowed window (120 to 2400) we fail the setup

kaizakin: > aren't we checking during discover that cdc is supported or not ? and based on that only we get the supported sync modes

yes, during `olake discover` it calls two functions in sequence `setup()` and `discover()` 
setup runs whatever CDC-prerequisite checks that driver has, then sets a `CDCSupport` bool field 
discover just reads the CDCSupport flag that Setup() already computed

setup() isn't just limited to discover it's also seem to be called by `sync` and `check`

as for the mysql side before the current changes
```go
cdcSupported, err := m.IsCDCSupported(ctx)
if err != nil {
    logger.Warnf("failed to check CDC support: %s", err)
}
if !cdcSupported {
    logger.Warnf("CDC is not supported")
}
m.CDCSupport = cdcSupported
```
which means Discover would quietly drop CDC from SupportedSyncModes and default to incremental, no error or failure only a warning

now it has been hardened. if the checks fail we instantly fail the setup instead of a warning.

as for the postgres side before the current changes
there was no `IsCDCSupported` check at all, The old Setup() only checked "does the replication slot exist"
That means Postgres would report CDCSupport = true even on a server with `wal_level=replica` or a user lacking REPLICATION then fail later.

Please correct me if im wrong @vaibhav-datazip

kaizakin:
@vaibhav-datazip i went thru the code a bit, it seems like mongodb doesn't do permission check as well.

and as for MSSQL it has a proper check but it doesn't fail the setup, does the same warning and move on.

and as for `Oracle` and `DB2` i assume CDC is unimplemented, because they're hardcoded to return false.

If these are relevant and within the scope i would like to help with a follow up PR. Please let me know :)

drivers/mysql/internal/mysql.go
vaibhav-datazip: we can't risk the permissions as this can cause issues in existing sync as well. can you please check and test all the scenarios here in postgres

some of them which I found are 

SELECT rolsuper OR rolreplication FROM pg_roles WHERE rolname = current_user

On RDS/Aurora both columns are false

please verify others as well

drivers/postgres/internal/postgres.go
```diff
@@ -133,16 +190,25 @@ func (p *Postgres) Setup(ctx context.Context) error {
...
 
-		// check if initial wait time is valid or not
-		if cdc.InitialWaitTime < 120 {
-			return fmt.Errorf("the CDC initial wait time must be at least 120 seconds")
+		// distinguish "not provided" from "explicitly set to 0"
+		waitTimeProvided, _ := utils.IsOfType(p.config.UpdateMethod, "initial_wait_time")
+		if !waitTimeProvided {
+			cdc.InitialWaitTime = 1200
+		} else if cdc.InitialWaitTime < 120 || cdc.InitialWaitTime > 2400 {
+			return fmt.Errorf("initial_wait_time must be between 120 and 2400 seconds, got %d", cdc.InitialWaitTime)
```
vaibhav-datazip, line 198: why are we failing if its <120 , why not just pin it 120 if its less, also do we really need an upper limit

drivers/postgres/internal/postgres.go
```diff
@@ -77,6 +79,61 @@ func (p *Postgres) CDCSupported() bool {
...
+			return false, fmt.Errorf("failed to check %s: %s", settingName, err)
+		}
+		if value != expectedValue {
+			logger.Warnf("%s is set to '%s', expected '%s'", settingName, value, expectedValue)
+			return false, nil
+		}
+		return true, nil
+	}
+
+	if ok, err := checkSetting(jdbc.PostgresWalLevelQuery(), "wal_level", "logical"); err != nil || !ok {
```
vaibhav-datazip, line 98: please check if these can throw fatal

drivers/mysql/internal/mysql.go
kaizakin: sure @vaibhav-datazip I'll test em all and report back with the results.

kaizakin:
1. test for `wal_level`

https://github.com/user-attachments/assets/191aef86-fe5c-4bd6-b1d3-a966f0c74489


2. test for `replication_slots`

https://github.com/user-attachments/assets/593dbb25-5009-41aa-99d9-5e3bdadc5a7c

3. test for `wal_senders`

https://github.com/user-attachments/assets/7bd06aed-6daa-40be-9a65-0d1aafa99073

kaizakin:
@vaibhav-datazip however I found a gap in my implementation when i was testing this `SELECT rolsuper OR rolreplication FROM pg_roles WHERE rolname = current_user` 

since `setup()` is being called in `discover` and `sync`, when i run `discover` or a `full refresh` with a non privileged user i get thrown a CDC related error
and not to mention it's not letting me move forward, failing right away.

which is not the expected behavior. we want CDC related errors to be thrown only during CDC & it should let me move forward without failing for this scenario

image

for now I'm converting this PR to draft,
I'll reopen it once this along with the changes you've requested are Implemented. 

Thank you.

kaizakin:
@vaibhav-datazip My bad I went thru the codebase a bit. I was getting that error because i ran `discover` with `update_method` field set in `source.json` 

I referred the docs https://olake.io/docs/connectors/postgres/?config-type=olake-cli#2-provide-configuration-details-1 as well 

I was wondering why the CDC based checks run after I myself have set the field in `source.json`

kaizakin:
4. test for `SELECT rolsuper OR rolreplication FROM pg_roles WHERE rolname = current_user`

https://github.com/user-attachments/assets/0ee93246-f3b8-4d78-a903-d75b6b7e6a08

drivers/postgres/internal/postgres.go
kaizakin: I've made it this way because the existing code follows it https://github.com/datazip-inc/olake/blob/master/drivers/postgres/internal/config.go#L32

I'm leaning towards your opinion on the pinning to <120 one. 
the changes i've made fails the sync if the `initial_wait_time` is set to 0 because the issue description explicitly asks for it.

my opinion is that we pin the  `initial_wait_time` to 120 if the given input is <120 (including 0)

---
> do we really need an upper limit

upper limit is not really needed but its good to have i think, to prevent someone from mistakenly entering a large value.

drivers/postgres/internal/postgres.go
kaizakin: Query errors from `pg_settings` are unlikely for normal Postgres users. However it can still fail from connection/query/scan problems.

kaizakin: if it throws fatal its okay because if we cannot verify CDC prerequisites we should not continue in CDC mode right?

kaizakin:
@vaibhav-datazip I've replied to your comments PTAL :)

drivers/mysql/internal/mysql.go
```diff
@@ -135,31 +135,51 @@ func (m *MySQL) Setup(ctx context.Context) error {
...
 		cdc := &CDC{}
 		if err := utils.Unmarshal(m.config.UpdateMethod, cdc); err != nil {
 			return err
 		}
-		if cdc.InitialWaitTime == 0 {
-			// default set 10 sec
-			cdc.InitialWaitTime = 10
+		if cdc.InitialWaitTime <= 0 {
+			// Fail the setup if initial_wait_time is 0 or negative
+			return fmt.Errorf("failed to setup CDC: Initial wait time must be greater than 0")
```
vaibhav-datazip: for <=0 we should cap it to minimum value which is allowed as we are doing in ui, instead of failing

drivers/mysql/internal/mysql.go
```diff
@@ -135,31 +135,51 @@ func (m *MySQL) Setup(ctx context.Context) error {
...
+			return fmt.Errorf("failed to get grants for current user: %s", err)
+		}
+		grantStr := strings.ToUpper(strings.Join(grants, " "))
+
+		if !strings.Contains(grantStr, "ALL PRIVILEGES") {
+			permissions := []string{"SELECT", "REPLICATION SLAVE", "REPLICATION CLIENT"}
+			for _, perm := range permissions {
+				if !strings.Contains(grantStr, perm) {
+					return fmt.Errorf("missing required permissions for CDC: SELECT, REPLICATION SLAVE, or REPLICATION CLIENT")
+				}
```
vaibhav-datazip: this might fail for mariaDB 10.5+ , please check 

also this will fail for role based permissions , please check if we can handle this or we will have to remove it

drivers/postgres/internal/postgres.go
```diff
@@ -132,16 +189,25 @@ func (p *Postgres) Setup(ctx context.Context) error {
...
-		// check if initial wait time is valid or not
-		if cdc.InitialWaitTime < 120 {
-			return fmt.Errorf("the CDC initial wait time must be at least 120 seconds")
+		// distinguish "not provided" from "explicitly set to 0"
+		waitTimeProvided, _ := utils.IsOfType(p.config.UpdateMethod, "initial_wait_time")
+		if !waitTimeProvided {
+			cdc.InitialWaitTime = 1200
+		} else if cdc.InitialWaitTime < 120 || cdc.InitialWaitTime > 2400 {
+			return fmt.Errorf("initial_wait_time must be between 120 and 2400 seconds, got %d", cdc.InitialWaitTime)
 		}
```
vaibhav-datazip, line 149: lets set a default time if time is not provided or not in the range , just like we do in olake ui

drivers/postgres/internal/postgres.go
```diff
@@ -77,6 +79,61 @@ func (p *Postgres) CDCSupported() bool {
 	return p.CDCSupport
 }
 
+// IsCDCSupported checks that the Postgres server and connecting role satisfy CDC
+// prerequisites: wal_level=logical, sufficient max_replication_slots/max_wal_senders,
+// and REPLICATION privilege.
+func (p *Postgres) IsCDCSupported(ctx context.Context, client *sqlx.DB) (bool, error) {
```
vaibhav-datazip: why do we need to pass the client separately , why wont p.client work here ?

pkg/jdbc/jdbc.go
```diff
@@ -221,6 +221,27 @@ func PostgresWalLSNQuery() string {
...
+func PostgresMaxWalSendersQuery() string {
+	return `SELECT name, setting FROM pg_settings WHERE name = 'max_wal_senders'`
+}
+
+// PostgresReplicationPrivilegeQuery returns the query to check if the connecting role
+// has the REPLICATION privilege (or superuser) required for CDC
+func PostgresReplicationPrivilegeQuery() string {
+	return `SELECT rolsuper OR rolreplication FROM pg_roles WHERE rolname = current_user`
+}
+
```
vaibhav-datazip: all these permissions rely on 

```
REVOKE SELECT ON pg_catalog.pg_settings FROM PUBLIC;
REVOKE SELECT ON pg_catalog.pg_roles FROM PUBLIC;
```

and if these two are revoked there wont be any affect on cdc , but our discover will incorrectly not show cdc mode, hence you can check and maybe remove these

drivers/postgres/resources/spec.json
```diff
@@ -167,7 +167,9 @@
                             "type": "integer",
                             "title": "Initial Wait Time",
                             "description": "Idle timeout for wal catchup to Saved LSN",
-                            "minimum": 120
+                            "minimum": 120,
+                            "maximum": 2400,
```
vaibhav-datazip: you shouldn't put a upper cap on initial wait time anywhere

drivers/mysql/internal/mysql.go
kaizakin: I've read more about it @vaibhav-datazip all of em in a way leads to false-positives so i think its better to remove it.

pkg/jdbc/jdbc.go
kaizakin: same case here as well, let's remove it

kaizakin:
PTAL @vaibhav-datazip

vaibhav-datazip:
@kaizakin , will try to review it before Thursday

drivers/mysql/internal/mysql.go
```diff
@@ -139,31 +139,35 @@ func (m *MySQL) Setup(ctx context.Context) error {
...
+			cdc.InitialWaitTime = constants.MinCDCInitialWaitTime
+		}
+
+		// Enable CDC support if binlog is configured
+		cdcSupported, err := m.IsCDCSupported(ctx)
+		if err != nil {
+			return err
+		}
+		if !cdcSupported {
+			return fmt.Errorf("failed to setup CDC")
```
vaibhav-datazip: we have changed how we throw errors now, so if the error is due to some config you can wrap it in precondition function , you can check in the codebase now,

also please resolve the conflict. once that is done , will try to merge this pr soon

kaizakin:
@vaibhav-datazip it's ready to merge :)

constants/constants.go
```diff
@@ -54,6 +54,12 @@ const (
 	// CdcTimestamp is the column name olake writes the CDC event timestamp into.
 	CdcTimestamp = "_cdc_timestamp"
 
+	// MinCDCInitialWaitTime is the floor (in seconds) for a CDC driver's initial_wait_time;
+	// values below this are clamped up to it rather than failing setup, matching the UI's minimum.
+	MinCDCInitialWaitTime = 120
+	// DefaultCDCInitialWaitTime is used when initial_wait_time is not provided.
+	DefaultCDCInitialWaitTime = 1200
```
vaibhav-datazip: you can remove these constants from here as we do have separate time being set for different drivers right ?

for mysql we can clamp it to 120 sec if its not provided or less than 120, for postgres if its not provided or less than 120 we can make it 10800 (as it was earlier).

drivers/mysql/internal/mysql.go
```diff
@@ -139,31 +140,35 @@ func (m *MySQL) Setup(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+		if !cdcSupported {
+			return errs.Precondition(errs.CDCPreconditionFailed, codeCDCUnsupported, fmt.Errorf("failed to setup CDC: binlog is not configured correctly"))
 		}
+
+		logger.Infof("CDC setup done with initial wait time %d", cdc.InitialWaitTime)
+
+		m.CDCSupport = true
```
vaibhav-datazip: this log can be removed 

and although at this point cdc supported will be true only still we can map 

m.cdcSupport = cdcSupported

drivers/postgres/internal/postgres.go
```diff
@@ -134,13 +135,14 @@ func (p *Postgres) Setup(ctx context.Context) error {
...
-		if cdc.InitialWaitTime < 120 {
-			return errs.Precondition(errs.ConfigInvalid, codeCDCWaitTimeTooLow,
-				fmt.Errorf("the CDC initial wait time must be at least 120 seconds"))
+		// distinguish "not provided" from "explicitly set to 0"
+		waitTimeProvided, _ := utils.IsOfType(p.config.UpdateMethod, "initial_wait_time")
+		if !waitTimeProvided {
+			cdc.InitialWaitTime = constants.DefaultCDCInitialWaitTime
+		} else if cdc.InitialWaitTime < constants.MinCDCInitialWaitTime {
+			logger.Warnf("initial_wait_time %d is below the minimum of %d seconds; using %d", cdc.InitialWaitTime, constants.MinCDCInitialWaitTime, constants.MinCDCInitialWaitTime)
+			cdc.InitialWaitTime = constants.MinCDCInitialWaitTime
```
vaibhav-datazip: what's the point of this ? , we can directly get cdc initial_wait_time .

drivers/postgres/resources/spec.json
```diff
@@ -167,7 +167,8 @@
                             "type": "integer",
                             "title": "Initial Wait Time",
                             "description": "Idle timeout for wal catchup to Saved LSN",
-                            "minimum": 120
+                            "minimum": 120,
+                            "default": 1200
```
vaibhav-datazip: you can set this to 10800, also have you tested this on olake-ui ?

vaibhav-datazip (commented):
the pr is almost ready to merge, there are some changes required , please go through them thoroughly and make changes.

also please refrain from force pushing , whenever you get that error you should instead follow these steps:

1. checkout to your base branch, and pull the latest changes (for fork: update  you fork)
2. checkout back to your feature branch and merge the base branch in it
3. resolve conflicts (if any)


Also, please don't resolve the reviewer comments on your own .

kaizakin:
@vaibhav-datazip comments are addressed PTAL

drivers/mysql/internal/mysql.go
```diff
@@ -149,9 +149,9 @@ func (m *MySQL) Setup(ctx context.Context) error {
 		if err := utils.Unmarshal(m.config.UpdateMethod, cdc); err != nil {
 			return err
 		}
-		if cdc.InitialWaitTime < constants.MinCDCInitialWaitTime {
-			logger.Warnf("initial_wait_time %d is below the minimum of %d seconds; using %d", cdc.InitialWaitTime, constants.MinCDCInitialWaitTime, constants.MinCDCInitialWaitTime)
-			cdc.InitialWaitTime = constants.MinCDCInitialWaitTime
+		if cdc.InitialWaitTime <= 0 || cdc.InitialWaitTime < minCDCInitialWaitTime {
```
vaibhav-datazip: do you think first condition is even required ?

drivers/postgres/internal/postgres.go
```diff
@@ -136,17 +136,11 @@ func (p *Postgres) Setup(ctx context.Context) error {
...
 		}
 
-		// distinguish "not provided" from "explicitly set to 0"
-		waitTimeProvided, _ := utils.IsOfType(p.config.UpdateMethod, "initial_wait_time")
-		if !waitTimeProvided {
-			cdc.InitialWaitTime = constants.DefaultCDCInitialWaitTime
-		} else if cdc.InitialWaitTime < constants.MinCDCInitialWaitTime {
-			logger.Warnf("initial_wait_time %d is below the minimum of %d seconds; using %d", cdc.InitialWaitTime, constants.MinCDCInitialWaitTime, constants.MinCDCInitialWaitTime)
-			cdc.InitialWaitTime = constants.MinCDCInitialWaitTime
+		if cdc.InitialWaitTime <= 0 || cdc.InitialWaitTime < minCDCInitialWaitTime {
```
vaibhav-datazip: here as well

vaibhav-datazip (commented):
will suggest that you should look at your changes and self review your pr as well before requesting for review, unnecessary code can be removed at that step 

also it will help your pr to get merged quicker , please look at your pr again in depth before requesting next review

kaizakin:
my bad @vaibhav-datazip I must have done that in a hurry 

Thanks for your patience and help in reviewing this PR

kaizakin:
I've amended the changes

this time i've read thru the changes, everything is on place

vaibhav-datazip:
> I've amended the changes
> 
> this time i've read thru the changes, everything is on place

we had also discussed about taking merge from staging instead of force pushing please keep that into account as well. As we will be doing squash and merge only one commit of this pr will be merged to staging.

drivers/postgres/resources/spec.json
```diff
@@ -167,7 +167,8 @@
                             "type": "integer",
                             "title": "Initial Wait Time",
                             "description": "Idle timeout for wal catchup to Saved LSN",
-                            "minimum": 120
+                            "minimum": 120,
+                            "default": 10800
```
vaibhav-datazip: you can remove default from here completely

drivers/mysql/internal/config.go
```diff
@@ -27,6 +27,8 @@ type Config struct {
 	SSHConfig        *utils.SSHConfig  `json:"ssh_config"`
 }
 
+const minCDCInitialWaitTime = 120
```
vaibhav-datazip: as these constants are being used in mysql.go you can define them there only

kaizakin:
@vaibhav-datazip nitpicks has been addressed. you can take a look now

drivers/mysql/internal/mysql.go
```diff
@@ -28,6 +29,8 @@ import (
 // MEDIUMINT's 3 bytes are the one MySQL integer width Go has no constant for.
 const maxUint24 = 1<<24 - 1
 
+const minCDCInitialWaitTime = 120
```
vaibhav-datazip: nit

```suggestion
const (
maxUint24 = 1<<24 - 1
 minCDCInitialWaitTime = 120
 )
```

whenever we define constant we can group them together like this as you have done in postgres side also adding a brief and short one liner comment helps for what that particular constant is used for

drivers/postgres/internal/postgres.go
```diff
@@ -24,6 +24,11 @@ import (
...
 )
 
+const (
+	minCDCInitialWaitTime     = 120
+	defaultCDCInitialWaitTime = 10800
+)
+
 const (
 	// get all schemas and table
 	getPrivilegedTablesTmpl = `SELECT nspname as table_schema,
```
vaibhav-datazip, line 36: you can keep these ones as they are like now

as the lower block has all the query statements, current implementation is more readable.

drivers/postgres/internal/postgres.go
```diff
@@ -134,17 +140,12 @@ func (p *Postgres) Setup(ctx context.Context) error {
 		if err := utils.Unmarshal(p.config.UpdateMethod, cdc); err != nil {
 			return err
 		}
-		// set default value
-		cdc.InitialWaitTime = utils.Ternary(cdc.InitialWaitTime == 0, 10800, cdc.InitialWaitTime).(int)
 
-		// check if initial wait time is valid or not
-		if cdc.InitialWaitTime < 120 {
-			return errs.Precondition(errs.ConfigInvalid, codeCDCWaitTimeTooLow,
```
vaibhav-datazip, line 142: is this error code still used , if not can we remove it from errors.go as well

drivers/mysql/internal/mysql.go
```diff
@@ -25,8 +26,13 @@ import (
 	"github.com/go-sql-driver/mysql"
 )
 
-// MEDIUMINT's 3 bytes are the one MySQL integer width Go has no constant for.
-const maxUint24 = 1<<24 - 1
+const (
+	// MEDIUMINT's 3 bytes are the one MySQL integer width Go has no constant for.
+	maxUint24 = 1<<24 - 1
+
+	// minCDCInitialWaitTime is the minimum wait time in seconds for CDC synchronization.
```
vaibhav-datazip: as you are going to push the commit for removing the unused error code, just add this minor change in wording as well

synchronization -> sync

wherever you have added synchronization in comment

kaizakin:
@vaibhav-datazip PTAL

vaibhav-datazip (approved):
LGTM

PR #1030, opened 2026-07-13
feat: log OLake version on startup

sahitya-datazip:
# Description

Adds OLake version to startup logs so that running version is visible when an application starts.
- Added a shared version utility under utils/version.
- Updated protocol/sync.go to log the OLake version during startup.

OL-2972

## Type of change

# How Has This Been Tested?

- [x] Executed the sync command and verified that OLake version is logged during startup.

# Screenshots or Recordings

### Discover:
image

### Sync:
image


mihir-datazip:
@sahitya-datazip Can you add a screenshot showing the log appearing when we run Discover and Sync.
Also update the target branch to be `staging` instead of `master`

mihir-datazip (approved):

PR #1024, opened 2026-07-09
chore: docker optimizations for integration tests

mihir-datazip:
# Description
Updating Olake docker image to use a custom base image for all drivers to reduce the time of spawning a fresh driver image.

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings
### Old test flow (~350 seconds)
image
**Note: No optimisation for multiple runs of same test**

### New flow with downloading base image (~129 second)
image

### New flow with cached base image (~17 second)
image

## Documentation
- [x] N/A (bug fix, refactor, or test changes only)

## Related PR's (If Any):

PR #1021, opened 2026-07-06
chore: add make targets for test databases, dev builds and integration tests

mihir-datazip:
# Description
Adds `make` targets wrapping the local dev/test infrastructure so contributors can start the database stacks, build driver binaries, and run the integration / 2PC / unit suites the same way CI does. Driver-specific bits (readiness probes, one-time init, build guards) live in per-driver `drivers/<driver>/driver.mk` fragments that the root `Makefile` pulls in via `-include`. `CONTRIBUTING.md` gains a "Running tests locally" pointer to `make help`.

- `db.<driver>.start|stop|restart|teardown|refresh` / `db.source.all.*` — start/stop/restart/teardown/refresh a source DB (compose up + readiness wait + one-time init, e.g. the idempotent MSSQL init SQL); `teardown` and `refresh` also drop volumes
- `db.destination.all.start|stop|restart` and `db.destination.all.teardown|refresh` — the MinIO + mc + Iceberg JDBC catalog (postgres) + Spark Connect stack the tests verify against (`teardown`/`refresh` also delete `destination/iceberg/local-test/data`); only the services the tests need are started (the unused lakekeeper REST catalog + migrate job are skipped)
- `db.all.start|stop|restart|teardown|refresh` — everything above (sources + destination)
  - `dev.<driver>.build` — host binary at `drivers/<driver>/olake` for every driver module (all sources plus `s3`), with the same version/commit/channel ldflags as `build.sh`; `db2` is gated by a build guard that checks for the IBM clidriver env
- Per-driver `drivers/<driver>/driver.mk` fragments declare each driver's readiness probe, retry overrides, recovery/init hooks and build guard, plus driver-only targets like `test.rebalance.kafka`
- The Iceberg writer JAR is now a first-class file rule — rebuilt only when `pom.xml` or the writer sources change, with a dockerized-Maven fallback when `mvn` is absent, and it refreshes the repo-root JAR copy to avoid the stale-JAR trap. It's a prerequisite of the test targets, so they never run against a stale writer
- `test.integration[.<driver>]`, `test.2pc[.<driver>]`, `test.unit` — CI-equivalent commands that auto-provision the databases (and JAR) they need
- `lint` / `build` — local mirrors of CI's golangci-lint and root-module build-check
- Drops the obsolete top-level `version:` key from four compose files (destination iceberg stack, mysql, oracle, postgres) to silence the Compose v2 deprecation warning

Fixes # (issue): N/A

## Type of change
[...]

PR #1033, opened 2026-07-14
refactor: de-couple Integration Test module from codebase

mihir-datazip:
# Description
Moving the IT framework completely outside the olake codebase as a separate module.
This helps in
- Decoupling the Testing module so we can compile the tests are independent of our codebase
- Dependency separation for our main driver builds
- Opens path to run tests even when the main codebase is not supported in current Arch
- CI doesn't need to maintain a name-based logic separation for UTs and ITs

## Type of change

# How Has This Been Tested?
Testing running all the tests locally once to verify, also CI output confirms all the tests running correctly

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):
https://github.com/datazip-inc/olake/pull/1024
https://github.com/datazip-inc/olake/pull/1021


.github/workflows/integration-tests-runner.yml
```diff
@@ -36,127 +29,44 @@ jobs:
...
-        run: go build -v ./...
-
+          sudo mkdir -p destination/iceberg/local-test/data/{postgres-data,minio-data,ivy-cache}
+          sudo chown -R 999:999 destination/iceberg/local-test/data
+          sudo chmod -R 777 destination/iceberg/local-test/data
+
+      # Before the stacks start: hashFiles walks whatever is in the workspace, and
+      # the catalog postgres locks its data dir to 0700 uid 999 the moment it boots,
+      # which the runner user cannot read. Keyed on the one pom.xml by path rather
+      # than **/pom.xml for the same reason -- no full-workspace walk.
```
ImDoubD-datazip: is this comment needed? I mean there is only one pom.xml file existing in the iceberg java side, which we require to load the dependency of iceberg-java, so it is pre-understood.

.github/workflows/integration-tests-runner.yml
```diff
@@ -36,127 +29,44 @@ jobs:
...
 
+      # One compose up per source stack plus the destination stack, all at once,
+      # so the image pulls overlap instead of queueing. Containers boot in the
+      # background while the steps below run; nothing blocks until the wait step.
+      - name: Start Test Infrastructure
+        run: make -j --output-sync=target db.all.up
+
+      # Host build deps per driver (db2: the IBM clidriver; a no-op for the rest).
+      # The cgo env for it comes from GO_ENV.db2 in the test targets, so nothing
+      # needs exporting into $GITHUB_ENV here.
```
ImDoubD-datazip: the comment does not make sense, as compared to what is being done below. if prepare for now only works for db2, let's just mention that in a single liner, that this runs for db2 in form of note.

.github/workflows/kafka-rebalance-tests.yml
ImDoubD-datazip, line 26: does kafka rebalance test require this big runner? can we use a smaller one if possible here?

.github/workflows/kafka-rebalance-tests.yml
```diff
@@ -38,61 +40,48 @@ jobs:
...
-      - name: Wait for Spark Connect
-        uses: nick-fields/retry@v2
-        with:
-          timeout_minutes: 5
-          max_attempts: 30
-          retry_wait_seconds: 5
-          command: |
-            bash -c 'echo > /dev/tcp/127.0.0.1/15002'
-
+      # Before anything starts: the iceberg catalog's postgres runs as uid 999 and
```
ImDoubD-datazip: why the uid 999 mentioned here, does this signify something special?

.github/workflows/kafka-rebalance-tests.yml
```diff
@@ -38,61 +40,48 @@ jobs:
...
-      - name: Build Project
-        run: go build -v ./...
+          sudo mkdir -p destination/iceberg/local-test/data/{postgres-data,minio-data,ivy-cache}
+          sudo chown -R 999:999 destination/iceberg/local-test/data
+          sudo chmod -R 777 destination/iceberg/local-test/data
 
+      # Before the stacks start: hashFiles walks whatever is in the workspace, and
+      # the catalog postgres locks its data dir to 0700 uid 999 the moment it boots,
+      # which the runner user cannot read. Keyed on the one pom.xml by path rather
+      # than **/pom.xml for the same reason -- no full-workspace walk.
```
ImDoubD-datazip: no need for all of this, just comment that to build the iceberg java, dependencies are required which we get via pom file

.github/workflows/integration-tests-runner.yml
```diff
@@ -168,51 +78,31 @@ jobs:
...
-          OLAKE_TEST_PHASE=discover go test -v -p "$#" -parallel 1 "$@" -timeout 0 -run 'Integration'
-
+          docker run --rm \
+            --add-host host.docker.internal:host-gateway \
+            -v "$PWD:/test-olake" \
+            -v go-mod:/go/pkg/mod \
+            -v go-build:/root/.cache/go-build \
+            -w /test-olake \
+            -e TELEMETRY_DISABLED=true \
+            -e OLAKE_SKIP_MOD_TIDY=1 \
```
ImDoubD-datazip: what does `OLAKE_SKIP_MOD_TIDY` do, does setting this is required?

.github/workflows/integration-tests-runner.yml
```diff
@@ -168,51 +78,31 @@ jobs:
...
-        run: |
-          docker compose -f ./destination/iceberg/local-test/docker-compose.yml down
-          docker compose -f ./drivers/mysql/docker-compose.yml down
-          docker compose -f ./drivers/postgres/docker-compose.yml down
-          docker compose -f ./drivers/oracle/docker-compose.yml down
-          docker compose -f ./drivers/mongodb/docker-compose.yml down
-          docker compose -f ./drivers/db2/docker-compose.yml down
-          docker compose -f ./drivers/mssql/docker-compose.yml down
-          docker compose -f ./drivers/kafka/docker-compose.yml down
+        run: make -j db.all.stop
```
ImDoubD-datazip: here you are saying db.all.stop, but here iceberg is also stopped with other dbs source. shouldn't they be separate (the iceberg and db)?

.github/workflows/kafka-rebalance-tests.yml
```diff
@@ -38,61 +40,48 @@ jobs:
...
-        run: go test -v ./drivers/kafka/internal/... -timeout 0 -run TestKafkaRebalance
+        run: make test.kafka.rebalance
 
+      # stop, not teardown: teardown also deletes the root-owned data dir above.
       - name: Cleanup
         if: always()
-        run: |
-          docker compose -f ./destination/iceberg/local-test/docker-compose.yml down
-          docker compose -f ./drivers/kafka/docker-compose.yml down
+        run: make -j db.kafka.stop db.destination.all.stop
```
ImDoubD-datazip: here kafka is not a db, but it is a source.
can we make it like: `olake.source.kafka.stop` and `olake.destination.all.stop`. here db seems miss leading.

.github/workflows/performance-test.yml
```diff
@@ -121,20 +122,21 @@ jobs:
...
+        path: ./tests/${{ matrix.driver }}/testdata/
         run-id: ${{ steps.last_run.outputs.id }}
         github-token: ${{ github.token }}
-    
+
     - name: Run Performance Tests
-      run: go test -v ./drivers/${{ matrix.driver }}/internal/... -timeout 0 -run 'Performance'
-    
+      working-directory: ./tests
+      run: go test -v ./${{ matrix.driver }}/... -timeout 0 -run 'Performance'
```
ImDoubD-datazip: shouldn't we have make command for this as well? or are we keeping it for local builds only? if yes, then we can add a single line comment here why we are not using make command here.

constants/constants.go
```diff
@@ -51,29 +52,38 @@ const (
...
+	CdcTimestamp = libconstants.CdcTimestamp
 )
 
-type DriverType string
+// The following identifiers live in lib/constants so the decoupled test modules under tests/
+// and olake share a single definition, without re-pathing every call site in the repo. They are
+// re-exported here so existing constants.<Name> usages in olake stay unchanged.
+//
+// TODO: unify these onto lib/constants directly in a later refactor. Deferred for now because the
+// change spans most of the codebase.
```
ImDoubD-datazip: can we add this comment, of unifying the lib/root constant on the top of the file itself.

also is this decoupling really that necessary right now, because if we are later gonna put common things in lib, then we can refactor fully at once on these common config, and util functions, which can be used by other services as well. since for now, there is very less code, so lib itself won't be required.

drivers/kafka/internal/testdata
```diff
@@ -0,0 +1 @@
+../../../tests/kafka/testdata
```
ImDoubD-datazip: this is not required right, we can change the test data path inside the integration tests (wherever execute query reads from, and can change the TestConfig directly.

drivers/mssql/internal/config_test.go
```diff
@@ -257,7 +257,7 @@ func TestConfig_URI(t *testing.T) {
 		{
 			name: "with jdbc url params",
 			config: func() *Config {
-				certs := testutils.GenerateTestCerts()
+				certs := lib.GenerateTestCerts()
```
ImDoubD-datazip: other drivers above don't require the certifications?

lib/constants/constants.go
ImDoubD-datazip, line 1: we can remove the lib folder changes for now, just create a small utils file in the tests and for now have whatever general functions required in test, have there. later we can do a revamp for the common functions and configs required by other services outside olake like this in lib.

lib/utils/typeutils/typeutils.go
```diff
@@ -0,0 +1,62 @@
...
+package typeutils
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+)
+
+// ReformatFloat64 converts a value to float64.
+func ReformatFloat64(v interface{}) (float64, error) {
```
ImDoubD-datazip: we can use RefromatValue function. You can check in type utils reformat.go

tests/db2/config.go
ImDoubD-datazip, line 1: I am expecting that all the config changes are as it is, earlier present in the driver side.

tests/kafka/config.go
```diff
@@ -0,0 +1,10 @@
+package kafka
+
+// Config holds the connection fields the integration tests read from source.json; it
+// duplicates the driver's Config connection fields.
+//
+// TODO: unify this with the driver config via autogenerated code from the proto definitions,
+// instead of hand-maintaining a copy here.
+type Config struct {
+	BootstrapServers string `json:"bootstrap_servers"`
+}
```
ImDoubD-datazip, line 10: we might not require the such small config struct to have there separate file for all the drivers. either we can have a single config struct for all the drivers, persisting whatever is required by each of them (as you wrote in the TODO, i think it will be small change only), or we can move these in the driver's util file only.

lib/utils/concurrent.go
```diff
@@ -0,0 +1,47 @@
...
+		})
+	}
+
+	return executor.Wait()
+}
+
+// RetryOnBackoff retries f up to attempts times with doubling backoff. This is a simplified copy
+// for the tests; the product version additionally logs and honours ErrNonRetryable, which need
+// olake-only packages the leaf lib can't import.
+func RetryOnBackoff(ctx context.Context, attempts int, sleep time.Duration, f func(ctx context.Context) error) (err error) {
```
ImDoubD-datazip: this function is currently present in the utils.go file, we can create a specific one for testing only. same with concurrent function.

.gitignore
```diff
@@ -15,7 +15,7 @@ olake-iceberg-java-writer.jar
...
 destination/iceberg/olake-iceberg-java-writer/dependency-reduced-pom.xml
-drivers/*/internal/testdata/logs
-drivers/*/internal/testdata/state.json
-drivers/*/internal/testdata/stats.json
-drivers/*/internal/testdata/streams.json
\ No newline at end of file
+tests/*/testdata/**/logs
+tests/*/testdata/**/state.json
+tests/*/testdata/**/stats.json
+tests/*/testdata/**/streams.json
```
ImDoubD-datazip, line 21: can there exist more than one nested state, stats, streams file the testdata directory? if not then why **?

Dockerfile
ImDoubD-datazip, line 1: have you tested the dockerfile image build of all the drivers, like it happens during release?

Dockerfile
```diff
@@ -63,6 +58,10 @@ COPY --from=builder /home/app/drivers/${DRIVER_NAME}/resources/spec.json /driver
 COPY --from=builder /home/app/destination/iceberg/resources/spec.json /destination/iceberg/resources/spec.json
 COPY --from=builder /home/app/destination/parquet/resources/spec.json /destination/parquet/resources/spec.json
 
+# Driver-specific runtime files staged by `make prepare.<driver>` (empty for most drivers)
```
ImDoubD-datazip: just add in comment, not empty for db2 due to clidriver

tests/go.work
ImDoubD-datazip, line 1: this is not required right, we can remove this.

release-tool.sh
ImDoubD-datazip, line 1: since we have done significant changes in the dockerfile and release.sh file, do you think we require this release.sh file? can we have the logic of pushing the build image directly from the dockerfile itself?

Makefile
```diff
@@ -262,7 +299,7 @@ $(ICEBERG_JAR): $(ICEBERG_JAR_SRCS)
 define DEV_BUILD_template
 .PHONY: dev.$(1).build
 dev.$(1).build: prepare.$(1)
-	$$(GO_ENV.$(1)) cd drivers/$(1) && go mod tidy && go build -ldflags="-w -s -X constants/constants.version=$$(GIT_VERSION) -X constants/constants.commitsha=$$(GIT_COMMITSHA) -X constants/constants.releasechannel=$$(RELEASE_CHANNEL)" -o olake main.go
+	export CGO_ENABLED=0; $$(GO_ENV.$(1)) cd drivers/$(1) && go mod tidy && go build -ldflags="-w -s -X constants/constants.version=$$(GIT_VERSION) -X constants/constants.commitsha=$$(GIT_COMMITSHA) -X constants/constants.releasechannel=$$(RELEASE_CHANNEL)" -o olake main.go
```
ImDoubD-datazip, line 310: here is `export CGO_ENABLED=0;` required? only in db2 we require CGO to be enabled, otherwise we don't require it right, and for db2, you have added in the driver make file right? here i think it won't be necessary then.

.github/workflows/integration-tests-runner.yml
```diff
@@ -168,51 +78,31 @@ jobs:
       # created iceberg JdbcCatalog race each other's CREATE TABLE on the shared catalog bookkeeping
       # tables (iceberg_tables, iceberg_namespace_properties). Bootstrap it once, serially, before
       # any tests run, so every later sync finds it initialized. The catalog is shared across all
-      # drivers, so one check (the first driver in the list) suffices.
+      # drivers, so one check with any single driver suffices.
       - name: Bootstrap Iceberg Catalog
```
ImDoubD-datazip, line 78: should we add a wait for destination step before this bootstrap step?

ImDoubD-datazip:
@mihir-datazip I experimented the drivers build with root go mod, and yes it is taking almost double the time as there are lot of dependencies and build happens twice (by go test and then go build). so keeping the tests separated is required with its own build.

.github/workflows/kafka-rebalance-tests.yml
mihir-datazip: @saksham-datazip can we reduce the size of this machine?

.github/workflows/kafka-rebalance-tests.yml
mihir-datazip: We use s3 as both source and destination... If we do source.s3.stop Will this also make the destination s3 stop??
Also if we are using different instances of source s3 and destination s3 it can be confusing maintaining 2 instances for namesake.

I can do only `olake.kafka.stop` `olake.s3.stop` if that is fine. these docker compose components might not always have 1:1 mapping with source destination...

.github/workflows/performance-test.yml
mihir-datazip: We can add it here

drivers/kafka/internal/testdata
mihir-datazip: This is for cases where olake-docs mention a command that points to the older path.
We can remove it once olake-docs is updated. For now we have to keep it

drivers/mssql/internal/config_test.go
mihir-datazip: I didn't see the logic for it... If we want to add more config tests for other drivers, I think it should be a separate PR

lib/constants/constants.go
mihir-datazip: There still will be a lib for absolute important cases like sharing constants.go and state_version.go for backward compatibility tests.
If you wish I can move the constants and configs to lib in this PR itself but that will create a lot more file changes. I can keep lib minimal but it has to share at least constants and state_version with current codebase

tests/db2/config.go
mihir-datazip: Yes they are untouched

lib/utils/typeutils/typeutils.go
mihir-datazip: Can't import utils from olake, this would add indirect dependencies of packages in the root go.mod

.gitignore
mihir-datazip: because they can be directly in `testdata -> logs` or in nested case like `kafke -> json -> testdata -> logs`, `kafka -> avro -> testdata -> logs`

Dockerfile
mihir-datazip: Yes tested

Dockerfile
mihir-datazip: Okay

.github/workflows/integration-tests-runner.yml
mihir-datazip: Got it

.github/workflows/integration-tests-runner.yml
mihir-datazip: Will update

.github/workflows/kafka-rebalance-tests.yml
mihir-datazip: Because when it is copied during docker build. We can get permission errors so bind mount needs to be owned.
This is same as before. I will remove this comment as it was before

.github/workflows/kafka-rebalance-tests.yml
mihir-datazip: Got it

.github/workflows/integration-tests-runner.yml
mihir-datazip: mod tidy runs during Docker build and does not need to run again and again as we mount the go cache.
This will be a no-op either way so we can remove it... but I prefer to have the control

.github/workflows/integration-tests-runner.yml
mihir-datazip: We can make it olake.all.stop as discussed above we cannot have 1:1 mapping of sources and docker-compose we run after having s3 as driver

constants/constants.go
mihir-datazip: We have to keep the constants, config and state ideally but it will cause a lot of changes.
I will remove the utils and shift it to test utils, but constants have to remain...

tests/kafka/config.go
mihir-datazip: Per driver config will still remain in their respective packages as code changes in a drivers config should not recompile other drivers tests...

lib/utils/concurrent.go
mihir-datazip: Yeah this will be handled in creating copies of functions removed from lib

tests/go.work
mihir-datazip: Okay

release-tool.sh
mihir-datazip: release.sh is still needed. Also this is not a good practice to tightly couple docker build logic with push.

Makefile
mihir-datazip: If we remove that and someone runs `make dev.build.db2` and then runs `make dev.build.postgres` that will create postgres with CGO which is not intended.
We can have a post step in db2 for disabling CGO but it someone manually cancels the db2 build before the CGO is disabled, we still face the same issue.
It is needed here

.github/workflows/integration-tests-runner.yml
mihir-datazip: Good catch!! That needs to be added actually.

.github/workflows/kafka-rebalance-tests.yml
mihir-datazip: We can keep this as this will be removed soon

.github/workflows/integration-tests-runner.yml
ImDoubD-datazip: lets remove this. if required can be added later.

mihir-datazip: Removed

.github/workflows/kafka-rebalance-tests.yml
ImDoubD-datazip: the change is good for now.

constants/constants.go
ImDoubD-datazip: let's keep the constants for now, we will later look for moving the common functionalities here, so that it can be used in various other services.

drivers/kafka/internal/testdata
ImDoubD-datazip: we have removed this I believe

mihir-datazip: Yes

lib/constants/constants.go
ImDoubD-datazip: so we have to move common utils out to lib as other outer services will require them, and since this will happen incrementally so we need to keep track of what has been moved where.

lib/utils/concurrent.go
ImDoubD-datazip: can you just check the retrybackoff function you have written, will it work same as the retrybackoff we have in utils?

release-tool.sh
ImDoubD-datazip: we can add a TODO later if possible.

ImDoubD-datazip (approved):
LGTM

PR #1035, opened 2026-07-15
fix: Preserve record order during concurrent filtering

### LINKED ISSUE #1014: Bug: Concurrent record filtering causes out-of-order data corruption
## Description
Concurrent filtering in `utils/typeutils/filter.go` scrambles record order. 
**Impact:** Processing CDC events out-of-order leads to data corruption (e.g., applying an older `UPDATE` after a newer one, or `DELETE` before `INSERT`).
## Steps to reproduce
1. Configure a pipeline with a filter.
2. Ingest a batch of ordered records.
3. Observe the output from `FilterRecords`—the records are shuffled due to non-deterministic goroutine execution and mutex locking.
## Expected behavior
Filtered records must strictly retain their original sequence.
## Actual behavior
Records are appended to the `filtered` slice randomly as goroutines complete.
## Proposed Solution
Instead of appending directly via a mutex, either:
1. Process records sequentially (safest/simplest).
2. Concurrently evaluate conditions into an index-mapped boolean array, then sequentially build the final filtered slice.

sahitya-datazip:
# Description

Preserves original order of records after concurrent filtering.
Previously records were appended to filtered slice as goroutines finished execution. Since goroutines complete in a non-deterministic order, the filtered output could have a different ordering than the input potentially resulting in out of order CDC  events.
This fix evaluates filter conditions concurrently while maintaining the order when reconstructing filtered records.

Fixes #1014

## Type of change

# How Has This Been Tested?

- [x] Tested locally with PostgreSQL CDC pipeline and Iceberg destination. 
- [x] Verified that records satisfying the filter conditions are returned in same order as input. 
- [x]  Added regression unit test (`TestFilterRecords_PreserveOrder`)
  
# Screenshots or Recordings

filter-bug

corrected-filter

## Documentation


vikaxsh:
please add unit test for order-assertion as well

vikaxsh (approved):

PR #1037, opened 2026-07-16
feat(tests): run integration tests against the real driver docker images

mihir-datazip:
# Description

This PR changes how the integration tests execute: instead of spinning up a disposable
`golang:bookworm` container that bind-mounts the whole repo and compiles each driver from
source with `build.sh` before running it, the tests now run the **real, shipped driver docker
images** (`olake/source-<driver>:local`) via a plain `docker run`, exercising the image's
actual `ENTRYPOINT` (`./olake`) exactly as a user would.

**Motivation:** the old harness tested source built on the fly inside a generic toolchain image,
so it never exercised the artifact we ship — the driver image, its runtime stage, its entrypoint,
its baked-in dependencies. Regressions in the image itself (missing runtime libs, entrypoint
wiring, platform issues) were invisible to CI. Running the tests against the same image users pull
closes that gap and, as a side effect, removes the per-test in-container `go build` (and the
`testcontainers-go` dependency stack) entirely.

**Key changes:**

- **New execution path (`tests/testutils/docker.go`).** `runOlake` runs
  `docker run --rm -v <testdata>:/testdata olake/source-<driver>:local <args>` and returns the
  container's exit code + combined output. `ensureDriverImage` builds the image via
  `make docker.build.<driver>` when it's missing locally (guarded by `sync.Once` so parallel tests
  trigger the slow build at most once); CI pre-builds it. `OLAKE_DRIVER_IMAGE` overrides the ref.
  `dockerRunArgs` / `syncArgs` / `discoverArgs` build the arg vectors; `dockerExitResult`
  normalizes a non-zero olake exit into a code rather than an error.
- **Config edits moved from in-container `jq`/shell to host-side Go (`tests/testutils/test_utils.go`).**
  `streams.json`, `state.json`, `iceberg_destination.json` are now edited directly on the host
  (`editJSONFile`, `updateSelectedStreams`, `updateStreamConfig`, `resetStateFile`,
  `saveStateFile`/`restoreStateFile`, `toggleArrowIcebergWrites`); the container sees the changes
  through the `/testdata` bind mount. `preSetupCommands []string` became `preSetup []func(*TestConfig) error`.
  `writeHostFile` unlinks-then-writes so the host user can overwrite files the (root) container left behind on Linux CI.
- **`TestConfig` restructured** into explicit host-side paths (read/written by the harness) vs.
  container-side paths (passed as olake args). Testdata now lives under
  `tests/<driver>/testdata[/<format>]`.
- **Removed the now-dead machinery:** `tests/testutils/exec.go` (`ExecCommand`), the `runInTestContainer`
  lifecycle-hook harness, and the `build.sh` timing-marker helpers (`logBuildRunTimings`) — there is
  no in-container build to time anymore.
- **`base.Dockerfile` + `Makefile`:** added a `runtime` stage (slim debian + JRE, kept in sync with
  the shipped `Dockerfile`) and `BASE_RUNTIME_IMAGE`; `docker.base.build` now builds it too, so a
  later `make docker.build.<driver>` reuses the cached apt layer instead of re-running it.
- **CI (`integration-tests-runner.yml`):** dropped the separate "Build Test Base Image" step. The
  "Bootstrap Iceberg Catalog" step now builds and runs the driver's real image
  (`make docker.build.<driver>` → `docker run … olake/source-<driver>:local check`), which doubles
  as warming the shared BuildKit go caches for the driver images the tests build lazily during the
  `-p 7` run.
- **`go.mod` slimdown across every `tests/*` module:** dropped the entire `testcontainers-go` /
  moby / docker / containerd / OpenTelemetry transitive dependency tree now that testcontainers is gone.
- **`tests/kafka/kafka_test.go`:** re-enabled `t.Parallel()` — the old `./olake: Text file busy`
  came from concurrent in-container recompiles; with a prebuilt image, parallel discover is safe.
- **`.dockerignore`:** exclude `destination/iceberg/local-test/data` (root/uid-owned compose runtime
  data) from the build context so `docker build .` no longer fails with "permission denied".

Fixes # (issue)

## Type of change

> Test / CI infrastructure change only — no shipped driver or product code behavior changes.

# How Has This Been Tested?

The full integration suite runs against the real driver images via the
`integration-tests-runner.yml` workflow (per-driver matrix, `-p 7`).

- [x] Scenario A — `TestIntegration` (Discover + Sync: Iceberg Legacy/Arrow + Parquet, Full-load / CDC / Incremental) against `olake/source-<driver>:local`
- [x] Scenario B — `Test2PCIntegration` (2PC CDC & Incremental state-recovery) and `TestRebalance` (Kafka) against the same images
- [x] Local run: `make docker.build.<driver>` (auto-triggered on first run) then `go test ./tests/<driver>/...`; optionally point at an existing image with `OLAKE_DRIVER_IMAGE=<ref>`

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):


tests/testutils/test_utils.go
```diff
@@ -60,23 +64,254 @@ type SyncSpeed struct {
...
-	HostTestDataPath       string
-	HostCatalogPath        string
-	HostTestCatalogPath    string
-	HostStatsPath          string
-	DataFormat             string
+}
+
+// applySuite moves every file a suite REWRITES into a per-suite directory so two suites for one
+// driver can run at once; read-only fixtures stay shared. A no-op for the integration suite.
+func applySuite(t *testing.T, c *TestConfig, suite string) {
```
ImDoubD-datazip, line 102: this separates the streams/state/stats/source per suite but leaves the destinations config files?
the issue which can be is the when through the t.parallel the general integration, 2pc, rebalance runs like in kafka, is the race b/w legacy and arrow, the flip of arrow writes can happen when 2pc integration test is running.

tests/testutils/test_utils.go
```diff
@@ -192,131 +428,184 @@ func GetTestConfig(driver string, extraParams ...string) *TestConfig {
...
-// pass flags as `--flag1, flag1 value, --flag2, flag2 value...`
-func discoverCommand(config TestConfig, flags ...string) string {
-	baseCmd := fmt.Sprintf("/test-olake/build.sh driver-%s discover --config %s", config.Driver, config.SourcePath)
-	if len(flags) > 0 {
-		baseCmd = fmt.Sprintf("%s %s", baseCmd, strings.Join(flags, " "))
-	}
-	return baseCmd
+// writeHostFile writes to the shared /testdata mount, unlinking first: the container runs as root,
+// so on Linux CI the test user cannot truncate a file a previous run left behind, only replace it.
+func writeHostFile(path string, data []byte, perm os.FileMode) error {
```
ImDoubD-datazip: when containers starting, config files may not exist briefly, as config edit passes through this function where remove happens first and then write file happens. this might give `error on entity` which is not related to the tests itself.

tests/testutils/test_utils.go
```diff
@@ -192,131 +428,184 @@ func GetTestConfig(driver string, extraParams ...string) *TestConfig {
...
+	_ = os.Remove(path)
```
ImDoubD-datazip, line 538: also should we check for error as well, as the path might not exist due to some early code error in test or any race if occurs?

destination/iceberg/java_client.go
```diff
@@ -141,6 +142,16 @@ func writeGCPCredsTempFile(credsJSON string) (string, error) {
...
+// sharedArchivePath returns the AppCDS archive next to jarPath (the driver image bakes one in), or
+// "" when there is none. Callers pair it with -Xshare:on, so a stale archive fails loudly.
+func sharedArchivePath(jarPath string) string {
+	archive := strings.TrimSuffix(jarPath, filepath.Ext(jarPath)) + ".jsa"
+	if info, err := os.Stat(archive); err != nil || info.Size() == 0 {
+		return ""
+	}
+	return archive
+}
+
```
ImDoubD-datazip: why is this needed?

destination/iceberg/java_client.go
```diff
@@ -171,21 +182,19 @@ func startServer(config *Config) (*serverInstance, error) {
...
-	} else {
-		serverCmd = exec.Command("java",
-			"-XX:+UseG1GC",
-			"-XX:MaxRAMPercentage=75.0",
-			"-XX:+ExitOnOutOfMemoryError",
-			"-jar", config.JarPath, string(configJSON))
+		javaArgs = append(javaArgs, "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005")
 	}
+	javaArgs = append(javaArgs, "-jar", config.JarPath, string(configJSON))
+	serverCmd := exec.Command("java", javaArgs...)
```
ImDoubD-datazip: these changes have to be confirmed with Vikash once, as he wrote the single jvm instance code and you have done changes on the same code.

destination/iceberg/java_client.go
```diff
@@ -171,21 +182,19 @@ func startServer(config *Config) (*serverInstance, error) {
...
 
 	// need to do some research on the following flags
-	var serverCmd *exec.Cmd
+	javaArgs := []string{
+		"-XX:+UseG1GC",
+		"-XX:MaxRAMPercentage=75.0",
+		"-XX:+ExitOnOutOfMemoryError",
+	}
+	if archive := sharedArchivePath(config.JarPath); archive != "" {
+		javaArgs = append(javaArgs, "-XX:SharedArchiveFile="+archive, "-Xshare:on")
```
ImDoubD-datazip: also shouldn't this be `-Xshare:auto` instead of `-Xshare:on`, as this might fail when not mappable. For example JDK patch bump in the base image, a different arch, or different heap/GC flags all make the JVM refuse to start so during this time, so iceberg writer might not boot, in this case fall back should happen that class must be loaded normally, which will be possible by xShare auto.

Dockerfile
```diff
@@ -48,6 +50,14 @@ ENV DRIVER_VERSION=${DRIVER_VERSION}
 # First try to copy from the source location (works after Maven build)
 COPY destination/iceberg/olake-iceberg-java-writer/target/olake-iceberg-java-writer-0.0.1-SNAPSHOT.jar /home/olake-iceberg-java-writer.jar
 
+# Pre-load the Iceberg writer's startup classes into an AppCDS archive.
+RUN out=$(java -XX:+UseG1GC -XX:ArchiveClassesAtExit=/home/olake-iceberg-java-writer.jsa \
+      -jar /home/olake-iceberg-java-writer.jar '{' 2>&1 || true); \
+    echo "$out" | grep -q 'OlakeRpcServer.main' || { \
+      echo 'AppCDS dump never reached OlakeRpcServer.main - is the jar shaded?' >&2; \
+      echo "$out" | head -20 >&2; exit 1; }; \
+    test -s /home/olake-iceberg-java-writer.jsa
```
ImDoubD-datazip: why is this required? can't we load the iceberg writer's startup classes everytime? how much of a difference does it makes?

Dockerfile
```diff
@@ -48,6 +50,14 @@ ENV DRIVER_VERSION=${DRIVER_VERSION}
 # First try to copy from the source location (works after Maven build)
 COPY destination/iceberg/olake-iceberg-java-writer/target/olake-iceberg-java-writer-0.0.1-SNAPSHOT.jar /home/olake-iceberg-java-writer.jar
 
+# Pre-load the Iceberg writer's startup classes into an AppCDS archive.
+RUN out=$(java -XX:+UseG1GC -XX:ArchiveClassesAtExit=/home/olake-iceberg-java-writer.jsa \
+      -jar /home/olake-iceberg-java-writer.jar '{' 2>&1 || true); \
+    echo "$out" | grep -q 'OlakeRpcServer.main' || { \
+      echo 'AppCDS dump never reached OlakeRpcServer.main - is the jar shaded?' >&2; \
+      echo "$out" | head -20 >&2; exit 1; }; \
```
ImDoubD-datazip: also this returns stderr (exit 1), which might fail the build. is this necessary?

Dockerfile
```diff
@@ -32,6 +32,8 @@ RUN apt-get update && \
     ca-certificates \
     libpam-modules \
     libcrypt1 \
+    iproute2 \
+    lsof \
```
ImDoubD-datazip, line 48: why this is required? where is it used?

utils/ticker.go
ImDoubD-datazip, line 1: this lib folder itself is not required now. what we had originally the ticker timing for initial lsn, i think 3 sec, we can stick to that only. just for that 100 ms ticking, we have introduced this, which i think is not necessary,

pkg/waljs/replicator.go
```diff
@@ -159,7 +160,7 @@ func AcknowledgeLSN(ctx context.Context, db *sqlx.DB, socket *Socket, fakeAck bo
 	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
 	defer cancel()
 
-	ticker := time.NewTicker(3 * time.Second)
+	ticker := libutils.NewTicker(100*time.Millisecond, 3*time.Second)
```
ImDoubD-datazip: let's stick to `time.NewTicker(3 * time.Second)` only.

tests/testutils/test_utils.go
```diff
@@ -192,131 +428,184 @@ func GetTestConfig(driver string, extraParams ...string) *TestConfig {
...
+	}
+
+	var filter interface{} = map[string]interface{}{}
+	if filterConfig != "" {
+		if err := json.Unmarshal([]byte(filterConfig), &filter); err != nil {
+			return fmt.Errorf("failed to parse filter config: %s", err)
+		}
+	}
+
+	return editJSONFile(config.HostCatalogPath, func(doc map[string]interface{}) error {
```
ImDoubD-datazip, line 566: here catalog rewrite happen, any mismatch or validation issue can they occur here? Or it will be always correct?

tests/testutils/test_utils.go
```diff
@@ -192,131 +428,184 @@ func GetTestConfig(driver string, extraParams ...string) *TestConfig {
...
+							if fmt.Sprint(col) != columnToExclude {
+								remaining = append(remaining, col)
+							}
+						}
+						selectedColumns["columns"] = remaining
+					}
+				}
+			}
+			kept = append(kept, stream)
+		}
```
ImDoubD-datazip, line 592: is this the only way for updation of streams.json, iterate through the map (even in nested)? Is there a better way to do this?

tests/testutils/test_utils.go
```diff
@@ -1093,148 +1354,88 @@ func (cfg *IntegrationTest) testIceberg2PCIncrementalRecovery(
...
 
-			if !slices.Contains(constants.SkipCDCDrivers, constants.DriverType(cfg.TestConfig.Driver)) {
-				for _, wt := range writerTypes {
-					t.Run(fmt.Sprintf("Iceberg (%s) 2PC CDC Recovery tests", wt.name), func(t *testing.T) {
-						if err := cfg.testIcebergWriter(ctx, t, c, currentTestTable, wt.useArrow, cfg.testIceberg2PCCDCRecovery); err != nil {
-							t.Fatalf("Iceberg (%s) 2PC CDC Recovery tests failed: %v", wt.name, err)
-						}
-					})
-				}
+		if err := updateSelectedStreams(cfg.TestConfig, cfg.Namespace, cfg.PartitionRegex, cfg.FilterConfig, []string{currentTestTable}, cfg.ColumnToExclude); err != nil {
```
ImDoubD-datazip, line 1443: if it fails, then we return fatal statement, shouldn't we retry this process `updateSelectedStreams` wherever called as it updates the streams.json?

tests/testutils/docker.go
```diff
@@ -1,73 +1,132 @@
...
+		args = append(args, "--platform", cfg.ImagePlatform)
+	}
+	args = append(args, extraFlags...)
+	args = append(args, driverImageRef(cfg.Driver))
+	return append(args, olakeArgs...)
+}
+
+// runOlake runs the driver image once as a user would and returns the container's exit code and
+// combined output. err is non-nil only when docker itself fails to launch.
+func runOlake(ctx context.Context, t *testing.T, cfg *TestConfig, olakeArgs ...string) (int, []byte, error) {
```
ImDoubD-datazip, line 83: no cancellable context, since context.background is passed from `TestIntegration` function. So if ci is cancelled or test panicked or in local if command + c is clicked, then sync container still keeps on running.

tests/testutils/test_utils.go
```diff
@@ -1296,205 +1495,180 @@ func (cfg *IntegrationTest) testKafkaRebalance(
...
-		})
+var (
+	sharedSparkOnce sync.Once
+	sharedSpark     sql.SparkSession
+	sharedSparkErr  error
+)
+
+// sparkSession returns the shared Spark Connect session, building it on first use and warming it
+// so the one-off server bootstrap is timed here instead of inflating whichever verify runs first.
+func sparkSession(ctx context.Context, t *testing.T) (sql.SparkSession, error) {
```
ImDoubD-datazip, line 1750: here i am believing the fact that the spark session here is goroutine safe as Integration, 2PC and Rebalance call the spark sql concurrently.

tests/testutils/test_utils.go
```diff
@@ -192,131 +428,184 @@ func GetTestConfig(driver string, extraParams ...string) *TestConfig {
...
+	if err := edit(doc); err != nil {
+		return err
 	}
-	return baseCmd
+	out, err := json.MarshalIndent(doc, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal %s: %s", path, err)
+	}
+	return writeHostFile(path, out, 0644)
 }
```
ImDoubD-datazip, line 533: here if there any number inside the json provided, that will explicitly turn to float64 right? Can we use json decoder for them so they retain what they originally were?

tests/testutils/test_utils.go
```diff
@@ -192,131 +428,184 @@ func GetTestConfig(driver string, extraParams ...string) *TestConfig {
...
-		baseCmd = fmt.Sprintf("%s %s", baseCmd, strings.Join(flags, " "))
+	if err := edit(doc); err != nil {
+		return err
 	}
-	return baseCmd
+	out, err := json.MarshalIndent(doc, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal %s: %s", path, err)
+	}
+	return writeHostFile(path, out, 0644)
```
ImDoubD-datazip: we should not pass the file permissions directly like this, the gosec will fail this. we should keep a default file permission constants and call this where ever the file permissions are provided.

tests/testutils/constants/constants.go
```diff
@@ -28,6 +28,10 @@ const (
 // CdcTimestamp is the column name olake writes the CDC event timestamp into.
 const CdcTimestamp = "_cdc_timestamp"
 
+// LatestStateVersion is the state file format olake writes today. A state file without it reads as
+// version 0, which puts the sync on legacy type mapping -- so the harness has to name the version.
+const LatestStateVersion = 6
```
ImDoubD-datazip: is this now required? shouldn't latest always be picked and for backward compatibility, all versions must be checked?

tests/postgres/postgres_util_test.go
```diff
@@ -266,6 +269,18 @@ func ExecuteQuery(ctx context.Context, t *testing.T, streams []string, operation
...
+		// streams[0] is the suite's replication slot. Drop a stale inactive slot left by a local
+		// rerun (fresh CI containers have none), then create it; falls through to the exec below.
+		_, _ = db.ExecContext(ctx, fmt.Sprintf(`SELECT pg_drop_replication_slot(slot_name) FROM pg_replication_slots WHERE slot_name = '%s' AND NOT active`, streams[0]))
+		query = fmt.Sprintf(`SELECT pg_create_logical_replication_slot('%s', 'pgoutput')`, streams[0])
+
+	case "drop-slot":
+		// Best-effort cleanup: never fail the suite on slot teardown (the CI container is discarded
+		// anyway). Skips a still-active slot; the next run's create-slot drops it once inactive.
+		_, _ = db.ExecContext(ctx, fmt.Sprintf(`SELECT pg_drop_replication_slot(slot_name) FROM pg_replication_slots WHERE slot_name = '%s' AND NOT active`, streams[0]))
+		return
```
ImDoubD-datazip: like the olake_slot, can we create another slot when we start the postgres container? this is used in 2pc so we can crate olake_slot_2pc when docker compose of postgres is up. these won't be needed then.

utils/telemetry/telemetry.go
```diff
@@ -56,8 +70,7 @@ type LocationInfo struct {
 func Init() {
 	go func() {
 		// check for disable
-		disabled, _ := strconv.ParseBool(os.Getenv("TELEMETRY_DISABLED"))
-		if disabled {
+		if Disabled() {
```
ImDoubD-datazip, line 73: is all of this required? can't we simply check for the telemetry if disabled before starting the go func?
```
func Init() {
		disabled, _ := strconv.ParseBool(os.Getenv("TELEMETRY_DISABLED"))
		if disabled {
			return
		}
        go func() {
		    ip := getOutboundIP()
		    telemetry = &Telemetry{
                  .....
```

tests/kafka/kafka_util_test.go
ImDoubD-datazip: is this required now?

.github/workflows/integration-tests.yml
```diff
@@ -1,21 +1,317 @@
...
+  preflight:
+    name: Detect changes
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      drivers: ${{ steps.drivers.outputs.drivers }}
+      testutils: ${{ steps.drivers.outputs.testutils }}
```
ImDoubD-datazip: is this required?

.github/workflows/integration-tests.yml
```diff
@@ -1,21 +1,317 @@
...
+      - 'tests/**'
+      - '.github/workflows/integration-tests.yml'
+      - '.github/actions/detect-drivers/action.yml'
 
 jobs:
+  # Ungated, like the three cache jobs below it: the environment prompts once per wave of jobs that
+  # reach it, so gating these too would cost three approvals a run. Cheap: a file list and 3 lookups.
+  preflight:
+    name: Detect changes
+    runs-on: ubuntu-latest
```
ImDoubD-datazip, line 34: here ubuntu-latest is used? this stage does not require high processing then right?

Makefile
```diff
@@ -307,22 +310,56 @@ iceberg.jar: $(ICEBERG_JAR)
...
-	@echo "Built drivers/$(1)/olake (version $$(GIT_VERSION), commit $$(GIT_COMMITSHA))"
+	export CGO_ENABLED=0; $$(GO_ENV.$(1)) cd drivers/$(1) && go build -o olake main.go
+	@echo "Built drivers/$(1)/olake"
 endef
 $(foreach d,$(DRIVERS),$(eval $(call DEV_BUILD_template,$(d))))
 
 # --- tests --------------------------------------------------------------------
+# Concurrency of the aggregate targets below: one job per driver by default, since every suite
+# drives its own database stack plus an olake container per sync. CI overrides it (TEST_JOBS=5).
+TEST_JOBS ?=
```
ImDoubD-datazip: what is this required for? is this set to 5? what does 5 mean here?

base.Dockerfile
ImDoubD-datazip, line 1: what's the benefit of keeping the olake base image stage in separate dockerfile? it was in main dockerfile, so no extra dockerfile was required to be managed.

.dockerignore
```diff
@@ -2,3 +2,10 @@
 test/
 destination/iceberg/local-test/data/
 tests/
+drivers/*/olake
+# Other gitignored local artifacts (native-run outputs, debugger binaries,
+# editor and Finder junk).
+olake-data
+**/__debug_bin*
+**/.DS_Store
+.vscode
```
ImDoubD-datazip, line 11: required?

.github/actions/detect-drivers/action.yml
```diff
@@ -0,0 +1,60 @@
...
+            case " $known " in *" $d "*) continue ;; esac
+            selected="$tested"
+            break
+          done
+        fi
+
+        drivers=$(printf '%s\n' $selected | sort -u | jq -Rc '[., inputs] | map(select(. != ""))')
+        echo "drivers=$drivers" >> "$GITHUB_OUTPUT"
+        echo "testutils=$testutils" >> "$GITHUB_OUTPUT"
+        echo "Affected drivers: $drivers (harness changed: $testutils)"
```
ImDoubD-datazip: ideally we should have a separate sh file for this and reference it here, because if any changes done in shell, it will be easy to understand and debug.

utils/logger/timing.go
ImDoubD-datazip, line 1: why is this required? if required, this should be kept in integration tests only, as per events like discover time taken total and sync time taken total. main code should not be changed for integration tests.

protocol/discover.go
```diff
@@ -69,11 +69,13 @@ var discoverCmd = &cobra.Command{
 		types.LogCatalog(streams, catalog, connector.Type())
 
 		// Discover Telemetry Tracking
-		defer func() {
-			telemetry.TrackDiscover(len(streams), connector.Type())
-			logger.Infof("Discover completed, wait 5 seconds cleanup in progress...")
-			time.Sleep(5 * time.Second)
-		}()
+		if !telemetry.Disabled() {
```
ImDoubD-datazip, line 73: no need to add telemetry disabled check for every event out there, just add in telemetry init when set in root.

protocol/sync.go
```diff
@@ -127,11 +127,17 @@ var syncCmd = &cobra.Command{
 		// Build the writer pool up front: it starts destination-owned resources
 		// (e.g. the Iceberg shared JVM) and validates the connection. pool.Close
 		// tears them down on exit (normal return or signal-canceled context).
+		stopInit := logger.TrackTiming("sync", "writer pool init")
```
ImDoubD-datazip, line 130: can we remove this track timing thing from everywhere. i don't think it is required as it will be tracking time in testing environment, we should not alter the main code for testing.

tests/testutils/docker.go
```diff
@@ -1,73 +1,132 @@
...
-	require.NoError(t, err, "read go.mod to derive the base image tag")
-	f, err := modfile.Parse(workFile, data, nil)
-	require.NoError(t, err, "parse go.mod to derive the base image tag")
-	require.NotNil(t, f.Go, "no go directive found in %s", workFile)
-	return "olakego/base:build-go" + f.Go.Version
+	ref := driverImageRef(cfg.Driver)
+	if os.Getenv(driverImageEnvVar) != "" {
+		return ref
+	}
+	ensureImageOnce.Do(func() {
```
ImDoubD-datazip, line 49: is this required? because now every driver has its own ci, so does the sync once required here. each driver will be build in its own vm.

.github/actions/detect-drivers/action.yml
ImDoubD-datazip, line 1: this detects which driver file have been changed, and runs the integration tests for that driver. if any other file are changed, then every driver integration tests runs. 
rather than detecting file changes, can we simplify it like when approval gate for integration tests come on commits, then we can multi select the driver we want to run the integration tests for. won't this make things very simplified.

also for common files, like if destination is changed, then running all driver against destination might not be required, you can test this from one driver as well, so asking which driver test to run is much more simplified i think.

destination/writers.go
```diff
@@ -105,7 +105,10 @@ func NewWriterPool(ctx context.Context, config *types.WriterConfig, syncStreams
 		batchSize:  batchSize,
 	}
 
-	if err := adapter.Check(ctx); err != nil {
+	stopCheck := logger.TrackTiming("destination", "check")
```
ImDoubD-datazip, line 110: no need let's remove this.

tests/testutils/timing.go
ImDoubD-datazip, line 1: no need for a separate file for this, can add in tests_utils only, with a simple comment of what it does.

utils/logger/logger.go
```diff
@@ -363,8 +363,14 @@ func SetupProcessLogger(processName string) (*ProcessOutputReader, *ProcessOutpu
 // SetupAndStartProcess creates and starts a process with stdout and stderr logged via the logger.
 // It handles the complete process lifecycle including starting the command and managing pipes.
 func SetupAndStartProcess(processName string, cmd *exec.Cmd) error {
+	// Readiness is observed by a reader goroutine, so this span is "child ready AND a reader got
+	// scheduled to notice"; the phases below separate that from the child's own startup cost.
+	defer TrackTiming(processName, "setup+start total")()
```
ImDoubD-datazip, line 368: i don't think all these changes are required, if only for testing , then this can be removed. let's not change the main code files for testing until absolute necessity.

protocol/sync.go
```diff
@@ -142,15 +148,20 @@ var syncCmd = &cobra.Command{
...
 		// Sync Telemetry tracking
-		telemetry.TrackSyncStarted(syncID, selectedStreamsMetadata.SelectedStreams, selectedStreamsMetadata.FullLoadStreams, selectedStreamsMetadata.CDCStreams, connector.Type(), destinationConfig, catalog)
-		defer func() {
-			stats := pool.GetStats()
-			telemetry.TrackSyncCompleted(syncID, err == nil, stats.ReadCount.Load(), stats.BytesRead.Load())
-			logger.Infof("Sync completed, wait 5 seconds cleanup in progress...")
-			time.Sleep(5 * time.Second)
-		}()
 
+		if !telemetry.Disabled() {
```
ImDoubD-datazip, line 153: no need to add telemetry disabled check for every event out there, just add in telemetry init when set in root. we can revert this back to what it was.

tests/testutils/test_utils.go
```diff
@@ -60,23 +64,254 @@ type SyncSpeed struct {
...
+			for _, entry := range entries {
+				selected, ok := entry.(map[string]interface{})
+				if !ok {
+					continue
+				}
+				out[fmt.Sprintf("%v.%v", namespace, selected["stream_name"])] = selected
+			}
+		}
+		return out
+	}
```
ImDoubD-datazip, line 250: these map parsing seems lot of work, can;t we simplify this? are there any other ways to do these things?

ImDoubD-datazip (commented):
I think a good part of the code is ai written, which I think at this instant of review is not self-reviewed. we should self review first line by line, remove all unnecessary code which are not required, or don't add any value of such in the codebase. cc @mihir-datazip

ImDoubD-datazip:
I think more reviews from other maintainers will be required as well, as a lot of those code changes has happened which other people wrote. cc @mihir-datazip @vaibhav-datazip @saksham-datazip @vikaxsh

tests/testutils/test_utils.go
mihir-datazip: Makes sense,  as our configs should be read only...
I will add a change that will separate out the destination configs instead of the toggle function we have atm

tests/testutils/test_utils.go
mihir-datazip: Separated out the config files to read only and each suite has its own set of configs so the framework remains race free on config updates

tests/testutils/test_utils.go
mihir-datazip: File not being present is also expected that is why error is intentionally ignored. We can log this error instead

destination/iceberg/java_client.go
mihir-datazip: Because the go code can run without AppCDS. That is a Dockerfile optimization

destination/iceberg/java_client.go
mihir-datazip: This have ~30sec benefit per IT run... If too complex then I can drop it if it touches more sensitive parts of code

destination/iceberg/java_client.go
mihir-datazip: Our clients doesn't build the docker image we do so it should not be that high of a blast radius to our customers, but we if we are planning to remove this optimization all together then it shouldn't matter

Dockerfile
mihir-datazip: We still load all the classes... AppCDS just caches the metadata to cut the redundant latency by half. Each JVM boot will take ~1 sec in CI if we don't include it...

Dockerfile
mihir-datazip: The point of the optimisation is lost if we don't hard fail...

Dockerfile
mihir-datazip: This is used in `java_client.go`
```
	// Forcefully kill any existing process on this port before starting
	reclaimPort(port)
```
This will fail silently if lsof and iproute is not present.

utils/ticker.go
mihir-datazip: I still think it causes substantial improvement. Below are the times taken by old ticker vs New ticker

## Old Ticker (3sec)
```
=== NAME  TestPostgresIntegration/Sync
    timing.go:34: [timing] postgres query "drop": 6ms
    test_utils.go:1647: postgres sync test cleanup
--- PASS: TestPostgresIntegration (76.42s)
    --- PASS: TestPostgresIntegration/Discover (0.98s)
    --- PASS: TestPostgresIntegration/Sync (75.44s)
        --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_CDC_tests (22.48s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_CDC_tests/Full-Refresh (6.02s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_CDC_tests/CDC_-_insert (5.36s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_CDC_tests/CDC_-_update (5.45s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_CDC_tests/CDC_-_delete (5.60s)
        --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_CDC_tests (22.92s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_CDC_tests/Full-Refresh (6.19s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_CDC_tests/CDC_-_insert (5.60s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_CDC_tests/CDC_-_update (6.03s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_CDC_tests/CDC_-_delete (5.07s)
        --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_CDC_tests (14.12s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_CDC_tests/Full-Refresh (3.63s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_CDC_tests/CDC_-_insert (3.54s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_CDC_tests/CDC_-_update (3.63s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_CDC_tests/CDC_-_delete (3.32s)
        --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_Incremental_tests (6.60s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_Incremental_tests/Full-Refresh (2.56s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_Incremental_tests/Incremental_-_insert (2.04s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_Incremental_tests/Incremental_-_update (1.98s)
        --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_Incremental_tests (6.91s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_Incremental_tests/Full-Refresh (2.33s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_Incremental_tests/Incremental_-_insert (2.07s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_Incremental_tests/Incremental_-_update (2.50s)
        --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_Incremental_tests (1.51s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_Incremental_tests/Full-Refresh (0.60s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_Incremental_tests/Incremental_-_insert (0.46s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_Incremental_tests/Incremental_-_update (0.44s)
        --- PASS: TestPostgresIntegration/Sync/Parquet_Rolling (0.86s)
PASS
ok      github.com/datazip-inc/olake/tests/postgres     76.889s
```


## New Ticker(100ms initial, then 3sec)
```
--- PASS: TestPostgresIntegration (43.72s)
    --- PASS: TestPostgresIntegration/Discover (1.39s)
    --- PASS: TestPostgresIntegration/Sync (42.33s)
        --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_CDC_tests (12.07s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_CDC_tests/Full-Refresh (3.74s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_CDC_tests/CDC_-_insert (2.67s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_CDC_tests/CDC_-_update (3.07s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_CDC_tests/CDC_-_delete (2.56s)
        --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_CDC_tests (10.10s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_CDC_tests/Full-Refresh (2.88s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_CDC_tests/CDC_-_insert (2.78s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_CDC_tests/CDC_-_update (2.38s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_CDC_tests/CDC_-_delete (2.03s)
        --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_CDC_tests (2.39s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_CDC_tests/Full-Refresh (0.66s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_CDC_tests/CDC_-_insert (0.47s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_CDC_tests/CDC_-_update (0.63s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_CDC_tests/CDC_-_delete (0.58s)
        --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_Incremental_tests (7.19s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_Incremental_tests/Full-Refresh (2.84s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_Incremental_tests/Incremental_-_insert (2.12s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Legacy)_Full_load_+_Incremental_tests/Incremental_-_update (2.19s)
        --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_Incremental_tests (7.33s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_Incremental_tests/Full-Refresh (2.46s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_Incremental_tests/Incremental_-_insert (2.09s)
            --- PASS: TestPostgresIntegration/Sync/Iceberg_(Arrow)_Full_load_+_Incremental_tests/Incremental_-_update (2.75s)
        --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_Incremental_tests (2.25s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_Incremental_tests/Full-Refresh (1.04s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_Incremental_tests/Incremental_-_insert (0.47s)
            --- PASS: TestPostgresIntegration/Sync/Parquet_Full_load_+_Incremental_tests/Incremental_-_update (0.49s)
        --- PASS: TestPostgresIntegration/Sync/Parquet_Rolling (0.97s)
PASS
ok      github.com/datazip-inc/olake/tests/postgres     44.165s
```

There is a clear 45% improvement in just integration tests due to this change. So I would definitely classify this change as important.

pkg/waljs/replicator.go
mihir-datazip: Addressed my concerns above... I think we should use it

tests/testutils/test_utils.go
mihir-datazip: Yes, deliberately we have no lib now so we just have to either have copy of the structure in tests which will not be a good solution. Regarding the concerns over nested... I don't think it will ever be slow(even milliseconds). If you want to refactor it, we can but that will be a purely cosmetic change.

tests/testutils/test_utils.go
mihir-datazip: Every failure mode of updateSelectedStreams is deterministic. Retry will not benefit here

tests/testutils/docker.go
mihir-datazip: You are confusing meaning of context here... An attached docker run proxies SIGINT to the container and --rm reaps it and CI cancellation goes through the same SIGINT.
The context is for Docker client, cancelling the client doesn't kills the container anyways.

I will add the test context but we still are left with some temporary containers until sync happens.

tests/testutils/test_utils.go
mihir-datazip: Yes it is concurrent safe
- sparkSessionImpl is {sessionId string, client} — immutable after construction. The client struct (sparkConnectClientImpl) likewise: all four fields set at construction, never written after.
- The only mutable state in the execution path (done, schema, the response stream) lives on ExecutePlanClient, which is created fresh per query — each Sql() call gets its own operation ID and stream object, so nothing mutable is ever shared between goroutines.
- The transport underneath is a gRPC ClientConn, which is documented goroutine-safe and multiplexes concurrent streams.
- Construction itself is sharedSparkOnce-guarded, so concurrent first-use from Integration/2PC/Rebalance is also safe.

utils/logger/logger.go
mihir-datazip: I still think we should include these...

protocol/sync.go
mihir-datazip: This needs due to bad design of the telemetry functions. If it is disable we don't need have sleep logic.
I can fix this by having a channel based wait... but I would still argue that exposing disabled check is helpful

tests/testutils/test_utils.go
mihir-datazip: The indexing is the cheap part — the payoff is that compare then does whole-entry comparison, so the test covers every field discover produces including ones added later. A typed struct would narrow the comparison to declared fields and silently absorb contract changes, which is the opposite of what this verification exists for

Dockerfile
ImDoubD-datazip: yes, this `reclaimPort(port)` is dead in prod for now. let's add these.

utils/telemetry/telemetry.go
ImDoubD-datazip: we need to remove 5 sec sleep after every discover and sync  telemetry call.

Makefile
ImDoubD-datazip: lets remove this.

tests/testutils/test_utils.go
mihir-datazip: Added

tests/testutils/test_utils.go
mihir-datazip: Will add it

Dockerfile
mihir-datazip: We can remove APP CDS optimization all together and move to a separate PR

tests/testutils/constants/constants.go
mihir-datazip: Keeping it....

tests/kafka/kafka_util_test.go
mihir-datazip: Removed it

.github/workflows/integration-tests.yml
mihir-datazip: Yes it is more than enough for this

Makefile
mihir-datazip: Removed

.dockerignore
mihir-datazip: Yes

.github/actions/detect-drivers/action.yml
mihir-datazip: Added a bash file

utils/logger/timing.go
mihir-datazip: We should keep these logs, they help a lot with performance observability

protocol/discover.go
mihir-datazip: Had to add it for the sleep

protocol/sync.go
mihir-datazip: As discussed we can keep these logs

tests/testutils/docker.go
mihir-datazip: As Tests run in parallel for drivers, we should build image once only

.github/actions/detect-drivers/action.yml
mihir-datazip: This was discussed and is the intended behaviour of CI

tests/testutils/timing.go
mihir-datazip: Separate files are fine... tests_utils is already bit large IMO

utils/ticker.go
```diff
@@ -0,0 +1,67 @@
...
+		C:       ch,
+		stop:    stop,
+		stopped: stopped,
+	}
+}
+
+func (t *Ticker) Stop() {
+	close(t.stop)
+	<-t.stopped
+}
```
ImDoubD-datazip: if concurrently called let's say 2 times, won't this panic? we can use sync.Once or select based close?

utils/ticker.go
```diff
@@ -0,0 +1,67 @@
...
+
+func NewTicker(initialDelay, interval time.Duration) *Ticker {
+	ch := make(chan time.Time, 1)
+	stop := make(chan struct{})
+	stopped := make(chan struct{})
+
+	go func() {
+		defer close(ch)
+		defer close(stopped)
```
ImDoubD-datazip: here close stopped will run first then close ch right, so if Stop() unblocks immediately as stopped closes. what can happen is ticker consumer checking C might still think its open. i think it should be swapped, first defer of stopped and then c.

tests/testutils/constants/constants.go
ImDoubD-datazip: we will change this later then.

tests/postgres/postgres_util_test.go
```diff
@@ -76,7 +76,10 @@ func ExecuteQuery(ctx context.Context, t *testing.T, streams []string, operation
 				col_point POINT,
 				col_polygon POLYGON,
 				col_circle CIRCLE,
-				CONSTRAINT unique_custom_key UNIQUE (col_bigserial),
+				-- Unnamed so Postgres derives a table-qualified index name
+				-- (<table>_col_bigserial_key); a fixed name like "unique_custom_key" is
+				-- schema-scoped and collides when concurrent suites create their tables at once.
+				UNIQUE (col_bigserial),
```
ImDoubD-datazip, line 85: nice finding.

tests/testutils/constants/constants.go
```diff
@@ -28,6 +28,10 @@ const (
 // CdcTimestamp is the column name olake writes the CDC event timestamp into.
 const CdcTimestamp = "_cdc_timestamp"
 
+// LatestStateVersion is the state file format olake writes today. A state file without it reads as
+// version 0, which puts the sync on legacy type mapping -- so the harness has to name the version.
```
vaibhav-datazip, line 32: the tests will pass if old version and backward compatibility is done, what if latest version here is not upgraded ?

tests/kafka/kafka_util_test.go
```diff
@@ -521,6 +530,40 @@ func encodeAndWriteAvro(ctx context.Context, t *testing.T, writer *kgo.Client, c
...
+		require.NoError(t, verr, "failed to validate-create while waiting for topic '%s' deletion", topic)
+		select {
+		case <-waitCtx.Done():
+			t.Fatalf("topic %q deletion did not complete within %s", topic, topicAdminTimeout)
+		case <-time.After(topicPollInterval):
+		}
+	}
+}
+
 // JSON data format resources
```
saksham-datazip, line 561: Their are too many if else condition so ig we can avoid that can you please check it for once please

tests/kafka/kafka_util_test.go
```diff
@@ -73,6 +71,7 @@ var (
 	// rebalance trigger
 	rebalanceTriggerCancel context.CancelFunc
 	rebalanceTriggerDone   chan struct{} // closed when the trigger goroutine has fully exited
+	rebalanceSyncStatsPath string
 
 	// JSON
```
saksham-datazip: Regarding this approach where we are defining this variable and populating inside kafka_test.go and than again using inside kafka_test_utils.go seems kinda workaround so can we please remove it and if needed we can simply create a function directly in test_utils.go?

saksham-datazip (commented):
As per our discussion can you please have a discussion and let me know about that regarding discover command checks ?
Only checked Kafka_test_utils.go and kafka_test.go ,please let me know if i have to check for some other files as well.

mihir-datazip:
> As per our discussion can you please have a discussion and let me know about that regarding discover command checks ? Only checked Kafka_test_utils.go and kafka_test.go ,please let me know if i have to check for some other files as well.

Added discover in isolated suite that does not run in parallel with anything... That fixes your concern for presence assertion limitations.

tests/kafka/kafka_util_test.go
mihir-datazip: Done

tests/kafka/kafka_util_test.go
mihir-datazip: Done

drivers/mssql/docker-entrypoint-initdb.d/01-init.sql
```diff
@@ -79,3 +79,23 @@ IF EXISTS (SELECT 1 FROM sys.databases WHERE name = 'olake_mssql_test' AND is_cd
...
+-- the database -- one capture job and one fn_cdc_get_max_lsn() shared by every
+-- table in it -- so suites sharing a database cannot run concurrently.
+-------------------------------------------------------------------------------
+IF DB_ID('olake_mssql_test_2pc') IS NULL
+BEGIN
+    CREATE DATABASE olake_mssql_test_2pc;
+END;
+GO
+
+USE olake_mssql_test_2pc;
```
vaibhav-datazip, line 96: why wont we need a separate 2pc db or table for other drivers?

tests/testutils/docker.go
```diff
@@ -1,73 +1,151 @@
...
-			args = append(args, "PLATFORMS="+platform)
-		}
-		build := exec.Command("make", args...)
-		build.Dir = rootPath
-		if out, err := build.CombinedOutput(); err != nil {
-			buildBaseImageErr = fmt.Errorf("failed to build base image %s: %w\n%s", image, err, out)
+	getOrBuildDriverImage(t, cfg)
+	defer trackPhaseTiming(t, cfg.Driver, olakeArgs[0]+" run")()
+
+	name := fmt.Sprintf("olake-it-%s-%d-%d", cfg.Driver, os.Getpid(), containerSeq.Add(1))
```
vaibhav-datazip, line 88: do we need both pid and seq in name, or can't we just replace both of these with oalekArgs[0]

tests/testutils/docker.go
```diff
@@ -1,73 +1,151 @@
...
+	}
+	if useState {
+		args = append(args, "--state", config.StatePath)
+	}
+	return append(args, flags...)
+}
+
+// discoverArgs builds the `olake discover ...` argument vector run against the driver image.
+func discoverArgs(config TestConfig, flags ...string) []string {
+	return append([]string{"discover", "--config", config.SourcePath}, flags...)
```
vaibhav-datazip, line 150: if these both functions are used for test utils why are they in docker.go

tests/testutils/constants/constants.go
vaibhav-datazip: isn't there a way we couple it with the state_version file , otherwise we will have to remember every time to upgrade it right ?

utils/logger/timing.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// TimingEnvVar gates phase timing, emitted as `[timing] <scope> <phase>: <duration>`. Unset -- the
+// default -- a sync pays nothing for it; the integration harness sets it on every container.
+const TimingEnvVar = "OLAKE_TIMING"
+
+var timingEnabled = os.Getenv(TimingEnvVar) != ""
+
+// LogTiming emits one timing line. No-op unless OLAKE_TIMING is set.
+func LogTiming(scope, phase string, d time.Duration) {
```
vaibhav-datazip, line 15: there are timed flags in build.sh which were used for logging the time of sync only. can you check those once and remove if no more needed

tests/mssql/mssql_util_test.go
```diff
@@ -14,9 +16,38 @@ import (
...
+	var err error
+	for attempt := 1; attempt <= 5; attempt++ {
+		if _, err = db.ExecContext(ctx, query); err == nil {
+			return nil
+		}
+		if !strings.Contains(err.Error(), "was deadlocked on lock resources") {
+			return err
+		}
+		t.Logf("CDC metadata statement lost a deadlock (attempt %d/5), retrying: %s", attempt, err)
+		time.Sleep(time.Duration(attempt) * time.Second)
```
vaibhav-datazip: shouldn't we handle error retry similar in all places , like we are checking context cancellation as well during db2

.github/actions/detect-drivers/action.yml
```diff
@@ -0,0 +1,31 @@
...
+    required: false
+    default: ${{ github.token }}
+
+outputs:
+  drivers:
+    description: JSON array of drivers to run; empty when a change touches no tested driver.
+    value: ${{ steps.resolve.outputs.drivers }}
+  testutils:
+    description: true when the shared test harness (tests/testutils) changed.
+    value: ${{ steps.resolve.outputs.testutils }}
```
vaibhav-datazip: are we utilizing this somewhere ?

tests/db2/db2_util_test.go
```diff
@@ -38,7 +45,38 @@ func buildDSN(config testutils.SourceConfig) string {
...
+	}
+	db, err := sqlx.ConnectContext(ctx, "go_ibm_db", dsn)
+	require.NoError(t, err, "failed to connect to db2")
+	dbConns[dsn] = db
+	return db
+}
+
+// exec runs one autocommit statement, retrying only the DB2 deadlock victim (SQL0911N,
+// SQLSTATE 40001): TestDB2Integration and TestDB22PC run in parallel and their DDL contends
+// on the system catalog, so either side can be rolled back and has to replay the statement.
```
vaibhav-datazip: db2integration -> db2sync

tests/testutils/constants/constants.go
mihir-datazip: We will move this to external secrets and this is a temporary copy
Initially I added this in a common place, but that idea was  I will add a TODO for it in both the files

drivers/mssql/docker-entrypoint-initdb.d/01-init.sql
mihir-datazip: Yes, I think we can remove it...

tests/testutils/docker.go
mihir-datazip: Yes, as discussed we are just making it unique... readability is not an issue as these are split second containers

tests/testutils/docker.go
mihir-datazip: They generate args for runOlake so we should group them under same file

tests/testutils/constants/constants.go
mihir-datazip: Not as of today

mihir-datazip: There is a plan for it but it is not the scope of this PR

utils/logger/timing.go
mihir-datazip: Yes we can remove them...

tests/mssql/mssql_util_test.go
mihir-datazip: Yes we should check the context done in retries, good catch

.github/actions/detect-drivers/action.yml
mihir-datazip: No, this needs to be removed, I think I did not commit it by mistake

tests/db2/db2_util_test.go
mihir-datazip: Will Add

vaibhav-datazip (dismissed):

saksham-datazip (commented):
Checked Kafka related files and LGTM

saksham-datazip (approved):

PR #1038, opened 2026-07-21
feat(ci): add Java dependency CVE scanning with Trivy

vikaxsh:
# Description
Adds a `trivy-java` job to the security CI that scans the Java Iceberg writer for known HIGH/CRITICAL vulnerabilities. Replaces the long-disabled TODO block that was removed because raw `pom.xml` filesystem scans hung for hours.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/destination/iceberg/olake-iceberg-java-writer/pom.xml
+++ b/destination/iceberg/olake-iceberg-java-writer/pom.xml
@@ -17,8 +17,13 @@
     <packaging>jar</packaging>
 
     <properties>
-        <!-- Debezium Version -->
-        <version.debezium>2.5.2.Final</version.debezium>
+        <!--
+          Dependency policy: declare only what we use; do not pin transitive
+          versions. Clear CVEs by upgrading the parent that ships them; when the
+          vulnerable copy is shaded inside a third-party jar no pin can reach it,
+          so record it in .trivyignore. netty-bom/jackson-bom are deliberate
+          exceptions that keep each family on a single version.
+        -->
 
         <!-- Instruct the build to use only UTF-8 encoding for source code -->
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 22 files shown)


shubham19may:
Compared this against a parallel CVE-cleanup pass I did independently on the same module (Docker Scout on the published image + `mvn dependency:tree` verification, not just pom reading). Java source edits are byte-identical between the two efforts (dead CDI code, `DebeziumException`→`RuntimeException`, etc. — only one correct way to do that cleanup). All divergence is in dependency strategy. Posting what I found, verified by actually running `mvn dependency:tree` on both trees, not just eyeballing diffs.

**Three things this branch gets right that mine didn't:**

1. **hadoop 3.3.6 → 3.5.0 fixes the shaded-protobuf HIGHs at the root.** Confirmed via `mvn dependency:tree`: 3.5.0 resolves `hadoop-shaded-protobuf_3_25:1.5.0`, replacing the old `hadoop-shaded-protobuf_3_7:1.1.1` (protobuf 3.7.1, vulnerable). My branch kept hadoop at 3.3.6 and reported those 4 HIGH findings as unfixable/shaded — that was wrong; a hadoop version bump reaches them.

2. **Kerby CVSS 9.8 (CVE-2023-25613) — removed, not just re-pinned.** I patched by pinning `org.apache.kerby:*` to 2.0.3 (kerby stays on the classpath, patched). This branch's hadoop 3.5.0 dependency tree has **zero** kerby artifacts at all — confirmed via `mvn dependency:tree`. Cleaner: nothing to patch if it's not there.

3. **Jetty HIGH findings — actually dead code, correctly removed.** Grepped the Java source: zero `org.eclipse.jetty` imports anywhere. The explicit `jetty-http/jetty-server/jetty-servlets/jetty-xml` deps in the base pom are pure inherited cruft from the original Debezium fork. This branch drops them and excludes jetty transitively from hadoop-common/hive-metastore/hadoop-client (headless gRPC server, no web UI — jetty is dead everywhere). My branch left them in and re-pinned to the latest available 9.4.x (still vulnerable, EOL line) instead of removing.

Net: what I reported as "9 unfixable HIGH" was actually 6 fixable (this branch fixes all 6) + 3 genuinely stuck (parquet-jackson's shaded jackson-databind copy — same on both branches, since parquet hasn't re-shaded past the fix version). The `.trivyignore` here documents that correctly: dated, links Apache Iceberg's own trivyignore precedent for the identical issue, states the exact clearing condition. Good pattern, worth keeping.

**Questions / things to double check before merge:**

- AWS SDK module list is missing `netty-nio-client`. If nothing on this path ever uses `S3AsyncClient` that's fine, but worth an explicit "confirmed unused" rather than a silent omission.
- `s3-transfer-manager` and `lakeformation` are included with a comment that `hadoop-aws`'s `S3AFileSystem` needs the transfer manager — can you cite where that call path is? If it's real, that's a legit catch that a pure source-import grep (what I used to decide what's safe to drop) can't see, since it's the *library's* internal need, not OLake's own code calling it.
- Trivy identifies dependencies via embedded `pom.properties`/MANIFEST metadata in nested jars. A genuinely relocated/shaded bundle (like `iceberg-aws-bundle`, which this PR already drops) strips that metadata, so Trivy silently misses CVEs bundled inside it — it wouldn't fail the build, it just wouldn't know. Worth a one-line comment in the pom or workflow recording *why* no shaded bundle deps are allowed here, so nobody reintroduces one thinking Trivy will catch problems in it.
- Scanning the actual built shaded jar (`mvn clean package` → scan the jar) rather than the pom is the right call — it matches what ships, and it's how this caught the parquet-jackson issue that a pom-only read would've missed. Just confirm `severity: HIGH,CRITICAL` + `ignore-unfixed: true` is a deliberate, documented gate — MEDIUM/LOW and "vulnerable but no patch yet" CVEs are invisible to it by design.

CI on this branch already reflects the above — `trivy-java` passed clean on `86a899c1` (only the 2 documented parquet-jackson CVEs ignored), which is a stronger verification than what I had.

destination/iceberg/olake-iceberg-java-writer/pom.xml
```diff
@@ -234,29 +163,141 @@
...
         <dependency>
             <groupId>javax.annotation</groupId>
             <artifactId>javax.annotation-api</artifactId>
             <version>1.3.2</version>
         </dependency>
 
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+            </dependency>
```
shubham19may, line 187: formatting?

destination/iceberg/olake-iceberg-java-writer/pom.xml
```diff
@@ -287,34 +324,10 @@
...
         </dependency>
-        <dependency>
-            <groupId>org.apache.iceberg</groupId>
-            <artifactId>iceberg-aws-bundle</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.iceberg</groupId>
-            <artifactId>iceberg-dell</artifactId>
-        </dependency>
-        <dependency>
```
shubham19may, line 306: if we add them, how much will it affect vulnerabilities?

destination/iceberg/olake-iceberg-java-writer/pom.xml
```diff
@@ -287,34 +324,10 @@
...
         <dependency>
             <groupId>org.apache.iceberg</groupId>
             <artifactId>iceberg-aws</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.iceberg</groupId>
-            <artifactId>iceberg-aws-bundle</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.iceberg</groupId>
```
shubham19may, line 303: if we add them, how much will it affect vulnerabilities?

vikaxsh: Removed Azure and AWS libraries to eliminate around six shaded Netty vulnerabilities

destination/iceberg/olake-iceberg-java-writer/pom.xml
vikaxsh: dell/aliyun  Removed as dead code

destination/iceberg/olake-iceberg-java-writer/pom.xml
vikaxsh: fixed in pr #1065

shubham19may (approved):

PR #1045, opened 2026-07-25
feat(mongodb): add explicit authentication mechanism support

### LINKED ISSUE #777: add authentication mechanism for MongoDB source
Currently OLake does not support authentication mechanism when configuring MongoDB as a source. This prevents users from connecting to MongoDB clusters that required authentication (eg. PLAINTEXT), causing source set up fail.

krishanu7:
# Description

This PR adds explicit authentication mechanism support for the MongoDB source connector.

Previously, the connector did not allow users to explicitly control the authentication mechanism when building the MongoDB connection URI. This change introduces an `auth_mechanism` configuration option in the MongoDB source spec, wires it into URI generation, and adds a corresponding UI dropdown in the MongoDB source form.

### Supported mechanisms

| Mechanism | Config | Notes |
|-----------|--------|-------|
| Default SCRAM | `auth_mechanism: ""` (omit) | Driver defaults to SCRAM-SHA-256 |
| SCRAM-SHA-1 | `"SCRAM-SHA-1"` | Explicit |
| SCRAM-SHA-256 | `"SCRAM-SHA-256"` | Explicit |
| PLAIN (LDAP) | `"PLAIN"` | Sets `authSource=$external` |
| MONGODB-X509 | `"MONGODB-X509"` | Password optional; sets `authSource=$external` |
| MONGODB-OIDC | `"MONGODB-OIDC"` | Password optional; requires `ENVIRONMENT` + `TOKEN_RESOURCE` via **Additional Connection Parameters** |
| MONGODB-AWS (IAM) | `use_iam: true` | **Not** in `auth_mechanism` enum — use **IAM Authentication** toggle |

For external mechanisms (`PLAIN`, `MONGODB-X509`, `MONGODB-OIDC`), the connector sets `authSource=$external` as required by MongoDB.

**Not included:** `GSSAPI` (Kerberos) — removed from the enum because it requires `CGO_ENABLED=1` and `-tags gssapi`, which conflicts with OLake's `CGO_ENABLED=0` build setup.

Fixes #777

## Type of change

# How Has This Been Tested?

### Test environment
- MongoDB 8.0 replica set (`rs0`) via local Docker (`make olake.mongodb.start`)
- OLake UI (localhost:8000) with driver image `v0.9.0`
- Destination: Parquet (local path / MinIO)
- MongoDB host for UI/Docker: `host.docker.internal:27017` (replica set member reconfigured accordingly)

### CLI testing

| Auth mechanism | `check` | `discover` | `sync` (full_refresh) | Result |
|----------------|---------|------------|------------------------|--------|
| Default SCRAM (`""`) | ✅ | ✅ | ✅ | 3 records synced |
| SCRAM-SHA-256 | ✅ | ✅ | ✅ | 3 records synced |
| SCRAM-SHA-1 | ✅ | — | ✅ | 3 records synced |

CLI Video Testing: https://drive.google.com/file/d/1ZE0p01ZO6LA15b7MBpKRMC6nt44n4MKX/view?usp=drive_link

UI  Video Testing: https://drive.google.com/file/d/14OLuHcxhm4WCYubvkJ7HjzF3eznjDDnP/view?usp=drive_link


vaibhav-datazip:
Hi @krishanu7 , please target your pr on staging branch also please sign the claim

drivers/mongodb/internal/config.go
```diff
@@ -2,18 +2,23 @@ package driver
...
+	"slices"
 	"strings"
 
 	"github.com/datazip-inc/olake/constants"
 	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/logger"
 )
 
+// externalAuthMechanisms are MongoDB auth mechanisms that require authSource="$external"
+var externalAuthMechanisms = []string{"PLAIN", "MONGODB-X509", "GSSAPI"}
```
vaibhav-datazip: can you check [this](https://www.mongodb.com/docs/manual/reference/connection-string-options/) documentation once, for auth mechanism 

there are some more of them

krishanu7:
@vaibhav-datazip  only MONGODB-OIDC was missing, i added that also in the externalAuthMechanisms slice

image

drivers/mongodb/resources/spec.json
```diff
@@ -7,6 +7,21 @@
...
+            "type": "string",
+            "title": "Authentication Mechanism",
+            "description": "MongoDB authentication mechanism. Leave empty for default (SCRAM). Use PLAIN for LDAP, SCRAM-SHA-256 for explicit SCRAM, MONGODB-X509 for certificate auth, GSSAPI for Kerberos, or MONGODB-OIDC for OpenID Connect.",
+            "enum": [
+                "",
+                "SCRAM-SHA-1",
+                "SCRAM-SHA-256",
+                "PLAIN",
+                "MONGODB-X509",
+                "GSSAPI",
```
vaibhav-datazip: GSSAPI might not work yet, would recommend you test this . as it might require a build tag for the mongo driver to support this

drivers/mongodb/resources/spec.json
```diff
@@ -7,6 +7,21 @@
...
+        "auth_mechanism": {
+            "type": "string",
+            "title": "Authentication Mechanism",
+            "description": "MongoDB authentication mechanism. Leave empty for default (SCRAM). Use PLAIN for LDAP, SCRAM-SHA-256 for explicit SCRAM, MONGODB-X509 for certificate auth, GSSAPI for Kerberos, or MONGODB-OIDC for OpenID Connect.",
+            "enum": [
+                "",
+                "SCRAM-SHA-1",
+                "SCRAM-SHA-256",
+                "PLAIN",
+                "MONGODB-X509",
```
vaibhav-datazip, line 19: if the mode here don't require password and someone tries to configure it using ui it might give a failure, please check that

drivers/mongodb/resources/spec.json
```diff
@@ -7,6 +7,21 @@
             "description": "Authentication database (mostly admin)",
             "default": "admin"
         },
+        "auth_mechanism": {
+            "type": "string",
+            "title": "Authentication Mechanism",
+            "description": "MongoDB authentication mechanism. Leave empty for default (SCRAM). Use PLAIN for LDAP, SCRAM-SHA-256 for explicit SCRAM, MONGODB-X509 for certificate auth, GSSAPI for Kerberos, or MONGODB-OIDC for OpenID Connect.",
```
vaibhav-datazip: what about the `MONGODB-AWS`, shouldn't we mention to toggle `use-iam`

drivers/mongodb/internal/config.go
```diff
@@ -2,18 +2,23 @@ package driver
...
 
+// externalAuthMechanisms are MongoDB auth mechanisms that require authSource="$external"
+var externalAuthMechanisms = []string{"PLAIN", "MONGODB-X509", "GSSAPI", "MONGODB-OIDC"}
+
 type Config struct {
 	Hosts            []string          `json:"hosts"`
 	Username         string            `json:"username"`
 	Password         string            `json:"password"`
 	AuthDB           string            `json:"authdb"`
+	AuthMechanism    string            `json:"auth_mechanism"`
```
vaibhav-datazip, line 19: this change should also be done for ui schema

drivers/mongodb/resources/spec.json
vaibhav-datazip: also need to discuss if authmechanism is added as separate field should we correct the description here

drivers/mongodb/internal/config.go
```diff
@@ -2,18 +2,23 @@ package driver
...
+	"slices"
 	"strings"
 
 	"github.com/datazip-inc/olake/constants"
 	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/logger"
 )
 
+// externalAuthMechanisms are MongoDB auth mechanisms that require authSource="$external"
+var externalAuthMechanisms = []string{"PLAIN", "MONGODB-X509", "GSSAPI", "MONGODB-OIDC"}
```
vaibhav-datazip: would suggest that you test the mechanisms being mentioned here (whichever possible), and mention which you were unable to test.

also as we haven't added mongodb-aws here, the comment with its reason

drivers/mongodb/resources/spec.json
```diff
@@ -7,6 +7,21 @@
...
+            "enum": [
+                "",
+                "SCRAM-SHA-1",
+                "SCRAM-SHA-256",
+                "PLAIN",
+                "MONGODB-X509",
+                "GSSAPI",
+                "MONGODB-OIDC"
```
vaibhav-datazip, line 20: wont oidc also require environment and token resource

also mention the environment it will support in olake

drivers/mongodb/resources/spec.json
krishanu7: Yes, I tested this locally and you are correct go.mongodb.org/mongo-driver requires CGO_ENABLED=1 and -tags gssapi to work, which conflicts with OLake's current CGO_ENABLED=0 build setup. I will remove GSSAPI from the supported enum for now encountering this build error.

drivers/mongodb/resources/spec.json
krishanu7: Tested and verified. UI failing validation because the JSON schema dependency strictly required a password when use_iam is false. i will restructure the dependencies block in spec.json using oneOf to make the password field optional when password-less mechanisms (like MONGODB-X509 or MONGODB-OIDC) are selected.

drivers/mongodb/resources/spec.json
krishanu7: Makes sense. I will update the description of auth_mechanism in spec.json to explicitly state: "For MONGODB-AWS authentication, use the 'IAM Authentication' toggle instead."

drivers/mongodb/internal/config.go
krishanu7: Done. I will add the auth_mechanism dropdown field to the MongoDB UI grid layout in utils/spec/uischema.go so it properly renders on the frontend.

drivers/mongodb/internal/config.go
krishanu7: I was able to test SCRAM-SHA-1 and SCRAM-SHA-256 locally. PLAIN, MONGODB-X509, and MONGODB-OIDC require enterprise/Atlas infrastructure that isn't feasible to test locally right now.

 I will also add a comment in config.go explaining that MONGODB-AWS is intentionally omitted from the enum because it's handled via the separate UseIAM toggle.

drivers/mongodb/resources/spec.json
krishanu7: Yes, OIDC requires env and token resource properties. Since adding dedicated UI fields for these adds significant complexity, I will update the auth_mechanism description to instruct users to pass environment (e.g., azure or gcp) and token resource via the existing additional connection parameters section.

drivers/mongodb/internal/config.go
```diff
@@ -2,18 +2,24 @@ package driver
...
 	"strings"
 
 	"github.com/datazip-inc/olake/constants"
 	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/logger"
 )
 
+// externalAuthMechanisms are MongoDB auth mechanisms that require authSource="$external".
+// Note: MONGODB-AWS is handled separately via the UseIAM config flag and is not included here.
+var externalAuthMechanisms = []string{"PLAIN", "MONGODB-X509", "MONGODB-OIDC"}
```
mihir-datazip: We should define constants for them to avoid using the same string literal everywhere

drivers/mongodb/internal/config.go
mihir-datazip: You will also need to raise a PR in https://github.com/datazip-inc/olake-docs for this

drivers/mongodb/internal/config.go
mihir-datazip: Also this slice should be independent of downstream logic, we should not remove MONGODB-AWS as this slice can be used in some other places as well

mihir-datazip: Add a comment on why `GSSAPI` is not included now.
Also we can drop it because the market usage of this is striking low not because we enforce CGO disabled for builds in Olake(mentioned in PR description).

For your reference, we just add this in `drivers.mk` for mongo:
```
  GO_ENV.mongodb = export CGO_ENABLED=1 GOFLAGS=-tags=gssapi;

  # Driver-image builds (OVERLAY_DIR set: Debian builder, root) install the krb5 headers; hosts
  # are only verified — macOS needs nothing (GSS.framework), Linux users install their own -dev pkg.
  .PHONY: prepare.mongodb
  prepare.mongodb:
  ifneq ($(OVERLAY_DIR),)
        apt-get update && apt-get install -y --no-install-recommends libkrb5-dev
  else
        @if [ "$$(uname -s)" = Linux ] && [ ! -f /usr/include/gssapi/gssapi.h ]; then \
                echo "ERROR: Kerberos headers required for the mongodb GSSAPI build."; \
      @if [ "$$(uname -s)" = Linux ] && [ ! -f /usr/include/gssapi/gssapi.h ]; then \
              echo "ERROR: Kerberos headers required for the mongodb GSSAPI build."; \
              echo "  Debian/Ubuntu: apt install libkrb5-dev    RHEL/Fedora: dnf install krb5-devel"; \
              exit 1; \
      fi
endif
```

drivers/mongodb/internal/config.go
mihir-datazip: Setting auth source should not be in 2 places, we just set authMechanism based on `c.UseIAM || c.AuthMechanism` and then based on what the value of authMechanism, we can decide what should be `authSource` in a single place.

Cleaner to understand than what is done right now

drivers/mongodb/internal/config.go
```diff
@@ -72,7 +84,13 @@ func (c *Config) URI() string {
 	}
 
 	if !c.UseIAM {
-		u.User = utils.Ternary(c.Password != "", url.UserPassword(c.Username, c.Password), url.User(c.Username)).(*url.Userinfo)
```
mihir-datazip, line 75: Same here, the logic of c.UseIAM should be handled in a single place and then common code can handle it without if else blocks for it in all the places

drivers/mongodb/internal/config.go
mihir-datazip, line 1: You should also add a `config_test.go` where we can have test cases for these changes...
We are increasing our UT coverage, so you can include the Unit Tests with the following logic.

1. Add Unit Test on the current code in staging and cover all the cases. -> Verify it runs
2. Add your changes and verify that nothing breaks and add the new test cases for your changes.

Also this is necessary as ITs will not touch this code

drivers/mongodb/resources/spec.json
```diff
@@ -7,6 +7,20 @@
             "description": "Authentication database (mostly admin)",
             "default": "admin"
         },
+        "auth_mechanism": {
+            "type": "string",
+            "title": "Authentication Mechanism",
+            "description": "MongoDB authentication mechanism. Leave empty for default (SCRAM). Use PLAIN for LDAP, SCRAM-SHA-256 for explicit SCRAM, MONGODB-X509 for certificate auth, or MONGODB-OIDC for OpenID Connect (requires passing ENVIRONMENT and TOKEN_RESOURCE in Additional Connection Parameters). For MONGODB-AWS authentication, use the 'IAM Authentication' toggle instead.",
```
mihir-datazip: We should also specify we don't support `GSSAPI`

drivers/mongodb/resources/spec.json
```diff
@@ -7,6 +7,20 @@
...
+            "enum": [
+                "",
```
mihir-datazip, line 15: We can use constants here as well

mihir-datazip (changes requested):

drivers/mongodb/internal/config.go
krishanu7: Makes sense. I will refactor this by introducing auth mechanism constants, move them to a dedicated auth_mechanism.go file, will keep externalAuthMechanisms independent of downstream logic (including MONGODB-AWS) and maintain a separate list for the ui supported mechanisms with a note on why GSSAPI is excluded.

drivers/mongodb/internal/config.go
```diff
@@ -33,32 +30,53 @@ type Config struct {
 	AdditionalParams map[string]string `json:"additional_params"`
 }
 
+func (c *Config) resolvedAuthMechanism() string {
```
mihir-datazip: I meant we can just set c.AuthMechanism when we have c.UseIAM.
Also all the fields we need to set remains in a single function... not adding resolving logic for all the fields in different functions

drivers/mongodb/internal/config.go
```diff
@@ -33,32 +30,53 @@ type Config struct {
...
+	if c.UseIAM {
+		return AuthMechanismAWS
+	}
+	return c.AuthMechanism
+}
+
+func authSourceForMechanism(mechanism, authDB string) string {
```
mihir-datazip: Same thing here, we can just set c.AuthDB and other props that depend on some other field in a decorator function instead of making a different function for each field.

drivers/mongodb/internal/config_test.go
```diff
@@ -0,0 +1,374 @@
...
+	}{
+		{
+			name: "default SCRAM with username and password",
+			config: &Config{
+				Hosts:    []string{"localhost:27017"},
+				Username: "user",
+				Password: "pass",
+				AuthDB:   "admin",
+			},
+			expectedContains: []string{
```
mihir-datazip: We can just have full URI here itself and assert, tests don't need this kind of assertion logic.
Just have final URI and do exact match, will be easier to maintain and understand

drivers/mongodb/internal/config_test.go
```diff
@@ -0,0 +1,374 @@
...
+				t.Fatalf("expected validation error but got none")
+			}
+			if !tt.expectErr && err != nil {
+				t.Fatalf("expected no validation error but got: %v", err)
+			}
+		})
+	}
+}
+
+func TestSupportedAuthMechanismsMatchSpec(t *testing.T) {
```
mihir-datazip: We don't need to add this test here...
These can either come in ITs, which you don't need to add in this PR

drivers/mongodb/internal/config.go
```diff
@@ -101,15 +113,22 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("database is required")
 	}
```
vaibhav-datazip: you can restructure URI it like this, by fixing the c.auth mechanism and c.auth db in validation function

```
func (c *Config) URI() string {

	sslEnabled := c.SSLConfiguration != nil && c.SSLConfiguration.Mode != utils.SSLModeDisable

	query := url.Values{}
	// Additional params are written first so configured fields always take precedence.
	for key, value := range c.AdditionalParams {
		if sslEnabled && slices.Contains(tlsFileParams, key) {
			continue
		}
		query.Set(key, value)
	}
	query.Set("authSource", c.AuthDB)
	if c.AuthMechanism != "" {
		query.Set("authMechanism", c.AuthMechanism)
	}
	if c.ReplicaSet != "" {
		query.Set("replicaSet", c.ReplicaSet)
		query.Set("readPreference", utils.Ternary(c.ReadPreference != "", c.ReadPreference, constants.DefaultReadPreference).(string))
	}
	if sslEnabled && query.Get("tls") == "" {
		query.Set("tls", "true")
	}

	u := &url.URL{
		Scheme:   utils.Ternary(c.Srv, "mongodb+srv", "mongodb").(string),
		Host:     strings.Join(c.Hosts, ","),
		Path:     "/",
		RawQuery: query.Encode(),
	}

	switch {
	case c.Username == "" || c.AuthMechanism == AuthMechanismAWS:
		// MONGODB-AWS takes its credentials from the environment, never from URI userinfo.
	case c.Password == "" || slices.Contains(passwordlessAuthMechanisms, authMechanism):
		u.User = url.User(c.Username)
	default:
		u.User = url.UserPassword(c.Username, c.Password)
	}

	return u.String()
}
```

drivers/mongodb/internal/config.go
```diff
@@ -130,5 +149,6 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("failed to validate ssl config: %w", err)
 	}
 
+	c.applyAuthDefaults()
 	return utils.Validate(c)
```
vaibhav-datazip, line 227: in the start of validate function you can add

```
	authMechanism := utils.Ternary(c.UseIAM, AuthMechanismAWS, c.AuthMechanism).(string)
	c.AuthDB = utils.Ternary(slices.Contains(externalAuthMechanisms, authMechanism), externalAuthDB, c.AuthDB).(string)

```

krishanu7:
## Summary

This change centralizes MongoDB authentication normalization and validation in `Validate()`, while keeping `URI()` as a non-mutating serializer.

## Issues fixed

- Added typed `auth_mechanism` support while retaining legacy `additional_params.authMechanism` compatibility.
- Normalized external mechanisms to `authSource=$external`.
- Added mechanism-specific credential policies:
  - SCRAM and PLAIN require username/password.
  - X.509 and OIDC reject passwords and permit an optional username.
  - AWS IAM does not put credentials in URI userinfo.
- Prevented empty usernames from producing URIs such as `mongodb://@host/`.
- Ensured AWS IAM uses runtime/environment credentials instead of URI credentials.
- Enforced TLS for PLAIN and X.509.
- Enforced a client certificate for X.509.
- Explicitly rejected unsupported and unknown mechanisms, including GSSAPI.
- Made typed authentication fields take precedence over legacy URI parameters.
- Rejected case-insensitive duplicate reserved parameters and aliases, preventing nondeterministic Go map selection.
- Resolved `tls`/`ssl`, SRV defaults, and inline SSL consistently.
- Rejected contradictory inline SSL plus `tls=false` configurations.
- Rejected inline CA/client certificate fields when `ssl.mode=disable`, because those certificates would otherwise be ignored.
- Kept `URI()` non-mutating; defaults and normalization are now owned by `Validate()`.
- Fixed shared `require` TLS mode so it continues loading the client certificate required for X.509/mTLS.

### UI Testing Results

| Configuration | Result |
|---|---|
| **Default SCRAM** | Successful |
| **SCRAM-SHA-1** | Successful |
| **SCRAM-SHA-256** | Form and branch configuration tested; real successful authentication was unavailable on the current Atlas deployment |
| **X.509** | Successful |
| **AWS IAM** | Successful |
| **SSL `disable`** | Successful against local MongoDB |
| **SSL `require`** | Successful |
| **SSL `verify-ca`** | Successful |
| **SSL `verify-full`** | Successful |
| **OIDC** | Form and validation tested; no real workload identity connection was available |
| **PLAIN** | Form and validation tested; no real LDAP connection was available |
| **GSSAPI** | Not exposed as a supported option and correctly rejected by validation |

krishanu7:
utils/ssl.go at line 68  returns immediately for X.509 SSL mode require. Client certificates are only loaded later at line 125, so that code never runs. Atlas receives no client certificate
So i changed the code in ssl.go as well.

drivers/mongodb/internal/config.go
```diff
@@ -92,43 +275,90 @@ func (c *Config) buildTLSConfig() (*tls.Config, error) {
...
+		mechanism = AuthMechanismAWS
+	} else if mechanism == AuthMechanismAWS {
+		return fmt.Errorf("MONGODB-AWS must be configured through use_iam in this connector")
 	}
 
-	if c.MaxThreads <= 0 {
-		c.MaxThreads = constants.DefaultThreadCount
+	if mechanism == AuthMechanismGSSAPI {
+		return fmt.Errorf("GSSAPI is not supported by the OLake MongoDB connector")
 	}
```
vaibhav-datazip: instead of so many if else maybe you can consider a switch case

```
reserved, err := resolveReservedParams(c.AdditionalParams)
	if err != nil {
		return err
	}

	// The typed field wins; additional_params is the legacy way of naming the mechanism.
	mechanism := c.AuthMechanism
	if mechanism == "" {
		mechanism = reserved["authMechanism"]
	}
	switch {
	case c.UseIAM && mechanism != "" && mechanism != AuthMechanismAWS:
		return fmt.Errorf("auth_mechanism cannot be set when use_iam is enabled; IAM authentication uses MONGODB-AWS")
	case c.UseIAM:
		mechanism = AuthMechanismAWS
	case mechanism == AuthMechanismAWS:
		return fmt.Errorf("MONGODB-AWS must be configured through use_iam in this connector")
	case mechanism == AuthMechanismGSSAPI:
		return fmt.Errorf("GSSAPI is not supported by the OLake MongoDB connector")
	}
```

drivers/mongodb/internal/config.go
```diff
@@ -92,43 +275,90 @@ func (c *Config) buildTLSConfig() (*tls.Config, error) {
...
 		return fmt.Errorf("database is required")
 	}
 
-	if !c.UseIAM {
-		if c.Username == "" {
-			return fmt.Errorf("username is required")
-		}
-		if c.AuthDB == "" {
-			return fmt.Errorf("authdb is required")
+	if err := validateReservedAdditionalParams(c.AdditionalParams); err != nil {
```
vaibhav-datazip: we are walking through additional params multiple times to find various values utilizing the additionalParam() (nearly 8 times) can we somehow optimize this

drivers/mongodb/internal/config.go
```diff
@@ -29,58 +31,239 @@ type Config struct {
...
+		return fmt.Errorf("password must be empty for %s", mechanism)
+	}
+	if policy.RequireTLS && !caps.tlsEnabled {
+		return fmt.Errorf("TLS is required for %s authentication", mechanism)
+	}
+	if policy.RequireClientCert && !caps.hasClientCert {
+		return fmt.Errorf("a client certificate is required for MONGODB-X509")
+	}
+	return nil
+}
```
vaibhav-datazip: this is purely auth mechanism related function shouldn't this be moved to auth_mechanism.go as well

also can we add a precondition type of error here

drivers/mongodb/internal/config.go
```diff
@@ -29,58 +31,239 @@ type Config struct {
...
 	for key, value := range c.AdditionalParams {
-		if sslEnabled && (key == "tlsCAFile" || key == "tlsCertificateKeyFile") {
+		if inlineSSL && isTLSFileParam(key) {
+			continue
+		}
+		if c.AuthDB != "" && strings.EqualFold(key, "authSource") {
+			continue
+		}
+		if c.AuthMechanism != "" && strings.EqualFold(key, "authMechanism") {
 			continue
```
vaibhav-datazip: maybe you can shorted this as all three conditions are for continue only 

similarly various parts in this file can be made cleaner by removing unnecessary code blocks

utils/ssl.go
```diff
@@ -38,19 +38,28 @@ type SSLConfig struct {
...
 func (sc *SSLConfig) Validate() error {
-	// TODO: Add Proper validations and test
 	if sc == nil {
 		return errors.New("'ssl' config is required")
 	}
 
-	if sc.Mode == Unknown {
+	switch sc.Mode {
+	case Unknown:
 		return errors.New("'ssl.mode' is required parameter")
```
vaibhav-datazip: can't we fall it back to default

krishanu7: We can change ssl.go so ssl: {} works. That only creates a Postgres-specific bug because Postgres builds sslmode before Validate(). The small Postgres reorder is required together with the ssl.go change, not on its own.

drivers/mongodb/internal/config.go
```diff
@@ -29,58 +30,116 @@ type Config struct {
...
+		return true
+	}
+	return false
+}
 
-	host := strings.Join(c.Hosts, ",")
+// URI builds the MongoDB connection string from an already-validated config.
+// It does not mutate Config: call Validate() first so AuthMechanism, AuthDB, and defaults are set.
+func (c *Config) URI() string {
+	inlineSSL := inlineSSLEnabled(c)
```
vaibhav-datazip, line 100: can we rename the function and variable

```suggestion
	inlineSSLEnabled := isInlineSSLEnabled(c)
```

drivers/mongodb/internal/auth_mechanism.go
```diff
@@ -0,0 +1,234 @@
...
+	"tlsCertificateKeyFile",
+	"sslClientCertificateKeyFile",
+	"tlsCertificateFile",
+	"tlsPrivateKeyFile",
+}
+
+func authPolicyFor(mechanism string) (authPolicy, bool) {
+	policy, ok := mechanismPolicies[mechanism]
+	return policy, ok
+}
```
vaibhav-datazip, line 84: I don't think this wrapper function is required

utils/ssl.go
```diff
@@ -62,64 +96,64 @@ func BuildTLSConfig(host string, sc *SSLConfig) (*tls.Config, error) {
...
+			tlsConfig.InsecureSkipVerify = true
+			tlsConfig.VerifyPeerCertificate = func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+				if len(rawCerts) == 0 {
+					return fmt.Errorf("no server certificate provided")
+				}
+				cert, err := x509.ParseCertificate(rawCerts[0])
 				if err != nil {
-					logger.Warnf("failed to parse intermediate certificate at position %d: %s", i, err)
-					continue
+					return fmt.Errorf("failed to parse server certificate: %w", err)
```
vaibhav-datazip, line 131: you will have to add error. precondition like we have done at other places

drivers/mongodb/internal/config.go
```diff
@@ -29,58 +30,116 @@ type Config struct {
...
+	policy, _ := authPolicyFor(c.AuthMechanism)
 
-	sslEnabled := c.SSLConfiguration != nil && c.SSLConfiguration.Mode != utils.SSLModeDisable
+	query := url.Values{}
 	for key, value := range c.AdditionalParams {
-		if sslEnabled && (key == "tlsCAFile" || key == "tlsCertificateKeyFile") {
+		if skipAdditionalParamKey(key, inlineSSL, c.AuthDB, c.AuthMechanism) {
```
vaibhav-datazip, line 105:
```suggestion
		if  c.skipAdditionalParamKey(key, inlineSSL) {
```

similarly other functions can also be created methods of the code like inlineSSLEnabled

drivers/mongodb/internal/auth_mechanism.go
```diff
@@ -0,0 +1,234 @@
...
+		ExternalAuthDB:  true,
+		RequireUsername: true,
+		RequirePassword: true,
+		RequireTLS:      true,
+	},
+	AuthMechanismX509: {
+		ExternalAuthDB:    true,
+		ForbidPassword:    true,
+		RequireTLS:        true,
+		RequireClientCert: true,
```
vaibhav-datazip, line 59: client cert is tls related so maybe you can create just one enum to handle this all, maybe similar can be done for password

drivers/mongodb/internal/config.go
```diff
@@ -29,58 +30,116 @@ type Config struct {
...
 		}
 		query.Set(key, value)
 	}
-	if sslEnabled && query.Get("tls") == "" {
+	if c.AuthDB != "" {
+		query.Set("authSource", c.AuthDB)
+	}
+	if c.AuthMechanism != "" {
+		query.Set("authMechanism", c.AuthMechanism)
+	}
```
vaibhav-datazip, line 115: Why do we even need skipAdditionalParamKey function , can't we just skip the check for these fields in additional params as well, because we dont check replicate set or read preference there as well right ?

drivers/mongodb/internal/config.go
```diff
@@ -29,58 +30,116 @@ type Config struct {
...
+	if inlineSSL {
 		query.Set("tls", "true")
 	}
 
-	// Construct final URI using url.URL
+	scheme := "mongodb"
+	if c.Srv {
+		scheme = "mongodb+srv"
+	}
```
vaibhav-datazip, line 127: can this be ternary as well

drivers/mongodb/internal/auth_mechanism.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	for key, value := range params {
+		switch {
+		case strings.EqualFold(key, "authMechanism"):
+			if r.hasAuthMechanism {
+				return r, fmt.Errorf("additional parameter %q is configured more than once", "authMechanism")
+			}
+			r.AuthMechanism = value
+			r.hasAuthMechanism = true
```
vaibhav-datazip, line 125: shouldn't this be marked true even if the authmechanism is directly set in config , even for other fields like auth mechanism , tls etc, this will simplify this block as well and even the config uri construction part will not need unecessary checks as validate has already done the changes

drivers/mongodb/internal/auth_mechanism.go
```diff
@@ -0,0 +1,234 @@
...
+				return r, fmt.Errorf("additional parameter %q is configured more than once", "tlsPrivateKeyFile")
+			}
+			r.hasKeyFile = true
+		}
+	}
+
+	return r, nil
+}
+
+func resolveAuthMechanism(useIAM bool, fieldMechanism, legacyMechanism string) (string, error) {
```
vaibhav-datazip, line 169: I dont think we will even require multiple functions if we can add this check 
- check if auth mechanism is explicitly mentioned and other fields and mark them
- check additional params and discard those once 
- AuthMechanismIAM will internally use the useIAM path only , and we will support that as well so that backward compatibility is maintained , but also simplify it by showing just AuthMechanismAWS there

drivers/mongodb/internal/auth_mechanism.go
```diff
@@ -0,0 +1,234 @@
...
+
+	switch {
+	case useIAM && mechanism != "" && mechanism != AuthMechanismAWS:
+		return "", fmt.Errorf("auth_mechanism cannot be set when use_iam is enabled; IAM authentication uses MONGODB-AWS")
+	case useIAM:
+		return AuthMechanismAWS, nil
+	case mechanism == AuthMechanismAWS:
+		return "", fmt.Errorf("MONGODB-AWS must be configured through use_iam in this connector")
+	case mechanism == AuthMechanismGSSAPI:
+		return "", fmt.Errorf("GSSAPI is not supported by the OLake MongoDB connector")
```
vaibhav-datazip, line 183: you can add a explanation comment for this here as well

utils/spec/uischema.go
```diff
@@ -21,14 +21,26 @@ var uiSchemaMap = map[string]string{
...
-        { "read_preference": 12, "srv": 12 },
-        { "max_threads": 12, "backoff_retry_count": 12 },
-        { "chunking_strategy": 12, "use_iam": 12 },
-        { "ssl": 12, "additional_params": 12 },
+        { "authdb": 12, "auth_mechanism": 12 },
+        { "username": 12, "password": 12 },
+        { "replica_set": 12, "read_preference": 12 },
+        { "srv": 12, "max_threads": 12 },
+        { "backoff_retry_count": 12, "chunking_strategy": 12 },
+        { "use_iam": 12, "ssl": 12 },
```
vaibhav-datazip, line 29: this is ui side change, please discuss this with product once

vaibhav-datazip (commented):
Looks like we can validate once and not consider the repeated fields again . This will also reduce the handling during uri building as well . Also please make sure backward compatibility is being maintained

drivers/mongodb/internal/config.go
```diff
@@ -29,58 +30,116 @@ type Config struct {
...
+}
 
-func (c *Config) URI() string {
-	connectionPrefix := "mongodb"
-	if c.Srv {
-		connectionPrefix = "mongodb+srv"
+func isTLSFileParam(key string) bool {
+	for _, candidate := range tlsFileParams {
+		if strings.EqualFold(candidate, key) {
+			return true
```
vaibhav-datazip, line 45: tls and ssl do the same thing this can also be triggered by ssl path

vaibhav-datazip (commented):
we can discuss the changes once, it might simplify the code and logic a lot

drivers/mongodb/internal/config.go
krishanu-datazip: removed the block entirely, all three keys are now deleted from AdditionalParams during Validate(), so by the time URI() runs, the loop is a plain unconditional copy

drivers/mongodb/internal/config.go
krishanu-datazip: removed the wrapper function entirely. Validate() now computes it inline as a local bool: inlineSSL := c.SSLConfiguration.Mode != utils.SSLModeDisable. No naming collision left because there's only one thing left to name.

drivers/mongodb/internal/auth_mechanism.go
krishanu-datazip: Removed. Both call sites (URI() and resolveAuthMechanism) now read mechanismPolicies[...] directly.
authPolicyFor(mechanism) was a one-line wrapper around a map lookup with no extra logic.

utils/ssl.go
krishanu-datazip: Every remaining bare error in SSLConfig.Validate() now goes through errs.Precondition with a dedicated code

drivers/mongodb/internal/config.go
krishanu-datazip: Yes, Scheme: utils.Ternary(c.Srv, "mongodb+srv", "mongodb").(string),

drivers/mongodb/internal/auth_mechanism.go
krishanu-datazip: Yes, replaced with a single ordered tlsRequirement enum (tlsOptional < tlsRequired < tlsClientCert), and did the same for password with a passwordRule enum (passwordOptional/passwordRequired/passwordForbidden)

drivers/mongodb/internal/config.go
krishanu-datazip: Yeah. removed entirely. replicaSet and readPreference never needed special-casing in additional_params because Validate() doesn't leave conflicting copies of them lying around.

drivers/mongodb/internal/auth_mechanism.go
krishanu-datazip: Done, by removing the flag struct entirely rather than patching it. The typed field vs. additional_params precedence is now resolved exactly once

drivers/mongodb/internal/auth_mechanism.go
krishanu-datazip: True, This is exactly what resolveAuthMechanism + normalizeAdditionalParams now do together: typed field checked first, additional_params as fallback, use_iam mapped onto MONGODB-AWS internally so both routes produce identical results, and the legacy additional_params.authMechanism key is deleted (discarded) exactly once, in Validate().

drivers/mongodb/internal/auth_mechanism.go
krishanu-datazip: Added

drivers/mongodb/internal/config.go
krishanu-datazip: YEs, reservedAliases maps both "tls" and "ssl" to the single canonical key "tls", so Validate() only has to consult reserved["tls"] once and both spellings are covered.

PR #1053, opened 2026-07-28
chore: remove common lib dependency in IT and olake module

mihir-datazip:
# Description

Removes the last shared Go code between the integration-test suites and the olake product module.

**`lib` module deleted.** It existed only to serve `tests/` — root `constants/` merely re-exported it. Contents folded back into `constants/constants.go` and `constants/state_version.go`; the `require`/`replace` pairs dropped from the root module, all 8 `drivers/*/go.mod` (they carried it transitively as an indirect dep) and all 8 `tests/*/go.mod`, plus the `go.work` use entry.

**Per-driver test config structs deleted.** The 7 `tests/<driver>/config.go` files hand-duplicated each driver's connection config. Replaced by `testutils.SourceConfig` — `source.json` read as a plain map with `String`/`Int`/`Strings`/`StringMap`/`Sub` accessors. `Int()` converts from `float64` rather than type-asserting, since a JSON number decoded into an `any` is never an `int`. db2's `BuildDSN` and oracle's `connectionString` become plain functions in their `_util_test.go`.

**`tests/testutils/constants/`** holds the test-side `DriverType`, driver names, `CdcTimestamp` and `SkipCDCDrivers`. Deliberately restated rather than imported from olake: a black-box suite that shares the product's constant cannot detect a contract break — rename the CDC column and every test using the shared symbol recompiles clean and passes, sailing past the exact regression the suite exists to catch.

**`tests/go.work`** now holds the 8 test-module `use` entries, moved out of the root `go.work` (which returns to matching master's). Resolution never depended on this — every test module already `replace`s `tests/testutils => ../testutils` explicitly. The workspace only provides the multi-module pattern the Makefile uses (`cd tests && go test ./postgres/...`), so a tests-scoped workspace preserves it verbatim with no Makefile changes. Measured effect: the test modules were silently raising 4 dependencies in the product's build list (otelhttp `v0.46.1`→`v0.60.0`, logrus, creack/pty, x/time) on every local and CI `go build ./...`.

**`ENV GOWORK=off` removed from the Dockerfile** — worth a careful review, as it changes the shipped artifact. It is not on `master`; it entered via #1033, needed only because `go.work` had begun listing `tests/` dirs that `.dockerignore` excludes (a `use` entry with no directory is a hard error). Its side effect was unintended: master builds driver images *in workspace mode*, and turning the workspace off changes 16–18 dependency versions per driver (e.g. `drivers/postgres`: aws-sdk-go-v2/service/s3 `v1.43.0`→`v1.88.4`, azure azcore `v1.6.0`→`v1.18.0`). With `tests/` out of `go.work` the original reason is gone, so the line is removed. Verified: `drivers/postgres` now resolves identically to master's build list.

Fixes # (issue)

## Type of change

- [x] Refactor / test infrastructure (non-breaking, no product behaviour change)

# How Has This Been Tested?

Static verification only — the integration suites were not run locally and need a CI run to sign off.

- [x] `go build ./...` and `go vet ./...` clean on the root module
- [x] `go build ./...` clean on all driver modules except db2
- [x] `go vet ./...` clean on all test modules except db2
- [x] `gofmt -l -s` clean across `tests/` and `constants/`
- [x] `cd tests && go vet ./postgres/... ./mysql/... ./mongodb/... ./mssql/... ./oracle/... ./kafka/... ./testutils/...` — confirms the Makefile's multi-module invocation still works under `tests/go.work`
- [x] `drivers/postgres` build list compared with and without the workspace, confirming the Dockerfile change restores master's resolution
- [x] Integration / 2PC / performance suites

db2 is excluded throughout: it fails at `sqlcli1.h` on arm64 macOS (no IBM clidriver), pre-existing and unrelated to this change. Its module graph resolves and its call sites follow the same mechanical pattern as the other drivers.

# Screenshots or Recordings

N/A

## Documentation

## Related PR's (If Any):

#1033 — introduced `ENV GOWORK=off`, removed here.


.github/workflows/security-ci.yaml
```diff
@@ -69,7 +69,7 @@ jobs:
       - name: install gosec
         run: curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.22.11
       - name: Run Gosec Security Scanner
-        run: $(go env GOPATH)/bin/gosec -exclude=G115 -severity=high -confidence=medium ./...
+        run: $(go env GOPATH)/bin/gosec -exclude=G115 -exclude-dir=tests -severity=high -confidence=medium ./...
```
vikaxsh, line 72: why are we excluding this?

tests/testutils/source_config.go
```diff
@@ -0,0 +1,83 @@
...
+// The integration tests hand source.json to the driver verbatim and only ever need a handful
+// of connection fields back out of it to talk to the source directly. Reading it as a map
+// keeps the tests from restating the driver's config schema: a field the tests do not read is
+// none of their business, and one the driver adds or renames cannot silently drift out of
+// sync with a copy kept here.
+//
+// Missing keys and type mismatches yield the zero value rather than failing. The fixtures are
+// small and the immediate consequence is a connection error naming the source, which is where
+// you would look anyway.
+type SourceConfig map[string]any
```
vikaxsh, line 20: why do we need these helpers?

.github/workflows/security-ci.yaml
mihir-datazip: ITs are decoupled from main code, we should not fail go sec for tests, we can have a separate CI for this

tests/testutils/source_config.go
mihir-datazip: We cannot add source specific config structs in tests, we should ideally just pass on our config files to Dockerfile

.github/workflows/security-ci.yaml
mihir-datazip: Added a separate step that we will move later

vikaxsh (approved):

PR #1060, opened 2026-07-31
fix(lint): enable lint for all modules in olake

mihir-datazip:
# Description

Driver modules were never linted: CI's golangci-lint action only loads the root module, so the eight
`drivers/*` modules accumulated ~80 findings invisibly while CI stayed green.

This PR:
- adds `make olake.lint` — one golangci-lint run over every product module in `go.work`
  (db2 included; `prepare.db2` provisions the IBM clidriver on a fresh machine), with `lint` as its alias
- wires it into the golangci CI job so driver modules are linted on every PR
- fixes all existing findings across the 8 driver modules

No functional changes: mechanical fixes (unused params → `_`, builtin shadowing, naming, error-check
wrappers, misspellings), plus annotations where the code was already correct — intentional
`(nil, nil)` contracts marked `//nolint:nilnil` and deliberate signed→unsigned reinterprets marked
`#nosec G115`, each with a stated reason. Two small deletions: an unused struct field (mssql) and a
dead reassignment (mysql).

Tests modules are deliberately out of scope here; their lint state ships with the IT revamp work.

## Type of change

- [x] Code quality / tooling (no functional change)

# How Has This Been Tested?

- [x] `make lint` passes locally: root + all 8 driver modules, zero findings
- [x] `go build ./... && go vet ./...` clean in every driver module (db2 under its clidriver env)
- [x] CI path exercised: `make olake.lint` provisions the clidriver and lints all modules on a stock runner

## Documentation

- [x] N/A (lint fixes and tooling only)

## Related PR's (If Any):
Tests-module linting (`test.lint`) follows separately here https://github.com/datazip-inc/olake/pull/1053


drivers/db2/internal/cdc.go
```diff
@@ -10,10 +10,10 @@ import (
 // CDC is not supported yet
 func (d *DB2) ChangeStreamConfig() (bool, bool, bool) { return false, false, false }
 
-func (d *DB2) PreCDC(ctx context.Context, streams []types.StreamInterface) error { return nil }
+func (d *DB2) PreCDC(_ context.Context, _ []types.StreamInterface) error { return nil }
 
-func (d *DB2) StreamChanges(ctx context.Context, streamIndex int, metadataStates map[string]any, processFn abstract.CDCMsgFn) (any, error) {
-	return nil, nil
+func (d *DB2) StreamChanges(_ context.Context, _ int, _ map[string]any, _ abstract.CDCMsgFn) (any, error) {
+	return nil, nil //nolint:nilnil // CDC unsupported for DB2: nil state with nil error is the stub contract
```
vaibhav-datazip: instead of passing this no lint directive , shouldn't we just throw an error whenever stream changes for db2 is called, 

the error wont be triggered, because its called from abstract cdc and sync mode cdc will be returned false for db2

drivers/db2/internal/db2.go
```diff
@@ -14,6 +15,8 @@ import (
 	"github.com/datazip-inc/olake/types"
 	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/logger"
+
+	// registers the go_ibm_db DB2 driver with database/sql
 	_ "github.com/ibmdb/go_ibm_db"
```
vaibhav-datazip: maybe we can remove this completely as we are already importing in readbatch.go

drivers/db2/internal/db2.go
```diff
@@ -138,9 +141,17 @@ func (d *DB2) forwardConnections(listener net.Listener, remoteAddr string) {
...
 
+			// The second copy always exits with net.ErrClosed once teardown closes both conns,
+			// so only unexpected copy failures are logged.
 			done := make(chan struct{}, 2)
-			go func() { io.Copy(localConn, remoteConn); done <- struct{}{} }()
-			go func() { io.Copy(remoteConn, localConn); done <- struct{}{} }()
+			tunnelCopy := func(dst, src net.Conn, way string) {
+				if _, err := io.Copy(dst, src); err != nil && !errors.Is(err, net.ErrClosed) {
+					logger.Warnf("ssh tunnel %s copy failed: %s", way, err)
+				}
```
vaibhav-datazip, line 147: earlier we were ignoring all the errors from io.copy, can this introduce any breakage in code or unexpected log spamming

drivers/kafka/internal/cdc.go
```diff
@@ -82,7 +82,7 @@ func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates
 	// A successful recovery stops processing for this reader so the next run starts from the recovered offsets.
 	if isRecoveryPerformed {
 		logger.Infof("reader[%d]: recovery performed for this sync, skipping this reader", readerID)
-		return nil, nil
+		return nil, nil //nolint:nilnil // recovery already consumed this reader's work; nil state keeps the committed offsets
```
vaibhav-datazip: why not just do `return nil, err` , doing this we will be able to remove lint directive

 as err will be nil right

drivers/mongodb/internal/cdc.go
```diff
@@ -163,7 +163,7 @@ func (m *Mongo) handleStreamCatchup(_ context.Context, cursor *mongo.ChangeStrea
 	return token, nil
 }
 
-func (m *Mongo) handleChangeDoc(ctx context.Context, cursor *mongo.ChangeStream, stream types.StreamInterface, startingResumeToken string, OnMessage abstract.CDCMsgFn) error {
+func (m *Mongo) handleChangeDoc(ctx context.Context, cursor *mongo.ChangeStream, stream types.StreamInterface, _ string, OnMessage abstract.CDCMsgFn) error {
```
vaibhav-datazip: we can remove starting resume token from caller as well as it seems to be a dead argument .

drivers/mongodb/internal/mon.go
```diff
@@ -142,7 +141,7 @@ func (m *Mongo) Setup(ctx context.Context) error {
 	if m.sshDialer != nil {
 		opts.SetDialer(m.sshDialer)
 	}
-	opts.SetMaxPoolSize(uint64(m.config.MaxThreads))
+	opts.SetMaxPoolSize(uint64(m.config.MaxThreads)) // #nosec G115 -- MaxThreads is a non-negative config value
```
vaibhav-datazip: can we use this to avoid lint directive 

```
if maxPoolSize := m.config.MaxThreads; maxPoolSize > 0 {
		opts.SetMaxPoolSize(uint64(maxPoolSize))
	}
```

drivers/mssql/internal/backfill.go
```diff
@@ -149,8 +154,8 @@ func (m *MSSQL) GetOrSplitChunks(ctx context.Context, pool *destination.WriterPo
 // Sorting []uint64 with < is cheaper than sorting [][]byte with bytes.Compare.
 func physlocSortKey(fileID, pageID int32) uint64 {
 	var b [8]byte
```
vaibhav-datazip:
```suggestion
func physlocSortKey(fileID uint16, pageID uint32) uint64 {
	var b [8]byte
	binary.LittleEndian.PutUint32(b[0:4], pageID)
	binary.LittleEndian.PutUint16(b[4:6], fileID)
```

drivers/mssql/internal/backfill.go
vaibhav-datazip: can we do this , to fix the datatype , will have to check once this strategy is working fine

```suggestion
		var fileID uint16
		var pageID uint32
```

drivers/mssql/internal/cdc.go
```diff
@@ -119,7 +119,7 @@ func (m *MSSQL) StreamChanges(ctx context.Context, streamIndex int, metadataStat
 
 	// No changes yet
 	if lsnInState >= targetLSN {
-		return nil, nil
+		return nil, nil //nolint:nilnil // no changes past the stored LSN; nil state keeps the previous checkpoint
```
vaibhav-datazip:
```suggestion
		return nil, err
```

drivers/mssql/internal/mssql.go
```diff
@@ -146,7 +145,7 @@ func (m *MSSQL) Close() error {
 
 func setupSSH(sshCfg *utils.SSHConfig) (*ssh.Client, error) {
 	if sshCfg == nil || sshCfg.Host == "" {
-		return nil, nil
+		return nil, nil //nolint:nilnil // SSH is optional: no config means no client and no error
```
vaibhav-datazip: to handle this we can add a check for sshCfg from where this function is being called just like other drivers (preferred), or we can define a err variable and make the change as we will be doing in other

drivers/mysql/internal/cdc.go
```diff
@@ -30,7 +30,7 @@ func (m *MySQL) prepareBinlogConn(ctx context.Context, mySQLGlobalState MySQLGlo
 		ServerID:                mySQLGlobalState.ServerID,
 		Flavor:                  "mysql",
 		Host:                    m.config.Host,
-		Port:                    uint16(m.config.Port),
+		Port:                    uint16(m.config.Port), // #nosec G115 -- TCP ports fit uint16
```
vaibhav-datazip: similarly we can remove directive from here as well

drivers/mysql/internal/cdc.go
```diff
@@ -57,16 +57,14 @@ func (m *MySQL) PreCDC(ctx context.Context, streams []types.StreamInterface) err
 		if err != nil {
 			return fmt.Errorf("failed to get current binlog position: %s", err)
 		}
-		m.state.SetGlobal(MySQLGlobalState{ServerID: uint32(1000 + time.Now().UnixNano()%4294966295), State: binlog.Binlog{Position: binlogPos}})
+		m.state.SetGlobal(MySQLGlobalState{ServerID: uint32(1000 + time.Now().UnixNano()%4294966295), State: binlog.Binlog{Position: binlogPos}}) // #nosec G115 -- modulo bounds the value below 2^32
```
vaibhav-datazip: can we do something like this 


```
func newServerID() uint32 {
	offset := time.Now().UnixNano() % (math.MaxUint32 - minServerID)
	if offset < 0 || offset > math.MaxUint32-minServerID {
		return minServerID
	}
	return minServerID + uint32(offset)
}
```

drivers/mysql/internal/config_test.go
```diff
@@ -415,7 +415,7 @@ func TestConfig_buildTLSConfig(t *testing.T) {
 			name:       "fails with invalid CA pem",
 			config:     &Config{SSLConfiguration: &utils.SSLConfig{Mode: utils.SSLModeVerifyCA, ServerCA: "not-a-pem"}},
 			expectErr:  true,
-			assertions: func(t *testing.T, _ *tls.Config) {},
+			assertions: func(_ *testing.T, _ *tls.Config) {},
```
vaibhav-datazip: when expect err is true , maybe we can remove the assertion field altogether

drivers/db2/internal/db2.go
mihir-datazip: no breakage, also this won't spammy as we already to a warn log on dial faliure

drivers/kafka/internal/cdc.go
mihir-datazip: Done

drivers/db2/internal/cdc.go
mihir-datazip: Added

drivers/db2/internal/db2.go
mihir-datazip: Removed

drivers/mongodb/internal/cdc.go
mihir-datazip: Removed

drivers/mongodb/internal/mon.go
mihir-datazip: Okay, there is already handling of negative values so this can be done

drivers/mysql/internal/config_test.go
mihir-datazip: Done

drivers/mssql/internal/backfill.go
mihir-datazip: Done

drivers/mssql/internal/backfill.go
mihir-datazip: Done

drivers/mssql/internal/cdc.go
mihir-datazip: Done

drivers/mssql/internal/mssql.go
mihir-datazip: Added

drivers/mysql/internal/cdc.go
mihir-datazip: Okay removed

drivers/mysql/internal/cdc.go
mihir-datazip: Added

hash-data:
just thinking if we can add detection for defer in loop?

mihir-datazip:
> just thinking if we can add detection for defer in loop?

@hash-data we use `github.com/golangci/golangci-lint/cmd/golangci-lint@latest` and in that if I have to add a new rule, I have to include all default 23 rules by hand.

A better solution will be to upgrade to v2 golangci-lint which has a flag to have the default rules with incremental custom rules.

Should I do this change in the same PR or should the v2 upgrade of golangci-lint go in new PR?

Makefile
```diff
@@ -5,6 +5,9 @@ GOPATH = $(shell go env GOPATH)
 GO_VERSION = $(shell awk '/^go / {print "go"$$2; exit}' go.mod)
 # A driver is any drivers/ subdir with its own go.mod (excludes util folders like abstract).
 DRIVERS = $(notdir $(patsubst %/go.mod,%,$(wildcard drivers/*/go.mod)))
+# Product modules only: tests/ (and its lib glue) carry their own lint state and are linted on
+# their own branch; drop the filter once the tests workspace split lands.
+ROOT_MODULES := $(filter-out $(CURDIR)/lib $(CURDIR)/tests/%,$(shell go list -m -f '{{.Dir}}'))
```
vaibhav-datazip: can we remove the the filtering now

drivers/mysql/internal/mysql.go
```diff
@@ -275,19 +275,19 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
...
 			}
 		case "unsigned mediumint", "unsigned int", "unsigned integer":
 			if v, ok := value.(int32); ok {
-				value = uint32(v)
+				value = uint32(v) // #nosec G115 -- deliberate two's-complement reinterpretation for unsigned MySQL columns
 			}
 		case "unsigned bigint":
 			if v, ok := value.(int64); ok {
-				value = uint64(v)
+				value = uint64(v) // #nosec G115 -- deliberate two's-complement reinterpretation for unsigned MySQL columns
```
vaibhav-datazip, line 291: is there a way to remove these directives as well,

hash-data:
> > just thinking if we can add detection for defer in loop?
> 
> @hash-data we use `github.com/golangci/golangci-lint/cmd/golangci-lint@latest` and in that if I have to add a new rule, I have to include all default 23 rules by hand.
> 
> A better solution will be to upgrade to v2 golangci-lint which has a flag to have the default rules with incremental custom rules.
> 
> Should I do this change in the same PR or should the v2 upgrade of golangci-lint go in new PR?

let us do it separately then we can create issue as good first issue

drivers/mysql/internal/mysql.go
mihir-datazip: This can cause behaviour change and our current ITs don't have edge cases for int overflow.
I say we leave it as it is for now.

Also on adding the tests, I noticed `unsigned mediumint` seem to have some bug     
```
test_utils.go:479: Executing Parquet query: SELECT * FROM `mysql_test_table_olake_view_1785946679399921000` WHERE `_op_type` = 'c'
    test_utils.go:479: 
                Error Trace:    /Users/mihir/project/olake/tests/testutils/test_utils.go:1855
                                                        /Users/mihir/project/olake/tests/testutils/test_utils.go:479
                                                        /Users/mihir/project/olake/tests/testutils/test_utils.go:681
                Error:          Not equal: 
                                expected: 16777215
                                actual  : -1
                Test:           TestMySQLIntegration/Sync/Parquet_Full_load_+_CDC_tests/CDC_-_insert
```

Makefile
mihir-datazip: Good catch, not needed now...

.githooks/pre-commit
```diff
@@ -1,11 +1,32 @@
...
-# Check if the linter passed
-if [ $? -ne 0 ]; then
-  echo "❌ golangci-lint found issues. Commit aborted."
-  exit 1
-fi
\ No newline at end of file
+# Lints only the half that changed: any change under tests/ runs test.lint, .go changes
+# anywhere else run olake.lint. Staged or not -- golangci-lint reads the working tree.
+
+changed=$(git diff --name-only HEAD)
```
vaibhav-datazip, line 6: I think we should run lint only on staged changes , not unstaged once

.githooks/pre-commit
```diff
@@ -1,11 +1,32 @@
...
+status=0
+if [ -n "$product" ]; then
+	echo "Running lint..."
+	make olake.lint || status=1
+else
+	echo "⏭️  no .go changes outside tests/, skipping olake.lint"
+fi
+if [ -n "$tests" ]; then
+	echo "Running lint on tests modules..."
+	make test.lint || status=1
```
vaibhav-datazip, line 24: although lint dosent take much time to run , should we add this complexity ?
If so shouldn't we divide it for driver specific thing as well

.githooks/pre-commit
mihir-datazip: Done the change... If you think that is complex maybe we can do a lint for both modules every time...

.githooks/pre-commit
mihir-datazip: The division is there because these 2 are different modules with their own codebase

vaibhav-datazip (approved):

PR #1061, opened 2026-08-01
chore: added s3 flag

saksham-datazip:
# Description

This PR makes NFS optional for the UI/Helm workflow by adding S3-backed configuration handling for CLI commands.

When a command receives an `s3://` path, the CLI now:
- Detects the S3 URI.
- Reads S3 credentials and configuration from environment variables.
- Downloads the referenced configuration files to a local temporary directory.
- Rewrites the command to use the downloaded local paths.
- Executes the command normally without changing connector behavior.
- Uploads generated or updated files (such as `streams.json`, `state.json`, `stats.json`, and `difference_streams.json`) back to the corresponding S3 job folder after successful execution.

This allows the UI and Helm deployment to use S3 as the shared storage backend instead of requiring NFS while keeping the connector execution flow unchanged.

Fixes # (issue)

## Type of change

- [x] normal sync and check commands working from ui
- [x] clear destination working from ui
- [x] Ran two jobs at once with heavy log load(10logs/second)
- [x] Backward compatibility for nfs via first running normal sync on staging for all three repos and than running them on my branch
- [x] Ran a sync and added commands such that worker oom happens in two times and their was no log loss

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings
N/A

## Documentation

## Related PR's (If Any):
Olake helm=> https://github.com/datazip-inc/olake-helm/pull/189
Olake ui=> https://github.com/datazip-inc/olake-ui/pull/418


protocol/discover.go
```diff
@@ -43,12 +43,19 @@ var discoverCmd = &cobra.Command{
 
 		return nil
 	},
-	RunE: func(cmd *cobra.Command, _ []string) error {
+	RunE: func(cmd *cobra.Command, _ []string) (err error) {
+		defer func() {
```
vikaxsh, line 47: what if we wrap this in a common fn and eliminate 3x defer

protocol/s3.go
```diff
@@ -0,0 +1,101 @@
...
+	if err := resolveS3PathFlag(ctx, &destinationConfigPath); err != nil {
+		return err
+	}
+	if err := resolveS3PathFlag(ctx, &streamsPath); err != nil {
+		return err
+	}
+	if err := resolveS3PathFlag(ctx, &statePath); err != nil {
+		return err
+	}
+	if err := resolveS3PathFlag(ctx, &differencePath); err != nil {
```
vikaxsh, line 33: use a loop here

protocol/s3.go
```diff
@@ -0,0 +1,101 @@
...
+	configFolder := viper.GetString(constants.ConfigFolder)
+	if configFolder == "" {
+		return nil
+	}
+
+	files := []struct {
+		local string
+		name  string
+	}{
+		{filepath.Join(configFolder, "streams.json"), "streams.json"},
```
vikaxsh, line 71: wee can use constants.StreamsPath instead of rebuilding

protocol/s3.go
```diff
@@ -0,0 +1,101 @@
+package protocol
```
vikaxsh, line 1: Can we move these to utils? Since these aren't commands, it would be better to keep the protocol layer thin and shift as much of this logic as possible into utils

utils/s3_utils.go
```diff
@@ -0,0 +1,177 @@
...
+
+	if region := envFirst("OLAKE_S3_REGION", "AWS_REGION"); region != "" {
+		configOpts = append(configOpts, config.WithRegion(region))
+	}
+
+	accessKey := envFirst("OLAKE_S3_ACCESS_KEY_ID", "AWS_ACCESS_KEY_ID")
+	secretKey := envFirst("OLAKE_S3_SECRET_ACCESS_KEY", "AWS_SECRET_ACCESS_KEY")
+	if accessKey != "" && secretKey != "" {
+		sessionToken := envFirst("OLAKE_S3_SESSION_TOKEN", "AWS_SESSION_TOKEN")
+		configOpts = append(configOpts, config.WithCredentialsProvider(
```
vikaxsh, line 41: for using s3 a remote fs we should follow strict protocol for env vars

utils/s3_utils.go
```diff
@@ -0,0 +1,177 @@
...
+type S3PathMapping struct {
+	OriginalPath string
+	LocalPath    string
+	IsS3         bool
+	Bucket       string
+	Key          string
+}
+
+// newS3Client creates a new S3 client.
+func newS3Client(ctx context.Context) (*s3.Client, error) {
```
vikaxsh, line 29: Can we avoid creating the S3 client separately for upload and resolve? We can create it once and reuse it

protocol/s3.go
saksham-datazip: Changed

protocol/root.go
```diff
@@ -44,14 +45,15 @@ var (
 var RootCmd = &cobra.Command{
 	Use:   "olake",
 	Short: "root command",
-	RunE: func(cmd *cobra.Command, args []string) error {
-
+	PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
```
vikaxsh, line 49: because of current twice init, utils.ResolveS3Paths downloads configs twice, need to fix this

protocol/sync.go
```diff
@@ -91,6 +91,8 @@ var syncCmd = &cobra.Command{
 		return nil
 	},
 	RunE: func(cmd *cobra.Command, _ []string) (err error) {
+		defer utils.FinalizeS3Upload(cmd.Context(), &err, noSave)
```
vikaxsh: we can move this defer to root from sync, discover, clear

saksham-datazip: Moved to common root.go

protocol/root.go
saksham-datazip: Here we previously had a discussion why we are doing it twice and you were not able to recall that so we just left it as it is at that time but i changed it now and tested for these two commands as well along with testing for normal sync, discover, spec, check command.
```
cd /Users/apple/Desktop/olake && ./build.sh driver-mssql badcmd     
cd /Users/apple/Desktop/olake && ./build.sh driver-mssql sync --help
```

vikaxsh: Before the PR, setup ran only once by accident , root `RunE` fired on the first Execute, and was skipped when the actual subcommand ran.
The PR moved it to `PersistentPreRunE` + `EnableTraverseRunHooks`, so it now runs on both Executes.

protocol/root.go
```diff
@@ -117,6 +129,11 @@ func CreateRootCommand(_ bool, driver any) *cobra.Command {
 // Drivers may have source-specific checkpointing constraints, so this helper
 // should not be used as a substitute for driver-level cancellation safety.
 func signalAwareRootContext(parent context.Context) context.Context {
+	// CreateRootCommand runs before Execute, so Cobra has not yet defaulted
+	// nil to Background().
+	if parent == nil {
```
vikaxsh, line 134: can we directly pass context.Background directly as
CreateRootCommand is the only caller

saksham-datazip: this signalAwareRootContext function has three callers one  CreateRootCommand and other are from unit_tests so i dont think we should change it.

utils/s3_utils.go
```diff
@@ -0,0 +1,181 @@
...
+			}
+			*flagPath = configFolder + "/" + path.Base(*flagPath)
+		}
+	}
+}
+
+// ResolveS3Paths initializes storage and downloads s3:// flag paths into S3LocalPath().
+func ResolveS3Paths(ctx context.Context, flagPaths []*string, telemetryFiles []*string) error {
+	configFolder := os.Getenv(constants.ConfigFolder)
+	ApplyConfigFolder(configFolder, flagPaths, telemetryFiles)
```
mihir-datazip: **F1 (medium-high, backward compat): this env override is not gated on S3 mode, and it silently overrides explicit CLI flags.**

`ApplyConfigFolder` throws away the directory of `--config`, `--destination`, `--streams`, `--state` and `--difference`, keeping only the basename — and this runs on **every** invocation, not just when `OLAKE_STORAGE_MODE=s3`.

There is no `viper.AutomaticEnv` / `BindEnv` anywhere in the tree, so `CONFIG_FOLDER` was previously **never read from the environment**. This PR creates a new env contract that overrides explicit CLI arguments, silently, with no log line.

Failure scenario: a non-S3 user or CI job exports `CONFIG_FOLDER=/mnt/config` (a natural thing to set — it is the viper key name), then runs `olake sync --config /etc/olake/prod/config.json`. The flag becomes `/mnt/config/config.json`. Best case a confusing unmarshal error; worst case it silently loads a *different* config with the same basename.

Second scenario: `--config a/config.json --destination b/config.json` both collapse to `$CONFIG_FOLDER/config.json`, so the destination config becomes the source config.

Suggested fix: gate on `s3util.IsS3Job()` — which is also what the resolved "strict protocol for env vars" comment was asking for — and log the rewrite when it happens.

utils/s3_utils.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	resp, err := s3util.GetObject(ctx, bucket, key)
+	if err != nil {
+		return S3PathMapping{}, fmt.Errorf("failed to download %s: %s", s3Path, err)
+	}
+	defer resp.Body.Close()
+
+	localPath := filepath.Join(S3LocalPath(), path.Base(key))
+	file, err := os.Create(localPath)
```
mihir-datazip: **F5 (medium, security): downloaded configs land world-readable at a predictable path and are never cleaned up.**

`os.Create` opens with mode `0666 &^ umask`, typically **0644**. These files are the source config (DB host/user/password) and the destination config (S3 keys, catalog creds).

1. Any local user can read `/tmp/config.json`.
2. The name is predictable inside a `1777` directory, and `os.Create` follows symlinks — a pre-planted `/tmp/config.json -> /elsewhere` gets written as the olake user. The sticky bit stops others *deleting* your file, not *creating* the symlink first.
3. No cleanup, so credentials persist in /tmp after exit.

Fair caveat: in the single-user worker container the real exposure is low, and on macOS `os.TempDir()` is already a per-user private dir. The sharp edge is Linux CLI runs on shared/dev hosts.

Suggested fix:
```go
dir, err := os.MkdirTemp("", "olake-s3-")   // once per process, removed on exit
...
f, err := os.OpenFile(localPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
```
`0o600` is already the idiom here (`utils/telemetry/telemetry_test.go:516`).

utils/s3_utils.go
```diff
@@ -0,0 +1,181 @@
...
+const s3URIPrefix = "s3://"
+
+// S3PathMapping tracks an S3 URI and its downloaded local copy.
+type S3PathMapping struct {
+	OriginalPath string
+	LocalPath    string
+	IsS3         bool
+	Bucket       string
+	Key          string
+}
```
mihir-datazip, line 27: Only `.LocalPath` is used (`s3_utils.go:139`). `OriginalPath`, `IsS3`, `Bucket` and `Key` have zero readers anywhere 

Also `S3PathMapping`, `ResolveS3Path`, `S3LocalPath`, `IsS3Path`, `ParseS3URI` and `ApplyConfigFolder` are all exported but have only in-package callers. We should encapsulate the functions which are for package level and only export functions for callers

utils/s3_utils.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	*flagPath = s3PathMap.LocalPath
+	return nil
+}
+
+// FinalizeS3Upload uploads local artifacts after a successful run. Callers invoke it
+// from PersistentPostRunE, which Cobra skips when RunE failed, so a partial local
+// write cannot overwrite remote streams/state.
+func FinalizeS3Upload(ctx context.Context, noSave bool) error {
+	if noSave || s3util.JobBucket == "" {
```
mihir-datazip: **F2 (low): silent half-enabled S3 mode.**

If `OLAKE_STORAGE_MODE=s3` but no flag resolves to an `s3://` URI (e.g. `CONFIG_FOLDER` unset, so `ApplyConfigFolder` returns early and the paths stay local), `JobBucket` is never set. This guard then makes the whole finalize a silent no-op — while `logger.Init(true)` has already switched to stdout-JSON and **disabled file logging entirely**.

So the run produces no `olake.log` and uploads nothing, with no warning. Worth a hard error at startup, or at minimum a warn when `IsS3Job()` is true but nothing resolved to S3.

utils/s3/s3.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	return s3Client, nil
+}
+
+// GetObject downloads an object and records the job prefix from the first key.
+func GetObject(ctx context.Context, bucket, key string) (*awss3.GetObjectOutput, error) {
+	if JobBucket == "" {
+		JobBucket = bucket
+		JobPrefix = path.Dir(key)
+	}
```
mihir-datazip: **F4 (low, design): `GetObject` latching package globals is a hidden side effect.**

A function named `GetObject` silently decides where *all* artifacts get uploaded, based on whichever key happens to be fetched first. It is exported, so any future caller reorders the upload destination, and the dependency is invisible at the call site (`protocol/root.go:51`).

It also latches **before** the `getS3Client()` nil-check below, so a call that fails with `s3 storage not initialized` still leaves the globals dirty.

Suggested fix: set the job location once, explicitly, in `ResolveS3Paths` from the resolved `CONFIG_FOLDER` URI, and keep `GetObject` a pure getter.

utils/s3/s3.go
```diff
@@ -0,0 +1,112 @@
...
+		configOpts = append(configOpts, config.WithRegion(region))
+	}
+
+	accessKey := os.Getenv(constants.EnvS3AccessKeyID)
+	secretKey := os.Getenv(constants.EnvS3SecretAccessKey)
+	if accessKey != "" && secretKey != "" {
+		configOpts = append(configOpts, config.WithCredentialsProvider(
+			credentials.NewStaticCredentialsProvider(accessKey, secretKey, os.Getenv(constants.EnvS3SessionToken)),
+		))
+	}
```
mihir-datazip, line 44: This duplicates `pkg/objstorage/s3.go:36-72` almost line for line. Worth a look if we can use common code if possible

utils/logger/logger.go
```diff
@@ -174,25 +177,34 @@ func StatsLogger(ctx context.Context, statsFunc func() (int64, int64, int64, int
 				if err := FileLogger(stats, "stats", ".json"); err != nil {
 					Fatalf("failed to write stats in file: %s", err)
 				}
+				if s3.IsS3Job() {
+					statsPath := filepath.Join(viper.GetString(constants.ConfigFolder), "stats.json")
+					if err := s3.UploadFileToS3(ctx, statsPath, s3.JobBucket, path.Join(s3.JobPrefix, "stats.json")); err != nil {
+						Debugf("failed to upload stats.json: %s", err)
+					}
+				}
```
mihir-datazip, line 185: This ticker goroutine and `FinalizeS3Upload` (`utils/s3_utils.go:162`) both PUT the same key, `JobPrefix/stats.json`.

`StatsLogger`'s ctx is `cmd.Context()` (`protocol/sync.go:152`) which is **not** cancelled when `sync`'s `RunE` returns. So this goroutine is still live during `PersistentPostRunE`.

This way S3 can end up holding the pre-final stats. I think we can drop `stats.json` from one of the two (`FinalizeS3Upload` already covers it), or stop the goroutine before finalize.

utils/telemetry/telemetry.go
```diff
@@ -39,6 +39,13 @@ const (
 	maxEventPropsFileSize = 1 << 20 // 1 MiB
 )
 
+// TelemetryFiles returns optional job-dir files S3 mode should fetch so Init can read them.
+func TelemetryFiles() []*string {
+	eventProps := eventPropsFile
+	userID := userIDFile + ".txt"
+	return []*string{&eventProps, &userID}
+}
```
mihir-datazip, line 47: We return pointers here for settable flags, but you are returning address of `eventProps`, which is function local variable, not the global one...

saksham-datazip: These are just local variables to reuse resolveS3Path but lets discuss

utils/logger/logger.go
saksham-datazip: Regarding the `stats.json` file, it is intended to be a live progress/monitoring file. That’s why it is updated and uploaded every 2 seconds by the ticker. For the final batch, it may not necessarily get uploaded if the sync completes before the next ticker interval.

I’ve therefore removed `stats.json` from `FinalizeS3Upload` so that its behavior remains consistent with the local/NFS implementation, where the file is also only updated by the periodic ticker and has no separate finalization step.

utils/s3/s3.go
saksham-datazip: This same function exists in worker and ui repo so lets discuss.

PR #1068, opened 2026-08-04
fix(postgres): dynamically inject cdc metadata columns based on sync mode

krishanu7:
## Description

This PR fixes an issue where the CDC metadata columns (`_cdc_timestamp` and `_cdc_lsn`) were incorrectly exposed in the destination schema during **`incremental`** and **`full_refresh`** sync modes when CDC was enabled.

### Changes
- Removed `_cdc_timestamp` and `_cdc_lsn` from the schema generated during the **discover** phase.
- Ensured `streams.json` remains free of CDC-specific metadata columns.
- Dynamically inject `_cdc_timestamp` and `_cdc_lsn` into the stream schema **only at runtime** when the effective sync mode is **`cdc`** or **`strictcdc`**.

This ensures that CDC metadata is present only when required and is not leaked into non-CDC sync modes.

Fixes #<issue-number>

---

## Type of change

---

# How Has This Been Tested?

### 1. Verify `discover` omits CDC metadata columns

**Command**
```bash
go run drivers/postgres/main.go discover \
  --config test_source.json > streams.json
```

**Expected `streams.json` output**
```json
"selected_columns": {
  "columns": [
    "_op_type",
    "_olake_id",
    "_olake_timestamp",
    "id",
    "name"
  ],
  "sync_new_columns": true
}
```

**Verified**
- `_cdc_timestamp` is **not present**
- `_cdc_lsn` is **not present**

---

### 2. Verify `full_refresh` sync does not write CDC metadata

**Command**
```bash
go run drivers/postgres/main.go sync \
  --config test_source.json \
  --catalog streams_full_refresh.json \
  --destination test_dest.json \
  --state state.json
```

**Destination Parquet schema**
```text
Schema properties:
- _olake_id
- _olake_timestamp
- _op_type
- id
- name
```

**Verified**
- `_cdc_timestamp` is **not present**
- `_cdc_lsn` is **not present**

---

### 3. Verify `cdc` sync injects CDC metadata at runtime

**Command**
```bash
go run drivers/postgres/main.go sync \
  --config test_source.json \
  --catalog streams_cdc.json \
  --destination test_dest.json \
  --state state.json
```

**Destination Parquet schema**
```text
Schema properties:
- _cdc_lsn
- _cdc_timestamp
- _olake_id
- _olake_timestamp
- _op_type
- id
- name
```

**Verified**
- `_cdc_timestamp` is present
- `_cdc_lsn` is present
- CDC metadata is injected only for CDC-based sync modes


protocol/sync.go
```diff
@@ -249,6 +254,18 @@ func classifyStreams(catalog *types.Catalog, streams []*types.Stream, state *typ
...
+			if elem.StreamMetadata.SelectedColumns != nil {
+				elem.StreamMetadata.SelectedColumns.Columns = append(elem.StreamMetadata.SelectedColumns.Columns, constants.CdcTimestamp)
+			}
+			// Postgres specific CDC LSN
+			if connector.Type() == "postgres" {
+				elem.Stream.UpsertField("_cdc_lsn", types.String, true, true)
+				if elem.StreamMetadata.SelectedColumns != nil {
+					elem.StreamMetadata.SelectedColumns.Columns = append(elem.StreamMetadata.SelectedColumns.Columns, "_cdc_lsn")
+				}
+			}
```
ImDoubD-datazip, line 269: driver specific knowledge should not exist in sync protocol layer.

types/type_schema.go
```diff
@@ -136,12 +136,10 @@ func (t *TypeSchema) GetProperty(column string) (bool, *Property) {
 }
 
 func (t *TypeSchema) ToParquet(defaultColumns bool, stream StreamInterface) *parquet.Schema {
-	// keeping default columns parquet schema for backward compatibility for olake columns
```
ImDoubD-datazip, line 139: why this comment removed?

protocol/sync.go
```diff
@@ -249,6 +254,18 @@ func classifyStreams(catalog *types.Catalog, streams []*types.Stream, state *typ
 		switch elem.Stream.SyncMode {
 		case types.CDC, types.STRICTCDC:
 			classifications.CDCStreams = append(classifications.CDCStreams, elem)
+			// Inject CDC metadata columns dynamically because they are omitted during discover
+			elem.Stream.UpsertField(constants.CdcTimestamp, types.TimestampMicro, true, true)
+			if elem.StreamMetadata.SelectedColumns != nil {
+				elem.StreamMetadata.SelectedColumns.Columns = append(elem.StreamMetadata.SelectedColumns.Columns, constants.CdcTimestamp)
+			}
```
ImDoubD-datazip, line 262: data loss case (can't happen currently):
if discover would have allowed streams with 0 columns, and cdc is configured, then in streams.json, cdc columns will come. now if the user adds columns in that table and inserts some data in it, then runs sync with the same streams.json created earlier, then destination will only have these cdc columns, not the new added column in that sync.

but the thing is discover does not fetch tables with 0 column.

protocol/sync.go
```diff
@@ -249,6 +254,18 @@ func classifyStreams(catalog *types.Catalog, streams []*types.Stream, state *typ
...
+			// Postgres specific CDC LSN
+			if connector.Type() == "postgres" {
```
ImDoubD-datazip, line 264: what about other drivers like mongodb, mysql, mssql, they also have specific cdc columns which is required?

drivers/abstract/abstract.go
ImDoubD-datazip, line 1: rather than doing this, where we have to do driver specific changes in sync protocol, better suggestion can be:
- in abstract.DriverInterface, add a CDCColumns(), which will return driver-specific CDC metadata columns any driver emits.
- move each driver's existing `upsertfield` calls out of the produce schema into it.

Driver | Returns
-- | --
postgres | {waljs.CDCLSN: String}
mysql | {binlog.CDCBinlogFileName: String, binlog.CDCBinlogFilePos: Int64}
mongodb | {CDCResumeToken: String}
mssql | {CDCStartLSN: String, CDCSeqVal: String}
kafka, s3, db2, oracle | nil

- call this CDCColumns() from abstract read function. kafka should be exempted. also before doing this, confirm yourself is read function the right place to have this change.

This above can be one of the ways to solve this. If anything better you can find, that also will be nice.

krishanu7:
## Testing done

### Unit tests

```bash
go test ./drivers/abstract/ ./protocol/ ./types/ \
  ./drivers/postgres/internal/ ./drivers/mysql/internal/ \
  ./drivers/mongodb/internal/ ./drivers/mssql/internal/ \
  ./drivers/s3/internal/ -count=1
```

All passed.

### Postgres (CDC enabled on source, Parquet local destination)

Using `olake_postgres-test` on port `5433`, schema `pr1068`:

- **discover** — `_cdc_timestamp` and `_cdc_lsn` not in `streams.json` schema or `selected_columns`
- **full_refresh** — Destination Parquet: `_olake_id`, `_olake_timestamp`, `_op_type`, `id`, `name`, `created_at` (no CDC columns)
- **incremental** — Same as full_refresh (no CDC columns)
- **cdc** — Destination includes `_cdc_timestamp` and `_cdc_lsn`
- **Mixed catalog** (`cdc_tbl`=cdc, `fr_tbl`=full_refresh, `inc_tbl`=incremental in one sync) — CDC columns only on `cdc_tbl`; `fr_tbl` and `inc_tbl` unchanged

### MySQL (CDC enabled on source, Parquet local destination)

Using `olake_mysql-test` on port `3306`, table `pr1068_mysql`:

- **discover** — `_cdc_timestamp`, `_cdc_binlog_file_name`, `_cdc_binlog_file_pos` not in catalog
- **full_refresh** — No CDC columns in destination
- **cdc** — Destination includes `_cdc_timestamp`, `_cdc_binlog_file_name`, `_cdc_binlog_file_pos`

---


## Known limitations 

1. **Re-discover required for old catalogs** — `streams.json` created before this fix may still list CDC columns. Incremental/full_refresh on those files can still leak CDC metadata until `discover` is re-run and the catalog is replaced.

2. **Existing Iceberg tables are not cleaned up** — Tables already created with leaked CDC columns will keep them. Schema evolution is additive only; re-discover does not drop columns. Options: recreate the table, use a new destination DB, or ignore null legacy CDC columns.

3. **`SelectedColumns` append is not de-duped** — Old catalogs that already include `_cdc_lsn` / `_cdc_timestamp` may get duplicates appended on CDC sync. Harmless in practice (filtering uses a set), but worth cleaning up in a follow-up.

drivers/abstract/abstract.go
```diff
@@ -195,6 +195,10 @@ func (a *AbstractDriver) Read(ctx context.Context, pool *destination.WriterPool,
 		a.GlobalConnGroup = utils.NewCGroupWithLimit(ctx, a.driver.MaxConnections())
 	}
 
+	if len(cdcStreams) > 0 && a.driver.CDCSupported() {
+		injectCDCColumns(a.driver, cdcStreams)
```
vaibhav-datazip, line 199: you can move this condition in below if block

drivers/abstract/abstract.go
```diff
@@ -234,6 +238,30 @@ func (a *AbstractDriver) Read(ctx context.Context, pool *destination.WriterPool,
...
+		for name, typ := range driverColumns {
+			s.UpsertField(name, typ, true, true)
+		}
+
+		selectedCols := stream.Self().StreamMetadata.SelectedColumns
+		if selectedCols != nil && len(selectedCols.Columns) > 0 {
+			selectedCols.Columns = append(selectedCols.Columns, constants.CdcTimestamp)
+			for name := range driverColumns {
+				selectedCols.Columns = append(selectedCols.Columns, name)
+			}
```
vaibhav-datazip, line 260: This won't be need, as we have a check in iceberg writer side that olake metadata columns will always be present 


this is in types/type_schema.go, you can add it in `toParquet` function as well
```
if !prop.OlakeColumn && (!isSelected(outName) || (defaultColumns && !includeSet.Exists(colName))) {
			return true
		}
```

protocol/sync.go
```diff
@@ -194,6 +194,10 @@ var syncCmd = &cobra.Command{
 }
 
 func classifyStreams(catalog *types.Catalog, streams []*types.Stream, state *types.State) (*StreamClassification, error) {
+	return classifyStreamsInternal(catalog, streams, state)
+}
+
+func classifyStreamsInternal(catalog *types.Catalog, streams []*types.Stream, state *types.State) (*StreamClassification, error) {
```
vaibhav-datazip, line 200: please check if this wrapper can be removed

drivers/abstract/abstract.go
```diff
@@ -234,6 +238,30 @@ func (a *AbstractDriver) Read(ctx context.Context, pool *destination.WriterPool,
...
 }
 
+// injectCDCColumns adds _cdc_timestamp and driver-specific CDC metadata to stream schemas at sync time.
+// CDC columns are omitted during discover so they do not appear in streams.json for non-CDC modes.
+func injectCDCColumns(driver DriverInterface, streams []types.StreamInterface) {
+	driverColumns := driver.CDCColumns()
+	for _, stream := range streams {
+		if stream == nil {
+			continue
+		}
```
vaibhav-datazip, line 248: this check might be dead please check

drivers/abstract/interface.go
```diff
@@ -35,6 +35,9 @@ type DriverInterface interface {
 	StreamIncrementalChanges(ctx context.Context, stream types.StreamInterface, cb BackfillMsgFn) error
 	// specific to cdc
 	CDCSupported() bool
+	// CDCColumns returns driver-specific CDC metadata columns (name -> type).
+	// Return nil when the driver has none (Kafka, S3, DB2, Oracle).
+	CDCColumns() map[string]types.DataType
```
vaibhav-datazip, line 40:
```suggestion
	CDCMetadataColumns() map[string]types.DataType
```

PR #1070, opened 2026-08-04
fix(postgres): pick the slot LSN source by recovery state so CDC works on a read-replica

### LINKED ISSUE #1069: Postgres CDC cannot run against a read-replica: `pg_current_wal_lsn()` errors during recovery
### What happens

Point the Postgres driver at a physical standby and every CDC run dies before any data moves:

```
failed to get slot position: ERROR: recovery is in progress (SQLSTATE 25006)
```

### Why

`pkg/waljs/replicator.go` builds the slot lookup with `pg_current_wal_lsn()`:

```go
ReplicationSlotTempl = "SELECT plugin, slot_type, confirmed_flush_lsn, pg_current_wal_lsn() as current_lsn FROM pg_replication_slots WHERE slot_name = '%s'"
```

That is a WAL control function, and it raises `recovery is in progress` on a standby. The standby equivalent is `pg_last_wal_replay_lsn()`. Nothing branches on `pg_is_in_recovery()`, so there is no path through this that works on a replica.

`GetSlotPosition` runs from `Postgres.PreCDC` (`drivers/postgres/internal/cdc.go`) on every CDC run, and again in the LSN wait loop in `replicator.go`, so it fires immediately rather than partway through a sync.

### Why it matters

Postgres 16 added logical decoding on standbys, so the slot can live on a replica and keep decode load off a busy primary. The rest of the driver already looks compatible: `pg_replication_slot_advance()` is allowed on a standby for slots local to it, and `IdentifySystem` returns the replay position there. This one expression is the whole blocker.

The MSSQL driver picked up read-replica support in #949 and #991. This is the Postgres counterpart.

### Reproduction

PostgreSQL 18, a streaming replica built with `pg_basebackup -R`, `wal_level=logical` and `hot_standby_feedback=on` on both ends, slot created on the standby:

```
-- on the standby
SELECT plugin, slot_type, confirmed_flush_lsn, pg_current_wal_lsn() as current_lsn
FROM pg_replication_slots WHERE slot_name = 'probe';

ERROR:  recovery is in progress
HINT:  WAL control functions cannot be executed during recovery.
```

### Suggested fix

Choose the LSN source by recovery state. One expression, no Go changes:

```sql
SELECT plugin, slot_type, confirmed_flush_lsn,
       CASE WHEN pg_is_in_recovery()
            THEN pg_last_wal_replay_lsn()
            ELSE pg_current_wal_lsn()
       END AS current_lsn
FROM pg_replication_slots WHERE slot_name = '%s'
```

`pg_is_in_recovery()` is volatile, so the `CASE` is not constant-folded and only the applicable branch is evaluated.

Results on the same pair:

```
-- standby, patched query
pgoutput|logical|0/3000060|0/301A880

-- primary, original query
pgoutput|logical|0/301A8B8|0/301A8B8
-- primary, patched query
pgoutput|logical|0/301A8B8|0/301A8B8
```

The standby returns a usable position instead of erroring, and the primary is unchanged.

I also checked `pg_replication_slot_advance('probe', ...)` on the standby, since `AdvanceLSN` calls it: it succeeds, so that path needs no change.

### Environment

- OLake v0.9.0, and `master` as of 2026-08-04 (identical string on both)
- PostgreSQL 18, pgoutput, slot and publication pre-created

Happy to send a PR with this if you want it. I have verified the SQL on both ends of a real replica pair but have not yet run a full OLake sync against a standby, so treat that part as unproven until someone does.

erikcw:
# Description

`ReplicationSlotTempl` in `pkg/waljs/replicator.go` calls `pg_current_wal_lsn()`. That is a WAL control function and it raises `recovery is in progress` on a physical standby, so Postgres CDC against a read-replica fails in `PreCDC` before any data moves. Nothing branches on `pg_is_in_recovery()`, so there was no path through it that worked on a replica.

Postgres 16 allows a logical slot to live on a standby, which is the natural place to keep decode load off a busy primary. The rest of the driver already works there: `pg_replication_slot_advance()` (used by `AdvanceLSN`) is permitted on a standby for slots local to it, and `IdentifySystem` returns the replay position. This one expression was the blocker.

The fix selects `pg_last_wal_replay_lsn()` when `pg_is_in_recovery()` is true. `pg_is_in_recovery()` is volatile, so the `CASE` is not constant-folded and only the applicable branch is evaluated. Behaviour on a primary is unchanged.

Fixes #1069

## Type of change

# How Has This Been Tested?

Built a throwaway PostgreSQL 18 primary and streaming standby (`pg_basebackup -R`), `wal_level=logical` and `hot_standby_feedback=on` on both, with a logical slot created on the standby. Ran the old and new versions of the query against each end.

- [x] Scenario A: old query on the standby fails, new query succeeds

```
-- old
ERROR:  recovery is in progress
HINT:  WAL control functions cannot be executed during recovery.

-- new
pgoutput|logical|0/3000060|0/301A880
```

- [x] Scenario B: primary output is unchanged

```
-- old
pgoutput|logical|0/301A8B8|0/301A8B8
-- new
pgoutput|logical|0/301A8B8|0/301A8B8
```

- [x] Scenario C: `pg_replication_slot_advance('probe', ...)` on the standby succeeds, confirming `AdvanceLSN` needs no change.

The repro script that produces the above is here, if it is useful for regression purposes: it spins the pair up in Docker and tears it down on exit. Happy to contribute it as a test if you would like it in-tree.

`gofmt`, `go build ./pkg/waljs/` and `go vet ./pkg/waljs/` are all clean.

# Screenshots or Recordings

I know the guidelines ask for a validation video and I have not attached one. What I have verified is the SQL, on both ends of a real replica pair, as above. I have not yet run a full OLake sync against a standby, so I did not want to record something implying more coverage than I actually have. If you want the full sync demo before this is reviewable, say so and I will get it recorded, or I am glad for a maintainer to take the one-line change and validate it in your own harness.

## Documentation

## Related PR's (If Any):

None. Related: #949 and #991 added read-replica support for MSSQL; this is the Postgres counterpart.


erikcw:
Update: I built the patched binary and ran a full sync against a standby, so the "not yet run end to end" caveat in the description is closed.

Setup was a PG18 primary plus a streaming standby from `pg_basebackup -R`, `wal_level=logical` and `hot_standby_feedback=on` on both, publication on the primary, logical slot created on the standby. Destination is AWS S3 Tables through the REST catalog. The image is built from this branch, reusing the published `olakego/source-postgres:v0.9.0` JAR since the change does not touch the Java writer.

Stock v0.9.0, same config:

```
{"connectionStatus":{"message":"failed to validate cdc configuration for slot probe_slot: ERROR: recovery is in progress (SQLSTATE 55000)","status":"FAILED"},"type":"CONNECTION_STATUS"}
```

Patched:

```
{"connectionStatus":{"status":"SUCCEEDED"},"type":"CONNECTION_STATUS"}

# snapshot
Total records read: 10

# after DELETE id=5 and UPDATE id=3 on the primary
Total records read: 2
addedEqualityDeletes=CounterResult{unit=COUNT, value=2}
```

Reading the Iceberg table back:

```
live rows (_op_type <> 'd'): [1, 2, 3, 4, 6, 7, 8, 9, 10]
id 5: (5, None, 'd')
id 3: (3, 'row-3-UPDATED', 'u')
```

Two things worth mentioning:

The stock failure actually surfaces in `check`, from the slot validation in `postgres.go`, before `PreCDC` is reached. Both paths go through the same `ReplicationSlotTempl` constant, so the one change covers them.

Snapshot backfill and streaming CDC both worked against the standby with nothing else changed, so as far as I can tell there is no second standby-hostile call hiding further in.

I have this as one self-contained script that stands the pair up, runs both binaries against it and tears everything down, if a repro or a recording would help.

saksham-datazip:
@erikcw Can you please once check if you have signed the cla?

pkg/waljs/replicator.go
```diff
@@ -17,8 +17,14 @@ import (
...
 const (
-	ReplicationSlotTempl = "SELECT plugin, slot_type, confirmed_flush_lsn, pg_current_wal_lsn() as current_lsn FROM pg_replication_slots WHERE slot_name = '%s'"
-	CDCLSN               = "_cdc_lsn" // Postgres LSN
+	// current_lsn is selected by recovery state: pg_current_wal_lsn() raises
+	// "recovery is in progress" on a standby, where pg_last_wal_replay_lsn() is
+	// the equivalent position. pg_is_in_recovery() is volatile, so the CASE is
+	// not constant-folded and only the applicable branch is evaluated.
+	ReplicationSlotTempl = "SELECT plugin, slot_type, confirmed_flush_lsn, " +
+		"CASE WHEN pg_is_in_recovery() THEN pg_last_wal_replay_lsn() ELSE pg_current_wal_lsn() END as current_lsn " +
+		"FROM pg_replication_slots WHERE slot_name = '%s'"
```
saksham-datazip, line 26: Can you once confirm and do a research if confirmed_flush_lsn can be null here ?

erikcw:
@saksham-datazip signed!

hash-data (approved):

hash-data:
Thanks for the pr @erikcw

saksham-datazip (approved):
Thanks for raising the pr @erikcw

PR #1074, opened 2026-08-06
fix(mysql): correct unsigned mediumint values over CDC

mihir-datazip:
# Description

`MEDIUMINT UNSIGNED` values above 8388607 reach the destination as negative numbers over CDC — `16777215` syncs as `-1`.

go-mysql reads MEDIUMINT's 3 bytes into an `int32` and sign-extends them, so anything with bit 23 set comes back negative. #846 handled the unsigned flag by reinterpreting the whole `int32` as `uint32`, which is correct for the 4-byte types but not for mediumint: the sign extension sits in the top byte, so `int32(-1)` becomes `4294967295` instead of `16777215`, and the mapped `Int32` conversion then wraps it straight back to `-1`.

The reinterpret is replaced by a mask back to the column's storage width (`0xFFFFFF` for mediumint), which recovers the stored value at every width and can't overflow the mapped type. `unsigned bigint` leaves the switch — 8 bytes have no spare width, so its bits are already final. tinyint/smallint/int keep their current behaviour: masking into int16/int32/int64 yields exactly what the uint reinterpret did.

Backfill is unaffected (the server applies signedness there, so values arrive correct and pass through untouched), and the fix stays behind the existing `LoadedStateVersion > 3` gate, so streams on older state keep legacy semantics.

## Type of change

# How Has This Been Tested?

- [x] Unit tests — new `drivers/mysql/internal/mysql_test.go` covering `dataTypeConverter`: each unsigned width at zero / signed max / sign bit / max, the legacy (`<= 3`) state-version path, backfill-shaped values that arrive already unsigned or widened, signed columns that must *not* be masked, and geospatial WKB → WKT.
- [x] MySQL integration test — 5 UNSIGNED columns pinned at their maxima (including mediumint at `8388608` and `16777215`) added to the test table, seed inserts, CDC update and expected data.

## Documentation

- [x] N/A (bug fix only)

## Related PR's (If Any):

#846 — introduced the unsigned handling this corrects.


drivers/mysql/internal/datatype_conversion.go
```diff
@@ -70,3 +74,32 @@ var mysqlTypeToDataTypes = map[string]types.DataType{
...
+			return int16(v) & math.MaxUint8
+		}
+	case "unsigned smallint":
+		if v, ok := value.(int16); ok {
+			return int32(v) & math.MaxUint16
+		}
+	case "unsigned mediumint":
+		if v, ok := value.(int32); ok {
+			return v & maxUint24
+		}
```
vikaxsh: need to add state check

drivers/mysql/internal/datatype_conversion.go
```diff
@@ -70,3 +74,32 @@ var mysqlTypeToDataTypes = map[string]types.DataType{
...
 }
+
+// MEDIUMINT's 3 bytes are the one MySQL integer width Go has no constant for.
+const maxUint24 = 1<<24 - 1
+
+// stripSignExtension masks an UNSIGNED column's value back to its storage width, undoing the sign
+// extension the binlog parser applies. The result comes back in the narrowest signed Go type that
+// holds the column's whole range, so no case widens further than its own values need.
+// UNSIGNED BIGINT is absent on purpose: it has no spare width, so its bits are already final.
+func stripSignExtension(value any, columnType string) any {
```
vikaxsh: we should keep this in mysql.go only as datatype_conversion.go is only for type mapping not value conversion

drivers/mysql/internal/datatype_conversion.go
```diff
@@ -70,3 +74,32 @@ var mysqlTypeToDataTypes = map[string]types.DataType{
...
 	"geometrycollection": types.String,
 }
+
+// MEDIUMINT's 3 bytes are the one MySQL integer width Go has no constant for.
+const maxUint24 = 1<<24 - 1
+
+// stripSignExtension masks an UNSIGNED column's value back to its storage width, undoing the sign
+// extension the binlog parser applies. The result comes back in the narrowest signed Go type that
+// holds the column's whole range, so no case widens further than its own values need.
+// UNSIGNED BIGINT is absent on purpose: it has no spare width, so its bits are already final.
```
vikaxsh: what if BIGINT UNSIGNED holds value above 2^63-1

drivers/mysql/internal/datatype_conversion.go
mihir-datazip: This is value change only... do we need to maintain the wrong value via a version gate?

drivers/mysql/internal/datatype_conversion.go
mihir-datazip: BIGINT UNSIGNED already carried correct value, ie, -1 of BIG INT is already UNSIGNED MAX BIG INT

drivers/mysql/internal/datatype_conversion.go
mihir-datazip: Okay

drivers/mysql/internal/datatype_conversion.go
mihir-datazip: I checked and we only support at max int64 in our olakes type so this has to be like this for now.
Added a todo for it though

drivers/mysql/internal/datatype_conversion.go
mihir-datazip: Done

vikaxsh (approved):

PR #1082, opened 2026-08-08
fix(mongodb): support inline TLS certificates for UI/Docker deployments

### LINKED ISSUE #1080: MongoDB source: tlsCAFile/tlsCertificateKeyFile in additional_params can never resolve — connector container has no way to receive cert files
## Description
The MongoDB source connector's additional_params lets you pass arbitrary connection-string parameters, including tlsCAFile / tlsCertificateKeyFile. But there's no way to actually get a certificate file onto the filesystem of the container that runs the connector, so any value here fails with:

open /certs/root-ca.crt: no such file or directory
This makes it impossible to connect to a self-hosted MongoDB with a private/self-signed CA, or one requiring X.509 client-certificate auth — both fairly common in production.

##Environment
Deployment: Docker Compose (standard olake-ui + temporal-worker + postgresql + temporal stack)
Connector: source-mongodb, hit during Test Connection / discover / sync

## Steps to reproduce

1.Stand up OLake via the standard docker-compose.
2.Configure a MongoDB source against a self-hosted Mongo with a private CA / client-cert auth.
3.Mount the cert file(s) into temporal-worker via a docker-compose volume, then set tlsCAFile: /certs/root-ca.crt (and/or tlsCertificateKeyFile) in "Additional Connection Parameters".
4.Run Test Connection.
## Expected behavior

Either the connector accepts inline PEM content in its config (see "Suggested fix"), or there's a documented, supported way to get an external file onto the filesystem of the container that actually executes the connector command.

## Actual behavior

open /certs/root-ca.crt: no such file or directory

## Root cause (traced through olake-ui + olake-helm/worker)
additional_params is merged directly into the Mongo connection URI's query string (drivers/mongodb/internal/config.go, Config.URI()), so tlsCAFile=/path is passed through verbatim to the Mongo Go driver, which needs that path to exist on disk.
The connector command doesn't run inside temporal-worker. temporal-worker (olake-helm/worker/executor/docker/executor.go) spawns a separate, short-lived sibling container per check/discover/sync via the host Docker socket.
That sibling container gets exactly one bind mount — the per-workflow temp dir (holding config.json etc.) → /mnt/config (hostConfig.Mounts in executor.go). Nothing else, including any volumes configured on temporal-worker itself, is inherited by the sibling container.
So there is currently no supported path — via docker-compose or the UI — to get a certificate file reachable by the container that opens the MongoDB connection.
By contrast, SSH tunnel support (ssh_config.private_key) already solves the equivalent problem correctly: the private key is passed as a raw PEM string in the JSON config and parsed in-memory (utils/ssh.go, ssh.ParsePrivateKeyWithPassphrase) — no file or mount required.
## Suggested fix
Apply the same pattern used for ssh_config.private_key to MongoDB TLS: add fields for the CA cert and client cert/key as inline PEM strings (e.g. tls_ca_cert, tls_certificate_key) to the MongoDB source config, and in drivers/mongodb/internal/mon.go (Setup()), build an in-memory *tls.Config from them (x509.NewCertPool().AppendCertsFromPEM(...), tls.X509KeyPair(...)) and apply it via options.Client().SetTLSConfig(...), instead of relying on file-path-based tlsCAFile/tlsCertificateKeyFile params. This sidesteps the container-mount problem entirely and likely affects other connectors with the same additional_params pattern (Postgres, MySQL, etc.) worth checking too.

krishanu7:
# PR #1082 — MongoDB inline TLS (ssl block)

## Summary

- Adds an `ssl` configuration block (`utils.SSLConfig`) to the MongoDB connector, aligned with Postgres/MySQL — inline PEM via `ssl.server_ca` / `ssl.client_cert` / `ssl.client_key` instead of filesystem paths
- Builds `*tls.Config` in memory during connector setup using `utils.BuildTLSConfig` and applies it via `SetTLSConfig`
- Strips `tlsCAFile` / `tlsCertificateKeyFile` from the connection URI when SSL is enabled, since file paths are unavailable in UI/Docker ephemeral containers
- Updates `spec.json` and `uischema` with the shared SSL widget (`mode`, `server_ca`, client cert fields)
- Preserves backward compatibility for CLI usage with mounted cert files via `additional_params` when `ssl.mode` is `disable` or unset

### Review refactor

Per review feedback, replaced the initial `tls_ca_cert` / `tls_certificate_key` fields with the shared `ssl` block pattern used by other connectors. TLS URI logic is consolidated in `config.go`; standalone `tls.go` removed.

Fixes #1080

## Example config

```json
{
  "hosts": ["host.docker.internal:27017"],
  "username": "mongodb",
  "password": "secure_password123",
  "authdb": "admin",
  "database": "olake_mongodb_test",
  "replica_set": "rs0",
  "ssl": {
    "mode": "verify-ca",
    "server_ca": "-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----\n"
  },
  "additional_params": {
    "directConnection": "true",
    "tlsAllowInvalidHostnames": "true"
  }
}
```

## Documentation

- [x] N/A (refactor and test changes only; no user-facing config or behavior change)

## Test plan

- [x] `go test ./drivers/mongodb/internal/...` — URI param stripping, TLS config building, SSL validation
- [x] Local repro: file-path `tlsCAFile` fails with `no such file or directory` in Docker (issue #1080)
- [x] Local verify: inline `ssl.server_ca` with `verify-ca` connects successfully from CLI (`127.0.0.1`) and UI worker container (`host.docker.internal`, no cert mount)
- [ ] CI passes

**Video:** https://drive.google.com/file/d/1h1YXV4SviFREemTVfSDr1azQbN9iA3fq/view?usp=drive_link


drivers/mongodb/resources/spec.json
```diff
@@ -76,10 +76,20 @@
...
+        "tls_certificate_key": {
+            "type": "string",
+            "title": "TLS Client Certificate & Key",
+            "description": "Client certificate and private key in PEM format (combined in one field). Required for mutual TLS or MONGODB-X509 authentication. Use this instead of tlsCertificateKeyFile in Additional Connection Parameters."
+        },
         "additional_params": {
             "type": "object",
             "title": "Additional Connection Parameters",
-            "description": "Additional MongoDB connection string parameters (e.g., authMechanism). Each parameter will be added to the connection URI.",
+            "description": "Additional MongoDB connection string parameters (e.g., tls, connectTimeout). Each parameter will be added to the connection URI. File-path TLS parameters (tlsCAFile, tlsCertificateKeyFile) are not supported in UI/Docker — use the TLS CA Certificate and TLS Client Certificate & Key fields instead.",
```
vishalm0509: Let's revert this change

drivers/mongodb/internal/config.go
```diff
@@ -59,7 +61,7 @@ func (c *Config) URI() string {
 
 	host := strings.Join(c.Hosts, ",")
 
-	for key, value := range c.AdditionalParams {
+	for key, value := range c.uriAdditionalParams() {
```
vishalm0509: We can revert this as well.

drivers/mongodb/internal/config.go
```diff
@@ -10,21 +10,23 @@ import (
...
+	ServerRAM         uint              `json:"server_ram"`
+	MaxThreads        int               `json:"max_threads"`
+	Database          string            `json:"database"`
+	RetryCount        int               `json:"backoff_retry_count"`
+	ChunkingStrategy  string            `json:"chunking_strategy"`
+	UseIAM            bool              `json:"use_iam"`
+	TLSCACert         string            `json:"tls_ca_cert"`
+	TLSCertificateKey string            `json:"tls_certificate_key"`
+	SSHConfig         *utils.SSHConfig  `json:"ssh_config"`
+	AdditionalParams  map[string]string `json:"additional_params"`
```
vishalm0509: There's scope to remove the new file introduced `tls.go` and follow the same pattern used in Postgres here.

You can follow this doc for reference: https://www.mongodb.com/docs/drivers/go/current/security/tls/

```
// these are supported by mongodb as well. 
caFile := "<path to CA certificate>"
certFile := "<path to public client certificate>"
keyFile := "<path to private client key>"
```

```suggestion
	SSLConfiguration *utils.SSLConfig  `json:"ssl"`
```

utils/spec/uischema.go
```diff
@@ -26,6 +26,8 @@ const MongoDBUISchema = `{
         { "read_preference": 12, "srv": 12 },
         { "max_threads": 12, "backoff_retry_count": 12 },
         { "chunking_strategy": 12, "use_iam": 12 },
+        { "tls_ca_cert": 24 },
+        { "tls_certificate_key": 24 },
         { "additional_params": 12, "ssh_config": 12 }
```
vishalm0509: Let's use 12
```suggestion
        { "ssl": 12, "additional_params": 12 },
```

drivers/mongodb/resources/spec.json
krishanu7: This text change is not needed?

drivers/mongodb/internal/config.go
krishanu7: func (c *Config) uriAdditionalParams() map[string]string 

Returns connection URI params. When inline PEM is set file-path TLS params are stripped and tls=true is ensured. Otherwise params are returned unchanged for CLI backward compatibility.

drivers/mongodb/internal/config.go
krishanu7: Ok i will check this

utils/spec/uischema.go
krishanu7: Okay

drivers/mongodb/resources/spec.json
vishalm0509: Not required. AdditionalParams is generic and not specific to TLS. So not needed in my opinion.

drivers/mongodb/internal/config.go
vishalm0509: We can handle this in the `URI()` itself. With the refactor, we can remove the `tls.go` and we can see how we can ensure this case. 

An example for reference:
```go
sslEnabled := c.SSLConfiguration != nil && c.SSLConfiguration.Mode != utils.SSLModeDisable
for key, value := range c.AdditionalParams {
    if sslEnabled && (key == "tlsCAFile" || key == "tlsCertificateKeyFile") {
        continue // inline PEM via ssl block takes precedence
    }
    query.Set(key, value)
}
```

drivers/mongodb/internal/config.go
```diff
@@ -78,7 +88,36 @@ func (c *Config) URI() string {
...
+func (c *Config) buildTLSConfig() (*tls.Config, error) {
+	if c.SSLConfiguration == nil || c.SSLConfiguration.Mode == utils.SSLModeDisable {
+		return nil, nil
+	}
+
+	return utils.BuildTLSConfig(mongoTLSHost(c.Hosts), c.SSLConfiguration)
+}
+
 // TODO: Add go struct validation in Config
 func (c *Config) Validate() error {
```
vishalm0509, line 112: currently in MongoDB Validate is not called anywhere.  Check other drivers on the use of `Validate()` and make changes acc.

drivers/mongodb/internal/config.go
```diff
@@ -78,7 +88,36 @@ func (c *Config) URI() string {
...
+	if len(hosts) == 0 {
+		return ""
+	}
+
+	host := hosts[0]
+	if h, _, ok := strings.Cut(host, ":"); ok {
+		return h
+	}
+	return host
+}
```
vishalm0509, line 101: `verify-full` will break on multi-host replica sets.

> "For a 3-node replica set with distinct hostnames (very common for self-hosted Mongo), every connection to node 2/3 will use node 1's hostname for SNI + hostname verification and fail the TLS handshake unless all nodes share a wildcard cert"

Please test the above mentioned case and provide a fix as mongo can have multiple hosts.

vishalm0509 (commented):
Please test with the mentioned edge-case.

drivers/mongodb/internal/config.go
krishanu7: Ok i will check and fix

vishalm0509:
@krishanu7 this is an adoption issue. You can work on this issue on priority than the other issues you are working on.

drivers/mongodb/internal/config.go
krishanu7: Addressed both review points:

1. **Validate()** — now called at the start of Setup() with hosts/database/auth checks and SSL defaults (similar to MySQL/Postgres).

2. **verify-full + multi-host** — we were pinning tls.Config.ServerName to hosts[0], so every node was verified against the first hostname.

   **Fix:** leave ServerName empty. The official Go driver fills it from the **dialed address** on each connection:

   https://github.com/mongodb/mongo-go-driver/blob/v1.17.3/x/mongo/driver/topology/connection.go  
   (configureTLS: if ServerName == "", set it from addr)

### Local verification
Server cert SAN = `mongo2.olake.local` (+ `127.0.0.1`) only — **not** `mongo1`.
| Test | Config | Result |
|------|--------|--------|
| 1 | `ssl.mode=verify-full`, host=`mongo2` only | `SUCCEEDED` |
| 2 | `ssl.mode=verify-full`, host=`mongo1` only | `FAILED` — `certificate is valid for mongo2..., not mongo1...` (correct) |
| 3 | `ssl.mode=verify-full`, hosts=`[mongo1, mongo2]` | `SUCCEEDED` |
| 4 | `ssl.mode=verify-ca`, hosts=`[mongo1, mongo2]` | `SUCCEEDED` |

drivers/mongodb/internal/config.go
```diff
@@ -78,7 +81,54 @@ func (c *Config) URI() string {
...
+	if c.SSLConfiguration == nil || c.SSLConfiguration.Mode == utils.SSLModeDisable {
+		return nil, nil
+	}
+	return utils.BuildTLSConfig("", c.SSLConfiguration)
+}
+
 func (c *Config) Validate() error {
+	if !c.Srv && len(c.Hosts) == 0 {
+		return fmt.Errorf("hosts is required")
+	}
```
vishalm0509: Even in case of `srv` we need atleast one host. So the check is wrong. 


```suggestion
	if len(c.Hosts) == 0 {
		return fmt.Errorf("hosts is required")
	}
```

drivers/mongodb/internal/config.go
```diff
@@ -78,7 +81,54 @@ func (c *Config) URI() string {
 	return u.String()
 }
 
-// TODO: Add go struct validation in Config
+func (c *Config) buildTLSConfig() (*tls.Config, error) {
+	if c.SSLConfiguration == nil || c.SSLConfiguration.Mode == utils.SSLModeDisable {
+		return nil, nil
+	}
+	return utils.BuildTLSConfig("", c.SSLConfiguration)
```
vishalm0509: Let's add a comment why we are setting hostname to `""`

**Example comment**
```suggestion
	// Pass "" so we don't hardcode one hostname for TLS verify-full. The mongo driver
	// fills ServerName from whichever host this connection is dialing
	return utils.BuildTLSConfig("", c.SSLConfiguration)
```

drivers/mongodb/internal/config_test.go
```diff
@@ -0,0 +1,305 @@
...
+import (
+	"crypto/tls"
+	"net/url"
+	"testing"
+
+	"github.com/datazip-inc/olake/utils"
+	"github.com/datazip-inc/olake/utils/testutils"
+)
+
+func TestConfigURI_StripsTLSFileParamsWhenSSLEnabled(t *testing.T) {
```
vishalm0509: Can we align this with other drivers’ `config_test.go` ? Prefer one test per function (TestConfig_URI, TestConfig_Validate, TestConfig_buildTLSConfig) with cases in a slice using expectedContains (or similar) instead of a separate top-level test for each scenario. You can refer the other drivers, you will understand.

drivers/mongodb/internal/config.go
```diff
@@ -78,7 +81,54 @@ func (c *Config) URI() string {
...
+		}
+	}
+
+	if err := c.SSLConfiguration.Validate(); err != nil {
+		return fmt.Errorf("failed to validate ssl config: %w", err)
+	}
+
+	if _, err := c.buildTLSConfig(); err != nil {
+		return err
+	}
```
vishalm0509: Why are we doing this in `Validate()` we can drop this.

vishalm0509 (changes requested):
@krishanu7  CI is failing

utils/spec/uischema.go
```diff
@@ -26,14 +26,46 @@ const MongoDBUISchema = `{
...
     "srv": {
         "ui:widget": "boolean"
     },
     "use_iam": {
         "ui:widget": "boolean"
     },
+    "ssl": {
+        "ui:options": {
+            "title": false
+        },
```
vishalm0509, line 41: Please update it to this

```go
        "ui:options": {
            "title": false, "label": false, "description": false
        },
```

vishalm0509 (dismissed):
lgtm

drivers/mongodb/internal/config_test.go
```diff
@@ -0,0 +1,349 @@
...
+	"github.com/datazip-inc/olake/utils/testutils"
+)
+
+func TestConfig_URI(t *testing.T) {
+	certs := testutils.GenerateTestCerts()
+
+	tests := []struct {
+		name             string
+		config           *Config
+		expectedContains []string
```
mihir-datazip: We should directly assert the URI here...

mihir-datazip (changes requested):
@krishanu7 LGTM, just a small change needed

mihir-datazip:
@krishanu7 a general comment... Keep the description formatting intact; we are used to checking the screen shots and videos under the section in the PR description template... changing it might cause reviewer to miss it

drivers/mongodb/internal/config_test.go
krishanu7: understood, the current expectedContains approach is a bit loose cause it checks the URI indirectly, directly asserting the query parameters will ensure we aree actually validating the URI structure and values. i will do the change.

drivers/mongodb/internal/config_test.go
```diff
@@ -0,0 +1,355 @@
...
+			expectedQuery: map[string]string{
+				"tls":        "true",
+				"authSource": "admin",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			uri := tt.config.URI()
```
mihir-datazip: Why are we parsing the URI? We can just assert the URI string right?

Our function signature is `func (c *Config) URI() string` so UTs should follow the same signature...
Why are we adding new functions in tests to parse the string and then asserting the data returned by those functions?

drivers/mongodb/internal/config_test.go
```diff
@@ -0,0 +1,355 @@
...
+}
+
+func TestConfig_buildTLSConfig(t *testing.T) {
+	certs := testutils.GenerateTestCerts()
+
+	tests := []struct {
+		name       string
+		config     *Config
+		wantNil    bool
+		assertions func(t *testing.T, tlsCfg *tls.Config)
```
mihir-datazip: Same here, why are we doing custom assertions? The function signature is `func (c *Config) buildTLSConfig() (*tls.Config, error)` so we should just assert the expected `tls.Config` struct and error right?

drivers/mongodb/internal/config_test.go
krishanu7: got := cfg.URI()

want := "mongodb://user:pass@mongo.example.com:27017/?authSource=admin&connectTimeoutMS=5000&tls=true"

if got != want {
    t.Fatalf("URI() = %q, want %q", got, want)
}
we should do like this?

mihir-datazip: Yes, that is how we should always do unit tests...

krishanu7: okay

krishanu7: I was referring to drivers/mysql/internal/config_test.go, the TestConfig_URI uses expectedContains / strings.Contains, and TestConfig_buildTLSConfig uses per-case assertions func(...). Same patterns you flagged on Mongo

mihir-datazip (dismissed):

drivers/mongodb/internal/config_test.go
mihir-datazip: In mysql we do this because mysql's DSN embeds a random TLS config name ("mysql_" + utils.ULID() at drivers/mysql/internal/config.go:64), so exact equality is impossible there.

mihir-datazip: The practice used in TestConfig_buildTLSConfig in mysql is not good. We should assert similarly for all cases... with custom assertions how do we make sure the fields we don't need are actually nil? skipping assertions is not a good way to write tests.

drivers/mongodb/internal/config_test.go
mihir-datazip, line 1:
```
		{
			name: "x509 client cert auth has no password",
			config: &Config{
				Hosts:    []string{"mongo.internal:27017"},
				Database: "analytics",
				AuthDB:   "$external",
				Username: "CN=olake-client,OU=Data,O=Acme",
				AdditionalParams: map[string]string{
					"authMechanism": "MONGODB-X509",
					"tls":           "true",
				},
			},
			wantURI: "mongodb://CN=olake-client,OU=Data,O=Acme@mongo.internal:27017/?authMechanism=MONGODB-X509&authSource=%24external&tls=true",
		},
		{
			name: "username without password",
			config: &Config{
				Hosts:    []string{"mongo.internal:27017"},
				Database: "analytics",
				AuthDB:   "admin",
				Username: "appuser",
			},
			wantURI: "mongodb://appuser@mongo.internal:27017/?authSource=admin",
		}
```

This was the behaviour in staging that we are changing in this PR. @krishanu7 @vishalm0509  can you confirm if this is expected regression before I merge the PR?

mihir-datazip: Basically the URIs that used to be passed before allowed
`x509 client cert auth has no password` and `username without password` which mongo itself supports and we are not allowing after this PR.

mihir-datazip (approved):

PR #1085, opened 2026-08-10
feat: s3 xml support

sahitya-datazip:
# Description

Added XML support for S3 source
**Normalization OFF:** stores entire XML content as a JSON string in `data` column
**Normalization ON:** level-0 flattening - top level fields as columns, nested maps as JSON columns

- XML Parser
     -    Added `XMLConfig` with `record_tag` configuration
     -    Implemented schema inference and record streaming for XML files
     -    Added unit tests for XML parser
- S3 Source Integration
    -    Added `xml` as a supported S3 format (`.xml` / `.xml.gz`)
    -  Updated `spec.json`  and `uischema.go` for XML configuration
- Integration Tests
   -   Added XML test variant to S3 Integration tests
   -  Added XML builders, schemas, expected data and filter configuration
   -  Added XML test data for Iceberg and Parquet destinations

[PRD - XML File Support in S3 Connector](https://datazip.atlassian.net/wiki/x/BACMJg)

## Type of change

# How Has This Been Tested?

- [x] Unit tests: `pkg/parser` XML Parser 
- [x] S3 Integration tests with XML source - with Iceberg and Parquet destinations

# Screenshots or Recordings

N/A

## Documentation

- [x] Documentation Link: [drivers/s3/README.md](https://github.com/datazip-inc/olake/blob/s3-xml-support/drivers/s3/README.md)


drivers/s3/internal/config.go
```diff
@@ -76,18 +78,18 @@ func (c *Config) Validate() error {
 
 	// Validate file format
 	if c.FileFormat == "" {
-		return fmt.Errorf("file_format is required (csv, json, or parquet)")
+		return fmt.Errorf("file_format is required (csv, json, parquet, or xml)")
 	}
 
 	validFormat := false
-	for _, format := range []FileFormat{FormatCSV, FormatJSON, FormatParquet} {
+	for _, format := range []FileFormat{FormatCSV, FormatJSON, FormatParquet, FormatXML} {
```
mihir-datazip, line 85: We can define this slice globally and use it here as well as logs so we don't have to repeatedly log `csv, json, parquet, or xml`

drivers/s3/internal/config.go
```diff
@@ -132,6 +134,12 @@ func (c *Config) Validate() error {
 		}
 	}
 
+	if c.FileFormat == FormatXML && c.XML == nil {
+		// Initialize with defaults if not provided
+		c.XML = &parser.XMLConfig{
+			RecordTag: "",
```
mihir-datazip, line 140: We don't need to initialise zero values, `c.XML = &parser.XMLConfig{}` is fine as well

drivers/s3/internal/config_test.go
```diff
@@ -96,7 +96,7 @@ func TestConfigValidate(t *testing.T) {
 				Region:          "us-east-1",
 				AccessKeyID:     "test-access-key",
 				SecretAccessKey: "test-secret-key",
-				FileFormat:      "xml",
+				FileFormat:      "yaml",
```
mihir-datazip, line 99: We can just name it "invalide-format" so we don't have to change it if yaml also comes in future...

drivers/s3/internal/config_test.go
```diff
@@ -281,7 +281,7 @@ func TestConfigPathPrefixNormalization(t *testing.T) {
 }
 
 func TestConfigAllFileFormats(t *testing.T) {
-	formats := []FileFormat{FormatCSV, FormatJSON, FormatParquet}
+	formats := []FileFormat{FormatCSV, FormatJSON, FormatParquet, FormatXML}
```
mihir-datazip, line 284: This can utilise the same global variable

drivers/s3/internal/s3.go
```diff
@@ -273,6 +273,9 @@ func (s *S3) matchesFileFormat(key string) bool {
 			(s.config.Compression == CompressionGzip && (strings.HasSuffix(lowerKey, ".json.gz") || strings.HasSuffix(lowerKey, ".jsonl.gz")))
 	case FormatParquet:
 		return strings.HasSuffix(lowerKey, ".parquet")
+	case FormatXML:
+		return strings.HasSuffix(lowerKey, ".xml") ||
+			(s.config.Compression == CompressionGzip && strings.HasSuffix(lowerKey, ".xml.gz"))
```
mihir-datazip, line 278: We can define a global map which keeps the supported extentions fo each format (maybe in the same global variable). My aim is to make changes in a single place and the functions should support dynamically.
Then we can loop for the extensions for formats for functions where we are using type switches.
We can also make a common util function for these conditions where the params can take the file suffixes like `csv`, `json`, `jsonl`, `xml`

drivers/s3/internal/s3.go
```diff
@@ -306,6 +309,8 @@ func (s *S3) ProduceSchema(ctx context.Context, streamID types.StreamID) (*types
 		inferredStream, err = s.inferSchemaForJSON(ctx, firstFile, stream)
 	case FormatParquet:
 		inferredStream, err = s.inferSchemaForParquet(ctx, firstFile, stream)
+	case FormatXML:
```
mihir-datazip, line 312: Same here, we can make this dynamic

pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+	// Loop through XML tokens and extract records with specified record tag
+	for len(records) < maxSamples {
+		token, err := decoder.Token()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			// check if we have some records, enough for schema inference - break
+			if len(records) > 0 {
+				logger.Warnf("Stopped reading XML after %d records due to error: %v", len(records), err)
```
mihir-datazip, line 168: I think we should fail here instead of passing the partial records...
I don't think we should silently continue despite an unexpected error.
Is there any case where there is failure here but the full parsing of the file will not fail?

pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+				break
+			}
+			return nil, fmt.Errorf("xml token error: %s", err)
+		}
+
+		// process start elements matching record tag
+		startElement, ok := token.(xml.StartElement)
+		if !ok || startElement.Name.Local != p.config.RecordTag {
```
mihir-datazip, line 176: Return with error instead of continue for !ok unless there is a reason for it.

Also do we skip something in the file if `startElement.Name.Local != p.config.RecordTag`?
In that case we should either add a small log saying we are skipping the record or in case `startElement.Name.Local == p.config.RecordTag` we can log that we found the record.

If the number of these logs can be spammy than we can leave it....

pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+			continue
+		}
+
+		// parse XML element into a record
+		record, err := p.parseXMLElement(decoder, startElement)
+		if err != nil {
+			if len(records) > 0 {
+				logger.Warnf("stopped reading XML after %d records due to error: %v", len(records), err)
```
mihir-datazip, line 184: Do we want to silent the error here?

pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+
+		// convert parsed record to a map
+		recordMap, err := p.parseXMLRecordAsMap(record, startElement.Name.Local)
+		if err != nil {
+			return nil, err
+		}
+		records = append(records, recordMap)
+	}
+
+	logger.Infof("Parsed %d records from XML for schema inference", len(records))
```
mihir-datazip, line 198: We can also add numbers of skipped records if any...

pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+// parseXMLDocumentAsMap parses the entire XML document as a single record and returns it as a map
+func (p *XMLParser) parseXMLDocumentAsMap(reader io.Reader) (map[string]any, error) {
+	decoder := xml.NewDecoder(reader)
+
+	// find root element and parse it as a single record
+	for {
+		token, err := decoder.Token()
+		if err == io.EOF {
+			return nil, fmt.Errorf("empty XML document")
+		}
```
mihir-datazip, line 211: Can move this check out of the loop? Empty document checking doesn't seem to make sense in for loop

pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+		record, err := p.parseXMLElement(decoder, startElement)
+		if err != nil {
+			return nil, err
+		}
+
+		if s, ok := record.(string); ok {
+			return map[string]any{startElement.Name.Local: s}, nil
+		}
+
+		content, err := p.parseXMLRecordAsMap(record, startElement.Name.Local)
```
mihir-datazip, line 230: Can we combine the logic in a single function for better readability?
Basically we can dynamically get string or map from a single function which return string when the record element happens to be string or map otherwise

This will avoid repetition of 
`return map[string]any{startElement.Name.Local: <variable>}, nil`

pkg/parser/parser.go
```diff
@@ -41,3 +41,8 @@ type JSONConfig struct {
 type ParquetConfig struct {
 	StreamingEnabled bool `json:"streaming_enabled"` // Default: true - use range requests
 }
+
+// XMLConfig holds XML-specific parsing configuration
+type XMLConfig struct {
```
mihir-datazip, line 46: I saw in [PRD](https://datazip.atlassian.net/wiki/spaces/PD/pages/646709252/XML+File+Support+in+S3+Connector#4.2-Case-2%3A-Normalization-ON-(Level%E2%80%910-Flattening)) this is mentioned as `row_identifier`

Please confirm this with product once

drivers/s3/resources/spec.json
```diff
@@ -165,6 +166,25 @@
...
+          "properties": {
+            "file_format": {
+              "const": "xml"
+            },
+            "xml": {
+              "type": "object",
+              "title": "XML Configuration",
+              "description": "XML format-specific settings",
+              "properties": {
+                "record_tag": {
```
mihir-datazip, line 180: Also `row_identifier "only used when normalization = true"` was mentioned in PRD, can you confirm this behaviour as well with Product team once

pkg/parser/parser.go
```diff
@@ -41,3 +41,8 @@ type JSONConfig struct {
...
+	RecordTag string `json:"record_tag"` //Default: ""
```
mihir-datazip, line 47: [Example 2 – Normalization ON (Level‑0 Flattening)](https://datazip.atlassian.net/wiki/spaces/PD/pages/646709252/XML+File+Support+in+S3+Connector#6.2-Example-2-%E2%80%93-Normalization-ON-(Level%E2%80%910-Flattening))

Here I think product expects a record level path support...

pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+
+		case xml.EndElement:
+			if t.Name.Local != startElement.Name.Local {
+				continue
+			}
+			if hasChildren {
+				return fields, nil
+			}
+			textValue := strings.TrimSpace(text.String())
+			if hasAttributes {
```
mihir-datazip, line 286: I think there should be a configurable flag for `Include_attributes: true` in config as asked by product...
Can you discuss with them if always including attributes is the intended behaviour?

pkg/parser/xml_test.go
```diff
@@ -0,0 +1,275 @@
...
+	callback := func(_ context.Context, record map[string]any) error {
+		records = append(records, record)
+		return nil
+	}
+	err := parser.StreamRecords(ctx, reader, callback)
+	require.NoError(t, err)
+	require.Len(t, records, 1)
+
+	//one row per order, attributes and nested columns as fields
+	assert.Equal(t, "1001", records[0]["id"])
```
mihir-datazip, line 214:
```
{
  "orders": {
    "order": [
      {
        "@id": "1001",
        "@status": "NEW",
        "order_date": "2024-08-01",
        "customer": {
          "id": "CUST_55",
          "name": "Alice"
        }
      }
    ]
  }
}
```
PRD mentions the attributes with a `@` prefix. Need to confirm this behaviour

pkg/parser/parser.go
mihir-datazip: Also check if there is need for the metadata fields like `_file_path, _file_size, _ingested_at`
Image

pkg/parser/xml.go
mihir-datazip: Also I think the product expects the scalar columns and timestamps to be supported. Currently I see we only have map and string which is not the intended behaviour

pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+			return textValue, nil
+
+		case xml.Comment, xml.Directive, xml.ProcInst:
+			// ignore
+		}
+	}
+}
+
+// setXMLField adds a value to the fields map for a given key, handling multiple values as slices
+func (p *XMLParser) setXMLField(fields map[string]any, key string, value any) {
```
mihir-datazip, line 301:
```
<order><item>a</item></order>            -> {"item":"a"}
<order><item>a</item><item>b</item></order> -> {"item":["a","b"]}
```

InferSchema over both -> item type = string      # array shape discarded
Downstream, the parquet destination's ReformatRecord → ReformatValue(String, []any{"a","b"}) hits the default: fmt.Sprintf("%v")
array -> String column: "[a b]"       # Go syntax, not JSON — unparseable downstream
map   -> String column: "map[k:v]"

Repeated sibling elements are the most common XML shape we should have "always an array" rule (even for a single child)

pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+func (p *XMLParser) InferSchema(_ context.Context, reader io.Reader) (*types.Stream, error) {
+	logger.Debug("Inferring XML schema from sample data")
+
+	//TODO : implement sampling of records from first and last files to get more accurate schema
+	maxSamples := 100
+
+	// Limit data read for schema inference to prevent OOM on large files
+	// 10MB should be enough to get 100 sample records for most XML files
+	const maxBytesForInference = 10 * 1024 * 1024 // 10MB
+	limitedReader := io.LimitReader(reader, maxBytesForInference)
```
mihir-datazip, line 41: This will truncate the reader and with record_tag set, parseXMLContent tolerates the resulting syntax error (records > 0 → warn + break).
But with record_tag: "" there is no such escape

pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+}
+
+// parseXMLElement recursively parses an XML element and its children into a map or string
+// Attributes are added as fields
+func (p *XMLParser) parseXMLElement(decoder *xml.Decoder, startElement xml.StartElement) (any, error) {
+	fields := make(map[string]any)
+
+	for _, attr := range startElement.Attr {
+		fields[attr.Name.Local] = attr.Value
```
mihir-datazip, line 245:
```
<order id="attrval"><id>childval</id></order> -> {"id":["attrval","childval"]}   # silent array
<order><name name="attr">textval</name></order> -> {"name":{"name":"textval"}}   # attr "attr" silently LOST
```
Fix this and also add UT case for it

pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+			return fmt.Errorf("failed to parse XML document: %s", err)
+		}
+
+		if err := callback(ctx, record); err != nil {
+			return fmt.Errorf("failed to process record: %s", err)
+		}
+		recordCount++
+	} else {
+		// Stream XML records based on specified record tag
+		decoder := xml.NewDecoder(reader)
```
mihir-datazip, line 96: xml.NewDecoder is used with no CharsetReader
This will make Non-UTF-8 XML fail hard. Handle this and also add UTs with non-utf-8 xml

```
<?xml version="1.0" encoding="ISO-8859-1"?> -> error: encoding "ISO-8859-1" declared but Decoder.CharsetReader is nil
```

pkg/parser/xml.go
mihir-datazip, line 1: Every empty form collapses to "", never nil

```
<order><a/><b></b><c>  </c></order> -> {"a":"","b":"","c":""}
```

CSV deliberately maps ""/"null" → nil via errNullValue (csv.go:283). XML has no equivalent. Consequence on a timestamp-typed column: 
ReformatValue(Timestamp, "") -> err "string does not start with date pattern (YYYY-MM-DD)"
ReformatRecord returns that error (reformat.go:51), so one empty optional date element fails the parquet-destination write. This is why the IT's XML variant has no null_col while the CSV variant does.

mihir-datazip, line 1: Namespaces: prefixes collapse, xmlns declarations become data

Only Name.Local is used, and Go keeps namespace declarations in Attr:
```
<h:order id="1"><h:price>10</h:price><f:price xmlns:f="http://f.ns">20</f:price></h:order>
-> {"id":"1", "price":["10", {"f":"http://f.ns","price":"20"}]}
```
Two distinct namespaces merged into one heterogeneous array, and xmlns:f surfaced as a column f. Related: record_tag: "ns:order" can never match, since only the local name is compared.

mihir-datazip (changes requested):

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+	// Loop through XML tokens and extract records with specified record tag
+	for len(records) < maxSamples {
+		token, err := decoder.Token()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			// check if we have some records, enough for schema inference - break
+			if len(records) > 0 {
+				logger.Warnf("Stopped reading XML after %d records due to error: %v", len(records), err)
```
sahitya-datazip: we use limitReader which will truncate after 10MB
so it can cause parse error when it truncates in middle of a record

pkg/parser/xml.go
sahitya-datazip: `!ok` should continue, all tokens are not startElement (like chardata, endElement also included in token)
skip logs might be spammy as we check every tag

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+		startElement, ok := token.(xml.StartElement)
+		if !ok || startElement.Name.Local != p.config.RecordTag {
+			continue
+		}
+
+		// parse XML element into a record
+		record, err := p.parseXMLElement(decoder, startElement)
+		if err != nil {
+			if len(records) > 0 {
+				logger.Warnf("stopped reading XML after %d records due to error: %v", len(records), err)
```
sahitya-datazip: yes, as we might have enough sample data to return

drivers/s3/internal/config.go
sahitya-datazip: added supportedFileFormats
[refactor: add dynamic file format in s3](https://github.com/datazip-inc/olake/pull/1085/changes/8bf40f3a740c4d9d3a78810325696e1508cbf0cd)

drivers/s3/internal/config.go
sahitya-datazip: removed initialisation
[refactor: add dynamic file format in s3](https://github.com/datazip-inc/olake/pull/1085/changes/8bf40f3a740c4d9d3a78810325696e1508cbf0cd)

drivers/s3/internal/config_test.go
sahitya-datazip: used supportedFileFormats 
[refactor: add dynamic file format in s3](https://github.com/datazip-inc/olake/pull/1085/changes/8bf40f3a740c4d9d3a78810325696e1508cbf0cd)

drivers/s3/internal/s3.go
sahitya-datazip: formatExtensions map in config.go
[refactor: add dynamic file format in s3](https://github.com/datazip-inc/olake/pull/1085/changes/8bf40f3a740c4d9d3a78810325696e1508cbf0cd)

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+
+		// convert parsed record to a map
+		recordMap, err := p.parseXMLRecordAsMap(record, startElement.Name.Local)
+		if err != nil {
+			return nil, err
+		}
+		records = append(records, recordMap)
+	}
+
+	logger.Infof("Parsed %d records from XML for schema inference", len(records))
```
sahitya-datazip: this is for sample data and token doesn't exactly point to any record instead elements

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+// parseXMLDocumentAsMap parses the entire XML document as a single record and returns it as a map
+func (p *XMLParser) parseXMLDocumentAsMap(reader io.Reader) (map[string]any, error) {
+	decoder := xml.NewDecoder(reader)
+
+	// find root element and parse it as a single record
+	for {
+		token, err := decoder.Token()
+		if err == io.EOF {
+			return nil, fmt.Errorf("empty XML document")
+		}
```
sahitya-datazip: token should be consumed first and when no root element it throws empty document
(skip non startElement)

pkg/parser/xml_test.go
sahitya-datazip: confirmed @ prefix is not required for attributes

pkg/parser/xml.go
sahitya-datazip: [chore: avoid limit reader for whole document inferSchema in xml parser](https://github.com/datazip-inc/olake/pull/1085/changes/edf938e61afa6a1d0edbeebba18f62ea533b3914)

drivers/s3/internal/config_test.go
sahitya-datazip: renamed in [chore: avoid limit reader for whole document inferSchema in xml parser](https://github.com/datazip-inc/olake/pull/1085/changes/edf938e61afa6a1d0edbeebba18f62ea533b3914)

pkg/parser/parser.go
sahitya-datazip: Confirmed with product, just a row_identifier is needed not the path
and metadata fields are OLake fields

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+
+		case xml.EndElement:
+			if t.Name.Local != startElement.Name.Local {
+				continue
+			}
+			if hasChildren {
+				return fields, nil
+			}
+			textValue := strings.TrimSpace(text.String())
+			if hasAttributes {
```
sahitya-datazip: Discussed on this, attribute fields need to be included by default

drivers/s3/internal/pkg/parser/parser.go
```diff
@@ -41,3 +41,8 @@ type JSONConfig struct {
 type ParquetConfig struct {
 	StreamingEnabled bool `json:"streaming_enabled"` // Default: true - use range requests
 }
+
+// XMLConfig holds XML-specific parsing configuration
+type XMLConfig struct {
```
sahitya-datazip: confirmed and renamed in [refactor: rename record_tag to row_identifier ifor xml](https://github.com/datazip-inc/olake/pull/1085/changes/41f4e964b7e74f11edb6cacebb2dff9bda6e01c8)

pkg/parser/xml.go
sahitya-datazip: simplified function in [chore: simplify document root as map in XML](https://github.com/datazip-inc/olake/pull/1085/changes/365a93e4377e30551da7a143b7f8e196b4f4dbd3)

drivers/s3/resources/spec.json
sahitya-datazip: fixed in [fix: apply row_identifier only when normalization is ON for XML](https://github.com/datazip-inc/olake/pull/1085/changes/a96bac03f29199a43034289ed2cc1fa3a3a4c04b)

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,329 @@
...
+			return textValue, nil
+
+		case xml.Comment, xml.Directive, xml.ProcInst:
+			// ignore
+		}
+	}
+}
+
+// setXMLField adds a value to the fields map for a given key, handling multiple values as slices
+func (p *XMLParser) setXMLField(fields map[string]any, key string, value any) {
```
sahitya-datazip: [fix: always store xml sibling children as an array](https://github.com/datazip-inc/olake/pull/1085/changes/8937eb0f92aa1e854bffb27c763d9a9eb057cdce)

pkg/parser/xml.go
sahitya-datazip: discussed with product and added underscore for attribute fields 
[chore: prefix attribute with _ and map empty leaves as null in xml parser](https://github.com/datazip-inc/olake/pull/1085/changes/e7000dc9e7dc24c8092827385c2f1fbb167bf445)

pkg/parser/xml.go
sahitya-datazip: added in [fix: decode non UTF-8 xml via Charset reader](https://github.com/datazip-inc/olake/pull/1085/changes/f342b9d0e3de9f06d7de6d4784d9d860cde16196)

drivers/s3/internal/pkg/parser/xml.go
sahitya-datazip: [chore: prefix attribute with _ and map empty leaves as null in xml parser](https://github.com/datazip-inc/olake/pull/1085/changes/e7000dc9e7dc24c8092827385c2f1fbb167bf445)

sahitya-datazip: [fix: skip xmlns declarations in XML attributes](https://github.com/datazip-inc/olake/pull/1085/changes/61e17c83f529d58e24113aea36790e068aa0eb28)

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,380 @@
...
+		return nil, fmt.Errorf("empty XML file")
+	}
+
+	sampleRecords, err := p.parseXMLContent(trimmed, maxSamples)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse XML: %s", err)
+	}
+
+	if len(sampleRecords) == 0 {
+		return nil, fmt.Errorf("no records found in XML file")
```
mihir-datazip, line 67: Can we update this error then saying the xml records are too large to infer schema....

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	// Resolve schema for each sample Record and update stream schema
+	for i, record := range sampleRecords {
+		if err := typeutils.Resolve(p.stream, record); err != nil {
+			return nil, fmt.Errorf("failed to resolve schema for record %d: %s", i, err)
+		}
+	}
+
+	logger.Infof("Inferred schema from XML file")
```
mihir-datazip: Can this log be spammy? If yes we can skip it or make it debug log...

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,380 @@
...
+
+		record, err := p.parseXMLDocumentAsMap(reader)
+		if err != nil {
+			return fmt.Errorf("failed to parse XML document: %s", err)
+		}
+
+		if err := callback(ctx, record); err != nil {
+			return fmt.Errorf("failed to process record: %s", err)
+		}
+		recordCount++
```
mihir-datazip: move this in default section only....

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,380 @@
...
+			startElement, ok := token.(xml.StartElement)
+			if !ok || startElement.Name.Local != p.config.RowIdentifier {
+				continue
+			}
+
+			// Parse XML element into a record
+			record, err := p.parseXMLElement(decoder, startElement)
+			if err != nil {
+				if !errors.Is(err, errNullValue) {
+					logger.Warnf("Error reading XML record %d: %v", recordCount, err)
```
mihir-datazip: Clarify the message that we are skipping record due to null value error...

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,380 @@
...
+			}
+			recordCount++
+		}
+	}
+	logger.Infof("Processed %d records from XML file", recordCount)
+	return nil
+}
+
+// parseXMLContent parses XML data and returns a slice of records based on specified record tag or entire document
+func (p *XMLParser) parseXMLContent(data []byte, maxSamples int) ([]map[string]any, error) {
```
mihir-datazip: rename this as `parseSampleXMLContent` so not hard failing is expected on truncated output

drivers/s3/internal/config.go
```diff
@@ -76,18 +87,18 @@ func (c *Config) Validate() error {
...
-	for _, format := range []FileFormat{FormatCSV, FormatJSON, FormatParquet} {
+	for _, format := range supportedFileFormats {
 		if c.FileFormat == format {
 			validFormat = true
 			break
 		}
 	}
 	if !validFormat {
-		return fmt.Errorf("invalid file_format: must be csv, json, or parquet")
+		return fmt.Errorf("invalid file_format: must be csv, json, parquet, or xml")
```
mihir-datazip: Use the `supportedFileFormats` variable to log as well

drivers/s3/internal/pkg/parser/xml_test.go
```diff
@@ -0,0 +1,434 @@
...
+	callback := func(_ context.Context, record map[string]any) error {
+		records = append(records, record)
+		return nil
+	}
+	err := parser.StreamRecords(ctx, reader, callback)
+	require.NoError(t, err)
+	require.Len(t, records, 1)
+
+	//one row per order; attrs scalar, children always arrays
+	assert.Equal(t, "1001", records[0]["_id"])
```
mihir-datazip, line 321: Are we sure we need all as string? Has product confirmed this behaviour?

pkg/parser/xml.go
mihir-datazip: I can't see this commit in final change

tests/s3/s3_util_test.go
mihir-datazip, line 1: We should make sure to cover the following cases in Integration tests.

1. Record extraction & modes
- Normalization ON + row_identifier at depth 1 (<orders><order>) — one row per record tag.
- Normalization OFF — backfill zeroes row_identifier, whole file → one row, everything in the data column as a JSON string.
- row_identifier nested deeper than root child (<feed><batch><order>) — still matched at any depth.
- Record tag that also exists as a nested element name (<order> containing <order>) — only the outer one starts a record; inner becomes a field.
- File with root but zero matching record tags — sync emits 0 rows and succeeds; discover on such a file errors "no records found"
- row_identifier set but normalization ON with records that are pure text (<order>text</order>)

We can cover these in a single sync run as well, no need to add different sync for each test case

2. Data types (element values → destination columns)

All of these exist in the builder already but their expected destination type is ⚠:
- ⚠ string, bool (true/false), int (int32/int64 range), float (incl. exponent form 1.5e3), plain date 2024-08-01, RFC3339 timestamp, milli/micro/nano timestamps — per PRD these should land as scalar/timestamp columns; today the always-array wrap makes them all array.
- Timestamp variants: with Z, with +05:30 offset, without timezone, and the precision-per-writer split (legacy=millis, arrow/parquet=micros — reuse textWriterExpectedData).
- Extreme timestamps (pre-1900, far future) — the OL-2946 nano-overflow class.
- Numeric-looking strings that must stay strings: leading zeros (007), +91... phone-style, 1e999 overflow — pin the inference policy (CSV's policy is the reference).
- Mixed-type column across rows/files (row 1 int, row 2 string) — type promotion at discover, no sync failure. (mixed_col exists — keep, but assert the promoted type explicitly)
- Same column typed differently in file 1 vs file 2 of one stream — schema union across files, later file wins/widens.
- Long text value (≥ a few KB, multiline) round-trips intact.
- Attribute-backed columns (_id etc.): string attr, numeric attr, timestamp-looking attr — attributes stay scalar even under always-array; pin whether a timestamp-looking attr infers timestamp (TypeFromValue will try).

3. Null / empty shapes (the CSV-parity gap)
- Self-closing <a/>, empty <b></b>, whitespace-only <c>  </c> → null, on a string column and on a timestamp column — the timestamp one is the original ReformatValue(Timestamp, "") write-failure case; this is exactly the null_col the CSV variant has and XML lacks. (Unit tests for this were added in 30433f5c and then lost in later test rewrites — restore those too.)
- Absent element (optional_col) vs empty element — both null at destination, and distinguishable from ""? Pin the policy.
- Empty attribute id="" — currently passes through as "", never null-mapped. Pin intended behavior.
- Record that is entirely empty (<order/>) mid-file — all-null row, sync continues.

4. Test Compressed file flow as well

pkg/parser/xml.go
sahitya-datazip: [fix: decode non UTF-8 xml via Charset reader](https://github.com/datazip-inc/olake/pull/1085/changes/2353b1d79f2d881ed8dcef45196d68ec7270b67c)

drivers/s3/internal/pkg/parser/xml.go
mihir-datazip: I meant this commit.
I can't see this commit in final change.

mihir-datazip: Unit tests for this case...

tests/s3/testdata/xml/parquet_destination.json
mihir-datazip, line 1: We can remove this file, we use the common config

tests/s3/testdata/xml/iceberg_destination.json
mihir-datazip, line 1: We can remove this file as well

drivers/s3/internal/pkg/parser/xml.go
sahitya-datazip: this error can also be caused when row_identifier provided is not in file

mihir-datazip: `no records found in XML file` can be misleading, you can decide what should be the best error here...

drivers/s3/internal/pkg/parser/xml.go
sahitya-datazip: renamed in [chore: address review comments](https://github.com/datazip-inc/olake/pull/1085/commits/16ef7453c6198845a3ebde61015be5fb73da16d5)

drivers/s3/internal/pkg/parser/xml.go
sahitya-datazip: [chore: address review comments](https://github.com/datazip-inc/olake/pull/1085/commits/16ef7453c6198845a3ebde61015be5fb73da16d5)

drivers/s3/internal/pkg/parser/xml.go
sahitya-datazip: [chore: address review comments](https://github.com/datazip-inc/olake/pull/1085/commits/16ef7453c6198845a3ebde61015be5fb73da16d5)

drivers/s3/internal/pkg/parser/xml.go
sahitya-datazip: changed to debug log [chore: address review comments](https://github.com/datazip-inc/olake/pull/1085/commits/16ef7453c6198845a3ebde61015be5fb73da16d5)

drivers/s3/internal/config.go
sahitya-datazip: [chore: use supportedFileFormats](https://github.com/datazip-inc/olake/pull/1085/commits/da8f6a85f87d2ee6a4043718876f523a8f3232a8)

drivers/s3/internal/pkg/parser/xml_test.go
sahitya-datazip: yes confirmed

tests/s3/testdata/xml/parquet_destination.json
sahitya-datazip: [chore: remove redundant IT destination jsons](https://github.com/datazip-inc/olake/pull/1085/changes/4158003056d0187c3b305f6845b95418f0475f4c)

tests/s3/testdata/xml/iceberg_destination.json
sahitya-datazip: [chore: remove redundant IT destination jsons](https://github.com/datazip-inc/olake/pull/1085/changes/4158003056d0187c3b305f6845b95418f0475f4c)

tests/s3/s3_util_test.go
sahitya-datazip: [chore: XML IT](https://github.com/datazip-inc/olake/pull/1085/changes/4b7b209bba4fc50c3b54cc6a99f7a0befa9f8e27)

- Normalization OFF couldn't be implemented in IT as updateSelectedStreams keeps it true
- Zero matching tags discover fails
- Data types - expected as arrays and strings

tests/s3/testdata/xml/parquet_destination.json
sahitya-datazip: [chore: remove redundant IT destination jsons](https://github.com/datazip-inc/olake/pull/1085/changes/4158003056d0187c3b305f6845b95418f0475f4c)

drivers/s3/internal/pkg/parser/xml.go
sahitya-datazip: [fix: skip truncated XML in sampling schema](https://github.com/datazip-inc/olake/pull/1085/changes/0452db94fe237290ba7ac5cd04f7046441d2b8d5)

drivers/s3/internal/pkg/parser/xml.go
sahitya-datazip: [fix: skip truncated XML in sampling schema](https://github.com/datazip-inc/olake/pull/1085/changes/0452db94fe237290ba7ac5cd04f7046441d2b8d5)

drivers/s3/internal/pkg/parser/xml.go
mihir-datazip: I still see the code is outside the `default` section of the switch statement

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,379 @@
...
+
+	// Loop through XML tokens and extract records with specified record tag
+	for len(records) < maxSamples {
+		token, err := decoder.Token()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			// check if we have some records, enough for schema inference - break
+			if len(records) > 0 && strings.Contains(strings.ToLower(err.Error()), "unexpected eof") {
```
mihir-datazip: For error comparisons always use the `errors` package. ALso as it is used in multiple places, make a function for it.
Something like
```
func isTruncatedXMLErr(err error) bool {
	var syntaxErr *xml.SyntaxError
	return errors.Is(err, io.ErrUnexpectedEOF) ||
		(errors.As(err, &syntaxErr) && strings.Contains(syntaxErr.Msg, "unexpected EOF"))
}
```

drivers/s3/internal/pkg/parser/xml.go
```diff
@@ -0,0 +1,379 @@
...
+
+	var data []byte
+	var err error
+
+	if p.config.RowIdentifier == "" {
+		data, err = io.ReadAll(reader)
+	} else {
+		// Limit data read for schema inference to prevent OOM on large files
+		// 10MB should be enough to get 100 sample records for most XML files
+		const maxBytesForInference = 10 * 1024 * 1024 // 10MB
```
mihir-datazip: Make this configurable with default of 10MB. The reason I want this configurable is that we need to add  UTs for the truncation logic.
Set the truncation limit to very small like a few bytes and test the following cases:
1. No truncation (truncation size >= the xml size)
2. Truncation should sample the schema correctly
3. Any other failures should fail the InferSchema as expected

mihir-datazip:
`Parquet / Incremental - update` can flake `PATH_NOT_FOUND`: the s3 cursor is whole-second with a strict `>`, so `update_1`/`evolve_1` uploaded in the cursor's second are silently skipped and the sync writes nothing.

Replace lines 784-805 of `tests/s3/s3_util_test.go` with:

```go
			sleepPastCursorSecond()
			variant.putFile(ctx, t, src, prefix, "insert_1", variant.BuildFile, 7, seedValues, false)

		case "update":
			// Object stores have no in-place update: changed data arrives as another file.
			sleepPastCursorSecond()
			variant.putFile(ctx, t, src, prefix, "update_1", variant.BuildFile, 10, updatedValues, false)

		case "evolve-schema":
			// An object store's ALTER TABLE: a file whose rows carry a column discover has
			// never seen. It is stamped after the previous sync like the "update" file, so
			// the update sync reads both and must evolve the destination to fit this one.
			// The rows carry updatedValues because that sync asserts every row against
			// ExpectedUpdatedData; evolvedColumn itself is asserted through the schema, not
			// per row, since the "update" file's rows sync a null there.
			if variant.BuildEvolvedFile != nil {
				sleepPastCursorSecond()
				variant.putFile(ctx, t, src, prefix, "evolve_1", variant.BuildEvolvedFile, 13, updatedValues, false)
			}

		default:
			t.Fatalf("unsupported operation: %s", operation)
		}
	}
}

// sleepPastCursorSecond crosses a second boundary before cursor-sensitive uploads: the s3 cursor
// keeps LastModified at whole seconds with a strict >, so a same-second upload is silently skipped.
// TODO: fix in the s3 driver (>= plus tracking keys already synced in the cursor's second).
func sleepPastCursorSecond() {
	time.Sleep(time.Until(time.Now().Truncate(time.Second).Add(1100 * time.Millisecond)))
}
```

```diff
@@ -781,8 +781,10 @@
 		case "insert":
 			// A file the incremental cursor has not seen: it is stamped after the previous
 			// sync, so only its rows are re-read.
+			sleepPastCursorSecond()
 			variant.putFile(ctx, t, src, prefix, "insert_1", variant.BuildFile, 7, seedValues, false)
 
 		case "update":
 			// Object stores have no in-place update: changed data arrives as another file.
+			sleepPastCursorSecond()
 			variant.putFile(ctx, t, src, prefix, "update_1", variant.BuildFile, 10, updatedValues, false)
@@ -797,9 +799,17 @@
 			if variant.BuildEvolvedFile != nil {
+				sleepPastCursorSecond()
 				variant.putFile(ctx, t, src, prefix, "evolve_1", variant.BuildEvolvedFile, 13, updatedValues, false)
 			}
 
 		default:
 			t.Fatalf("unsupported operation: %s", operation)
 		}
 	}
 }
+
+// sleepPastCursorSecond crosses a second boundary before cursor-sensitive uploads: the s3 cursor
+// keeps LastModified at whole seconds with a strict >, so a same-second upload is silently skipped.
+// TODO: fix in the s3 driver (>= plus tracking keys already synced in the cursor's second).
+func sleepPastCursorSecond() {
+	time.Sleep(time.Until(time.Now().Truncate(time.Second).Add(1100 * time.Millisecond)))
+}
```

drivers/s3/internal/pkg/parser/parser.go
```diff
@@ -44,5 +44,6 @@ type ParquetConfig struct {
 
 // XMLConfig holds XML-specific parsing configuration
 type XMLConfig struct {
-	RowIdentifier string `json:"row_identifier"` //Default: "" - applied when stream normalization ON
+	RowIdentifier        string `json:"row_identifier"` //Default: "" - applied when stream normalization ON
+	MaxBytesForInference int64  `json:"-"`              //Default: 10MB
```
mihir-datazip: We should use unexported field here like `maxBytesForInference int64` as we just don't want to restrict json mashal/unmarshal, we genuinely want this config at package level only

sahitya-datazip: [fix: empty-only XML fields as nullable strings](https://github.com/datazip-inc/olake/pull/1085/changes/8d26b3ea28c1f28b30cdb977be331f0aeb19b165)

mihir-datazip (approved):
LGTM

PR #1112, opened 2026-08-13
fix: issue with flush last batch

hash-data:
# Description
Fixing an issue in the flush writer
  - a bug that is not stopping commits on error in the last batch flush
  - the error message gets overridden by other errors. 

Blast Radius of bug:
 - if there is an error in schema evolution that is thrown by the Go side (e.g. int to string conversion), it might miss the records.
Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/destination/writers.go
+++ b/destination/writers.go
@@ -302,7 +302,7 @@ func (wt *WriterThread) Close(closeCtx context.Context, finalMetadataState any) 
 	default:
 		defer wt.stats.ThreadCount.Add(-1)
 
-		// for cancelling on flush error
+		// for canceling on flush error
 		ctx, cancel := context.WithCancel(closeCtx)
 		defer cancel()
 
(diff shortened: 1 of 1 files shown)


destination/writers.go
```diff
@@ -286,28 +286,38 @@ func (wt *WriterThread) rollbackStats() {
...
 }
 
-func (wt *WriterThread) Close(ctx context.Context, finalMetadataState any) (err error) {
+func (wt *WriterThread) Close(closeCtx context.Context, finalMetadataState any) (err error) {
 	select {
-	case <-ctx.Done():
+	case <-closeCtx.Done():
 		// Wait for in-flight flushes so rollback below can't race their stat updates.
 		// Block's error is ctx cancellation that brought us here, so ignored.
 		_ = wt.group.Block()
```
mihir-datazip, line 294: We can still ignore error from this right??

mihir-datazip (commented):
LGTM

destination/writers.go
hash-data: Yes, as the main context is already completed, we are just making sure no goroutine becomes dangling

mihir-datazip (approved):

PR #1117, opened 2026-08-14
feat: add telemetry for error events

vaibhav-datazip:
# Description
Two additions to telemetry, so a run can be understood after the fact without asking the user for logs.
[Telemetry Additions for Adoption](https://datazip.atlassian.net/wiki/spaces/PD/pages/643825675/Telemetry+Additions+for+Adoption)

**1. What a sync was made of.** Seven counters, added to both `Sync Started - CLI` and `Sync Completed - CLI`. They are counted inside `classifyStreams`, the function that already decides which streams are synced, so the numbers cannot disagree with the run.

| key | meaning |
|---|---|
| `full_refresh_streams_count` | streams synced full-refresh |
| `incremental_streams_count` | streams synced incrementally |
| `cdc_streams_count` | streams synced via CDC |
| `strict_cdc_streams_count` | streams synced via strict CDC |
| `selected_streams_count` | streams that passed selection and validation |
| `normalized_streams_count` | streams with normalization on |
| `partitioned_streams_count` | streams with a partition regex |

Only streams that actually sync are counted, so the four sync-mode counters sum to `selected_streams_count`.

**2. Why a run failed.** A new `Failure - CLI` event, sent by every failing command — `sync`, `check`, `discover`, `spec` and `clear-destination`.

```
Failure - CLI  category=auth_failed  classified_by=vendor  command=check  code=28P01  error_source=postgres
```

| key | always sent | meaning |
|---|---|---|
| `category` | yes | one of 23 fixed buckets — `auth_failed`, `network_unreachable`, `cdc_position_lost`, … |
| `classified_by` | yes | evidence used: `vendor`, `stdlib`, `precondition`, `default` |
| `command` | yes | subcommand that ran |
| `error_source` | yes | component that classified it — `postgres`, `iceberg`, … |
| `code` | when one exists | the server's own code (`28P01`, `1045`, `AccessDenied`) or a constant for a condition we detected (`postgres.replication_slot_missing`) |
| `error_type` | only when unclassified | root cause's Go type, which is what a new rule gets written from |
| `sync_id` | `sync` only | joins this event to the two sync events above |

Every field is either a constant in this repository or a code the vendor defines. No log line, config value, server message or user input reaches the payload, and absent fields are omitted rather than sent empty. No new common properties were added — a successful run sends no failure event at all.

## Type of change

# How Has This Been Tested?

- [x] ran sync with artificial failures and verified on mixpanel. 

# Screenshots or Recordings
image
image
image
image

## Documentation

## Related PR's (If Any):


destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeFailures.java
```diff
@@ -0,0 +1,109 @@
...
+        // Bounded so a self-referencing cause cannot loop.
+        for (int depth = 0; depth < 16; depth++) {
+            Throwable next = root.getCause();
+            if (next == null || next == root) {
+                break;
+            }
+            root = next;
+        }
+        return root;
+    }
```
mihir-datazip, line 84: If we already use  Apache Commons Lang, we can utilise that here... If not, then we can use the current logic

`ExceptionUtils.getRootCause(throwable)` from `org.apache.commons.lang3.exception.ExceptionUtils`

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeFailures.java
mihir-datazip, line 1: Also if we can add unit tests for this class it will be helpful

utils/errs/errs.go
```diff
@@ -0,0 +1,363 @@
...
+	InternalError Category = "internal_error"
+)
+
+// How a category was decided. The share arriving as ClassifiedByDefault is the coverage metric:
+// a rise means rules are missing, not that failures changed.
+const (
+	ClassifiedByPrecondition = "precondition" // a condition OLake detected itself
+	ClassifiedByVendor       = "vendor"       // from the vendor's own error code
+	ClassifiedByStdlib       = "stdlib"       // from a Go standard library error type
+	ClassifiedByDefault      = "default"      // nothing matched
```
mihir-datazip: Shouldn't this classifier loudly say something like unclassified or unmatched instead of saying default?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
mihir-datazip, line 1: Got the following review from AI. Check these once and address all the ones that are applicable

```
The startup-error contract (the new catch block)

1. The named exception is never classified — java_client.go:220-221 still wraps every SetupAndStartProcess failure as errs.Precondition(errs.InternalError, codeJVMStartFailed, ...), and Go never parses the stderr line. Worse than first stated: because errs.Classify short-circuits on From(err).Category != Unclassified (utils/errs/errs.go:219-221), the precondition also suppresses the shared standard() rules, not just the iceberg classifier. Wrong catalog credentials, an unreachable REST/JDBC catalog, a bad warehouse bucket — all land as internal_error, the one category errs documents as "a defect in the writer, not a condition a user can fix". The tables in destination/iceberg/errors.go are reachable only through fromGRPCStatus, which needs a live status that cannot exist pre-listen.

2. Outermost class + raw getMessage() — OlakeRpcServer.java:104 still emits t.getClass().getName() / t.getMessage(). CatalogUtil.buildIcebergCatalog (:63) funnels real causes through Iceberg's DynConstructors, which rethrows as IllegalArgumentException("Cannot initialize Catalog implementation ...") — so the class name identifies nothing, and exceptions constructed without a message print : null. OlakeFailures.rootCause() exists for exactly this but is private at OlakeFailures.java:73 and in a different package (...iceberg.rpc vs ...iceberg), so it is unreachable from here.

3. Stderr-drain race — logger.go sets cmd.Stderr to an *os.File, so os/exec dups the fd and cmd.Wait() (logger.go:431) creates no copier goroutine to wait on. Nothing synchronizes the errCh send with the reader goroutine's LoadOrStore at logger.go:322, and getFirstErrorLine (:415) is called the instant the select wakes. System.exit(1) at OlakeRpcServer.java:106 follows the println immediately. Under CPU contention the tail is empty and the feature degrades silently to "Iceberg[50051] exited before ready".

4. First-writer-wins slot — logger.go:322 is LoadOrStore under a key that is the process name, and for the stderr reader p.IsError == true makes every line qualify — no marker matching. The first stderr line of the JVM's life wins permanently. Note the project's own log4j2.xml is <Configuration status="WARN">, so any log4j StatusLogger complaint goes to stderr ahead of the crafted line; Picked up JAVA_TOOL_OPTIONS: and JDK reflective-access warnings do the same.

Session / shutdown correctness

5. CLOSE_SESSION leaks the open writer — still open, confirmed, but out of scope for this PR. OlakeRowsIngester.java:69 is still a bare sessions.remove(threadId) with no complete/abort of the operator's BaseTaskWriter, and Iceberg.Close (iceberg.go:152-179) still returns early on cancellation while its defer sends CLOSE_SESSION regardless. Correcting my earlier parenthetical: this PR's only change to that file is the one-line onError swap at :191 — the CLOSE_SESSION handler is pre-existing shared-JVM code. Worth a separate issue rather than a blocker here.

6. Shutdown hook isn't graceful — pre-existing. OlakeRpcServer.java:95 registers server::shutdown, which is non-blocking, with no awaitTermination(). The Go side's 10s SIGTERM wait (java_client.go:275-289) is therefore waiting on nothing. Untouched by this PR, but the new catch block draws the eye to it.

7. The try spans awaitTermination() — The try opens at :36 and awaitTermination() sits at :99, inside it. A post-startup interrupt, or SIGTERM landing between start() and addShutdownHook, prints "Iceberg writer failed to start" hours into a healthy sync and exits 1; the interrupt flag is swallowed too. Ending the try at .start() fixes the mislabel.

8. The missing-args branch bypasses the new contract — :38 reports via LOGGER.error, and src/main/resources/log4j2.xml pins the only appender to target="SYSTEM_OUT". getFirstErrorLine reads stderrReader.errorLines only, so the file's one explicit failure check produces nothing Go can attach.

Smaller / cleanup

9. Arrow service capability mismatch — pre-existing, low value. java_client.go:241-242 builds both stubs unconditionally. Since arrow-writer-enabled and the Go-side UseArrowWrites gate come from the same config field, this only bites on a stale jar — where UNIMPLEMENTED maps to unsupported_feature (errors.go:104) after the legacy path has already created tables, naming it a capability gap instead of a version skew.

10. Dead static state — hadoopConf, icebergProperties, icebergCatalog (:31-33) have no reader outside main — the icebergCatalog hits elsewhere in the tree are IcebergUtil method parameters that merely share the name. icebergProperties is still a byte-identical putAll copy of stringConfigMap (:59). All three can be locals.

11. Duplicated failure header — still open, plausible. printStackTrace() at :105 re-prints the class and message the println at :104 just wrote.
```

utils/telemetry/telemetry.go
```diff
@@ -104,67 +114,152 @@ func Init() {
...
+// contained: one here would otherwise kill the command being reported on.
+func send(name string, build func()) {
+	inflight.Add(1)
 	go func() {
+		defer inflight.Done()
+		defer func() {
+			if r := recover(); r != nil {
+				logger.Debugf("recovered from panic while sending %s event: %v", name, r)
+			}
+		}()
```
mihir-datazip: Can't we use single defer here?

utils/telemetry/telemetry.go
```diff
@@ -69,6 +78,7 @@ type LocationInfo struct {
 
 func Init() {
 	go func() {
```
mihir-datazip, line 81: We don't need to make init async...
We can avoid using 
```
    initDone = make(chan struct{})
	initOnce sync.Once
```

utils/telemetry/telemetry.go
```diff
@@ -104,67 +114,152 @@ func Init() {
...
-func TrackDiscover(streamCount int, sourceType string) {
+// send runs an event in the background, recording it so Flush can wait for it. Panics are
+// contained: one here would otherwise kill the command being reported on.
+func send(name string, build func()) {
+	inflight.Add(1)
 	go func() {
+		defer inflight.Done()
+		defer func() {
+			if r := recover(); r != nil {
+				logger.Debugf("recovered from panic while sending %s event: %v", name, r)
```
mihir-datazip: We can make this info log, it is not spammy and it might be important log for telemetry

utils/telemetry/telemetry.go
```diff
@@ -104,67 +114,152 @@ func Init() {
...
+// exiting: check and discover leave through logger.Fatal, dropping anything still in flight.
+func Flush(timeout time.Duration) {
+	if Disabled() {
+		return
+	}
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		<-initDone
+		inflight.Wait()
```
mihir-datazip, line 145: what is this `inflight.Wait()` guarding? It is in its own go routine so this wait will not affect anything...

utils/telemetry/telemetry.go
```diff
@@ -104,67 +114,152 @@ func Init() {
...
+	}
+	return destinationType, catalogType
+}
+
+func TrackSyncStarted(syncID string, mix types.StreamMix, sourceType string, destinationConfig *types.WriterConfig, configuredStreams int) {
+	destinationType, catalogType := destinationShape(destinationConfig)
+
+	send("sync started", func() {
 		props := map[string]interface{}{
-			"sync_start":          time.Now(),
```
mihir-datazip, line 132: Existing dashboards will have regression after this change... example: "Sync Started - CLI" silently read zero, and the normalized/partitioned series steps down at the release boundary with no schema change to explain it. Does product team expects this regression?

selected_count→selected_streams_count, full_load_streams→full_refresh_streams_count, cdc_streams→cdc_streams_count, normalized_streams→normalized_streams_count, partitioned_streams→partitioned_streams_count
Also in last two whole-catalog counts became selected-only counts so make sure that product team understands what each parameter is actually computing...

utils/errs/errs.go
```diff
@@ -0,0 +1,363 @@
...
+	return Failure{Category: Unclassified, ClassifiedBy: ClassifiedByDefault}
+}
+
+// maxUnwrapDepth bounds the walk so a cyclic chain cannot hang a failure report.
+const maxUnwrapDepth = 100
+
+// classificationOf finds the classification closest to the root cause. Two chain shapes exist:
+// %w unwraps one error at a time, so the walk keeps descending and the deepest wins; errors.Join
+// unwraps to a slice, which errors.Unwrap does not follow, so branches are searched by hand.
+func classificationOf(err error, depth int) (Failure, bool) {
```
mihir-datazip: The "innermost classification wins" rule is clear for a linear %w chain, but errors.Join creates a tree. Here, we return the first branch containing a classification, regardless of whether another branch contains a classification closer to its leaf.
I think as our telemetry is not gated via ITs, we should include UTs for all these cases to make sure that changes or updates to these parts of code is still gated in CI

utils/errs/errs.go
```diff
@@ -0,0 +1,363 @@
...
+	case syscall.ETIMEDOUT:
+		return "timed_out"
+	}
+	return ""
+}
+
+// rootType names the concrete type at the bottom of the chain. The outermost is almost always
+// *fmt.wrapError, which says nothing; the root identifies the failure and is what someone reads
+// when deciding which rule to add next.
+func rootType(err error) string {
```
mihir-datazip, line 368: classificationOf searches every branch of a joined error, while rootType always follows branches[0]. This means the classification and the reported ErrorType can be derived from different branches of the same error tree.
It would be good to either share a common traversal strategy

protocol/root.go
```diff
@@ -154,3 +156,49 @@ func init() {
...
 	}
 }
+
+const (
+	// Codes for conditions the CLI detects itself, before any connector is reached.
+	codeFlagMissing    = "config.flag_missing"
+	codeNoValidStreams = "catalog.no_valid_streams"
+	// telemetryFlushTimeout bounds how long a failing command waits for its report to leave the
+	// process: one HTTP round trip, and no more, so a hung endpoint cannot delay the exit.
+	telemetryFlushTimeout = 5 * time.Second
```
mihir-datazip: `telemetryFlushTimeout` should not be configured in this package...

protocol/root.go
```diff
@@ -154,3 +156,49 @@ func init() {
...
+func ReportFailure(err error) {
+	if err == nil {
+		return
+	}
+	// Classified here because this is the only point every failure reaches. A command's RunE
+	// misses its own PreRunE hooks, where the config is read and decrypted, and spec and
+	// clear-destination have no wrapper at all.
+	f := errs.From(errs.Classify(err))
+
+	if telemetry.Disabled() {
```
mihir-datazip: Just shift this check above...
Why do `f := errs.From(errs.Classify(err))` before this check?

drivers/s3/internal/pkg/parser/errors.go
mihir-datazip, line 1: parser is s3 specific. Do we need this logic specifically for s3 driver only?
I think we can move this somewhere common.

mihir-datazip:
@vaibhav-datazip we should also attach the screenshots/recording for the telemetry failures re created in most common cases we are expecting to catch.

Do this AFTER all the review comments and merge conflicts are resolved. So I can verify before merge....

protocol/sync.go
```diff
@@ -151,10 +154,10 @@ var syncCmd = &cobra.Command{
 
 		// Added this check to avoid the sleep when tracking telemetry is disabled
 		if !telemetry.Disabled() {
-			telemetry.TrackSyncStarted(syncID, selectedStreamsMetadata.SelectedStreams, selectedStreamsMetadata.FullLoadStreams, selectedStreamsMetadata.CDCStreams, connector.Type(), destinationConfig, catalog)
+			telemetry.TrackSyncStarted(syncID, selectedStreamsMetadata.Mix, connector.Type(), destinationConfig, len(catalog.Streams))
 			defer func() {
 				stats := pool.GetStats()
-				telemetry.TrackSyncCompleted(syncID, err == nil, stats.ReadCount.Load(), stats.BytesRead.Load())
+				telemetry.TrackSyncCompleted(syncID, selectedStreamsMetadata.Mix, destinationConfig, err == nil, stats.ReadCount.Load(), stats.BytesRead.Load())
```
mihir-datazip, line 160: A panicking sync reports sync_status: "SUCCESS". RunE has an unnamed return, so the defer closes over the local err declared at :95. Defers run during panic unwinding — so if connector.Read panics on the main goroutine, the assignment at :167 never completes, err is still nil, and TrackSyncCompleted(..., err == nil, ...) sends SUCCESS. It then sleeps 5s and lets the panic continue


We should cover the panic recover cases as well in Unit Tests

drivers/abstract/abstract.go
```diff
@@ -262,7 +265,7 @@ func generateThreadID(streamID, hash string) string {
 //   - map[string]*destination.WriterThread for multiple writers keyed by stream ID
 func handleWriterCleanup(ctx context.Context, cancel context.CancelFunc, err *error, writer any, threadID string, mtState *any, dedupInserts *bool) {
 	if r := recover(); r != nil {
-		*err = utils.Ternary(*err == nil, fmt.Errorf("panic recovered: %v", r), fmt.Errorf("%s: panic recovered: %v", *err, r)).(error)
+		*err = utils.Ternary(*err == nil, fmt.Errorf("panic recovered: %v", r), fmt.Errorf("%w: panic recovered: %v", *err, r)).(error)
```
mihir-datazip: A recovered thread panic lands as unclassified, and this PR's %w change can bury it. drivers/abstract/abstract.go:267-269. With no prior error, fmt.Errorf("panic recovered: %v", r) carries no classification → unclassified / error_type: *errors.errorString. And switching %s→%w means a panic occurring after an error now inherits that error's category — a panic behind a network blip reports network_unreachable.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeFailures.java
vaibhav-datazip: we aren't currently using apache commons

utils/telemetry/telemetry.go
mihir-datazip: As discussed, keep it async and just use `initDone = make(chan struct{})` no need of `initOnce`

utils/telemetry/telemetry.go
mihir-datazip: As discussed, confirm with @shubham19may or @hash-data that we can use info logs for telemetry or not...

utils/telemetry/telemetry.go
mihir-datazip: Resolve this yourself once you discuss these with prod team

vaibhav-datazip: Discussed with product about these changes

utils/telemetry/telemetry.go
vaibhav-datazip: discussed this , removed initDone as it was redundant

utils/telemetry/telemetry.go
vaibhav-datazip: discussed about this

.github/workflows/unit-tests.yml
```diff
@@ -22,3 +22,20 @@ jobs:
...
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: "17"
+          cache: maven
+
+      - name: Run Java Unit Tests
+        working-directory: destination/iceberg/olake-iceberg-java-writer
+        run: mvn -B --no-transfer-progress -Dtest=OlakeFailuresTest -DfailIfNoTests=false test
```
mihir-datazip: Why are we specifically adding test files? Can't we use something like `run: mvn -B --no-transfer-progress test` to run everything?

drivers/abstract/abstract_test.go
```diff
@@ -0,0 +1,425 @@
...
+	"github.com/datazip-inc/olake/types"
+	"github.com/datazip-inc/olake/utils"
+	"github.com/datazip-inc/olake/utils/errs"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// cleanupResult reproduces the classification handleWriterCleanup applies to a recovered panic,
+// including the thread wrap that follows. Kept identical to abstract.go so a change there fails this test.
+func cleanupResult(prior error, r any, threadID string) error {
```
mihir-datazip: this functions re-implements the classification, so it is not true of anything in the file.
If we revert abstract.go:296 from %w back to %s and all 8 cases of TestHandleWriterCleanupClassification plus all 3 of TestHandleWriterCleanupMessage stay green.

handleWriterCleanup is callable directly so we should use that instead. If it looks like too much work, you can remove this test all together for now

drivers/abstract/abstract_test.go
```diff
@@ -0,0 +1,425 @@
...
+		err = fmt.Errorf("thread[%s]: %w", threadID, err)
+	}
+	return err
+}
+
+func networkReset() error {
+	return fmt.Errorf("read failed: %w", &net.OpError{Op: "read", Net: "tcp", Err: syscall.ECONNRESET})
+}
+
+func TestHandleWriterCleanupClassification(t *testing.T) {
```
mihir-datazip, line 44: Drive the real function instead of a copy:

```go
err = prior
func() {
    defer handleWriterCleanup(ctx, cancel, &err, writer, threadID, nil, nil)
    panic(r)
}()
```

It recovers without re-panicking, so the call returns normally and the test can assert on `err`.

drivers/abstract/abstract_test.go
```diff
@@ -0,0 +1,425 @@
...
+			assert.Equal(t, tc.expectedCategory, got.Category, "category")
+			assert.Equal(t, tc.expectedBy, got.ClassifiedBy, "classified_by")
+			assert.Equal(t, tc.expectedCode, got.Code, "code")
+			assert.Equal(t, tc.expectedType, got.ErrorType, "error_type")
+			assert.Equal(t, tc.expectedComponent, got.Component, "component")
+		})
+	}
+}
+
+func TestHandleWriterCleanupMessage(t *testing.T) {
```
mihir-datazip, line 171: Same call shape as above. Note that with `writer = nil` the real function falls into the `default:` branch, so the message becomes `thread[x]: unsupported writer type: prev error: ...` — that whole close/metadata half of the function is what the copy was hiding.

protocol/sync_test.go
```diff
@@ -0,0 +1,237 @@
...
+const (
+	orderCorrect deferOrder = iota // registered last, so LIFO runs it first — what sync.go does
+	orderWrong                     // registered first, so it runs after telemetry has read err
+	orderAbsent                    // no recoverToError at all, the state before the fix
+)
+
+// runRunE mirrors syncCmd's RunE: a named result, the TrackSyncCompleted defer that reads
+// err == nil, and recoverToError placed according to order. It reports the sync_status the
+// telemetry defer would send, and the value that continued unwinding.
+func runRunE(order deferOrder, body func() error) (status string, panicked any) {
```
mihir-datazip: This is a hand-written copy of `syncCmd.RunE`'s defer stack, so nothing here is anchored to the real code — let's drop this file.

What would be worth having instead is a `TestClassifyStreamsMix`: `classifyStreams` is in-package and takes plain args, and its `Mix` counting (CDC vs STRICTCDC split, counters sitting past the skip branches so the mode counters sum to `Selected`) is the actual new telemetry payload in this PR and is currently untested.

protocol/sync_test.go
```diff
@@ -0,0 +1,237 @@
...
+		return body()
+	}
+	_ = inner()
+	return status, panicked
+}
+
+// TestRunEStatus covers every way syncCmd's RunE can finish and the sync_status each produces.
+// How the panic is classified is recoverToError's contract, covered in root_test.go; what
+// matters here is that no way of dying can be reported as a success.
+func TestRunEStatus(t *testing.T) {
```
mihir-datazip: Remove — the panic shapes are already covered against the real `recoverToError` in `root_test.go:149`.

protocol/sync_test.go
```diff
@@ -0,0 +1,237 @@
...
+			// the panic must continue so safego.Recovery still logs the stack and exits non-zero
+			assert.Equal(t, tc.expectedPanic, panicked != nil, "panic continued")
+		})
+	}
+}
+
+// TestRunEDeferOrder covers all three registrations against all three outcomes. The wrong and
+// absent orders must still reproduce the defect: if either stops doing so, the LIFO reasoning
+// the fix in sync.go depends on has changed.
+func TestRunEDeferOrder(t *testing.T) {
```
mihir-datazip: This is not needed as this asserts the LIFO ordering in the mock, so reordering the defers in `sync.go:165` (the exact regression it names) leaves it green.

protocol/sync_test.go
```diff
@@ -0,0 +1,237 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			status, _ := runRunE(tc.order, tc.body)
+			assert.Equal(t, tc.expectedStatus, status, "sync_status")
+		})
+	}
+}
+
+// TestRunEIsDeterministic runs each outcome repeatedly. Nothing here may depend on goroutine
+// scheduling, so the same input must always produce the same sync_status.
+func TestRunEIsDeterministic(t *testing.T) {
```
mihir-datazip: Remove this test.
200 iterations of a synchronous function, there's no scheduling here to shake out.

protocol/sync_test.go
```diff
@@ -0,0 +1,237 @@
...
+				status, _ := runRunE(orderCorrect, tc.body)
+				require.Equal(t, first, status, "run %d disagreed on sync_status", i)
+			}
+		})
+	}
+}
+
+// TestRunEReportsTheClassifiedError covers the join between the two files: the error
+// recoverToError writes is the same one the telemetry defer reads to derive FAILED.
+func TestRunEReportsTheClassifiedError(t *testing.T) {
```
mihir-datazip: You can remove this as well as it duplicates `TestRecoverToError`.

utils/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,576 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			destination, catalog := destinationShape(tc.config)
+			assert.Equal(t, tc.expectedDestination, destination)
+			assert.Equal(t, tc.expectedCatalog, catalog)
+		})
+	}
+}
+
+// TestAddStreamMix pins the seven per-sync counters and the values they carry.
+func TestAddStreamMix(t *testing.T) {
```
mihir-datazip: This only pins a struct-to-map copy; the counting it implies belongs in a `classifyStreams` test.

utils/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,576 @@
...
+			"Flush must not wait when telemetry is disabled")
+
+		close(release)
+		drain()
+	})
+}
+
+// TestDisabledReadsTheEnvOnce covers the switch every call site gates on. The value is cached,
+// so this asserts consistency rather than re-reading the environment.
+func TestDisabledReadsTheEnvOnce(t *testing.T) {
```
mihir-datazip: Delete — `assert.Equal(Disabled(), Disabled())` asserts a value against itself.

utils/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,576 @@
...
+		// Flush short-circuits before touching the WaitGroup when telemetry is off.
+		start := time.Now()
+		Flush()
+		assert.Less(t, time.Since(start), 50*time.Millisecond)
+	}
+}
+
+// TestFlushTimeoutIsSizedToOneRoundTrip pins the relationship the constant's comment claims:
+// the flush budget matches what a single sendEvent is allowed to take.
+func TestFlushTimeoutIsSizedToOneRoundTrip(t *testing.T) {
```
mihir-datazip: Delete — this compares the constant to its own literal. It never reads the `sendEvent` client/context budgets it claims to pin, so bumping those to 10s leaves this green.

utils/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,576 @@
...
+
+	err := telemetry.sendEvent("Bad - CLI", map[string]any{"ch": make(chan int)})
+
+	require.Error(t, err)
+	assert.Empty(t, capture.bodies, "nothing may be sent when the payload cannot be built")
+}
+
+// TestFlushTimesOut covers the deadline branch: a hung endpoint must not delay the exit past
+// flushTimeout, which is the whole reason the wait runs in a goroutine.
+func TestFlushTimesOut(t *testing.T) {
```
mihir-datazip: Delete, or make `flushTimeout` a package var the test can shrink. As written it costs 5s on every CI run (`make test.unit` passes neither `-short` nor `-race`) and forces the whole package to need `-short` under `-race`.

utils/errs/errs_test.go
```diff
@@ -0,0 +1,473 @@
...
+	t.Run("classifying twice is stable", func(t *testing.T) {
+		once := Classify(context.Canceled)
+		assert.Equal(t, From(once), From(Classify(once)))
+	})
+}
+
+// TestRegister covers the per-connector classifiers: the component is stamped from the
+// registration, a nil classifier is ignored, and an unrecognized error still reaches the
+// shared rules.
+func TestRegister(t *testing.T) {
```
mihir-datazip, line 406: Add a `t.Cleanup` that restores the `classifiers` slice — right now this leaves a panicking classifier installed for the rest of the test binary.

utils/errs/errs_test.go
```diff
@@ -0,0 +1,473 @@
...
+				"classified_by=%q must only ever accompany an unclassified category", ClassifiedByDefault)
+		}
+	}
+}
+
+// TestMultierrorIsNotWalked documents a known gap rather than asserting desired behavior:
+// hashicorp/go-multierror unwraps to a `chain`, whose Unwrap returns the next chain rather than
+// the error itself, so the by-hand walk never sees the *Error inside. errors.As does find it.
+// Live only if ErrExecSequential gains a caller on a path that reaches ReportFailure.
+func TestMultierrorIsNotWalked(t *testing.T) {
```
mihir-datazip: Delete, or flip it to assert the fixed behaviour. As written it pins the bug, so it will fail whoever switches `classificationOf` to `errors.As` — which is what its own comment recommends doing.

utils/telemetry/telemetry_test.go
vaibhav-datazip: removed

utils/telemetry/telemetry_test.go
vaibhav-datazip: removed

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/OlakeRpcServer.java
vaibhav-datazip: 1. Fixed: this branch added the Precondition, which short-circuits Classify at errs.go:230 so nothing else ran; the call site now only marks the path with a sentinel and a new fromJVMStart rule reads the Java class off stderr into the existing javaExceptions table

2. Fixed: rootCause() made public and used, plus a toString() fallback for messageless exceptions; this is what makes 1 possible, since the wrapper class always mapped to internal_error.

3. Pushback: utils/logger/logger.go is byte-identical to staging

4. Pushback: same file, same provenance; reproduced it (a JAVA_TOOL_OPTIONS notice does take the slot), but OLake never sets that variable and our image doesn't either, so it needs an externally instrumented JVM to trigger, and the fallback holds. Patch ready for a logger PR.

5. Pushback: Close already had the early-return-on-cancel on staging (only %s→%w changed).

6. Pushback: addShutdownHook is byte-identical at staging line 93
7. Fixed: the try now ends at .start() and awaitTermination() sits outside it.

8. Pushback: the LOGGER.error line and log4j2.xml are both unchanged from staging, and the branch is unreachable in practice since OLake always passes the config JSON.

11. Resolved by 2 : the println now names the root cause while printStackTrace() starts from the outermost and prints the chain, so they no longer duplicate keeping both.

mihir-datazip (approved):
LGTM

PR #1124, opened 2026-08-18
test: add unit tests for types/state.go

### LINKED ISSUE #659: feat: Add Unit Tests for state.go
# **Problem**

The OLake codebase currently has unit tests for:

* `resolver`
* `flattener`
* `catalog`

However, the file `olake/types/state.go` does not have any test coverage.

This file handles critical state management logic for the OLake connector, including:

* **State Management Methods:**
  * `SetType()` - Sets the state type (GLOBAL, STREAM, MIXED)
  * `ResetStreams()` - Resets all stream states
  * `ResetCursor()` - Resets cursor values for a specific stream
  * `isZero()` - Checks if state is empty

* **Global State Operations:**
  * `GetGlobal()` - Retrieves global state with read lock
  * `SetGlobal()` - Sets or updates global state with stream tracking
  * `HasCompletedBackfill()` - Checks if backfill is completed for a stream (supports both GLOBAL and STREAM state types)

* **Cursor Management:**
  * `GetCursor()` - Retrieves cursor value for a stream and key
  * `SetCursor()` - Sets cursor value for a stream (creates new StreamState if needed)

* **Chunk Management:**
  * `GetChunks()` - Retrieves all chunks for a stream
  * `SetChunks()` - Sets chunks for a stream (skips for FULLREFRESH mode)
  * `RemoveChunk()` - Removes a specific chunk and returns remaining count

etc.

Without unit tests, bugs in state management, race conditions with concurrent access, incorrect serialization/deserialization, or edge cases with nil values, empty states, or chunk operations may slip through during development. This is particularly critical as the `State` struct is used extensively across the codebase including:

* Incremental sync operations
* Backfill operations
* CDC (Change Data Capture) processing etc.

---

#### **Proposed Solution**

Add comprehensive unit tests for the logic inside `state.go`.

This will help:

* Increase overall test coverage
* Improve state management reliability across different state types
* Catch bugs earlier during CI, especially around:
  * Concurrent access patterns (read/write locks)
  * State serialization/deserialization edge cases
  * Chunk management operations
  * Cursor value handling
  * Empty state handling
* Make contributions safer when modifying state management logic

For reference on how existing unit tests are structured, you can refer to **PR #273** and the existing `catalog_test.go`, `flatten_test.go` and `resolver_test.go` files.

---

#### **Notes for Contributors**

* Add tests covering most of the scenarios and edge cases which should indicate both pass and fail scenarios.

* Follow the existing test file structure and naming conventions to maintain consistency with other parts (see `catalog_test.go` for reference).

* Use `testify/assert` and `testify/require` for assertions (as seen in existing test files).

rishyym0927:
# Description

`types/state.go` holds the sync resume logic — backfill chunks, incremental cursors, the CDC global position, and `state.json` serialization — but had no test coverage, unlike `resolver`, `flattener` and `catalog`.

Bugs here fail silently: a wrong answer produces no error and no log line, it just re-copies data that was already synced or skips data that was not. This PR adds **108 table-driven cases** covering every function in the file.

Tests only — **no production code is changed.**

Fixes #659

cc @vaibhav-datazip @nayanj98

### Structure

Follows the architecture of the existing unit tests, `reformat_test.go` in particular:

- one `Test<Function>` per function under test — **all 18 functions in `state.go` map 1:1**
- a doc comment on every test function
- a flat, data-only `tests` table with `// ===== section =====` grouping and a single loop at the end
- `testify/assert` throughout, with `require` where continuing past a failure would nil-panic instead of reporting cleanly
- `t.Cleanup` restores the viper state path that the fixture overrides

### What is covered

- **Lifecycle** — `SetType`, `isZero`, `ResetStreams`
- **Global state** — `GetGlobal` / `SetGlobal`, including that `SetGlobal(nil, streamID)` (as called from `cdc.go:65`) preserves the stored position while still tracking the stream, and that a stream set is created when the decoded global has none
- **Cursors** — `GetCursor` / `SetCursor` / `ResetCursor`: empty-key no-ops, implicit `StreamState` creation on first write, primary vs. secondary cursor reset, and that streams are keyed on **(namespace, name)** together so the same table name in two schemas does not alias
- **Chunks** — `GetChunks` / `SetChunks` / `RemoveChunk`: the deliberate `FULLREFRESH` no-op, empty-set vs. no-chunks-recorded, and `RemoveChunk`'s `-1`-on-unknown-stream vs. remaining-count contract (the `0` return is what signals backfill completion at `backfill.go:61`)
- **`HasCompletedBackfill`** — exercised down both branches, since they are different code paths: `GLOBAL` (set membership) and `STREAM` (zero chunks remaining)
- **Serialization** — `State.MarshalJSON` dropping streams with `HoldsValue == false`; `StreamState.MarshalJSON` flattening the `sync.Map`; `StreamState.UnmarshalJSON` rehydrating a chunk array into a real `*Set[Chunk]`, asserted by type assertion rather than just by length
- **Concurrency** — 240 goroutines across the `RWMutex`, verified under `-race`
- **Current-behaviour pins** — `initStreamState` leaves `StreamState.SyncMode` empty, so a freshly created stream state serializes with `"sync_mode": ""`. Pinned so that changing it is a deliberate act rather than an accident.

### Two notes for anyone writing tests against this file

Both are documented in a comment at the top of the new file, since they cost real debugging time:

1. `State` embeds `*sync.RWMutex` — a **pointer** — so a bare `&State{}` compiles fine and then nil-panics on every locking method. Fixtures need `&State{RWMutex: &sync.RWMutex{}}`.
2. Every locking mutator — `ResetStreams`, `ResetCursor`, `SetGlobal`, `SetCursor`, `SetChunks`, `RemoveChunk` (`SetType` is the exception: it neither locks nor logs) — ends in `LogState()`, which calls `logger.Fatalf` → `os.Exit(1)` when `constants.StatePath` is unset. That kills the test binary with **no failure output at all**. It is latent, too: `LogState` returns early while `isZero()` is true, so tests pass right up until one stores real data. Fixtures point the path at a `t.TempDir()` file.

## Type of change

- [x] Test coverage only (no production code changed, no behaviour change)

# How Has This Been Tested?

- [x] `go test ./types/... -count=1 -v` — 108 cases, all passing
- [x] `go test ./types/... -race` — clean, no data races
- [x] `make test.unit` equivalent across the root module — all packages pass
- [x] `make olake.lint` — exit 0, no findings
- [x] `make build` — clean
- [x] `gofmt -l -s` and `go vet ./types/...` — clean

**Coverage of `state.go` after this PR** — every function is exercised; 15 of 18 are at 100%:

| Function | Coverage |
|---|---|
| `initStreamState`, `isZero`, `SetType`, `ResetStreams`, `ResetCursor` | 100% |
| `HasCompletedBackfill`, `GetGlobal`, `SetGlobal` | 100% |
| `GetCursor`, `SetCursor` | 100% |
| `GetChunks`, `SetChunks`, `RemoveChunk` | 100% |
| `State.MarshalJSON`, `LogWithLock` | 100% |
| `LogState` | 83.3% |
| `StreamState.MarshalJSON` | 88.9% |
| `StreamState.UnmarshalJSON` | 84.2% |

The three partial ones are short only of branches that cannot be reached from a test:

- `LogState` — the `logger.Fatalf` → `os.Exit(1)` arm, which terminates the test binary by definition
- `StreamState.MarshalJSON` — the non-string `sync.Map` key arm, unreachable because every writer in `state.go` keys on a string
- `StreamState.UnmarshalJSON` — the two `json.Marshal`/`json.Unmarshal` error arms on a value that was just successfully decoded

# Screenshots or Recordings

https://drive.google.com/file/d/1ySoqGA1kDrU8iR6W69hPZ_jaSQQut7pT/view?usp=sharing

## Documentation

- [x] N/A (test changes only)

## Related PR's (If Any):

None.

---

### One thing I would like a maintainer's call on

While writing these tests I found that `RemoveChunk` does not check its type assertion:

```go
// types/state.go:248
castedStateChunks, _ := stateChunks.(*Set[Chunk])
castedStateChunks.Remove(chunk)   // nil pointer dereference if the assertion failed
```

This is reachable from a malformed `state.json`. `StreamState.UnmarshalJSON` only converts `chunks` into a `Set` when the JSON value is an **array** — a state file containing `"chunks": {"min": 1, "max": 10}` leaves a `map[string]interface{}` under that key instead. `GetChunks` handles that safely (it uses the comma-ok form and returns `nil`); `RemoveChunk` panics.

`TestStreamStateUnmarshalJSON` documents the decode side of this — the case *"a chunk object is not converted into a chunk set"* asserts that a non-array value does not become a `*Set[Chunk]` — but it deliberately stops short of calling `RemoveChunk` on it, since that would assert a panic as intended behaviour.

I raised this on the issue before starting and did not get a call either way, so **I have kept this PR tests-only**. Happy to either add a small guard here or open a separate issue for it — just let me know which you prefer.


nayanj98:
hi @rishyym0927 , went through the video which you have attached. we expect that the first time contributor is able to understand the project and in case of unit test, he/she is able to assess the impact this file will make in the whole process, so i would recommend that you can attach video of complete sync from source to destination, and maybe place a debug pointer in stream.go file , this will also help you to think of better edge cases for unit tests and track the codeflow. 
 
will assign a reviewer as soon as you add that video

nayanj98:
Closing this PR as the video requirement was not fulfilled and there was no further reply from the contributor

PR #1145, opened 2026-08-24
chore: new streams format - available_streams and selected_streams

vishalm0509:
# Description

This PR adds two new optional files alongside it, without changing `streams.json`'s shape or
behavior:

- `streams.json` — unchanged. Still combined, still written every discover, byte-for-byte
  identical to current output.
- `available_streams.json` — new, `streams[]` only.
- `selected_streams.json` — new, `selected_streams` only, sparse (only `stream_name`,
  `partition_regex`, `sync_mode`, and `cursor_field` when incremental are auto-populated;
  everything else is a manual override).
- Move user-configurable stream fields (`sync_mode`, `cursor_field`, `destination_database`, `destination_table`) into `selected_streams`, with getters falling back to `streams[]` so old catalogs keep working.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

--- 

**Development doc:**
https://datazip.atlassian.net/wiki/x/AQAkLQ


types/catalog.go
```diff
@@ -151,33 +226,38 @@ func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
...
-		// NOTE: new streams are not added to selected_streams, user needs to manually enable them
-		// manipulate destination db in new streams according to old streams
-
-		// prefix == "" means old stream when db normalization feature not introduced
-		if constantValue {
-			newStream.Stream.DestinationDatabase = oldCatalog.Streams[0].Stream.DestinationDatabase
-		} else if prefix != "" {
-			newStream.Stream.DestinationDatabase = fmt.Sprintf("%s:%s", prefix, utils.Reformat(newStream.Stream.Namespace))
+		// set destination database to the stream
+		clearStreamConfigurableFields(newStream.Stream)
```
mihir-datazip: Wouldn't newly discovered streams lose sync_mode/cursor?

types/catalog.go
```diff
@@ -142,6 +216,7 @@ func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
 					oldStream := oldStreams[streamID].Stream
 					newStream := newStreams[streamID].Stream
 					MergeSelectedColumns(&metadata, oldStream, newStream)
+					migrateConfigurableFieldsFromStream(&metadata, oldStream)
```
mihir-datazip: Config migration only runs for selected streams under a non-nil selected_streams, but clearing (line 257) is unconditional — legacy and deselected streams might lose their config on a discover refresh.

types/catalog.go
```diff
@@ -98,11 +167,16 @@ func GetWrappedCatalog(streams []*Stream, driver string) *Catalog {
...
 		catalog.SelectedStreams[stream.Namespace] = append(catalog.SelectedStreams[stream.Namespace], StreamMetadata{
-			StreamName:      stream.Name,
-			AppendMode:      utils.Ternary(driver == string(constants.Kafka), true, false).(bool),
-			Normalization:   IsDriverRelational(driver),
-			SelectedColumns: selectedCols,
+			StreamName:          stream.Name,
+			AppendMode:          utils.Ternary(driver == string(constants.Kafka), true, false).(bool),
+			Normalization:       IsDriverRelational(driver),
+			SelectedColumns:     selectedCols,
+			SyncMode:            stream.SyncMode,
```
mihir-datazip: Baking discover defaults into metadata makes them indistinguishable from user choices, and metadata-first precedence silently kills the legacy streams[] editing surface

After this PR, discover writes e.g. `metadata.sync_mode="cdc"` (driver default) and clears `stream.sync_mode`. Anyone following the long-standing workflow of editing `streams[].stream.sync_mode`/`cursor_field` — including this repo's own IT helper `updateStreamConfig` (tests/testutils/test_utils.go:626, currently green only because sync suites seed from pre-PR fixtures) — has the edit silently ignored: `resolveConfigurableField` returns the baked metadata default. Consider only persisting fields into metadata that differ from defaults, or documenting/warning on the dead surface.

types/catalog.go
```diff
@@ -235,16 +315,45 @@ func MergeSelectedColumns(metadata *StreamMetadata, oldStream *Stream, newStream
...
+
+	var destDBs []string
+	for namespace, metadataList := range catalog.SelectedStreams {
+		for _, metadata := range metadataList {
+			streamID := fmt.Sprintf("%s.%s", namespace, metadata.StreamName)
+			// Stale selected_streams entries (no matching streams[]) would append "" and break prefix detection.
+			if _, ok := streamDestDB[streamID]; !ok {
+				continue
+			}
+			destDBs = append(destDBs, resolveConfigurableField(metadata.DestinationDatabase, streamDestDB[streamID]))
```
mihir-datazip: Deriving the dest-db pattern from selected streams only lets a single selected stream's custom database become the "pattern" stamped onto every newly discovered stream.

New-format catalogs have cleared dest DBs on unselected streams[], so with one selected stream whose metadata destination_database is `custom_db`, `destDBs==["custom_db"]` -> `(true, "custom_db")` -> mergeCatalogs stamps every new stream's destination_database with `custom_db`; when the user later enables one, migrate copies it into metadata and its data lands in `custom_db`. Pre-PR the comparison spanned ALL streams, saw the discover defaults (`prefix:namespace`) mismatch, and returned no pattern.

types/catalog.go
```diff
@@ -142,6 +216,7 @@ func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
 					oldStream := oldStreams[streamID].Stream
```
mihir-datazip: This might panic on nil-pointer when an old selected_streams entry has no matching old streams[] entry but the stream exists in the new discover. Please check once

types/catalog.go
```diff
@@ -235,16 +315,45 @@ func MergeSelectedColumns(metadata *StreamMetadata, oldStream *Stream, newStream
...
+				destDBs = append(destDBs, s.Stream.DestinationDatabase)
+			}
+		}
+	}
+
+	if len(destDBs) == 0 {
+		return false, ""
+	}
+
+	prefixOrConstValue := strings.Split(destDBs[0], ":")
```
mihir-datazip: Can you check if destDBs[0] is deterministic? It comes from ranging over the SelectedStreams map
The old code iterated the streams slice deterministically. Sorting the entries (or iterating streams[] and looking up metadata) restores determinism.

types/catalog.go
```diff
@@ -78,6 +105,48 @@ type StreamMix struct {
...
+		}
+		schemaCatalog := &Catalog{}
+		if err := utils.UnmarshalFile(schemaFilePath, schemaCatalog, false); err != nil {
+			return nil, fmt.Errorf("failed to read schema from %s: %s", schemaFilePath, err)
+		}
+		catalog.Streams = schemaCatalog.Streams
+		return catalog, nil
+	}
+
+	return nil, fmt.Errorf("streams file %s has no streams[] and no selected_streams; file may be empty or malformed", streamsFilePath)
```
mihir-datazip: I think hard error on an empty catalog can brick a split-layout job, and regresses `{}` placeholder streams files, check it once

protocol/discover.go
```diff
@@ -84,16 +85,17 @@ var discoverCmd = &cobra.Command{
...
-	if serr := utils.UnmarshalFile(streamsPath, &oldStreams, false); serr != nil {
-		return fmt.Errorf("failed to read old catalog: %w", serr)
+	oldStreams, err := types.ResolveCatalog(streamsPath, schemaPath)
+	if err != nil {
+		return fmt.Errorf("failed to read old catalog: %w", err)
 	}
 
-	if derr := utils.UnmarshalFile(differencePath, &newStreams, false); derr != nil {
-		return fmt.Errorf("failed to read new catalog: %w", derr)
+	newStreams, err := types.ResolveCatalog(differencePath, schemaPath)
```
mihir-datazip: Both catalogs share one --schema: in split layout the new catalog is joined against the OLD schema, so newly added streams silently vanish from the diff.

With `--streams old-streams.json --schema old-schema.json --difference new-streams.json` (written by a newer discover with its own schema.json), `newStreams` gets the old schema's streams[]; in `GetStreamsDelta` a new selected entry with no streams[] match hits `if !newStreamExists { continue }` (types/catalog.go:405) and the addition is silently missing from difference_streams.json. There is no flag to supply the second schema.

types/catalog.go
```diff
@@ -235,16 +315,45 @@ func MergeSelectedColumns(metadata *StreamMetadata, oldStream *Stream, newStream
...
+				continue
+			}
+			destDBs = append(destDBs, resolveConfigurableField(metadata.DestinationDatabase, streamDestDB[streamID]))
+		}
+	}
+
+	// Legacy catalogs: dest DB may live only on streams[] when selected_streams is empty or nil.
+	if len(destDBs) == 0 {
+		for _, s := range catalog.Streams {
+			if s.Stream != nil && s.Stream.DestinationDatabase != "" {
```
mihir-datazip: The legacy fallback now skips empty destination DBs, where the old code included them and bailed on mismatch

protocol/sync.go
```diff
@@ -274,7 +274,7 @@ func classifyStreams(catalog *types.Catalog, streams []*types.Stream, state *typ
 		if elem.StreamMetadata.PartitionRegex != "" {
 			classifications.Mix.Partitioned++
 		}
-		switch elem.Stream.SyncMode {
+		switch elem.GetSyncMode() {
 		case types.CDC, types.STRICTCDC:
 			// One read path, two counters: the sync treats them alike, telemetry does not.
 			if elem.Stream.SyncMode == types.STRICTCDC {
```
mihir-datazip: Strict-CDC telemetry still reads the cleared `elem.Stream.SyncMode` — StrictCDC is always 0 for new-format catalogs.

protocol/root.go
```diff
@@ -62,6 +63,10 @@ var RootCmd = &cobra.Command{
 			viper.Set(constants.StatePath, statePathEnv)
 			viper.Set(constants.StreamsPath, streamsPathEnv)
 			viper.Set(constants.DifferencePath, differencePathEnv)
+			if splitStreams || schemaPath != "" {
```
mihir-datazip: --split-streams / --schema derivation is gated on `!noSave`, so the flags are silently inert with --no-save.

types/stream_configured.go
```diff
@@ -163,12 +163,16 @@ func (s *ConfiguredStream) SupportedSyncModes() *Set[SyncMode] {
...
 func (s *ConfiguredStream) GetSyncMode() SyncMode {
-	return s.Stream.SyncMode
+	return resolveConfigurableField(s.StreamMetadata.SyncMode, s.Stream.SyncMode)
 }
 
 func (s *ConfiguredStream) GetDestinationDatabase(icebergDB *string) string {
-	if s.Stream.DestinationDatabase != "" {
-		return utils.Reformat(s.Stream.DestinationDatabase)
+	destDB := resolveConfigurableField(s.StreamMetadata.DestinationDatabase, s.Stream.DestinationDatabase)
+	if destDB == "" && s.Stream.DefaultStreamProperties != nil {
```
mihir-datazip: **[low] The DefaultStreamProperties fallback sits above the icebergDB config fallback, and new discovers always populate it — blanking a stream's destination database no longer routes it to the writer-level iceberg database.**

Pre-PR, an empty destination_database fell through to `*icebergDB` (destination/iceberg passes `&i.config.IcebergDatabase`). After any discover with this build, `default_stream_properties.destination_database` carries the discover default (the prefix defaults to the connector type, so `GenerateDestinationDetails` never returns empty), which now shadows the iceberg config.

types/catalog.go
```diff
@@ -78,6 +105,48 @@ type StreamMix struct {
...
+// Default layout: streams.json contains both streams[] and selected_streams.
+// This is the normal output of discover and is returned as-is; schemaPath is ignored.
+//
+// Split layout (opt-in via --schema): streams.json contains only selected_streams;
+// streams[] lives in a separate schema.json. When streams[] is absent, schemaPath must
+// point to the schema file or an error is returned.
+func ResolveCatalog(streamsFilePath, schemaFilePath string) (*Catalog, error) {
+	catalog := &Catalog{}
+	if err := utils.UnmarshalFile(streamsFilePath, catalog, false); err != nil {
+		return nil, fmt.Errorf("failed to read streams from %s: %s", streamsFilePath, err)
```
mihir-datazip: `%s` (here and line 135) severs the errs classification chain. we should use %w

types/catalog.go
```diff
@@ -78,6 +105,48 @@ type StreamMix struct {
...
+	// streams[] is absent — split layout (selected_streams populated, streams[] in schema.json)
+	if len(catalog.SelectedStreams) > 0 {
+		if schemaFilePath == "" {
+			return nil, fmt.Errorf("--schema required: streams.json contains only selected_streams (split layout). Pass --schema <path> to provide the stream metadata file")
+		}
+		schemaCatalog := &Catalog{}
+		if err := utils.UnmarshalFile(schemaFilePath, schemaCatalog, false); err != nil {
+			return nil, fmt.Errorf("failed to read schema from %s: %s", schemaFilePath, err)
+		}
+		catalog.Streams = schemaCatalog.Streams
```
mihir-datazip: The split join never reconciles selected_streams against schema.json — a selected entry missing from a stale schema is silently never synced.

`classifyStreams` iterates `catalog.Streams`, so a selection with no schema entry is never visited and no warning is logged; sync reports success while the stream is skipped. The silent skip predates this PR for combined files, but with the selection now living in a separately-edited file it becomes a routine hazard — worth at least a warning for selected entries with no matching stream.

drivers/abstract/abstract.go
```diff
@@ -143,9 +143,11 @@ func (a *AbstractDriver) Discover(ctx context.Context, maxDiscoverThreads int, i
 
 		// add default stream properties
 		convStream.DefaultStreamProperties = &types.DefaultStreamProperties{
-			Normalization: types.IsDriverRelational(a.driver.Type()),
-			AppendMode:    a.driver.Type() == string(constants.Kafka),
+			Normalization:       types.IsDriverRelational(a.driver.Type()),
+			AppendMode:          a.driver.Type() == string(constants.Kafka),
 			UpdateType:    types.UpdateTypeEquality,
```
vaibhav-datazip, line 150: there is a formatting issue here, you can setup pre-commit , it will check all such things

types/catalog.go
```diff
@@ -80,6 +107,48 @@ type StreamMix struct {
...
+
+	// streams[] is present — this is the default combined layout
+	if len(catalog.Streams) > 0 {
+		return catalog, nil
+	}
+
+	// streams[] is absent — split layout (selected_streams populated, streams[] in schema.json)
+	if len(catalog.SelectedStreams) > 0 {
+		if schemaFilePath == "" {
+			return nil, fmt.Errorf("--schema required: streams.json contains only selected_streams (split layout). Pass --schema <path> to provide the stream metadata file")
```
vaibhav-datazip: we can add error with precondition here , as we have done at other places

types/catalog.go
vishalm0509: If we are removing it from the `stream` then for new discover they will be lost. 

We can set them in the `DefaultStreamProperties` but I remember discussing it to be used only for **connector** level properties like `append_mode`, `normalization`. We can discuss regd. the approach we will be going forward with. 

I have added them to `DefaultStreamProperties` now to avoid losing the value.

types/catalog.go
vishalm0509: I have removed the clear from the `mergeCatalog` that should address this without any side effects. Also, earlier `sync_mode`, `cursor_field`, `destination_database`, `destination_table` were preserved for even if they are unselected and selected. Now they are not, and are stored in the defaults.

This shouldn't affect any Sync process.

types/catalog.go
vishalm0509: This has been removed now - as all the configurable fields have been moved to `selected_streams`. We are no longer checking for destination_database patterns.

types/catalog.go
vishalm0509: Removed this check completely. As earlier we had this field in the `[]streams`, now it's in `[]selected_streams` and user will have to edit it if a new streams has been selected. If not edited, we will use the value from the `DefaultStreamProperties`. 

We can discuss about this approach internally.

types/catalog.go
vishalm0509: Now I have changed the priority order in which the value will be read `[]streams` -> `[]selected_streams` -> `default_stream_properties`. 

We are sure `[]streams` will not have any of these fields from this version, if the user by mistake out-of-practice edits the field in the `[]streams` we will read that value only and omit the value set in the `[]selected_streams`

types/catalog.go
vishalm0509: It can result in nil pointer error. But that case should ideally never occur.

protocol/sync.go
vishalm0509: Fixed

types/catalog.go
```diff
@@ -80,6 +107,48 @@ type StreamMix struct {
...
+		}
+		schemaCatalog := &Catalog{}
+		if err := utils.UnmarshalFile(schemaFilePath, schemaCatalog, false); err != nil {
+			return nil, fmt.Errorf("failed to read schema from %s: %s", schemaFilePath, err)
+		}
+		catalog.Streams = schemaCatalog.Streams
+		return catalog, nil
+	}
+
+	return nil, fmt.Errorf("streams file %s has no streams[] and no selected_streams; file may be empty or malformed", streamsFilePath)
```
vaibhav-datazip: here as well , you can add a error preconditioned with a type

protocol/root.go
vishalm0509: `--no-save` is used when we want don't want the files to be written - so it is logged to the stdout. Now in case of `--split-streams` the stdout is same as that of normal discover so this shouldn't matter.

types/catalog.go
vishalm0509: Fixed

types/catalog.go
vishalm0509: Addressed.

types/catalog.go
vishalm0509: Moved to DefaultStreamProperties - so removed this code.

types/stream_configured_test.go
```diff
@@ -447,3 +447,197 @@ func TestConfiguredStream_GetFilter(t *testing.T) {
...
+			metadata: StreamMetadata{CursorField: "updated_at"},
+			wantPri:  "created_at",
+		},
+		{
+			name: "falls back to default_stream_properties when metadata and stream empty",
+			stream: &Stream{
+				DefaultStreamProperties: &DefaultStreamProperties{CursorField: "updated_at"},
+			},
+			wantPri: "updated_at",
+		},
```
vaibhav-datazip, line 536: can we add a case with multiple cursor fields as well

types/stream_configured_test.go
```diff
@@ -447,3 +447,197 @@ func TestConfiguredStream_GetFilter(t *testing.T) {
...
+
+func TestConfiguredStream_GetDestinationDatabase(t *testing.T) {
+	tests := []struct {
+		name     string
+		stream   *Stream
+		metadata StreamMetadata
+		want     string
+	}{
+		{
+			name:   "legacy format reads from stream",
```
vaibhav-datazip: can we add a case where we pass non nil icebergdb value so that stream_configured.go:177  branch is also tested

protocol/root.go
```diff
@@ -60,6 +60,9 @@ var RootCmd = &cobra.Command{
 			viper.Set(constants.StatePath, statePathEnv)
 			viper.Set(constants.StreamsPath, streamsPathEnv)
 			viper.Set(constants.DifferencePath, differencePathEnv)
+			if selectedStreamsPath != "" {
+				viper.Set(constants.SelectedStreamsPath, selectedStreamsPath)
+			}
```
vaibhav-datazip, line 66: as we had discussed about this , if we are passing --selected-streams without any path

 the discover should automatically result in two files separate for streams and selected streams, please check this 

this will reduce the step of first doing normal discover then passing the path of selected-stream to get the split

types/catalog.go
```diff
@@ -80,32 +96,65 @@ type StreamMix struct {
...
+		}
+		if len(selectedCatalog.SelectedStreams) == 0 {
+			return nil, fmt.Errorf("selected_streams file %s has no selected_streams", selectedStreamsFilePath)
+		}
+		catalog.SelectedStreams = selectedCatalog.SelectedStreams
+	}
+
+	if len(catalog.Streams) == 0 && len(catalog.SelectedStreams) > 0 {
+		return nil, fmt.Errorf("streams file %s has selected_streams but no streams[]", streamsFilePath)
+	}
```
vaibhav-datazip, line 137: maybe you forgot about this precondition block here is what you can use


```suggestion
		if len(selectedCatalog.SelectedStreams) == 0 {
			return nil, errs.Precondition(errs.CatalogError, codeSelectedStreamsEmpty,
				fmt.Errorf("selected_streams file %s has no selected_streams", selectedStreamsFilePath))
		}
		catalog.SelectedStreams = selectedCatalog.SelectedStreams
	}

	if len(catalog.Streams) == 0 && len(catalog.SelectedStreams) > 0 {
		return nil, errs.Precondition(errs.CatalogError, codeStreamsMissing,
			fmt.Errorf("streams file %s has selected_streams but no streams[]", streamsFilePath))
	}
```


these codes can be added

```
const (
	codeSelectedStreamsEmpty = "catalog.selected_streams_empty"
	codeStreamsMissing       = "catalog.streams_missing"
)
```

types/catalog.go
```diff
@@ -80,32 +96,65 @@ type StreamMix struct {
...
+
+	return catalog, nil
+}
+
+// splitCatalogForWrite returns two Catalog values for the opt-in split file layout:
+// streams[] only, and selected_streams only.
+func splitCatalogForWrite(catalog *Catalog) (streamsFile, selectedStreamsFile *Catalog) {
+	return &Catalog{Streams: catalog.Streams}, &Catalog{SelectedStreams: catalog.SelectedStreams}
+}
```
vaibhav-datazip: looks like dead code

```suggestion
func splitCatalogForWrite(catalog *Catalog) (*Catalog, *Catalog) {
	return &Catalog{Streams: catalog.Streams}, &Catalog{SelectedStreams: catalog.SelectedStreams}
}
```

protocol/sync.go
vaibhav-datazip, line 228: so now lets say we only pass streams but we have both streams and selected streams (streams only has stream related data)

in that case now our sync will run with all the streams right ? even clear destination can clear all ?

should we handle this ? a similar thing might happen in merge stream as well

types/catalog.go
vishalm0509: Done

types/catalog.go
vishalm0509: Removed

protocol/sync.go
vishalm0509: Done

protocol/root.go
vishalm0509: It is possible - but UX is not as expected. We can discuss.
> http://github.com/spf13/cobra/issues/818

types/catalog.go
vishalm0509: Done after second review

types/catalog.go
vishalm0509: Done

types/stream_configured_test.go
vishalm0509: Removed these changes

types/stream_configured_test.go
vishalm0509: removed

destination/parquet/parquet_test.go
```diff
@@ -17,13 +17,15 @@ import (
...
+func boolPtr(b bool) *bool { return &b }
+
 func newStream(normalized bool, partitionRegex string, cols map[string]types.DataType) *types.ConfiguredStream {
 	st := types.NewStream("users", "public", nil)
 	for name, dt := range cols {
 		st.UpsertField(name, dt, true, false)
 	}
 	return &types.ConfiguredStream{
-		StreamMetadata: types.StreamMetadata{Normalization: normalized, PartitionRegex: partitionRegex},
+		StreamMetadata: types.StreamMetadata{Normalization: boolPtr(normalized), PartitionRegex: partitionRegex},
```
mihir-datazip, line 28: We can just do `&normalised` right?

destination/parquet/parquet_2pc_test.go
```diff
@@ -761,7 +761,7 @@ func testConfiguredStream() *types.ConfiguredStream {
 	return &types.ConfiguredStream{
 		Stream: stream,
 		StreamMetadata: types.StreamMetadata{
-			Normalization: false,
+			Normalization: boolPtr(false),
```
mihir-datazip, line 764: just use `new(bool)` here

protocol/clear.go
```diff
@@ -26,16 +26,16 @@ var clearCmd = &cobra.Command{
 			return err
 		}
 
-		catalog = &types.Catalog{}
-		if err := utils.UnmarshalFile(streamsPath, catalog, false); err != nil {
+		catalog, err = types.ResolveCatalog(streamsPath, selectedStreamsPath)
+		if err != nil {
```
mihir-datazip, line 30: Just verify if we should use `errs.Precondition` here or not as this flow of resolving catalog might need telemetry.
cc: @vaibhav-datazip

protocol/sync_test.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func boolPtr(b bool) *bool {
```
mihir-datazip, line 13: I think we can remove this by using `&var` or `new(bool)`

protocol/sync.go
```diff
@@ -313,6 +318,14 @@ func classifyStreams(catalog *types.Catalog, streams []*types.Stream, state *typ
 
 		return false
 	})
+
+	// in case of split-write, streams.json & selected_streams.json are maintained separately.
+	// so we need to check if all the selected streams are present in the streams[]
+	for id := range selectedStreamsMap {
+		if _, ok := matchedSelected[id]; !ok {
+			logger.Warnf("Skipping; selected stream %s has no matching entry in streams[]. Rediscover or check streams.json.", id)
```
mihir-datazip: Is not failing the sync here intentional? Which case can this actually happen?
To me this seems like data corruption for which we should ideally hard fail

types/stream_configured_test.go
```diff
@@ -447,3 +447,285 @@ func TestConfiguredStream_GetFilter(t *testing.T) {
...
+	tests := []struct {
+		name     string
+		stream   *Stream
+		metadata StreamMetadata
+		want     bool
+	}{
+		{
+			name:     "explicit true in metadata",
+			stream:   &Stream{},
+			metadata: StreamMetadata{Normalization: boolPtr(true)},
```
mihir-datazip, line 655: Same, we can use `bool(true)`

types/catalog.go
```diff
@@ -80,32 +105,95 @@ type StreamMix struct {
...
+// ResolveCatalog loads a catalog from disk.
+// Combined layout: streams.json has both streams[] and selected_streams.
+// Split layout: pass --selected-streams so selected_streams is overlaid from that file.
+func ResolveCatalog(streamsFilePath, selectedStreamsFilePath string) (*Catalog, error) {
+	catalog := &Catalog{}
+	if err := utils.UnmarshalFile(streamsFilePath, catalog, false); err != nil {
+		return nil, fmt.Errorf("failed to read streams from %s: %w", streamsFilePath, err)
+	}
+
+	if selectedStreamsFilePath == "" && len(catalog.Streams) > 0 && len(catalog.SelectedStreams) == 0 {
```
mihir-datazip, line 117: This might break the older behaviour of `nil` selected_streams sync everything.
Please confirm once if this behaviour is no longer needed now

types/catalog.go
mihir-datazip, line 1: `Sync mode` and `cursor` edits from older UIs are silently ignored now.
As discover now stamps `sync_mode`, and `cursor_field` for incremental streams, into every selected_streams entry at types/catalog.go:188. Every released olake-ui, and current main, still edits sync mode and cursor on streams[].stream and passes `selected_streams` entries through untouched via a spread. The getter at types/stream_configured.go:173 gives selected_streams priority, so the stale discover default wins.
A user who flips CDC to full refresh in the UI still gets CDC at sync time.
The difference flow misses the change for the same reason, and mergeCatalogs carries the stale value across rediscovers. [PR](https://github.com/datazip-inc/olake-ui/pull/420) on olake-ui only persists a split file, so this bites the new UI too unless it also moves those edits.

types/catalog.go
```diff
@@ -50,16 +59,32 @@ type StreamMetadata struct {
...
 	// When enabled, source column names are preserved as-is; otherwise utils.Reformat() is applied to generate destination-safe lowercase column names.
-	UseSourceColumnNames bool `json:"use_source_column_names"`
+	UseSourceColumnNames bool `json:"use_source_column_names,omitempty"`
 	//legacy filter input
 	Filter string `json:"filter,omitempty"`
 	//new filter input
-	FilterConfig    *FilterConfig    `json:"filter_config,omitempty"`
-	SelectedColumns *SelectedColumns `json:"selected_columns"`
+	FilterConfig        *FilterConfig    `json:"filter_config,omitempty"`
+	SelectedColumns     *SelectedColumns `json:"selected_columns,omitempty"`
```
mihir-datazip, line 71: Make sure that column selection is not regressed in olake-ui after this change

types/catalog_test.go
mihir-datazip, line 294: Nothing in `types/` calls `MergeSelectedColumns` from a test. It is only reached through `mergeCatalogs` in `TestCatalogMergeCatalogs`, in two cases ("single stream merge -- old stream fields carried forward" and "use new cursor field when old cursor field is empty"), and both run with `sync_new_columns: false`. The merge case in `TestLogCatalog` sets `sync_new_columns: true` on the old file but never asserts on the resulting columns.

This PR changed the function's contract (nil or empty `selected_columns` now stays unset instead of being filled with the full schema, and `mergeCatalogs` gained an `oldConfigured != nil` guard), so the new behaviour is currently unverified. Cases worth adding to `TestCatalogMergeCatalogs`:

- `sync_new_columns: true`: schema drops one column and gains another, expect the removed one gone and the new one present.
- `sync_new_columns: true` with a deselected column that already existed in the old schema: it must stay out, only columns absent from the old schema join.
- OLake column deselected with `sync_new_columns: false`: it must be re-added (`prop.OlakeColumn` branch). `createSchemaFromTemplate` doesn't copy `OlakeColumn` today, so the helper needs that one field.
- Nil `selected_columns` on the old entry: must stay nil after merge, since sync treats nil as "all columns".
- Empty column list on the old entry: left untouched, including its `sync_new_columns` value.
- Selected stream with no matching old `streams[]` entry: metadata carried through unmerged, covering the new guard.

protocol/discover.go
mihir-datazip, line 1: We need to add test cases in Discover integration tests for coverage of the changes you did in CI

mihir-datazip: Also include the test case for merge

destination/parquet/parquet_2pc_test.go
vishalm0509: `new(bool)` works for this case because the value here is `false` so zero value is assigned. But if we want to initialize with `true` then this helper will be useful right.

destination/parquet/parquet_test.go
vishalm0509: Will address based on the decision of the below comment - https://github.com/datazip-inc/olake/pull/1145#discussion_r3954990462

destination/parquet/parquet_2pc_test.go
mihir-datazip: you can use `new(true)` in that case

protocol/clear.go
vishalm0509: we have already have `errs.Precondition` set in the `ResolveCatalog()` function

destination/parquet/parquet_2pc_test.go
vishalm0509: Done

destination/parquet/parquet_test.go
vishalm0509: Done

protocol/sync_test.go
vishalm0509: Done, removed it

types/stream_configured_test.go
vishalm0509: Done

types/catalog.go
vishalm0509: Yes, we read all the columns from `type_schema` and add it to `selected_columns` on selection.  It is handled in the UI

protocol/clear.go
vaibhav-datazip: the innermost function returns defined errors so we wont need the precondition here again

protocol/root.go
vaibhav-datazip: had discussed this with @mihir-datazip as well, we can run pass the flag to get separated files in one run only

protocol/root.go
```diff
@@ -158,13 +178,33 @@ func init() {
...
+	codeNoStreams              = "catalog.no_streams_discovered"
 	// recovered panic as an internal error
 	codePanicRecovered = "sync.panic_recovered"
 )
 
+// validateCatalogFlags rejects --streams combined with --available-streams/--selected-streams,
+// and, when required, requires at least one catalog source.
+func validateCatalogFlags(streamsFlagRequired bool) error {
+	hasLegacy := streamsPath != ""
+	hasNew := availableStreamsPath != "" || selectedStreamsPath != ""
```
vaibhav-datazip, line 193: shouldn't this be changed to && as we need both to be present for new, right ?

types/legacy_catalog.go
```diff
@@ -0,0 +1,178 @@
...
+		FilterConfig:         streamMetadata.FilterConfig,
+		SelectedColumns:      streamMetadata.SelectedColumns,
+	}
+}
+
+// legacyToCanonical converts a whole LegacyCatalog (streams.json shape) into the canonical
+// Catalog shape, so a legacy old-catalog can feed the single mergeCatalogs algorithm like any
+// new-format old catalog. Streams need no conversion (*Stream/*ConfiguredStream is shared);
+// only SelectedStreams' metadata shape differs.
+func legacyToCanonical(legacy *LegacyCatalog) *Catalog {
```
vaibhav-datazip, line 81: can this comment be corrected, as its also used in ResolveCatalog

protocol/discover.go
```diff
@@ -17,11 +17,21 @@ import (
 	"github.com/spf13/viper"
 )
 
+// isStreamDifferenceCommand reports whether both old and new streams flags were passed
+func isStreamDifferenceCommand() bool {
+	hasOldStreams := streamsPath != "" || availableStreamsPath != "" || selectedStreamsPath != ""
```
vaibhav-datazip, line 22: similarly here , shouldn't it be like this. similarly for difference

```suggestion
	hasOldStreams := streamsPath != "" || (availableStreamsPath != "" && selectedStreamsPath != "")
```

types/catalog.go
```diff
@@ -328,18 +409,30 @@ func GetStreamsDelta(oldStreams, newStreams *Catalog) *Catalog {
...
+				oldDestinationTable := resolveConfigurableField(oldMetadata.DestinationTable, oldStream.Stream.DestinationTable)
+				newDestinationTable := resolveConfigurableField(newMetadata.DestinationTable, newStream.Stream.DestinationTable)
+
 				// check cursor field if SyncMode is incremental
-				cursorDelta := utils.Ternary(newStream.Stream.SyncMode == INCREMENTAL, oldStream.Stream.CursorField != newStream.Stream.CursorField, false).(bool)
+				cursorDelta := utils.Ternary(newSyncMode == INCREMENTAL, oldCursorField != newCursorField, false).(bool)
 
-				return (oldMetadata.Normalization != newMetadata.Normalization) ||
+				return oldConfigured.NormalizationEnabled() != newConfigured.NormalizationEnabled() ||
+					oldConfigured.AppendModeEnabled() != newConfigured.AppendModeEnabled() ||
```
vaibhav-datazip, line 428: although adding parentheses around these wont make any difference , but it will make it consistent ?

types/catalog.go
```diff
@@ -351,9 +444,13 @@ func GetStreamsDelta(oldStreams, newStreams *Catalog) *Catalog {
...
-				// safely change for destination database and table if difference present
-				deltaStream.Stream.DestinationDatabase = oldStream.Stream.DestinationDatabase
-				deltaStream.Stream.DestinationTable = oldStream.Stream.DestinationTable
+				// keep the user's existing destination mapping in the diff output even when discover produced new values
+				oldDestinationDatabase := resolveConfigurableField(oldMetadata.DestinationDatabase, oldStream.Stream.DestinationDatabase)
+				oldDestinationTable := resolveConfigurableField(oldMetadata.DestinationTable, oldStream.Stream.DestinationTable)
+				deltaStream.Stream.DestinationDatabase = oldDestinationDatabase
+				deltaStream.Stream.DestinationTable = oldDestinationTable
+				newMetadata.DestinationDatabase = oldDestinationDatabase
+				newMetadata.DestinationTable = oldDestinationTable
```
vaibhav-datazip, line 453: we computed these in indifferent function as well, can we common out these variables so that we dont need to call `resolveConfigurableField` function again ?

types/catalog.go
```diff
@@ -80,37 +106,108 @@ type StreamMix struct {
...
+func (c *Catalog) sortByNamespaceStreamName() {
+	// sort []streams
+	slices.SortFunc(c.Streams, func(left, right *ConfiguredStream) int {
+		if cmp := strings.Compare(left.Stream.Namespace, right.Stream.Namespace); cmp != 0 {
+			return cmp
+		}
+		return strings.Compare(left.Stream.Name, right.Stream.Name)
+	})
+
+	// sort []selected_streams
```
vaibhav-datazip, line 164: instead of this comment you can frame it better that, values under the map keys are being sorted here, the map keys automatically get sorted while getting written in the file  as encoding/json always writes map keys in sorted order.

types/catalog.go
```diff
@@ -50,16 +60,32 @@ type StreamMetadata struct {
...
+// resolveConfigurableField returns the first non-zero value.
+// Callers pass values in priority order: selected_streams, then streams[].
+func resolveConfigurableField[T comparable](values ...T) T {
+	var zero T
+	for _, v := range values {
+		if v != zero {
+			return v
+		}
+	}
+	return zero
```
vaibhav-datazip, line 88: Can we drop the helper and call `cmp.Or` directly at the call sites? 

```go
syncMode := cmp.Or(s.StreamMetadata.SyncMode, s.Stream.SyncMode)
```

types/catalog.go
```diff
@@ -328,18 +409,30 @@ func GetStreamsDelta(oldStreams, newStreams *Catalog) *Catalog {
...
+				oldCursorField := resolveConfigurableField(oldMetadata.CursorField, oldStream.Stream.CursorField)
+				newCursorField := resolveConfigurableField(newMetadata.CursorField, newStream.Stream.CursorField)
+				oldDestinationDatabase := resolveConfigurableField(oldMetadata.DestinationDatabase, oldStream.Stream.DestinationDatabase)
+				newDestinationDatabase := resolveConfigurableField(newMetadata.DestinationDatabase, newStream.Stream.DestinationDatabase)
+				oldDestinationTable := resolveConfigurableField(oldMetadata.DestinationTable, oldStream.Stream.DestinationTable)
+				newDestinationTable := resolveConfigurableField(newMetadata.DestinationTable, newStream.Stream.DestinationTable)
+
 				// check cursor field if SyncMode is incremental
-				cursorDelta := utils.Ternary(newStream.Stream.SyncMode == INCREMENTAL, oldStream.Stream.CursorField != newStream.Stream.CursorField, false).(bool)
+				cursorDelta := utils.Ternary(newSyncMode == INCREMENTAL, oldCursorField != newCursorField, false).(bool)
```
vaibhav-datazip, line 425:
```suggestion
				cursorDelta := newSyncMode == INCREMENTAL && oldCursorField != newCursorField
```

types/catalog.go
```diff
@@ -80,37 +106,108 @@ type StreamMix struct {
...
+
+// sortByNamespaceStreamName orders streams[] by namespace then name, and each selected_streams
+// namespace slice by stream_name. encoding/json already sorts selected_streams map keys.
+func (c *Catalog) sortByNamespaceStreamName() {
+	// sort []streams
+	slices.SortFunc(c.Streams, func(left, right *ConfiguredStream) int {
+		if cmp := strings.Compare(left.Stream.Namespace, right.Stream.Namespace); cmp != 0 {
+			return cmp
+		}
+		return strings.Compare(left.Stream.Name, right.Stream.Name)
```
vaibhav-datazip, line 161: similarly here


```suggestion
	return cmp.Or(
    strings.Compare(left.Stream.Namespace, right.Stream.Namespace),
    strings.Compare(left.Stream.Name, right.Stream.Name),
)
```

types/catalog.go
```diff
@@ -80,37 +106,108 @@ type StreamMix struct {
...
+	})
+
+	// sort []selected_streams
+	for namespace := range c.SelectedStreams {
+		slices.SortFunc(c.SelectedStreams[namespace], func(a, b StreamMetadata) int {
+			return strings.Compare(a.StreamName, b.StreamName)
```
vaibhav-datazip, line 167: should we sort selected columns as well during such operation ?

protocol/root_test.go
```diff
@@ -117,6 +117,54 @@ func TestSignalAwareRootContextPreservesParentCancellation(t *testing.T) {
...
+		required     bool
+		expectErr    bool
+		expectedCode string
+	}{
+		{name: "legacy only, required", streams: "streams.json", required: true, expectErr: false},
+		{name: "new format pair, required", available: "available_streams.json", selected: "selected_streams.json", required: true, expectErr: false},
+		{name: "nothing passed, required", required: true, expectErr: true, expectedCode: codeFlagMissing},
+		{name: "nothing passed, not required (discover)", required: false, expectErr: false},
+		{name: "legacy and new format together", streams: "streams.json", available: "available_streams.json", selected: "selected_streams.json", required: true, expectErr: true, expectedCode: codeConflictingStreamFlags},
+		{name: "legacy and available only", streams: "streams.json", available: "available_streams.json", required: true, expectErr: true, expectedCode: codeConflictingStreamFlags},
```
vaibhav-datazip, line 148: you can add a test case where only one of available or selected stream is available

types/catalog_test.go
```diff
@@ -1101,3 +1225,684 @@ func validateBasicSchemas(t *testing.T, expected, actual *TypeSchema, testName s
...
+}
+
+func writeLegacyCatalogFile(t *testing.T, dir, name string, catalog *LegacyCatalog) string {
+	t.Helper()
+	path := filepath.Join(dir, name)
+	data, err := json.Marshal(catalog)
+	require.NoError(t, err)
+	require.NoError(t, os.WriteFile(path, data, 0600))
+	return path
+}
```
vaibhav-datazip, line 1537: these are identical maybe we can replace them with something like this 

```
 writeJSONFile(t, dir, name string, v any)
```

types/catalog_test.go
```diff
@@ -1101,3 +1225,684 @@ func validateBasicSchemas(t *testing.T, expected, actual *TypeSchema, testName s
...
+func TestLogCatalog(t *testing.T) {
+	t.Run("always writes all three files, new user (no prior catalog)", func(t *testing.T) {
+		dir := t.TempDir()
+		streamsPath, availablePath, selectedPath := setLogCatalogPaths(t, dir)
+
+		discovered := []*Stream{
+			{Name: "users", Namespace: "public", Schema: oldSchema(), SyncMode: INCREMENTAL, CursorField: "updated_at", DefaultStreamProperties: &DefaultStreamProperties{Normalization: true, UpdateType: UpdateTypeEquality}},
+		}
+
+		LogCatalog(discovered, nil, nil, "postgres")
```
vaibhav-datazip, line 1689: maybe you can create this function similar to how we write test case across the codebase , breaking it into test cases, also some of the code looks too repetitive so maybe a helper function can be used to simplify this

types/catalog_test.go
```diff
@@ -1101,3 +1225,684 @@ func validateBasicSchemas(t *testing.T, expected, actual *TypeSchema, testName s
...
+		delta := GetStreamsDelta(cat, cat)
+		assert.Empty(t, delta.Streams)
+		assert.Empty(t, delta.SelectedStreams)
+	})
+
+	t.Run("new stream not in old catalog added to delta", func(t *testing.T) {
+		old := &Catalog{
+			Streams:         []*ConfiguredStream{},
+			SelectedStreams: map[string][]StreamMetadata{},
+		}
```
vaibhav-datazip, line 1248: similarly here as well the test case array format is not being followed also in some cases we are just checking the len , maybe we can make that better by asserting actual streams

types/catalog_test.go
```diff
@@ -1101,3 +1225,684 @@ func validateBasicSchemas(t *testing.T, expected, actual *TypeSchema, testName s
...
+			},
+			SelectedStreams: map[string][]StreamMetadata{
+				"public": {{StreamName: "users", SyncMode: SyncMode("incremental")}},
+			},
+		}
+		delta := GetStreamsDelta(old, newCat)
+		require.Len(t, delta.Streams, 1)
+	})
+
+	t.Run("old dest DB preserved in delta output (metadata priority)", func(t *testing.T) {
```
vaibhav-datazip, line 1315: I can't find dest table case , please add if missed

protocol/root.go
vishalm0509: Yes, we need to make sure both are passed together as they are coupled. Added this check the validateCatalogFlags

types/legacy_catalog.go
vishalm0509: Done

protocol/discover.go
vishalm0509: Done

types/catalog.go
vishalm0509: Agreed. Done

types/catalog.go
vishalm0509: No, they are part of the same function and I feel leaving it as it is more readable

types/catalog.go
vishalm0509: I have updated the comment

types/catalog.go
vishalm0509: Helper is explicit for our use case. so that we use that everywhere we use any configurable fields.

types/catalog.go
vishalm0509: Cleaner. Updated it

types/catalog.go
vishalm0509: Okay, done

protocol/root_test.go
vishalm0509: Added, now we are failing loudly if either one is missing.

types/catalog_test.go
vishalm0509: I have added

types/catalog_test.go
vishalm0509: Done

types/catalog_test.go
vishalm0509: Done

types/catalog_test.go
vishalm0509: Removed it now. Using expected directly now

vishalm0509:
I have also added the following change:

`toLegacyCatalog` now writes each selected stream's `sync_mode`, `cursor_field`, `destination_database` and `destination_table` overrides onto a copy of its `streams[]` entry in `streams.json`, so the legacy file gives the same sync mode, cursor and destination as the split files, while available_streams.json keeps the discovered values.

PR #1161, opened 2026-08-26
feat: upsert mode in Kafka(source)

sahitya-datazip:
# Description

Added source side **Kafka Upsert** support with Iceberg positional deletes

When `append_mode` is `false` and `dedup_keys` are set on streams
- Records with at least one non-null selected dedup field upsert via `_olake_id` derived from those keys
- Missing selected dedup fields append (offset/partition identity)
- All selected dedup fields present but **null** fail the sync
- Mode A (`dedup_keys = ["_kafka_key"]` only): Upsert + Kafka tombstones deletes
- Mode B (`dedup_keys = selected fields`): Upsert only

Upsert streams force update_type=pos

[PRD](https://datazip.atlassian.net/wiki/spaces/PD/pages/654704668/Kafka+Upsert+Mode+Support)

## Type of change

# How Has This Been Tested?

- [x] Locally tested Upsert cases
- [x] Unit tests
- [x]  Integration tests for Upsert

## Documentation

### DIFF
--- a/destination/iceberg/arrow-writer/writer.go
+++ b/destination/iceberg/arrow-writer/writer.go
@@ -173,39 +173,41 @@ func (w *ArrowWriter) extract(ctx context.Context, records []types.RawRecord) er
 		}
 
 		writer.data = append(writer.data, rec)
-		recordOpType := rec.OlakeColumns[constants.OpType].(string)
 		recordOlakeID := rec.OlakeColumns[constants.OlakeID].(string)
 		filePosition := writer.dataWriter.currentRowCount + int64(len(writer.data)-1)
 
 		if w.indexThread != nil {
-			if err := w.indexRecord(writer, recordOlakeID, recordOpType, filePosition); err != nil {
+			if err := w.indexRecord(writer, recordOlakeID, rec.OlakeColumns, filePosition); err != nil {
 				return err
 			}
-		} else if w.upsertMode && (recordOpType == "d" || recordOpType == "u" || recordOpType == "i") {
-			if _, exists := writer.olakeIDPosition[recordOlakeID]; !exists {
-				// first time, add to equality deletes and track position
-				writer.equalityDeletes = append(writer.equalityDeletes, recordOlakeID)
-				writer.olakeIDPosition[recordOlakeID] = PositionalDelete{
-					FilePath: writer.dataWriter.filePath,
-					Position: filePosition,
-				}
-			} else {
-				// duplicates, add prev position to positional deletes (n-1 logic)
-				// the latest (nth) occurrence is kept in the map but not added to deletes
-				prev := writer.olakeIDPosition[recordOlakeID]
-				writer.positionalDeletes = append(writer.positionalDeletes, PositionalDelete{
-					FilePath: prev.FilePath,
-					Position: prev.Position,
-				})
-				writer.olakeIDPosition[recordOlakeID] = PositionalDelete{
-					FilePath: writer.dataWriter.filePath,
-					Position: filePosition,
+		} else if w.upsertMode {
+			recordOpType := rec.OlakeColumns[constants.OpType].(string)
+			if recordOpType == "d" || recordOpType == "u" || recordOpType == "i" {
+				if _, exists := writer.olakeIDPosition[recordOlakeID]; !exists {
+					// first time, add to equality deletes and track position
+					writer.equalityDeletes = append(writer.equalityDeletes, recordOlakeID)
+					writer.olakeIDPosition[recordOlakeID] = PositionalDelete{
+						FilePath: writer.dataWriter.filePath,
+						Position: filePosition,
+					}
+				} else {
+					// duplicates, add prev position to positional deletes (n-1 logic)
+					// the latest (nth) occurrence is kept in the map but not added to deletes
+					prev := writer.olakeIDPosition[recordOlakeID]
+					writer.positionalDeletes = append(writer.positionalDeletes, PositionalDelete{
+						FilePath: prev.FilePath,
+						Position: prev.Position,
+					})
+					writer.olakeIDPosition[recordOlakeID] = PositionalDelete{
+						FilePath: writer.dataWriter.filePath,
+						Position: filePosition,
+					}
 				}
 			}
 		}
 
 		// Normalise "i" → "c" in the data file so downstream consumers see a consistent op type.
-		if recordOpType == "i" {
+		if rec.OlakeColumns[constants.OpType] == "i" {
 			rec.OlakeColumns[constants.OpType] = "c"
 		}
 	}
@@ -214,8 +216,9 @@ func (w *ArrowWriter) extract(ctx context.Context, records []types.RawRecord) er
 }
 
 // search for index for the record and emmit pos when found
-func (w *ArrowWriter) indexRecord(writer *Writer, olakeID, opType string, filePosition int64) error {
-	if w.upsertMode && (opType != "r" && opType != "c") {
+func (w *ArrowWriter) indexRecord(writer *Writer, olakeID string, olakeColumns map[string]any, filePosition int64) error {
+	opType := olakeColumns[constants.OpType].(string)
+	if w.upsertMode && opType != "r" {
 		previous, found, err := w.indexThread.Lookup(olakeID)
 		if err != nil {
 			return fmt.Errorf("failed to look up row[%s] in index: %s", olakeID, err)
@@ -225,6 +228,11 @@ func (w *ArrowWriter) indexRecord(writer *Writer, olakeID, opType string, filePo
 				FilePath: previous.FilePath,
 				Position: previous.Position,
 			})
+			if opType != "d" {
+				olakeColumns[constants.OpType] = "u"
+			}
+		} else if opType == "u" {
+			olakeColumns[constants.OpType] = "c"
 		}
 	}
 
--- a/drivers/kafka/internal/cdc.go
+++ b/drivers/kafka/internal/cdc.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/datazip-inc/olake/constants"
@@ -28,6 +29,19 @@ func (k *Kafka) ChangeStreamConfig() (bool, bool, bool) {
 func (k *Kafka) PreCDC(ctx context.Context, streams []types.StreamInterface) error {
 	if len(streams) == 0 {
 		return fmt.Errorf("no valid streams found for CDC")
+	}
+
+	// Upsert: set SourceDefinedPrimaryKey from dedup_keys
+	for _, stream := range streams {
+		cfg, err := UpsertConfigFrom(stream.Self().StreamMetadata)
+		if err != nil {
+			return fmt.Errorf("stream[%s]: %s", stream.ID(), err)
+		}
+		if !cfg.Enabled {
+			continue
+		}
+		stream.GetStream().SourceDefinedPrimaryKey = types.NewSet(cfg.DedupKeys...)
+		logger.Infof("stream[%s]: upsert dedup keys %v", stream.ID(), cfg.DedupKeys)
 	}
 
 	var groupID string
@@ -116,14 +130,105 @@ func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates 
 				fmt.Errorf("missing partition Metadata for topic %s partition %d", record.Message.Topic, record.Message.Partition))
 		}
 
-		// process the change if data is present
-		if record.Data != nil {
-			// Raw wire bytes: len(Key) + len(Value). Headers are excluded — they
-			// carry protocol metadata (schema IDs, trace context), not user data.
-			err := processFn(ctx, abstract.NewCDCChange(currentPartitionMeta.Stream, record.Message.Timestamp, "create",
-				record.Data, nil, int64(len(record.Message.Key)+len(record.Message.Value))))
-			if err != nil {
-				return false, err
+		stream := currentPartitionMeta.Stream
+		cfg, err := UpsertConfigFrom(stream.Self().StreamMetadata)
+		if err != nil {
+			return false, err
+		}
+		bytesRead := int64(len(record.Message.Key) + len(record.Message.Value))
+		if !cfg.Enabled {
+			//append only
+			if record.Data != nil {
+				if err := processFn(ctx, abstract.NewCDCChange(stream, record.Message.Timestamp, "create", record.Data, nil, bytesRead)); err != nil {
+					return false, err
+				}
+			}
+		} else {
+			// upsert
+			kafkaKey := ""
+			if record.Data != nil {
+				if kKey, ok := record.Data[Key].(string); ok {
+					kafkaKey = kKey
+				}
+			}
+			if kafkaKey == "" && len(record.Message.Key) > 0 {
+				kafkaKey = k.canonicalizeKafkaKey(string(record.Message.Key))
+			}
+
+			appendByOffsetPartition := func(data map[string]any) error {
+				if data == nil {
+					data = map[string]any{}
+				}
+				data[Partition] = record.Message.Partition
+				data[Offset] = record.Message.Offset
+				if kafkaKey != "" {
+					data[Key] = kafkaKey
+				}
+				olakeID := utils.GetKeysHash(data, Offset, Partition)
+				extraColumns := map[string]any{
+					constants.OlakeID: olakeID,
+				}
+				logger.Warnf("appending with offset/partition - olake_id for topic=%s partition=%d offset=%d", record.Message.Topic, record.Message.Partition, record.Message.Offset)
+				return processFn(ctx, abstract.NewCDCChange(stream, record.Message.Timestamp, "create", data, extraColumns, bytesRead))
+			}
+
+			if record.Message.Value == nil {
+				if cfg.AllowTombstoneDeletes {
+					//dedup is only _kafka_key (present and not null) - tombstone deletes are valid
+					data, err := cfg.checkDedupKeysExist(record.Data, kafkaKey, record.KeyFields)
+					if err != nil {
+						//key - null
+						if errors.Is(err, errNullDedupKeys) {
+							return false, fmt.Errorf("stream[%s] offset=%d: %w", stream.ID(), record.Message.Offset, err)
+						}
+						// key - absent
+					} else {
+						extraColumns := map[string]any{
+							constants.OlakeID: cfg.generateOlakeIDFromExistingKeys(data),
+						}
+						if err := processFn(ctx, abstract.NewCDCChange(stream, record.Message.Timestamp, "delete", data, extraColumns, bytesRead)); err != nil {
+							return false, err
+						}
+					}
+				} else {
+					//upsert only -- null value is appended by _kafka_key (no deletes)
+					if kafkaKey != "" {
+						data := map[string]any{
+							Partition: record.Message.Partition,
+							Offset:    record.Message.Offset,
+							Key:       kafkaKey,
+						}
+						// olake_id - with kafka_key string
+						extraColumns := map[string]any{
+							constants.OlakeID: kafkaKey,
+						}
+						if err := processFn(ctx, abstract.NewCDCChange(stream, record.Message.Timestamp, "create", data, extraColumns, bytesRead)); err != nil {
+							return false, err
+						}
+					}
+					// null, null - avoid
+				}
+			} else if record.Data != nil {
+				// Non null value: Upsert when atleast one selected dedup key exist, all absent - append, all null - fail sync
+				data, err := cfg.checkDedupKeysExist(record.Data, kafkaKey, record.KeyFields)
+				if err != nil {
+					// all present dedup fields null - fail sync
+					if errors.Is(err, errNullDedupKeys) {
+						return false, fmt.Errorf("stream[%s] offset=%d: %w", stream.ID(), record.Message.Offset, err)
+					}
+					// no selected dedup fields is present - append
+					if err := appendByOffsetPartition(record.Data); err != nil {
+						return false, err
+					}
+				} else {
+					// atleast one non-null - upsert (partial null/empty/whitespace included in hash)
+					extraColumns := map[string]any{
+						constants.OlakeID: cfg.generateOlakeIDFromExistingKeys(data),
+					}
+					if err := processFn(ctx, abstract.NewCDCChange(stream, record.Message.Timestamp, "update", data, extraColumns, bytesRead)); err != nil {
+						return false, err
+					}
+				}
 			}
 		}
 
@@ -279,7 +384,7 @@ func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kgo.Client, st
 			}
 
 			message := iter.Next()
-			data, key, err := k.parseKafkaData(message)
+			data, key, keyFields, err := k.parseKafkaData(message)
 			if err != nil {
 				logger.Warnf("failed to parse message of topic: %s, partition: %d, offset %d, error: %s", message.Topic, message.Partition, message.Offset, err)
 			} else if data != nil {
@@ -293,7 +398,7 @@ func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kgo.Client, st
 				}
 			}
 
-			stopProcessing, err := stopProcessFn(types.KafkaRecord{Data: data, Message: message})
+			stopProcessing, err := stopProcessFn(types.KafkaRecord{Data: data, KeyFields: keyFields, Message: message})
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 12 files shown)


types/kafka_types.go
mihir-datazip, line 1: We should migrate Kafka types to the `drivers/kafka` having it in common code is not correct.
Please migrate it.

types/catalog.go
mihir-datazip, line 53: We should think of adding some enum here like `Mode` and have the upsert follow the enum naming

types/catalog.go
```diff
@@ -59,25 +59,15 @@ type StreamMetadata struct {
 	//new filter input
 	FilterConfig    *FilterConfig    `json:"filter_config,omitempty"`
 	SelectedColumns *SelectedColumns `json:"selected_columns"`
+
+	DedupKeys []string `json:"dedup_keys,omitempty"`
```
mihir-datazip, line 64: We need to Validate presence of at least dedup key in case of upsert mode.

Also do you think `dedup_columns` might make more sense given all other configs use column by convention?

drivers/kafka/internal/cdc.go
```diff
@@ -112,18 +125,108 @@ func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates
...
+		}
+		bytesRead := int64(len(record.Message.Key) + len(record.Message.Value))
+		if !cfg.Enabled {
+			//append only
+			if record.Data != nil {
+				if err := processFn(ctx, abstract.NewCDCChange(stream, record.Message.Timestamp, "create", record.Data, nil, bytesRead)); err != nil {
+					return false, err
+				}
+			}
+		} else {
```
mihir-datazip, line 146: Make a separate function that will resolve the olake_id from kafkaesque record, that will help us to write unit tests clearly for all cases.

types/catalog.go
sahitya-datazip: validation in [upsert.go](https://github.com/datazip-inc/olake/pull/1161#diff-1b22802d25962f8669d4e46e2184ef258ddd969ef83ae27ce75607d79d204481)
PRD mentions `dedup_keys`

types/kafka_types.go
sahitya-datazip: should we also have to migrate `pkg/kafka` into `drivers/kafka`

drivers/kafka/internal/upsert.go
```diff
@@ -0,0 +1,107 @@
...
+
+	anyPresent := false
+	anyNonNull := false
+	for _, pk := range c.DedupKeys {
+		val, ok := data[pk]
+		if !ok {
+			continue
+		}
+		anyPresent = true
+		if val != nil {
```
mihir-datazip: **Keyless messages collapse into one row.** A null/empty Kafka key is stored as `_kafka_key=""` (cdc.go:394). `""` passes this check as a present, non-null value, and `GetKeysHash` with one key returns `""`, so every keyless message gets `_olake_id=""` and deletes the previous one.

Example: `dedup_keys=["_kafka_key"]`, three null-key messages `{id:1}`, `{id:2}`, `{id:3}` → only `{id:3}` survives, with no error. The PRD says a record with no dedupe key should be **inserted**. `[_kafka_key, x]` with `x` absent collapses the same way.

Suggest treating an empty `_kafka_key` as *absent*, so it takes the append-by-offset path.

drivers/kafka/internal/cdc.go
```diff
@@ -116,14 +130,105 @@ func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates
...
+			// upsert
+			kafkaKey := ""
+			if record.Data != nil {
+				if kKey, ok := record.Data[Key].(string); ok {
+					kafkaKey = kKey
+				}
+			}
+			if kafkaKey == "" && len(record.Message.Key) > 0 {
+				kafkaKey = k.canonicalizeKafkaKey(string(record.Message.Key))
```
mihir-datazip: **Tombstones and upserts compute different keys when Schema Registry is used.** Upserts take the decoded key from `record.Data[Key]`, but a tombstone has `Data == nil`, so it falls back to the raw bytes. For Confluent wire-format keys (Avro/JSON-Schema), `canonicalizeKafkaKey` returns the raw bytes unchanged (`raw[0] != '{'`), the index lookup misses, and the delete never applies. A `d` row with a binary `_kafka_key` is appended and the entity stays live.

Suggest carrying the decoded key string that `parseKafkaData` already computes on `KafkaRecord`, and using it for both paths.

drivers/kafka/internal/cdc.go
```diff
@@ -116,14 +130,105 @@ func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates
...
+					//upsert only -- null value is appended by _kafka_key (no deletes)
+					if kafkaKey != "" {
+						data := map[string]any{
+							Partition: record.Message.Partition,
+							Offset:    record.Message.Offset,
+							Key:       kafkaKey,
+						}
+						// olake_id - with kafka_key string
+						extraColumns := map[string]any{
+							constants.OlakeID: kafkaKey,
```
mihir-datazip: **This tombstone row can overwrite a live entity.** The branch is labelled "no deletes", but `_olake_id = kafkaKey` can equal a column-based olake_id: `GetKeysHash` with one key is `fmt.Sprint(value)`. The writers now look up every non-`r` op, so the lookup hits.

Example: `dedup_keys=["customer_id"]`, the producer keys by id: `key="101"`, value `{customer_id:101,status:"active"}` → `_olake_id="101"`. A tombstone for key `"101"` → `create` with `_olake_id="101"` → index hit → the customer row is positionally deleted and replaced by a row with only `_kafka_*` columns (customer_id/status NULL).

Repeated tombstones for one key also replace each other instead of appending. If the intent is "append", the id needs to be offset/partition-based, like `appendByOffsetPartition`.

destination/iceberg/arrow-writer/writer.go
```diff
@@ -225,6 +228,11 @@ func (w *ArrowWriter) indexRecord(writer *Writer, olakeID, opType string, filePo
 				FilePath: previous.FilePath,
```
mihir-datazip: **Deletes are lost on partitioned tables.** Tombstone delete records carry only `_kafka_key`, so the partition is computed from nulls. The positional delete is queued and registered on the null partition, not on the partition that holds the row, and `registerPosDeleteFiles` attaches no referenced-file metadata. Iceberg never applies it to the live row.

Repro: `partition_regex: "/{int_value,identity}"` (the IT's own layout), key K in `int_value=100`, tombstone → K still returned by `SELECT`, next to the `d` row. The legacy writer applies it only in narrow cases (a single referenced data file, an unfiltered scan). Updates that change the partition value hit the same problem, e.g. `day(_olake_timestamp)` across midnight.

The delete should be routed to the partition of the row being deleted (`previous`), not to the incoming record's partition.

destination/iceberg/arrow-writer/writer.go
```diff
@@ -214,8 +216,9 @@ func (w *ArrowWriter) extract(ctx context.Context, records []types.RawRecord) er
 }
 
 // search for index for the record and emmit pos when found
-func (w *ArrowWriter) indexRecord(writer *Writer, olakeID, opType string, filePosition int64) error {
-	if w.upsertMode && (opType != "r" && opType != "c") {
+func (w *ArrowWriter) indexRecord(writer *Writer, olakeID string, olakeColumns map[string]any, filePosition int64) error {
+	opType := olakeColumns[constants.OpType].(string)
+	if w.upsertMode && opType != "r" {
```
mihir-datazip: **This changes behaviour for every driver, not just Kafka.** Looking up `c` ops (and relabelling hit→`u`, miss→`c`; mirrored at legacy-writer/writer.go:63) applies to every `update_type=pos` stream: Postgres/MySQL/Mongo included.

Concrete regression: a Postgres table without a PK (olake_id = hash of the whole row). After the first CDC run, inserting a row identical to an existing one hits the index and deletes the existing copy, so the destination has 1 row where the source has 2. Also:
- incremental pos streams, which always emit `u`, now store new rows as `c`
- a re-insert after a delete is stored as `u`
- every steady-state insert pays a Pebble `Get`
- the `extract()` comment at L162 is now stale

Suggest gating this on a Kafka-upsert flag (the stream's upsert config) rather than on `opType != "r"`, and adding a non-Kafka regression test.

drivers/kafka/internal/cdc.go
```diff
@@ -116,14 +130,105 @@ func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates
...
+						//key - null
+						if errors.Is(err, errNullDedupKeys) {
+							return false, fmt.Errorf("stream[%s] offset=%d: %w", stream.ID(), record.Message.Offset, err)
+						}
+						// key - absent
+					} else {
+						extraColumns := map[string]any{
+							constants.OlakeID: cfg.generateOlakeIDFromExistingKeys(data),
+						}
+						if err := processFn(ctx, abstract.NewCDCChange(stream, record.Message.Timestamp, "delete", data, extraColumns, bytesRead)); err != nil {
```
mihir-datazip: **Stream filters drop tombstones.** The delete record only has `{_kafka_key}`, and the stream filter runs before the writer (`FlattenAndCleanData`, iceberg.go:382) with no exemption for `d` ops. Any value-column condition other than `= null` / `!= value` evaluates false on the missing column (`evaluate(nil, 100, "<")` → false, utils/typeutils/filter.go:105), and the delete is silently dropped.

The PR's own `TestKafkaUpsert` config (`string_value != '' AND float_value < 100`) hits this; the IT passes only because it accepts 0 `d` rows. Deletes should bypass the filter, or be evaluated against the row they delete.

drivers/kafka/internal/cdc.go
```diff
@@ -116,14 +130,105 @@ func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates
...
+					// no selected dedup fields is present - append
+					if err := appendByOffsetPartition(record.Data); err != nil {
+						return false, err
+					}
+				} else {
+					// atleast one non-null - upsert (partial null/empty/whitespace included in hash)
+					extraColumns := map[string]any{
+						constants.OlakeID: cfg.generateOlakeIDFromExistingKeys(data),
+					}
+					if err := processFn(ctx, abstract.NewCDCChange(stream, record.Message.Timestamp, "update", data, extraColumns, bytesRead)); err != nil {
```
mihir-datazip: **Parallel readers can leave permanent duplicates.** With value-column dedup keys, one entity isn't tied to one Kafka partition. Each reader has its own writer thread and `StreamIndexThread`, and `Lookup` only sees that thread's pending writes.

Example: a 3-partition topic, `dedup_keys=["customer_id"]`, default `max_threads=3`. `{101,active}` lands on p0 (reader 0) and `{101,inactive}` on p1 (reader 1). Both lookups miss, both are written as `c`, and Pebble keeps only the last writer's location. In the next sync `{101,closed}` deletes one copy, so two live rows remain forever.

Before this PR olake_ids were unique per partition/offset, so this couldn't happen. We need either a shared index/lookup across the stream's writers, or to restrict value-column dedup to a single reader.

drivers/kafka/internal/upsert.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (c UpsertConfig) generateOlakeIDFromExistingKeys(data map[string]any) string {
+	existing := make([]string, 0, len(c.DedupKeys))
+	for _, pk := range c.DedupKeys {
+		if _, ok := data[pk]; ok {
+			existing = append(existing, pk)
+		}
+	}
+	return utils.GetKeysHash(data, existing...)
```
mihir-datazip: **The identity isn't stable across records.** Only the keys *present* in each record are hashed; `GetKeysHash` returns the raw value for one key, and otherwise md5 of the values joined by `|` with no key names and no escaping:
- `dedup [customer_id, order_id]`: `{customer_id:42}` and `{order_id:42}` both → `"42"`, so they merge.
- `{customer_id:42, order_id:null}` (`md5("42|<nil>|")`) vs `{customer_id:42}` (`"42"`) → two live rows for one entity.
- composite `{5,0}` → `md5("5|0|")`, the same id as an append-fallback row at offset 5 / partition 0.
- datatype change (PRD edge case): JSON `101` → `101.0`, or Avro long → double (`1.5e+06`), creates a new row instead of an update.

Suggest hashing **all** configured keys in a fixed order, with key names and typed/normalized values, and namespacing the append-fallback id so it can't collide.

drivers/kafka/internal/cdc.go
```diff
@@ -304,7 +409,29 @@ func (k *Kafka) processKafkaMessages(ctx context.Context, reader *kgo.Client, st
 	}
 }
 
-func (k *Kafka) parseKafkaData(message *kgo.Record) (map[string]interface{}, string, error) {
+// canonicalizeKafkaKey - normalize _kafka_key string
+func (k *Kafka) canonicalizeKafkaKey(raw string) string {
+	raw = strings.TrimSpace(raw)
```
mihir-datazip: `strings.TrimSpace` on arbitrary key bytes merges distinct binary keys. IntegerSerializer keys 9–13 and 32 all fail JSON decode and trim to the same three NUL bytes, so with `dedup [_kafka_key]` key 10's upsert deletes key 9's row. In a quick run, 1,538 of keys 0–65535 collided with another key; LongSerializer keys fall back the same way.

This parser is shared, so it also rewrites `_kafka_key` for existing **append-mode** streams (trimmed/re-marshalled where master stored the raw bytes), while the PRD says append behaviour stays unchanged. Suggest canonicalizing only valid JSON-object keys and leaving everything else byte-exact (or base64 for non-UTF-8).

drivers/kafka/internal/upsert.go
```diff
@@ -0,0 +1,107 @@
...
+	Enabled               bool
+	DedupKeys             []string
+	AllowTombstoneDeletes bool
+}
+
+func NewUpsertConfig(meta types.StreamMetadata, dedupKeys []string) (UpsertConfig, error) {
+	// case 1 - dedup key is only _kafka_key -- upsert + tombstone deletes on
+	// case 2 - dedup key selection [_kafka_key(choice) + columns] -- upsert only (no tombstones)
+	cfg := UpsertConfig{
+		Enabled:               !meta.AppendMode,
```
mihir-datazip: **This breaks existing Kafka jobs.** Upsert is enabled by `!meta.AppendMode`, and `append_mode` is `omitempty` (absent ⇒ false). The following synced fine on master as append, but now fail the **entire sync** in PreCDC with `upsert mode: dedup key is required`:
- catalogs written from `drivers/kafka/README.md:160-185`, which documents a non-existent key `append_only: true`
- catalogs with no `selected_streams` (`classifyStreams` assigns zero-value metadata)
- streams previously toggled to `append_mode=false`

The PRD says Append is the default and existing behaviour is unchanged. Suggest an explicit opt-in (e.g. `ingestion_mode: upsert`, or `len(DedupKeys) > 0`), and skipping/erroring the single stream in `classifyStreams` rather than aborting the whole job. The README needs fixing too.

drivers/kafka/internal/upsert.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func UpsertConfigFrom(meta types.StreamMetadata) (UpsertConfig, error) {
+	return NewUpsertConfig(meta, meta.DedupKeys)
+}
+
+// isKafkaKeyOnlyDedup: checks if user selects only _kafka_key as dedup key
+// need to enable upsert +  delete; anything else is upsert only
+func isKafkaKeyOnlyDedup(dedupKeys []string) bool {
+	return len(dedupKeys) == 1 && dedupKeys[0] == Key
```
mihir-datazip: **Question for product (PRD conflict):** tombstones delete only when `dedup_keys` is exactly `["_kafka_key"]`. With a column key like the PRD's own `customer_id`, or any composite that includes `_kafka_key`, a tombstone never deletes, even when the key can be read from the tombstone's JSON key (`record.KeyFields`).

PRD §7 says "the configured Deduplication Key is used to identify the record that needs to be deleted", but the edge-case table says "key exists but value is null → append". The two statements conflict, so we should settle on one before merging. At minimum, deriving the dedup key from `KeyFields` on tombstones seems to be what §7 intends.

drivers/kafka/internal/upsert.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (c UpsertConfig) Validate() error {
```
mihir-datazip: `Validate` only rejects empty names. It also accepts:
- `_kafka_partition` → one live row per partition, silently
- `_kafka_offset` → merges same-offset messages across partitions
- names missing from the schema → every record goes through the append fallback, with a WARN per record and no dedup
- duplicates, e.g. `["_kafka_key","_kafka_key"]`, which silently turns off tombstone deletes

Worth rejecting the `_kafka_partition`/`_kafka_offset`/`_kafka_timestamp` system columns and de-duplicating the list here, and validating the keys against the stream schema (depends on what the UI dropdown offers).

drivers/kafka/internal/cdc.go
```diff
@@ -116,14 +130,105 @@ func (k *Kafka) StreamChanges(ctx context.Context, readerID int, metadataStates
...
+					}
+					// null, null - avoid
+				}
+			} else if record.Data != nil {
+				// Non null value: Upsert when atleast one selected dedup key exist, all absent - append, all null - fail sync
+				data, err := cfg.checkDedupKeysExist(record.Data, kafkaKey, record.KeyFields)
+				if err != nil {
+					// all present dedup fields null - fail sync
+					if errors.Is(err, errNullDedupKeys) {
+						return false, fmt.Errorf("stream[%s] offset=%d: %w", stream.ID(), record.Message.Offset, err)
```
mihir-datazip: The all-null dedup-key failure is permanent, but it isn't wrapped as non-retryable (`ErrNonRetryable` / `errs.Precondition`). `RetryOnBackoff` reruns the reader 4× with 60s/120s/240s backoff, re-reading and re-writing everything up to the bad offset each time, so the sync fails only after 7+ minutes. The message also omits the partition, so `offset=1000` is ambiguous; please include `partition=%d`.

types/catalog.go
```diff
@@ -337,6 +339,7 @@ func GetStreamsDelta(oldStreams, newStreams *Catalog) *Catalog {
 					(oldMetadata.UseSourceColumnNames != newMetadata.UseSourceColumnNames) ||
 					!reflect.DeepEqual(oldMetadata.FilterConfig, newMetadata.FilterConfig) ||
 					(oldMetadata.AppendMode != newMetadata.AppendMode) ||
+					!reflect.DeepEqual(oldMetadata.DedupKeys, newMetadata.DedupKeys) ||
```
mihir-datazip: `reflect.DeepEqual` treats `nil` vs `[]`, and a pure reorder (which gives identical olake_ids, since `GetKeysHash` sorts keys), as a change. It also compares DedupKeys when `append_mode` is on, where they're irrelevant. A re-save of an existing append stream can therefore land it in `difference_streams.json` and clear the Iceberg table. `ClearState` keeps the Kafka consumer group, so the dropped rows are never re-read.

Suggest comparing sorted, non-empty sets, only when upsert is enabled, and adding a test for `GetStreamsDelta`, which has none.

tests/testutils/test_utils.go
```diff
@@ -692,7 +693,12 @@ func GetBackfillStreamsFromCDC(cdcStreams []string) []string {
 func (cfg *IntegrationTest) resetTable(ctx context.Context, t *testing.T) error {
 	cfg.ExecuteQuery(ctx, t, cfg.TestConfig, "drop")
 	cfg.ExecuteQuery(ctx, t, cfg.TestConfig, "create")
-	cfg.ExecuteQuery(ctx, t, cfg.TestConfig, "add")
+	cfg.ExecuteQuery(ctx, t, cfg.TestConfig, Ternary(len(cfg.DedupKeys) > 0, "upsert_add", "add").(string))
+	if len(cfg.DedupKeys) > 0 {
```
mihir-datazip: For upsert this resets only `state.json`. The Pebble index under `/testdata/olake-table-index` survives from the Legacy sub-test into the Arrow sub-test (same stream ID + destination DB ⇒ same `indexDir`). The dropped table has no snapshot, so `fillTableIndex` doesn't truncate, the stale `{"key":"upsert-key"}` entry is hit, the op stays `u`, and `VerifyIcebergSync(opSymbol "c")` finds 0 rows.

`prepareTableIndexSync` already `RemoveAll`s this directory for the Postgres index tests; the same is needed here. (Traced from the code, since CI never reached this test; see the review summary.)

tests/kafka/kafka_test.go
```diff
@@ -108,3 +108,11 @@ func TestKafkaRebalance(t *testing.T) {
 	t.Parallel()
 	kafkaJSONBaseConfig(t).TestRebalance(t)
 }
+
+func TestKafkaUpsert(t *testing.T) {
+	t.Parallel()
+	cfg := kafkaJSONBaseConfig(t)
+	cfg.DedupKeys = []string{"_kafka_key"}
```
mihir-datazip: **Test coverage:** upsert currently runs as one config (`_kafka_key`, JSON, normalization on, one filter, one partition regex, one message per sync). Please add cases for the stream-config combinations, since several of the bugs above only appear there:
- **Filter**: a tombstone passing through a filter, and an update that moves a row out of the filter (see cdc.go:189)
- **Partition regex**: an update that changes partition, and a tombstone delete on a partitioned table, asserting the old row is gone (see arrow-writer L228)
- **Normalization off**: `updateSelectedStreams` currently forces it on
- **Selected columns**: deselecting the dedup column / `_kafka_key`
- **Column dedup key** (`customer_id`), **multiple dedup keys**, and a **key read from a JSON Kafka key** (`KeyFields`)
- **Avro / Schema Registry** keys: `ExecuteQueryAvro` has no `upsert_*` ops
- **Append and upsert streams in the same sync**, including one invalid upsert stream not failing its append sibling
- **Several messages for one key in one sync**, and one key across partitions/readers
- **Crash recovery** with upsert on (`TestKafka2PC` is append-only)
- **PRD edge cases end-to-end**: missing key → insert, null key / all-null composite → sync fails, one key null → no failure, keyless messages, dedup key changed, dedup key type changed

Also, `TestKafkaUpsert` doesn't match `-run 'Sync'`, so `make test.sync` / `make test.sync.kafka` skip it; CI's `test.integration.kafka` does run it.

drivers/kafka/internal/upsert_test.go
```diff
@@ -0,0 +1,328 @@
...
+			},
+			data: map[string]any{
+				"id":   nil,
+				"name": "sam",
+				"age":  20,
+			},
+			wantErr: errNullDedupKeys,
+		},
+		{
+			name: "dedup key = kafka key(nil) -- fails",
```
mihir-datazip: Production never builds `{_kafka_key: nil}`: a null key becomes `""` (cdc.go:394). This case therefore passes while the real keyless path collapses rows (see upsert.go:85). The null-key / all-null → fail cases also only test classification; the `StreamChanges` error return (cdc.go:214-217) never runs in any test.

drivers/kafka/internal/upsert_test.go
```diff
@@ -0,0 +1,328 @@
...
+				Enabled:   true,
+				DedupKeys: []string{Key},
+			},
+			data: map[string]any{
+				Key: nil,
+			},
+			wantErr: errNullDedupKeys,
+		},
+		{
+			name: "selected dedup field value is empty string - works",
```
mihir-datazip: This asserts that an empty-string key is valid. For `_kafka_key` that's exactly the keyless-message case that should fall back to append (PRD: no dedupe key → insert).

drivers/kafka/internal/upsert_test.go
```diff
@@ -0,0 +1,328 @@
...
+				"id":  "101",
+				"age": 20,
+			},
+			check: func(t *testing.T, data map[string]any) {
+				_, ok := data["name"]
+				assert.False(t, ok, "absent name must stay absent")
+			},
+		},
+		{
+			name: "dedupe fields are partial null",
```
mihir-datazip: This only asserts that no error is returned. Please also assert the resulting `_olake_id`, and that `{id:42,name:null}` and `{id:42}` map to the same entity (they currently don't; see upsert.go:106).

drivers/kafka/internal/upsert_test.go
```diff
@@ -0,0 +1,328 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.wantModeOnlyKafkaKey, isKafkaKeyOnlyDedup(tt.dedupKeys))
+		})
+	}
+}
+
+func TestNewUpsertConfig(t *testing.T) {
```
mihir-datazip: Please add backward-compat cases: `append_mode` absent/false with no `dedup_keys` should keep append behaviour (or at least not fail the whole sync); see upsert.go:24.

tests/testutils/test_utils.go
```diff
@@ -948,6 +954,32 @@ func (cfg *IntegrationTest) testIcebergFullLoadAndCDC(
...
+			{
+				name:      "CDC - strict - insert",
+				operation: "",
+				useState:  false,
+				opSymbol:  "c",
+				expected:  cfg.ExpectedData,
+			},
+			{
+				name:      "CDC - strict - update",
+				operation: "upsert_update",
```
mihir-datazip: Naming this `upsert_update` bypasses `evolvedSchema := operation == "update"`, so the evolved-schema DESCRIBE check never runs for the upsert suite.

PR #1162, opened 2026-08-26
feat(iceberg): add v3 deletion vector support

### LINKED ISSUE #1168: implement dv writing in Olake (Difficulty: Medium)
https://github.com/datazip-inc/olake/blob/2967a8e86b2af2b81a00a567d1545666ca309ab4/types/delete_mode.go#L18-L23

---

###### This issue was generated by [todo-issue](https://github.com/Juulsn/todo-issue) based on a `TODO` comment in 2967a8e86b2af2b81a00a567d1545666ca309ab4. It's been assigned to @hash-data because they committed the code.

vikaxsh:
# Description

Adds `dv` as a third per-stream delete mode alongside `eq` and `pos`, writing Iceberg v3 Puffin deletion vectors instead of Parquet positional deletes.

Introduces `PositionalDeleteSink` so `PositionalDeltaWriter` can target either representation, plus `DeletionVectorConverter` and `PreviousDeleteLoader` to merge new deletes with existing Parquet positional deletes or deletion vectors without resurrecting previously deleted rows.

Updates `EqualityDeleteMigrator.migrate()` to accept a `targetMode`, allowing tables to switch from equality deletes directly to `pos` or `dv` encoding in a single atomic `RewriteFiles` commit.

Fixes #1168 

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):


hash-data:
AI review
```

P0

1. CI workflow hacked, comment says so. .github/workflows/integration-tests.yml — feat/iceberg-deletion-vectors and feat/interoperability added to both push and pull_request triggers, and the github.event_name == 'pull_request' gate dropped under a TEMP(hack) ... REVERT BEFORE MERGE comment. Merges as-is, 16gb-runner ITs fire on every push to those branches.

2. Proto field numbers reused, no reserved. records_ingest.proto, IcebergPayload.Metadata:

┌─────┬─────────────────────────────────┬──────────────────────────────────────────┐
│ tag │               was               │                   now                    │
├─────┼─────────────────────────────────┼──────────────────────────────────────────┤
│ 9   │ bool use_positional_deletes     │ repeated PartitionField partition_fields │
├─────┼─────────────────────────────────┼──────────────────────────────────────────┤
│ 10  │ repeated PartitionField         │ optional int64 base_snapshot_id          │
├─────┼─────────────────────────────────┼──────────────────────────────────────────┤
│ 11  │ optional int64 base_snapshot_id │ DeleteMode delete_mode                   │
└─────┴─────────────────────────────────┴──────────────────────────────────────────┘

sink_jar_path is a user config knob (destination/iceberg/config.go:69), so Go-binary/jar skew is a reachable deployment shape, not hypothetical. Tags 9 and 10 change wire type and fail loudly; tag 11 stays varint — an old client's base_snapshot_id parses as a garbage delete_mode, silently. Fix costs nothing: keep 9/10/11 where they were, reserved 9;, put delete_mode/declare_identifier_fields at 12/13. If you decide lockstep shipping makes skew impossible, say so in the PR — but decide it explicitly.

P1

3. validateNoConflictingDeleteFiles() is unconditional and mostly redundant. IcebergTableOperator.java:242. Iceberg's BaseRowDelta.validate() already calls validateAddedDVs(base, startingSnapshotId, filter, parent) unconditionally (1.10.2, line 168) — that is the DV-staleness guard the comment describes. What the added call buys on top is validateNoNewDeleteFiles with the default conflictDetectionFilter = alwaysTrue(): any delete file added anywhere in the table since baseSnapshotId now fails the commit. It sits inside baseSnapshotId != null, so it hits pos mode too, which never needed it — and the code 60 lines below (staged.parentId() != baseSnapshotId → return 0L) exists precisely because concurrent commits are expected and were previously tolerated. Gate on deleteMode == DeleteMode.DELETION_VECTOR.

4. pos→dv leaves Parquet positional deletes in a v3 table. ensureFormatVersion upgrades unconditionally, nothing rewrites existing pos-delete files. PreviousDeleteLoader only merges deletes for data files this commit touches, and BaseDVFileWriter.close() only retires previous files where ContentFileUtil.isFileScoped holds. Untouched data files keep Parquet deletes forever. Iceberg-java still reads them; v3 spec says a table must not carry position delete files, and other engines may reject. eq→dv gets a migration RPC, pos→dv gets nothing — and iceberg_dv.go tests eq→dv but not pos→dv.

5. Migrator writes Puffin blobs named .parquet. OlakeTableIndexer passes session.fileFactory into EqualityDeleteMigrator.migrate(..., DELETION_VECTOR) → BaseDVFileWriter. That factory is built from getTableFileFormat(icebergTable) = parquet (IcebergSession.java:34-35). This is exactly what ArrowDeletionVectorWriter:53-56 and IcebergTableWriterFactory.deleteSink both build a separate PUFFIN factory to avoid. Metadata carries FileFormat.PUFFIN so reads work; anything inferring format from the extension does not. Build the PUFFIN factory inside writeDeletionVectors.

6. Wrong partition on the vector — consequence understated. TODO'd in ArrowDeletionVectorWriter.add and PositionalDeltaWriter.write; both say the delete still applies because DeleteFileIndex matches by path. True after planning — but scan planning prunes delete manifests by partition first, so a partition-filtered read of the moved-from partition can miss the vector and resurrect the row. The ITs deliberately avoid it (dv-part-update-id1 updates amount, not the partition column). Shipping TODO'd is defensible; "partitioned table + partition-changing update" needs to reach user-facing docs, not just a code comment.

P2

7. PreviousDeleteLoader.planDeletes runs table.newScan().planFiles() — full table plan, once per writer lifecycle, i.e. per commit. Pos mode pays nothing equivalent. Consider planning from delete manifests only.

8. Append-mode stream with update_type: dv still sends DELETE_MODE_DELETION_VECTOR — iceberg.go Setup now sends stream.GetUpdateType() unconditionally where it previously sent options.TableIndex != nil, and destination/writers.go:129 leaves TableIndex nil in append mode. The destination table gets irreversibly upgraded to v3 for a stream that will never write a delete, and the server builds position maps the client has no index to store.

9. ensureFormatVersion is a silent one-way table-wide upgrade driven by a per-stream setting, logged only as a Java-side WARN. Worth surfacing to the Go side.

result: Reviewed PR #1162 (iceberg DV v3) — 2 P0 (CI REVERT BEFORE MERGE hack left in; proto tags 9/10/11 reused without reserved, tag 11 misparses silently), 4 P1 (redundant validateNoConflictingDeleteFiles regresses pos-mode concurrency; pos→dv leaves Parquet pos-deletes in a v3 table, untested; migrator writes Puffin as .parquet; partition mis-stamp can resurrect rows on partition-filtered reads), 3 P2; both builds pass.
```



w.r.t. v3 spec 

```

New P1: v3 row lineage is mandatory and this breaks it

No opt-out — row lineage is required for all v3 tables, no flag, no property. Spec, on a row moving to a different data file: "The row's existing non-null _row_id must be copied into the new data file."

Olake's upsert is "DV the old position, append a fresh row." The new row carries no _row_id, so it inherits a brand-new one from first_row_id. Every UPDATE mints a new row identity. Any consumer using row tracking — Databricks CDF-on-Iceberg, Dremio, incremental reads — sees delete+insert instead of update, on every table this PR upgrades.

Not cheaply fixable: carrying _row_id forward means the table index has to store it alongside RowLocation. Minimum bar is a documented limitation; it's a v3-only semantic that eq/pos at v2 never exposed.

Mechanics I checked and they're fine: row-id allocation advances by sum(recordCount) of added data files, and olake derives record count from ParquetUtil.fileMetrics on the real footer (IcebergTableOperator.java:483), so next-row-id accounting is correct. nextRowId is a primitive defaulting to INITIAL_ROW_ID = 0, and pre-upgrade files keep a null first_row_id — the v2→v3 upgrade neither NPEs nor corrupts.

New P1: the silent upgrade locks out most query engines, irreversibly

ensureFormatVersion flips a shared table to v3 from a per-stream setting, with a Java-side WARN as the only signal. Engine reality as of 2026:

┌──────────────────────────────────────────┬───────────────────────────────────────────────┐
│                  Engine                  │                 v3 / DV reads                 │
├──────────────────────────────────────────┼───────────────────────────────────────────────┤
│ Spark 4.0 + Iceberg 1.10.x               │ full                                          │
├──────────────────────────────────────────┼───────────────────────────────────────────────┤
│ EMR 7.12, Glue, S3 Tables, Databricks UC │ GA                                            │
├──────────────────────────────────────────┼───────────────────────────────────────────────┤
│ Athena                                   │ hard fail — Cannot read unsupported version 3 │
├──────────────────────────────────────────┼───────────────────────────────────────────────┤
│ Trino                                    │ experimental, no row-level DML                │
├──────────────────────────────────────────┼───────────────────────────────────────────────┤
│ ClickHouse                               │ DV reads still an open PR                     │
├──────────────────────────────────────────┼───────────────────────────────────────────────┤
│ Snowflake managed                        │ not public                                    │
└──────────────────────────────────────────┴───────────────────────────────────────────────┘

A user flipping update_type: dv on an existing table can lock their BI stack out of it, and format version only moves forward. This needs explicit opt-in — create-at-v3 is fine, upgrading someone's existing table shouldn't happen without them asking.

Softened: pos→dv leftovers (was P1 #4, now P2)

The spec is more forgiving than I assumed — files written before the upgrade stay valid; the obligation is to absorb lingering position deletes when that data file's deletes are next updated. This implementation does exactly that, and I traced the whole chain: BaseDeleteLoader.loadPositionDeletes tags the returned index with its source files, BitmapPositionDeleteIndex.merge preserves that list, BaseDVFileWriter.close() retires the file-scoped ones — and olake's granularity is file-scoped. Compliant for touched files, spec-permitted for untouched ones. Residual: orphan Parquet deletes linger on cold data files forever, and there's still no pos→dv IT.

Confirmed: rejecting eq on v3 is stricter than Iceberg

MergingSnapshotProducer.validateDeleteFileForVersion case 3 is content() == EQUALITY_DELETES || isDV(file) — equality deletes are legal in v3. The rejection in validateOrUpgradeFormatVersion is an olake writer limitation (BaseEqualityDeltaWriter emits a pos-delete for in-batch dedup), which the code comment states accurately. Worth surfacing as an operational consequence: once any stream upgrades a table to v3, every eq stream targeting that table fails permanently and the only recovery is dropping the table.

Also verified clean

- .parquet-named Puffin from the migrator still commits — ContentFileUtil.isDV tests format() == PUFFIN, not the extension. Stays a naming/ops issue (P2), not a commit failure.
- PreviousDeleteLoader's explicit FileContent.POSITION_DELETES filter is load-bearing, not defensive padding: a mid-migration table with eq deletes plus a DV on one data file would otherwise hand a Puffin blob to BaseDeleteLoader.openDeletes, whose format switch has no PUFFIN case.
- findDV asserts dv.dataSequenceNumber() >= data file seq — satisfied, both land in one commit.
- Old-DV retirement works end to end.

result: Deep v3 research on PR #1162 — partition mis-stamp escalated to P0 (verified DeleteFileIndex prunes delete manifests by partition before path matching, so partition-filtered reads resurrect superseded rows); two new P1s (mandatory v3 row lineage broken by delete+append upserts; silent one-way v3 upgrade locks out Athena/Trino/ClickHouse/Snowflake); pos→dv leftovers softened to P2 as spec-permitted.
```

destination/iceberg/index.go
```diff
@@ -25,16 +25,17 @@ func (i *Iceberg) reconcileTableIndex(ctx context.Context, index types.StreamInd
 	table := fmt.Sprintf("%s.%s", i.stream.GetDestinationDatabase(&i.config.IcebergDatabase), i.stream.GetDestinationTable())
 
 	if hasEqualityDeletes {
-		// first check for equality deletes and migrate them to positional deletes
+		// Rewrite straight into the stream's target encoding, one commit, not eq -> pos -> dv.
 		migrated, err := i.server.tableIndexClient.MigrateEqualityDeletes(ctx, &proto.MigrateEqualityDeletesRequest{
-			ThreadId: i.options.ThreadID,
+			ThreadId:   i.options.ThreadID,
+			TargetMode: protoDeleteMode(i.stream.GetUpdateType()),
```
hash-data, line 31: we are not migrating as of now right?

destination/iceberg/iceberg.go
```diff
@@ -196,7 +200,8 @@ func (i *Iceberg) Check(ctx context.Context) error {
 	ctx, cancel := context.WithTimeout(ctx, 300*time.Second)
 	defer cancel()
 
-	identifierField := utils.Ternary(i.config.NoIdentifierFields, "", constants.OlakeID).(string)
+	// Row-identity column, sent even where the catalog forbids declaring it.
+	identifierField := constants.OlakeID
```
hash-data, line 204: we should not remove no identifier flag as it will create issue in backward compatibility

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeRowsIngester.java
```diff
@@ -245,18 +250,64 @@ private void sendTableResponse(StreamObserver<RecordIngest.RecordIngestResponse>
...
+        }
+    }
+
+    /**
+     * Checks the requested delete mode against the table's actual format version
+     * before any writer is built, rather than letting Iceberg reject it at commit
+     * time after a whole sync's worth of work.
+     */
+    private void validateOrUpgradeFormatVersion(Table table, DeleteMode deleteMode) {
+        int current = ((org.apache.iceberg.HasTableOperations) table).operations().current().formatVersion();
```
hash-data, line 280:
```suggestion
        int curSpecVersion = ((org.apache.iceberg.HasTableOperations) table).operations().current().formatVersion();
```

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/rpc/OlakeRowsIngester.java
```diff
@@ -245,18 +250,64 @@ private void sendTableResponse(StreamObserver<RecordIngest.RecordIngestResponse>
...
+        int required = deleteMode.minimumFormatVersion();
+
+        if (current < required) {
+            // e.g. a stream reconfigured from eq/pos to dv against a table that
+            // already exists at v2. One-way; see IcebergUtil.ensureFormatVersion.
+            IcebergUtil.ensureFormatVersion(table, required);
```
hash-data, line 286: we can write logic here only, instead of creating a function

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableIndex/EqualityDeleteMigrator.java
```diff
@@ -128,25 +144,101 @@ public static Result migrate(Table table, String identifierField, OutputFileFact
...
+  private static WrittenDeletes writeDeletionVectors(Table table, OutputFileFactory fileFactory,
+      Map<String, PartitionGroup> groups) throws IOException {
+
+    PartitionSpec spec = table.spec();
+
+    // ONE writer for the WHOLE migration, not one per partition (contrast with
+    // writePositionDeletes below): the spec allows many vectors inside one Puffin
+    // file with no restriction on which data files they reference, and one writer
+    // per partition would recreate the small-files problem vectors exist to remove.
+    DVFileWriter writer = new BaseDVFileWriter(fileFactory, new PreviousDeleteLoader(table));
```
hash-data, line 222: we decided not to write any migration logic?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -217,23 +236,41 @@ public long commitThread(String threadId, String payload, Table table, Long base
...
+                // it, and a vector cannot be opened as Parquet if we tried.
+                String single = deleteFile.referencedDataFile();
+                if (single != null) {
+                  referencedDataFiles.add(single);
+                  continue;
+                }
+                // PARTITION-granularity Parquet positional delete files can span every
+                // data file in a partition, so referencedDataFile() is null; read the
+                // file to enumerate all of them.
+                try (CloseableIterable<Object> rows = TableIndexScanner.openParquet(table, deleteFile.location(), DeleteSchemaUtil.pathPosSchema())) {
```
hash-data: there is merge positional in iceberg, can we use that?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -33,6 +33,17 @@ type ArrowWriter struct {
 	createdFiles   map[string]*PartitionFiles
 	upsertMode     bool
 	indexThread    *types.StreamIndexThread
+	// deleteMode decides how superseded rows are expressed: under DeletionVector, positions stream to the server as Puffin vectors instead of a delete file.
+	deleteMode types.UpdateType
+	// pendingVectors buffers positions per data file until a batch is worth sending. Only used under DeleteModeDeletionVector.
+	pendingVectors     map[string]*pendingVector
+	pendingVectorCount int
```
hash-data, line 40: this variable seems not required

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -217,23 +236,41 @@ public long commitThread(String threadId, String payload, Table table, Long base
...
             if (deleteFiles != null) {
               for (DeleteFile deleteFile : deleteFiles) {
-                if (deleteFile.content() == FileContent.POSITION_DELETES) {
-                  try (CloseableIterable<Object> rows = TableIndexScanner.openParquet(table, deleteFile.location(), DeleteSchemaUtil.pathPosSchema())) {
-                    for (Object row : rows) {
-                      Object filePath = TableIndexScanner.getFieldValue(row, "file_path");
-                      if (filePath != null) {
-                        referencedDataFiles.add(filePath.toString());
-                      }
+                if (deleteFile.content() != FileContent.POSITION_DELETES) {
```
hash-data, line 251: we need validation for deletion vector as well right? like deletion vector is committed for the file which not exist?

hash-data: have we tested conflicting dv ?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableIndex/PreviousDeleteLoader.java
```diff
@@ -0,0 +1,95 @@
...
+ * files and Puffin deletion vectors transparently - do not hand-roll Puffin parsing,
+ * {@code org.apache.iceberg.DVUtil} is package-private in Iceberg 1.10.2.
+ *
+ * <p>The table is planned once, on first use, and reused for the life of this loader:
+ * one instance is created per writer session, and the plan describes the snapshot the
+ * caller's table index was built against - the same snapshot the commit refuses to move
+ * past. Not thread-safe; callers hold one instance per writing thread, same as every
+ * other writer-scoped object in this package.
+ */
+public final class PreviousDeleteLoader implements Function<String, PositionDeleteIndex> {
```
hash-data, line 39: why we need to create a new class for it?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/DeleteMode.java
```diff
@@ -0,0 +1,57 @@
+package io.debezium.server.iceberg.tableoperator;
+
+import io.debezium.server.iceberg.rpc.RecordIngest.IcebergPayload;
+
+/** How a writer represents the removal of a row that a later version supersedes. */
+public enum DeleteMode {
```
hash-data, line 6: the resolver will be in common part right, it should not depend on arrow right? let us simplify if possible we can discuss

hash-data (commented):
first review part 1

hash-data:
we need to run clear destination once someone change eq -> dv or pos -> dv

destination/iceberg/iceberg.go
vikaxsh: haven't removed identifierField, it is there

destination/iceberg/arrow-writer/writer.go
vikaxsh: it is required for checking delete batch size and flush

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/IcebergTableOperator.java
vikaxsh: POSITION_DELETES covers both dv and pos deletes files, rowDelta.validateDataFilesExist(referencedDataFiles) would handle that

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableIndex/PreviousDeleteLoader.java
vikaxsh: becasue dv rewrites existing deletes files either dv of pos parquet

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/DeleteMode.java
vikaxsh: how is it depending on arrow mode? could you please explain more on this

hash-data:
11. Conflict guard is conditional. validateNoConflictingDeleteFiles() runs only when baseSnapshotId != null (IcebergTableOperator.java:236-245). On the null path a concurrent writer's DV can be silently swallowed by the merge.

.github/workflows/integration-tests.yml
```diff
@@ -240,7 +244,9 @@ jobs:
     # this too -- so gate on "not failed" rather than on success. go-cache is exempt entirely: its
     # cache warm and db2 clidriver download must not block seven drivers, and its lint and gosec
     # failures already surface as their own check.
-    if: ${{ !cancelled() && github.event_name == 'pull_request' && needs.preflight.result == 'success' && needs.preflight.outputs.drivers != '[]' && needs.build-jar.result != 'failure' && needs.apt-warm.result != 'failure' }}
+    # TEMP(hack): dropped `github.event_name == 'pull_request' &&` to run ITs from push while the
+    # PR has merge conflicts. REVERT BEFORE MERGE.
+    if: ${{ !cancelled() && needs.preflight.result == 'success' && needs.preflight.outputs.drivers != '[]' && needs.build-jar.result != 'failure' && needs.apt-warm.result != 'failure' }}
```
hash-data: revert

destination/iceberg/arrow-writer/writer.go
```diff
@@ -148,7 +162,9 @@ func (w *ArrowWriter) getOrCreateWriter(ctx context.Context, pKey string, values
 				return nil, err
 			}
 		}
-		if writer.positionalDeleteWriter == nil {
+		// Deletion vectors are encoded server-side from streamed positions, so this
+		// mode writes no delete file of its own - see sendPendingVectors.
+		if writer.positionalDeleteWriter == nil && w.deleteMode != types.UpdateTypeDeletionVector {
```
hash-data: we can remove this check ? or let us also not create equality writer as well?

destination/iceberg/arrow-writer/writer.go
```diff
@@ -373,6 +393,11 @@ func (w *ArrowWriter) Close(ctx context.Context, finalMetadataState any) (err er
 		return fmt.Errorf("failed to close arrow writers: %w", err)
 	}
 
+	// The trailing partial batch has to reach the server before the commit, which is where the vectors holding it are published.
+	if err := w.sendPendingVectors(ctx); err != nil {
```
hash-data, line 396: do we need to complete writer of this as well when schema evolution happens? as per info it should not but just clarifying.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/rpc/IcebergSession.java
```diff
@@ -13,20 +15,26 @@ public class IcebergSession {
...
 
         FileFormat fileFormat = IcebergUtil.getTableFileFormat(icebergTable);
         this.fileFactory = IcebergUtil.getTableOutputFileFactory(icebergTable, fileFormat);
     }
 
+    /** Whether the table declares identifier fields; read from the table, since a catalog may refuse the declaration. */
     public boolean createIdentifierFields() {
-        return identifierField != null && !identifierField.isEmpty();
+        return !icebergTable.schema().identifierFieldIds().isEmpty();
     }
```
hash-data, line 39: ??

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -217,23 +236,41 @@ public long commitThread(String threadId, String payload, Table table, Long base
...
             if (deleteFiles != null) {
               for (DeleteFile deleteFile : deleteFiles) {
-                if (deleteFile.content() == FileContent.POSITION_DELETES) {
-                  try (CloseableIterable<Object> rows = TableIndexScanner.openParquet(table, deleteFile.location(), DeleteSchemaUtil.pathPosSchema())) {
-                    for (Object row : rows) {
-                      Object filePath = TableIndexScanner.getFieldValue(row, "file_path");
-                      if (filePath != null) {
-                        referencedDataFiles.add(filePath.toString());
-                      }
+                if (deleteFile.content() != FileContent.POSITION_DELETES) {
```
hash-data: in arrow writer as well?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/DeleteMode.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Maps the wire enum onto this one. UNSPECIFIED is rejected rather than defaulted:
+   * it means the sender did not set delete_mode, and guessing on its behalf would
+   * silently write a representation the caller never asked for - and, for a table
+   * created under that guess, at the wrong format version. Every caller sets it,
+   * including the destination check, which sends EQUALITY for its throwaway table.
+   * UNRECOGNIZED means the sender knows a mode this build does not.
+   */
+  public static DeleteMode resolve(IcebergPayload.DeleteMode deleteMode) {
```
hash-data, line 42: let us do clear destination when user changes back from dv to anything

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableIndex/EqualityDeleteMigrator.java
```diff
@@ -128,25 +144,101 @@ public static Result migrate(Table table, String identifierField, OutputFileFact
...
+
+    PartitionSpec spec = table.spec();
+
+    // ONE writer for the WHOLE migration, not one per partition (contrast with
+    // writePositionDeletes below): the spec allows many vectors inside one Puffin
+    // file with no restriction on which data files they reference, and one writer
+    // per partition would recreate the small-files problem vectors exist to remove.
+    DVFileWriter writer = new BaseDVFileWriter(fileFactory, new PreviousDeleteLoader(table));
+    try {
+      for (PartitionGroup group : groups.values()) {
```
hash-data, line 234: cant we use same writer for arrow ?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableIndex/EqualityDeleteMigrator.java
```diff
@@ -128,25 +144,101 @@ public static Result migrate(Table table, String identifierField, OutputFileFact
...
+        for (RowPosition row : group.positions) {
+          writer.delete(row.path, row.position, spec, partition);
+        }
+      }
+    } finally {
+      writer.close();
+    }
+
+    DeleteWriteResult result = writer.result();
+    return new WrittenDeletes(result.deleteFiles(), result.rewrittenDeleteFiles());
```
hash-data, line 249: we are sure that all older dv also merged? as well as pos got removed?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableIndex/PreviousDeleteLoader.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /**
+   * ONE manifest scan. Iceberg's planner already works out which delete files apply to
+   * which data files (partitions, sequence numbers taken into account); this records
+   * the pairing it reports rather than re-deriving it.
+   */
+  static Map<String, List<DeleteFile>> planDeletes(Table table) {
+    Map<String, List<DeleteFile>> byPath = Maps.newHashMap();
+    try (CloseableIterable<FileScanTask> tasks = table.newScan().planFiles()) {
```
hash-data, line 80: this is heavy operation and we already do it eq migrator, see if we can have some different logic here

destination/iceberg/arrow-writer/writer.go
vikaxsh: for eq there is already w.indexThread == nil check, it won't create eq writer

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/debezium/server/iceberg/tableoperator/DeleteMode.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Maps the wire enum onto this one. UNSPECIFIED is rejected rather than defaulted:
+   * it means the sender did not set delete_mode, and guessing on its behalf would
+   * silently write a representation the caller never asked for - and, for a table
+   * created under that guess, at the wrong format version. Every caller sets it,
+   * including the destination check, which sends EQUALITY for its throwaway table.
+   * UNRECOGNIZED means the sender knows a mode this build does not.
+   */
+  public static DeleteMode resolve(IcebergPayload.DeleteMode deleteMode) {
```
vikaxsh: old code uses to check did go send identifierField, now it check from table itself, as go always sends

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/IcebergTableOperator.java
vikaxsh: yeah, Iceberg has only 3 content types (DATA, POSITION_DELETES, EQUALITY_DELETES)

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/rpc/IcebergSession.java
vikaxsh: old code uses to check did go send identifierField, now it check from table itself, as go always sends

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/IcebergTableOperator.java
vikaxsh: added a comment

destination/iceberg/arrow-writer/writer.go
vikaxsh: queueVectorDeletes runs once per batch, and each call adds len(deletes). The send happens only once the total reaches deletionVectorBatchSize, which can take many calls. local would reset to 0 on every call, so it would only ever see one batch's count.

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/IcebergTableOperator.java
```diff
@@ -257,6 +295,12 @@ public long commitThread(String threadId, String payload, Table table, Long base
             deleteFiles.forEach(rowDelta::addDeletes);
           }
         }
+
+        // Deletion-vector replace semantics: a superseded vector must leave the table
+        // in the SAME commit the new one arrives in, or the table ends up with two
+        // vectors for one data file. Always empty outside DELETION_VECTOR mode.
+        rewrittenDeleteFiles.forEach(rowDelta::removeDeletes);
```
hash-data, line 302: did not get this

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/IcebergTableWriterFactory.java
```diff
@@ -99,4 +104,20 @@ private BaseTaskWriter<Record> deltaWriter(Table icebergTable, FileFormat format
...
+  private PositionalDeleteSink deleteSink(Table icebergTable, FileFormat format,
+      GenericAppenderFactory appenderFactory, OutputFileFactory fileFactory) {
+    if (deleteMode == DeleteMode.DELETION_VECTOR) {
+      // Deletion vectors are Puffin, not the table's data format. OutputFileFactory
+      // takes its file extension from the format it was built with, so reusing
+      // fileFactory here would name Puffin blobs ".parquet".
+      OutputFileFactory dvFileFactory = IcebergUtil.getTableOutputFileFactory(icebergTable, FileFormat.PUFFIN);
+      // A vector replaces the data file's previous one, so it has to be seeded with
+      // the positions already deleted or this commit would resurrect them.
+      return new PositionalDeleteSink.DeletionVectors(
```
hash-data, line 118: do we need to have previous loader in delta writer as well?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/IcebergTableWriterFactory.java
```diff
@@ -78,12 +81,14 @@ private BaseTaskWriter<Record> appendWriter(Table icebergTable, FileFormat forma
 
   private BaseTaskWriter<Record> deltaWriter(Table icebergTable, FileFormat format, GenericAppenderFactory appenderFactory, OutputFileFactory fileFactory, long targetFileSize) {
 
-    if (usePositionalDeletes) {
+    if (deleteMode.addressesPositions()) {
       // One writer for both layouts: an unpartitioned table is a single entry keyed
       // on the empty partition struct, so there is no partitioned/unpartitioned split.
+      // pos vs dv is entirely the sink's concern from here - the writer never branches.
       return new PositionalDeltaWriter(icebergTable.spec(), format, appenderFactory, fileFactory,
```
hash-data, line 89: let us separate the names, call dv writer explicit

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/PositionalDeleteSink.java
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * Writes positional delete files, one per referenced data file.
+   *
+   * <p>Iceberg wants a delete file's positions sorted by path then offset, which CDC
+   * order does not give us, so each partition gets a writer that buffers and sorts on
+   * close. Keeping one delete file per referenced data file is what lets Iceberg treat
+   * them as file-scoped and match them to data files by path rather than by partition.
+   */
+  final class PositionalFiles implements PositionalDeleteSink {
```
hash-data, line 75: this was not there in older version ?

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/IcebergTableOperator.java
vikaxsh: it is for removing old dv in same commit

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/IcebergTableWriterFactory.java
vikaxsh: iceberg api BaseDVFileWriter make's it mandatory

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/IcebergTableWriterFactory.java
vikaxsh: in deleteSink, calling explicit only

destination/iceberg/olake-iceberg-java-writer/src/main/java/io/olake/iceberg/tableoperator/PositionalDeleteSink.java
vikaxsh: let's discuss

tests/mysql/mysql_util_test.go
```diff
@@ -308,6 +308,77 @@ func ExecuteQuery(ctx context.Context, t *testing.T, conf *testutils.TestConfig,
 	case "evolve-schema":
 		query = fmt.Sprintf("ALTER TABLE %s MODIFY COLUMN id_int BIGINT, MODIFY COLUMN price_float DOUBLE, ADD COLUMN includedColumn INT;", integrationTestTable)
 
+	case "dv-create":
```
hash-data, line 311: do we need to add these for other drivers?

types/catalog.go
```diff
@@ -325,11 +325,16 @@ func GetStreamsDelta(oldStreams, newStreams *Catalog) *Catalog {
...
+			// NOTE: delete mode changes keep the table, except dv -> other and pos -> dv (see dvDelta)
 			// TODO: log the differences for user reference
 			isDifferent := func() bool {
 				// check cursor field if SyncMode is incremental
 				cursorDelta := utils.Ternary(newStream.Stream.SyncMode == INCREMENTAL, oldStream.Stream.CursorField != newStream.Stream.CursorField, false).(bool)
+				// leaving dv: v3 forbids the Parquet positional deletes eq/pos write; pos -> dv: not supported yet (only eq -> dv is migrated)
+				oldUpdateType := utils.Ternary(oldMetadata.UpdateType == "", UpdateTypeEquality, UpdateType(oldMetadata.UpdateType)).(UpdateType)
+				newUpdateType := utils.Ternary(newMetadata.UpdateType == "", UpdateTypeEquality, UpdateType(newMetadata.UpdateType)).(UpdateType)
+				dvDelta := (oldUpdateType == UpdateTypeDeletionVector && newUpdateType != UpdateTypeDeletionVector) ||
+					(oldUpdateType == UpdateTypePosition && newUpdateType == UpdateTypeDeletionVector)
```
hash-data, line 374: what about case of pos -> dv

PR #1179, opened 2026-09-02
fix(iceberg): preserve OLake metadata columns

### LINKED ISSUE #1177: Iceberg destination: selected_columns excludes OLake metadata columns from CREATE-time schema, breaking _olake_id identifier field
## Environment

- Olake version: v0.9.5 (also present on current `master` as of this report)
- Destination: Iceberg, REST catalog (AWS S3 Tables)
- Source: Postgres, CDC (pgoutput)

## Description

When a stream's `selected_columns.columns` is a non-empty allowlist, the Iceberg destination excludes OLake's own metadata columns (`_olake_id`, `_olake_timestamp`, `_op_type`, `_cdc_timestamp`) from the schema sent to the writer at `GET_OR_CREATE_TABLE` time, because they are never listed in the user's column allowlist. The table therefore gets created (or evolved) without `_olake_id`, with no identifier field at all.

Once real records are written, `_olake_id` (and the other three metadata columns) get merged into the row data regardless of `selected_columns` (`destination/iceberg/iceberg.go`, `FlattenAndCleanData`: `maps.Copy(records[idx].Data, record.OlakeColumns)`), so `_olake_id` shows up as a "new column" against the table's actual schema. Schema evolution then adds it as a new field on an existing table, and Iceberg requires new fields on an existing table to be optional. Right after that, OLake calls `UpdateSchema.setIdentifierFields(["_olake_id"])`, which throws:

```
java.lang.IllegalArgumentException: Cannot add field _olake_id as an identifier field: not a required field
	at org.apache.iceberg.Schema.validateIdentifierField(Schema.java:168)
```

This repeats on every retry (with backoff) because `GET_OR_CREATE_TABLE` loads the already-created, already-broken table rather than fixing it.

## Root cause (read from source, not just observed)

`types/type_schema.go`, `TypeSchema.ToIceberg()`:

```go
    prop := value.(*Property)
    colName := key.(string)
    outName := stream.ResolveColumnName(colName)
    if !isSelected(outName) || (defaultColumns && !prop.OlakeColumn && !includeSet.Exists(colName)) {
        return true
    }
    ...
})
```

`isSelected` comes from `ConfiguredStream.IsSelectedColumn()` (`types/stream_configured.go`), which applies the configured allowlist to every property, including ones with `OlakeColumn: true`. Discover registers the four metadata columns as properties with `OlakeColumn: true` for every stream (`drivers/abstract/abstract.go`, `DefaultColumns` plus `convStream.UpsertField(column, typ, true, true)`), so they end up subject to the same allowlist as real business columns.

Compare with the sibling function for the Parquet destination, `TypeSchema.ToParquet()`, one function above it in the same file. It hardcodes the four metadata columns into `groupNode` before the `Properties.Range` loop runs, so they never go through `isSelected` at all:

```go
groupNode := parquet.Group{
    constants.OlakeID:        parquet.String(),
    constants.OlakeTimestamp: parquet.Timestamp(parquet.Microsecond),
    constants.OpType:         parquet.String(),
    constants.CdcTimestamp:   parquet.Optional(parquet.Timestamp(parquet.Microsecond)),
}
```

`ToIceberg()` has no equivalent carve-out. The original column-selection feature spec (#121) actually calls for one: its "Edge Cases Handling" section says "All CDC operation columns will be mandatorily selected and disabled from editing." So the intent was for these columns to always pass through, but that never got implemented on the Iceberg code path.

## Steps to reproduce

1. Configure a Postgres CDC stream with `selected_columns.columns` set to a non-empty list of real business columns (any `sync_new_columns` value reproduces it -- both branches of `IsSelectedColumn` filter properties not in the list).
2. Point it at an Iceberg REST destination (S3 Tables in our case).
3. Run `sync`.

Example `selected_columns` block that reproduces it:

```json
"selected_columns": {
  "columns": ["id", "username", "email", "first_name", "last_name", "is_staff", "is_active", "is_superuser", "last_login", "date_joined"],
  "sync_new_columns": true
}
```

## Expected behavior

The four OLake metadata columns should always be included in the Iceberg schema regardless of `selected_columns`, the same way `ToParquet()` already does it for the Parquet destination. The resulting table should have `_olake_id` present and required from the very first `CREATE TABLE`.

## Actual behavior

- The table is created without `_olake_id` and without any identifier field (all columns nullable).
- The very first batch write triggers a schema-evolution add of `_olake_id` as an optional column, which then fails `setIdentifierFields` with the error above.
- The stream is permanently stuck retrying with backoff; zero rows ever land, because `GET_OR_CREATE_TABLE` keeps loading the same broken table rather than recreating it.

## Suggested fix

In `TypeSchema.ToIceberg()`, treat `prop.OlakeColumn` fields the same way `ToParquet()` does: bypass `isSelected()` for them (only enforce the allowlist on non-`OlakeColumn` properties), e.g.:

```go
if (!prop.OlakeColumn && !isSelected(outName)) || (defaultColumns && !prop.OlakeColumn && !includeSet.Exists(colName)) {
    return true
}
```

I can test a fix against our setup if that's useful. For now we're working around it by adding the four metadata column names to `selected_columns.columns` directly. That doesn't reintroduce the excluded source columns, since `RetainSelectedColumns` filters `record.Data` before `OlakeColumns` gets merged in.

copacabanaservice01:
# Description

Preserve OLake-owned metadata columns when `selected_columns` filters an Iceberg schema. User-selected filtering now applies only to non-OLake properties, matching the existing intent in catalog merging and the Parquet path.

This prevents `_olake_id` and the other OLake metadata fields from being omitted at table-creation time and later reintroduced as optional schema-evolution fields.

Fixes #1177

## Type of change

# How Has This Been Tested?

- [x] Added `TestTypeSchemaToIcebergAlwaysIncludesOlakeColumns` for both `sync_new_columns=false` and `sync_new_columns=true`.
- [x] Proved the regression test fails with the previous filter condition and passes with this change.
- [x] `go test ./types -count=1`
- [x] `go vet ./types`
- [x] `go build ./...`

`go test ./... -count=1` was also run on Windows. The touched `types` package passed; the existing `destination/parquet` path-separator tests failed identically on an untouched `master` checkout (`TestStagingPaths` and `TestDropStreamsRemoves2PCState`).

# Screenshots or Recordings

N/A

## Documentation

## Risk and scope

The production change is one condition in `TypeSchema.ToIceberg`. It bypasses `selected_columns` only for properties already marked `OlakeColumn`; ordinary source columns, default-column mode, and explicitly included partition columns retain their existing behavior.

## Related PR's (If Any):

None.


mihir-datazip (dismissed):

mihir-datazip (dismissed):

mihir-datazip (commented):
Placeholder comment

vaibhav-datazip (dismissed):

mihir-datazip (dismissed):

mihir-datazip (dismissed):

types/type_schema.go
```diff
@@ -175,7 +175,7 @@ func (t *TypeSchema) ToIceberg(defaultColumns bool, stream StreamInterface, incl
 		colName := key.(string)
 		outName := stream.ResolveColumnName(colName)
 		// skip non-olake columns in defaultColumns mode unless explicitly included (e.g. partition columns)
-		if !isSelected(outName) || (defaultColumns && !prop.OlakeColumn && !includeSet.Exists(colName)) {
+		if (!prop.OlakeColumn && !isSelected(outName)) || (defaultColumns && !prop.OlakeColumn && !includeSet.Exists(colName)) {
```
vaibhav-datazip:
```suggestion
		if !prop.OlakeColumn && (!isSelected(outName) || (defaultColumns && !includeSet.Exists(colName))) {
```

copacabanaservice01:
Thanks for the approval. The latest head is mergeable and the CLA check is green. When convenient, could a maintainer merge this, or let me know if another check is still required?

types/type_schema_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/datazip-inc/olake/constants"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTypeSchemaToIcebergAlwaysIncludesOlakeColumns(t *testing.T) {
+	for _, syncNewColumns := range []bool{false, true} {
+		t.Run(fmt.Sprintf("sync_new_columns_%t", syncNewColumns), func(t *testing.T) {
+			stream := NewStream("users", "public", nil)
+			stream.UpsertField("id", Int64, false, false)
+			stream.UpsertField("email", String, false, false)
```
vaibhav-datazip: also can you change this to follow the same structure as we have used in other unit test files. you can check schema_test or catalog_test

then you can add a few test cases as well, like including just one column in selected columns etc, or keeping it blank

vaibhav-datazip:
> Thanks for the approval. The latest head is mergeable and the CLA check is green. When convenient, could a maintainer merge this, or let me know if another check is still required?

@copacabanaservice01 ,  That was a test approval to check if the tests are working fine on external PRs or not.

 I've added few comments now please check those, after that we can proceed with merging this pr

types/type_schema.go
copacabanaservice01: Applied in b71b076. The factored condition is equivalent for regular columns and keeps OLake metadata outside both selection filters.

types/type_schema_test.go
copacabanaservice01: Updated in b71b076. I switched the test to table-driven cases and added nil/blank, one/all selected, `sync_new_columns`, default-column, and explicit-include coverage. The focused and full `types` tests, `go vet ./types`, and `go build ./...` all pass.

vaibhav-datazip (approved):
LGTM

PR #1182, opened 2026-09-03
fix: map Iceberg REST auth type

sahitya-datazip:
# Description

Iceberg treats `rest.auth.type` as an AuthManager scheme (`oauth2`, `none`, `sigv4`, `google`, `basic`) or a Java class name. 
This maps UI labels to Iceberg values in destination.
Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Locally tested for Unity auth type(PAT)

## Documentation

### DIFF
--- a/destination/iceberg/config.go
+++ b/destination/iceberg/config.go
@@ -116,7 +116,7 @@ func (c *Config) Validate() error {
 	if c.CatalogType == "biglake" {
 		c.RestAuthType = "org.apache.iceberg.gcp.auth.GoogleAuthManager"
 	}
-	c.RestAuthType = icebergRESTAuthType(c.RestAuthType)
+	c.RestAuthType = olakeAuthTypeToIcebergAuthType(c.RestAuthType)
 	if slices.Contains(constants.RESTCatalogs, string(c.CatalogType)) {
 		c.CatalogType = RestCatalog
 	}
@@ -204,10 +204,8 @@ func (c *Config) Validate() error {
 	return utils.Validate(c)
 }
 
-func icebergRESTAuthType(authType string) string {
+func olakeAuthTypeToIcebergAuthType(authType string) string {
 	switch strings.ToLower(strings.TrimSpace(authType)) {
-	case "":
-		return ""
 	case "oauth2", "oauth2 u2m", "oauth2 m2m", "token", "token federation",
 		"personal access token (pat)", "pat":
 		return "oauth2"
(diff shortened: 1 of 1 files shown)


destination/iceberg/config.go
```diff
@@ -202,3 +203,24 @@ func (c *Config) Validate() error {
 	}
 	return utils.Validate(c)
 }
+
+func icebergRESTAuthType(authType string) string {
```
mihir-datazip: Rename this function to `olakeAuthTypeToIcebergAuthType` for clarity

destination/iceberg/config.go
```diff
@@ -202,3 +203,24 @@ func (c *Config) Validate() error {
...
+	switch strings.ToLower(strings.TrimSpace(authType)) {
+	case "":
```
mihir-datazip: This case can be handled in default only

destination/iceberg/config.go
```diff
@@ -202,3 +203,24 @@ func (c *Config) Validate() error {
...
+		return "none"
+	case "sigv4":
+		return "sigv4"
+	case "google", "gcp":
+		return "google"
+	default:
+		if strings.Contains(authType, ".") && !strings.Contains(authType, " ") {
+			return authType
+		}
+		return ""
```
mihir-datazip, line 223: Add a warn log here printing the olake auth type as mapping is missing ideally for this

destination/iceberg/config.go
mihir-datazip, line 1: The olake-ui server's optimization mapper copies the stored destination config's rest_auth_type straight into the rest.auth.type catalog property at `server/internal/services/optimization/mapper.go:191`. That path never runs the driver's validation, so a Unity destination saved with "OAuth2 U2M" would still hand the raw label to whatever consumes those properties. The driver fix does not cover it.

Just confirm if this behaviour is fine...

mihir-datazip (changes requested):

destination/iceberg/config.go
sahitya-datazip: [chore: address review comments](https://github.com/datazip-inc/olake/pull/1182/changes/74cd534a040537bf047343059d936ef0164151b4)

destination/iceberg/config.go
sahitya-datazip: [chore: address review comments](https://github.com/datazip-inc/olake/pull/1182/changes/74cd534a040537bf047343059d936ef0164151b4)

destination/iceberg/config.go
sahitya-datazip: [chore: address review comments](https://github.com/datazip-inc/olake/pull/1182/changes/74cd534a040537bf047343059d936ef0164151b4)

destination/iceberg/config.go
sahitya-datazip: For Ingestion there won't be issue as change was made for iceberg writer

mihir-datazip (approved):
LGTM

PR #1137, opened 2026-08-21
test: backwards compatibility suite and refactor of the test infra

mihir-datazip:
# Description

Adds the backward-compatibility suite -- the same scenarios run twice, once entirely on a baseline image and once handing off to this build after the initial load, then the two destinations are compared -- and splits the test harness so it can host it.

`testutils` now holds only what every suite shares (`TestConfig`, docker plumbing, JSON/state helpers); the suites live in `integration/`, `performance/` and `compatibility/`. `TestConfig.Setup` resolves the repo root, creates the suite's own working directory, copies the shared fixtures in and lets a driver's own override them by name, so files are addressed by name rather than by a field each.

Also fixes a product bug this surfaced: `LoadedStateVersion` was initialised from `LatestStateVersion` by a package-level var initializer, which Go evaluates before `init()` assigns it -- so every command that does not load a state file ran at state version 0 and silently took the pre-v7 branch of the gated parquet behaviours.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [x] `make test.lint` -- clean across all nine modules, db2 included
- [x] `make test.discover.<driver>` -- passes for postgres, mysql, mongodb, mssql, oracle, kafka (JSON/AVRO) and s3 (all four variants)
- [x] `make test.sync.<driver>` -- passes for postgres, mysql, mongodb, mssql, oracle, kafka; s3 passes 3 of 4 variants

# Screenshots or Recordings

N/A

## Documentation

- [x] N/A (refactor and test changes only)

## Related PR's (If Any):

None

https://claude.ai/code/session_01KoSeYoWQWg4xsEg89NVsRE

Issue #1031, opened 2026-07-13
MySQL BINARY fields are not supported with Iceberg destination


miikka:
## Environment

- **Deploy method**: AWS ECS (Fargate) using `olakego/source-mysql` image
- **Olake version**:  0.8.2

## Description

I'm trying to capture CDC data from an Aurora MySQL database and to stream it to Apache Iceberg in AWS S3 Tables. I have a table with `binary(10)` primary key - simplified schema:

```sql
CREATE TABLE `items` (
  `id` binary(10) NOT NULL,
  PRIMARY KEY (`id`)
) DEFAULT CHARSET=utf8mb4;
```

The values in `id` field are random and they are not guaranteed to be valid UTF-8. `olake discover` maps the type to `string`, but when I actually try to run OLake, I get the following error in the logs:

```
failed to create new writer thread: failed to setup writer thread: failed to create writer thread: failed to load or create table: rpc error: code = Internal desc = grpc: error while marshaling: string field contains invalid UTF-8
```

I'm not sure exactly what goes wrong but `binary(10)` can have non-valid UTF-8 and Protobuf strings (used in gRPC) are required to be valid UTF-8. Looks similar to the MSSQL-related issue https://github.com/datazip-inc/olake/pull/845.

## Steps to reproduce

1. Have the table with primary key `binary(10)`.
2. Insert some rows with non-UTF-8 IDs - random bytes should do the trick
3. `olake discover`; configure Iceberg destination
4. `olake sync`

## Expected behavior

The writer works and the data lands as `binary` in Iceberg.

## Actual behavior

The writer does not even start. Nothing gets written to Iceberg.

vaibhav-datazip:
Hi @miikka , we are looking into this issue.

vaibhav-datazip:
@miikka , is this a blocker for your use case ?

PR #1183, opened 2026-09-03
feat: carry MySQL binary and fixed-length binary columns as bytes into Iceberg

mihir-datazip:
# Description

Adds: https://github.com/datazip-inc/olake/issues/1031

Source binary columns were cast to string on the way to the writers. Any byte that is not valid UTF-8 failed the Iceberg writer's gRPC marshal (`string field contains invalid UTF-8`), so a MySQL table with a `BINARY`, `VARBINARY` or `BLOB` column holding non-text bytes could not be synced at all. This adds binary types to the olake type model and carries bytes end to end for MySQL.

**Types**
- `Binary` DataType (iceberg `binary`, parquet `BYTE_ARRAY`).
- `FixedBinary` as a parameterised type: `FixedBinaryOf(16)` is `fixed_binary(16)`, mapping to iceberg `fixed[16]` and parquet `FIXED_LEN_BYTE_ARRAY(16)`; `IcebergTypeToDatatype` parses `fixed[n]` back. The length rides inside the type string, so the catalog JSON and `Set[DataType]` are unchanged.
- `ToIceberg`, `ToNewParquet` and `IcebergTypeToDatatype` are table driven (`destinationTypes`, `icebergToDataType`), with tests that every declared type has explicit mappings.
- The typecast tree is a general n-children tree: `Binary` above `String`, `FixedBinary` under `Binary`, `Object`/`Array` under `String`, `Null`/`Unknown` attached through the parent map. Two fixed binaries of different lengths widen to `Binary`.

**Destination pipeline**
- `TypeFromValue` maps `[]byte` to `Binary`; the flattener keeps byte slices; `ReformatValue` routes binary types through `ReformatBytes`, which zero-pads short `fixed_binary(n)` values (MySQL strips the padding from binlog row images) and rejects over-long ones before parquet-go panics.
- Legacy writer sends bytes through the proto `bytes_value` field; Arrow writer maps `binary` / `fixed[n]` to Arrow `Binary` / `FixedSizeBinary`.
- Java writer parses `fixed[n]` to `Types.FixedType` and converts `BINARY` (ByteBuffer) and `FIXED` (byte[]) values.
- Schema check accepts a value-detected `binary` into an existing `fixed[n]` column; `GetKeysHash` hex-encodes byte primary keys.

**MySQL**
- `binary`, `varbinary` and the blob family map to `Binary`; discover turns `BINARY(n)` into `fixed_binary(n)` from `COLUMN_TYPE`.
- Binlog: the binary collation tells `BINARY`/`VARBINARY`/`BLOB` apart from `CHAR`/`VARCHAR`/`TEXT`, which share wire types.

**Scope decisions**
- No state-version gate: an Iceberg table that already holds these columns as `string` fails schema validation on its next sync, since `string -> binary` is not a valid transition.
- MySQL only; other drivers keep their string mapping. Kafka Avro `bytes` values now flow as `Binary` through the shared pipeline.
- The compatibility rule excludes the binary data types below `v0.9.6`, a placeholder for the release that ships this.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Unit tests across `types`, `typeutils`, `utils`, both Iceberg writers, the transition check, the MySQL type map and the binlog helpers (`go test ./types/ ./utils/... ./destination/... ./pkg/binlog/` and `drivers/mysql`).
- [x] Integration suite with new seed columns `data_binary BINARY(16)`, `data_varbinary VARBINARY(64)`, `data_blob BLOB` holding non-UTF-8 bytes: `make test.integration.mysql` passes (Discover, Sync for legacy/arrow/parquet with CDC and incremental, 2PC).
- [x] Manual end to end: full refresh and CDC of a table with `BINARY(16)`, `BINARY(4)`, `VARBINARY(64)`, `BLOB` land byte-exact in Iceberg as `fixed[16]`, `fixed[4]`, `binary` (verified with Spark `hex()` against MySQL `HEX()`).
- [ ] `make test.compatibility.mysql COMPATIBILITY_BASELINE=<base sha>`: passes only with the harness change that dates a commit baseline by its release tag, which is not part of this PR yet; without it the base build seeds the binary columns and fails as expected.

# Screenshots or Recordings

N/A

## Documentation

- [ ] Documentation Link: the data types page in olake-docs needs `binary` and `fixed_binary(n)` added; not written yet.

## Related PR's (If Any):

- #1137 (base branch: backward-compatibility test infra this builds on).


drivers/mysql/internal/datatype_conversion.go
```diff
@@ -72,3 +76,53 @@ var mysqlTypeToDataTypes = map[string]types.DataType{
...
+		return mapped
+	}
+	params := getColumnTypeParams(columnType)
+
+	parameterisedType, ok := typeConstructor(params)
+	if !ok {
+		return mapped
+	}
+	return parameterisedType
+}
```
vaibhav-datazip: can we simplify this function , It will remove our dependency on regex and other functions as well, as parameterized types also has just binary in it and array value (the scale and precision of binary and decimals) also wont need regex things

```suggestion
func resolveColumnType(dataType, columnType string, mapped types.DataType) types.DataType {
	if dataType != "binary" {
		return mapped
	}
	_, rest, ok := strings.Cut(columnType, "(")
	if !ok {
		return mapped
	}
	nStr, _, _ := strings.Cut(rest, ")")
	n, err := strconv.Atoi(strings.TrimSpace(nStr))
	if err != nil || n <= 0 {
		return mapped
	}
	return types.FixedBinaryOf(n)
}
```

drivers/mysql/internal/datatype_conversion.go
```diff
@@ -40,12 +44,12 @@ var mysqlTypeToDataTypes = map[string]types.DataType{
...
-	"tinyblob":   types.String,
-	"blob":       types.String,
-	"mediumblob": types.String,
-	"longblob":   types.String,
+	"binary":     types.Binary,
+	"varbinary":  types.Binary,
+	"tinyblob":   types.Binary,
+	"blob":       types.Binary,
+	"mediumblob": types.Binary,
+	"longblob":   types.Binary,
```
vaibhav-datazip, line 52: shouldn't we add this same support for other drivers as well


Driver | File | Still String
-- | -- | --
Postgres | drivers/postgres/internal/datatype_conversion.go | bytea (line 40)
Oracle | drivers/oracle/internal/datatype_conversion.go | raw, longraw, blob, bfile (31–32, 45–46)
MSSQL | drivers/mssql/internal/datatype_conversion.go | binary, varbinary, image, rowversion, timestamp (46–50)
DB2 | drivers/db2/internal/datatype_conversion.go | blob, binary, varbinary (41–43)

drivers/mysql/go.mod
```diff
@@ -12,7 +12,10 @@ require (
 	github.com/go-sql-driver/mysql v1.8.1
 )
 
-require github.com/jmoiron/sqlx v1.4.0
+require (
+	github.com/jmoiron/sqlx v1.4.0
+	github.com/stretchr/testify v1.11.1
```
vaibhav-datazip, line 17: testify is added in mod but its never imported by any file in this folder , maybe we can remove this

types/type_schema.go
```diff
@@ -246,77 +248,133 @@ func (p *Property) Nullable() bool {
...
+//					    /			  \
+//	            (Float64)             (TimestampNano)
+//		        /	    \            		/
+//			   / 	     \				   /
+//	    (Int64)       (Float32)       (TimestampMicro)
+//			             /                  /
+//			            /                  /
+//			         (Int32)         (TimestampMilli)
+//			          /                  /
+//			         /                  /
```
vaibhav-datazip, line 295: this tree representation might be wrong , as int 32 is in 64 branch right ?

utils/typeutils/flatten.go
```diff
@@ -70,7 +71,12 @@ func (f *FlattenerImpl) flatten(key string, value any, destination types.Record)
 	case bool, int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64, float32, float64, string, time.Time, json.Number:
 		destination[outKey] = v
 	case []byte:
-		destination[outKey] = string(v)
+		switch {
+		case constants.LoadedStateVersion < 8:
+			destination[outKey] = string(v)
+		default:
+			destination[outKey] = v
+		}
```
vaibhav-datazip, line 79:
```suggestion
destination[outKey] = utils.Ternary(constants.LoadedStateVersion < 8, string(v), v)
```

types/type_utils.go
```diff
@@ -0,0 +1,204 @@
...
+	for _, f := range families {
+		if d == f.pattern {
+			return f, nil
+		}
+		if params, ok := f.parse(string(d)); ok {
+			return f, params
+		}
+	}
+	return nil, nil
+}
```
vaibhav-datazip, line 117: do we need to check for  various families as we are currently only using this for binary , also this function can then be simplified to something like this , this might also reduce some other functions like Params

```
func instanceOf(d DataType) (isPattern bool, width int, ok bool) {
	s := string(d)
	if s == string(FixedBinary) { // "fixed_binary(%d)"
		return true, 0, false // family, no params
	}
	var n int
	if _, err := fmt.Sscanf(s, "fixed_binary(%d)", &n); err != nil || n <= 0 {
		return false, 0, false
	}
	return false, n, true // instance
}
```

utils/utils.go
```diff
@@ -267,11 +267,24 @@ func IsJSON(str string) bool {
 }
 
 // GetKeysHash returns md5 hashsum of concatenated map values (sort keys before)
```
vaibhav-datazip, line 269: this comment needs to moved down

utils/utils.go
vaibhav-datazip, line 293: in this get hash function as well, shouldn't we hex encode the binary values before appending ?

destination/iceberg/iceberg.go
```diff
@@ -509,18 +498,26 @@ func parsePartitionRegex(pattern string, resolveColumnName func(string) string)
...
 
-// isValidTransition checks if type transition is valid using lookup table
-func isValidTransition(oldType, newType string) bool {
-	if oldType == newType {
+// isValidTypeForColumn reports whether a column of columnType can take a value of incomingType
+func isValidTypeForColumn(columnType string, incomingType types.DataType) bool {
+	olakeTypeForColumn := types.IcebergTypeToDatatype(columnType)
+	if olakeTypeForColumn.ToIceberg() != columnType {
+		return false
+	}
```
vaibhav-datazip, line 506: isn't this just round tripping the datatype , will this branch ever get fired.

types/type_utils.go
```diff
@@ -0,0 +1,204 @@
...
+// value cannot reveal parameters: a fixed_binary(16) value is detected as binary. ok is false for
+// a type that carries none, which detection can already name exactly.
+func ParameterlessForm(d DataType) (DataType, bool) {
+	if _, parameterised := d.Params(); !parameterised {
+		return "", false
+	}
+	if parent, inTree := typeParent[BaseOf(d)]; inTree {
+		return parent, true
+	}
+	return String, true
```
vaibhav-datazip, line 158: shouldn't binary be returned from here as it sits on top of the tree ?

drivers/mysql/internal/mysql.go
```diff
@@ -293,7 +293,7 @@ func (m *MySQL) dataTypeConverter(value interface{}, columnType string) (interfa
 		}
 	}
 
-	olakeType := typeutils.ExtractAndMapColumnType(columnType, mysqlTypeToDataTypes)
+	olakeType := types.ForLoadedState(typeutils.ExtractAndMapColumnType(columnType, mysqlTypeToDataTypes))
```
vaibhav-datazip, line 296: even for incremental state saving the binary cursor field , we should add these checks as they might throw error while re reading in next sync

destination/iceberg/iceberg.go
```diff
@@ -509,18 +498,26 @@ func parsePartitionRegex(pattern string, resolveColumnName func(string) string)
 	return partitionInfo, nil
 }
 
-// isValidTransition checks if type transition is valid using lookup table
-func isValidTransition(oldType, newType string) bool {
-	if oldType == newType {
+// isValidTypeForColumn reports whether a column of columnType can take a value of incomingType
+func isValidTypeForColumn(columnType string, incomingType types.DataType) bool {
```
vaibhav-datazip, line 502: This runs once per value per record (iceberg.go:261) on every normalized Iceberg sync.

Each call does a regex match (IcebergTypeToDatatype → icebergInstance) and two ToIceberg() calls that each fmt.Sprintf, even for "long". Micro-benchmark for a 50-column record: 1.3 µs / 40 allocs before, 6.2 µs / 190 allocs now.

Two small changes bring it to 1.66 µs / 40 allocs:
- Fast path at the call site: finalSchema[key] == detectedIcebergType || (that value is already computed on line 258).
- In ToIceberg, return mapping.icebergType without Sprintf when len(params) == 0.

constants/state_version.go
```diff
@@ -39,10 +39,17 @@ import (
...
-//   - Version 7: (Current Version) Parquet INT96 and unsigned 32-bit columns map to their correct types.
+//   - Version 7: Parquet INT96 and unsigned 32-bit columns map to their correct types.
 //     * INT96: earlier the raw 96-bit integer was emitted as a string, which disagreed with the inferred Timestamp schema and collapsed the column to String.
 //     * Unsigned 32-bit: earlier read as a signed int32 and mapped to Int32, so values above 2^31-1 wrapped negative. Now widened to Int64, matching pg/mysql.
 //     * Older state keeps both previous behaviors so existing destination columns do not change type on upgrade.
+//
+//   - Version 8: (Current Version) MySQL binary columns keep their bytes.
+//     * BINARY/VARBINARY/BLOB map to Binary, and BINARY(n) to fixed_binary(n), where they were String.
+//     * A byte value detects as Binary and survives flattening instead of being cast to a string.
+//     * The binlog keeps a binary-collation column's raw bytes and reports its BINARY type name.
```
vaibhav-datazip, line 50: in newer unit tests as well, shouldn't there be case for version set <8 and passing binary

types/fixed_binary.go
```diff
@@ -0,0 +1,26 @@
...
+
+import "github.com/parquet-go/parquet-go"
+
+// fixedBinaryFamily registers fixed_binary(%d): one width, and it must be positive. A fixed width
+// is a promise about every value, so two different widths cannot meet in a third; instances that
+// disagree fall to the family's parent in the typecast tree, which is also the type a value
+// reveals when its width cannot be seen.
+var fixedBinaryFamily = newTypeFamily(FixedBinary, func(p []int) bool { return len(p) == 1 && p[0] > 0 })
+
+func fixedBinaryNode(params ...any) parquet.Node {
```
vaibhav-datazip, line 11: a DDL which upscales binary values ex. binary(8) to binary (16) might fail in schema evolution. why dont we map all binary types  to just binary

tests/testutils/integration/integration.go
```diff
@@ -69,17 +70,17 @@ func (th *TestHandler) runSyncAndVerify(
 	case "iceberg":
 		{
 			if evolvedSchema {
-				VerifyIcebergSync(t, testTable, th.TestConfig.DestinationDB, th.UpdatedDestinationDataTypeSchema, th.DefaultCDCColumnsSchema, schema, opSymbol, th.TestConfig.PartitionRegex, th.TestConfig.Driver, isCDC, th.TestConfig.ColumnToExclude)
+				VerifyIcebergSync(t, testTable, th.TestConfig.DestinationDB, th.UpdatedDestinationDataTypeSchema, th.TypeMapping, th.DefaultCDCColumnsSchema, schema, opSymbol, th.TestConfig.PartitionRegex, th.TestConfig.Driver, isCDC, th.TestConfig.ColumnToExclude)
 			} else {
-				VerifyIcebergSync(t, testTable, th.TestConfig.DestinationDB, th.DestinationDataTypeSchema, th.DefaultCDCColumnsSchema, schema, opSymbol, th.TestConfig.PartitionRegex, th.TestConfig.Driver, isCDC, th.TestConfig.ColumnToExclude)
+				VerifyIcebergSync(t, testTable, th.TestConfig.DestinationDB, th.DestinationDataTypeSchema, th.TypeMapping, th.DefaultCDCColumnsSchema, schema, opSymbol, th.TestConfig.PartitionRegex, th.TestConfig.Driver, isCDC, th.TestConfig.ColumnToExclude)
```
vaibhav-datazip, line 75: Partitioning using binary fields will fail, should we add a explicit check for this instead of failing when trying to commit , also we can mention this in docs

destination/iceberg/iceberg.go
```diff
@@ -509,18 +498,26 @@ func parsePartitionRegex(pattern string, resolveColumnName func(string) string)
...
+	if incomingType.ToIceberg() == columnType {
 		return true
 	}
-	if transitions, ok := validTypeTransitions[oldType]; ok {
-		if transitions[newType] {
-			return true
-		}
+
+	// value detection cannot reveal a width, so a parameterised column takes the parameterless form
+	// on trust; ReformatBytes enforces the parameters per value when the record is written
```
vaibhav-datazip, line 515: here are some more performance regressions flagged by ai please check those , if these are relevant and can degrade performance for billions of records transfer

| path | old | new | where |
|---|---|---|---|
| legacy writer `toProtoFieldValue`, 3 common values | 116 ns | 284 ns (+56 ns/value) | `legacy-writer/writer.go:250` does `IcebergTypeToDatatype(iceType)` (regex) and up to 7 `SameType` calls per value |
| legacy writer, one `fixed[16]` value | — | 812 ns, 24 allocs | `IcebergTypeToDatatype`, `ReformatBytes` → `Params()` regex |
| arrow `FixedSizeBinaryBuilder` | — | same shape as above | `arrow-writer/utils.go:353`: `FixedBinaryOf(width)` (`Sprintf`) plus a regex parse, per value |
| parquet `Fields.Process`, 30 columns | 981 ns / 41 allocs | 1574 ns / 81 allocs | `fields.go:96,77`: `TypeFromValue` now runs **twice** per value (string values try every date format) |
| parquet batch fast path | skips `Process` when types match | **never skips** for a table with BINARY(n) | `parquet.go:537-539` compares `elem == detectedType`; a detected `Binary` never equals `fixed_binary(16)`, so every batch runs the sequential `Process` loop |

**Fixes**
- Legacy writer: keep the string `switch iceType` and add `case iceType == "binary" || strings.HasPrefix(iceType, "fixed[")`. The comment on `toArrowType` ("fixed[n] carries a width and so cannot be a string case") is answered by `HasPrefix`.
- `valueFitsColumn(value, detected, column)`: pass in the already-detected type.
- Parquet fast path: also accept `detectedType == types.Binary && FixedBinaryWidth(elem) > 0`.
- `ReformatBytes(v, width int)`.
- These may get fixed if fixed[n] is dropped.

utils/typeutils/flatten.go
mihir-datazip: Switch case should be standard practice for state version gates because the code blocks for different state is separated cleanly and in future if new gate comes, it will be just a new case.

types/type_utils.go
mihir-datazip: In future decimal family will also come, we need to write logic accordingly

drivers/mysql/internal/datatype_conversion.go
mihir-datazip: We have discussed this with @shubham19may and for this change we wish to only support MySQL, we can add support for other drivers later

types/type_utils.go
mihir-datazip: Ideally adding new parameterised type should have minimal changes. For a common function like `instanceOf `, we should not need to add if else blocks for type specific logic.

types/fixed_binary.go
mihir-datazip: The reason is that people usually use fixed binary as primary key columns so we should conserve the width as well in destination.
Also I think it is fine to fail schema evolution for this case but I will confirm with @shubham19may once

drivers/mysql/go.mod
mihir-datazip: Removed

types/type_schema.go
mihir-datazip: Fixed

utils/utils.go
mihir-datazip: Moved

utils/utils.go
mihir-datazip: We are appending keys which are always strings....
What do you mean here?

destination/iceberg/iceberg.go
mihir-datazip: It does fire: for any Iceberg type OLake doesn't model, like `date`, `time`, `uuid`, `timestamp`, `decimal(10, 2)`, `list<..>` or `struct<..>`

types/type_utils.go
mihir-datazip: Not all drivers support Binary. This case is fallback for types outside our tree as not all the drivers will support binary, this case is handled as string

drivers/mysql/internal/mysql.go
mihir-datazip: Added

destination/iceberg/iceberg.go
mihir-datazip: Done

constants/state_version.go
mihir-datazip: Added UTs

destination/iceberg/iceberg.go
mihir-datazip: Fixed the performance issues

utils/utils.go
vaibhav-datazip: This was related to how hash is calculated currently when no pk is present in table, not hex encoding the binary values might create some issues as well.

vaibhav-datazip: olakeID differs backfill vs CDC for a BINARY(n) primary key

olakeID computed on the raw driver record, before the writer: abstract/backfill.go:80, cdc.go:145, incremental.go:141.

- Backfill: SELECT returns BINARY(16) fully 0x00-padded → keyString hex of 16 bytes.
- CDC: MySQL strips trailing 0x00 in binlog row image — stated in this PR's own typeutils/reformat.go:126-127 and seed comment tests/mysql/mysql_util_test.go:83-84. And binaryTypeName drops the width (CHAR→BINARY, pkg/binlog/filter.go), so converter runs ReformatValue(Binary,…)→ReformatBytes(v, 0)` → no padding.
- Padding restored only in the writer. olakeID already computed by then.

Same row → two different _olake_id → CDC update/delete never matches backfilled row → duplicates/orphans in upsert mode.

Integration suite does not catch it: PK is id INT; data_fixed_binary is a data column only.

drivers/abstract/incremental.go
```diff
@@ -164,9 +165,34 @@ func ReformatCursorValue(cursorField string, cursorValue any, stream types.Strea
...
+
+// DecodeCursorValue returns the value a cursor's state entry stands for. JSON cannot carry bytes, so
+// FormatCursorValue keeps a binary cursor as hex and it decodes back to its bytes here; every other
+// cursor is kept in state as its value.
+func DecodeCursorValue(cursorField string, stateValue any, stream types.StreamInterface) (any, error) {
+	encoded, isString := stateValue.(string)
+	if !isString || !IsBinaryCursor(cursorField, stream) {
+		return stateValue, nil
+	}
+	decoded, err := hex.DecodeString(encoded)
```
vaibhav-datazip, line 189: shouldn't in this function we just take the value of cursorColType we have already computed earlier and using that run decode ?

destination/iceberg/iceberg.go
vaibhav-datazip: renaming is needed

destination/iceberg/iceberg.go
vaibhav-datazip: **Still slower:**

| Case | base | head | head + §5.1 patch |
|---|---|---|---|
| Iceberg validity, per value, `fixed[16]` column | — | **252 ns, 9 allocs** | 21 ns, 0 allocs |
| Iceberg batch with 10 `fixed[16]` + 10 `binary` columns | (11.6 ms with those columns as string) | 12.2 ms, **1.58M allocs** | 11.9 ms, 680k allocs |
| Iceberg validity, per value, long column ← int32 | 9 ns | **114 ns** | 14 ns |
| Iceberg batch, Mongo/Kafka-JSON shape (long←int32, double←int64, string←bool) | 11.2 ms | **12.1 ms (+8%)** | 11.3 ms |
| Parquet batch, 10 BINARY(16) columns | — | 22.1 ms, 1.85M allocs (vs 20.5 ms / 1.45M as `binary`) | allocs 1.45M (same as `binary`) |
| Legacy, per `fixed[16]` / `binary` value | — | 45 ns / 43 ns | unchanged, already fine |


Where the remaining cost comes from:
- Fixed column, Iceberg: every value round-trips the column type. IcebergTypeToDatatype("fixed[16]") runs scan (ReplaceAll + Split) and Of (Sprintf) to build fixed_binary(16). ToIceberg() then parses it again (instanceOf → scan + asAny) and Sprintfs "fixed[16]" back. That is the same round trip you flagged at iceberg.go:506.
- Mixed types (non-MySQL drivers too): isValidTypeForColumn has no lookup table. Every widening goes through IcebergTypeToDatatype → Accepts → lowestCommonAncestor, which builds an ancestors map on each call (type_schema.go:369). #657 replaced exactly this kind of per-value work with a map.
- Parquet fixed columns: FixedBinaryWidth is parsed twice per value, once in ValueFitsColumn (fields.go:80-82) and once in ReformatValue's default case (reformat.go:104), each time through instanceOf → scan.

types/data_types.go
```diff
@@ -84,69 +135,43 @@ func GetIcebergRawSchema() []*proto.IcebergPayload_SchemaField {
...
-		return Float64
-	case "timestamptz":
-		return TimestampMilli
-	default:
+// ForLoadedState returns the type a column carries for the state version this sync is pinned at.
+// Binary columns were carried as text before state version 8, so state written by such a build
+// keeps them as String and an existing destination column does not change type on upgrade.
+func ForLoadedState(d DataType) DataType {
+	switch {
+	case constants.LoadedStateVersion < 8 && (d == Binary || BaseOf(d) == FixedBinary):
```
vaibhav-datazip, line 165: can we make binary assertion function consistent everywhere

| Where | Spelling |
|---|---|
| `ForLoadedState` (`data_types.go:165`) | `d == Binary \|\| BaseOf(d) == FixedBinary` |
| `IsBinaryCursor` (`incremental.go:178`) | `SameType(t, Binary) \|\| SameType(t, FixedBinary)` |
| `ReformatValue` (`reformat.go:66,104`) | `case types.Binary` + default `FixedBinaryWidth(d) > 0` |
| `ValueFitsColumn` (`fields.go:80`) | `BaseOf(column) == FixedBinary` |
| writers (`writer.go:306`, `transforms.go`) | `IcebergBytesWidth(iceType)` |
| arrow `toArrowType` (`arrow-writer/utils.go:162-182`) | `IcebergTypeToDatatype` + a `SameType` chain + `Params()`; the legacy writer uses a string switch for the same job |

utils/typeutils/fields.go
```diff
@@ -72,6 +72,19 @@ func (f Fields) Header() (header []string) {
 	return
 }
 
+// ValueFitsColumn checks if a value of detectedType fits the column type, so the column need not change.
+func ValueFitsColumn(value any, detectedType, column types.DataType) bool {
```
vaibhav-datazip, line 76: similarly here the third argument can be changed to `columnType` that might be a better name

drivers/mysql/internal/datatype_conversion.go
```diff
@@ -72,3 +75,20 @@ var mysqlTypeToDataTypes = map[string]types.DataType{
...
+	_, rest, ok := strings.Cut(columnType, "(")
+	if !ok {
+		return mapped
+	}
+	nStr, _, _ := strings.Cut(rest, ")")
+	n, err := strconv.Atoi(strings.TrimSpace(nStr))
+	if err != nil || n <= 0 {
+		return mapped
+	}
+	return types.FixedBinaryOf(n)
```
vaibhav-datazip, line 93: there is one more column in this query `INFORMATION_SCHEMA.COLUMNS.CHARACTER_MAXIMUM_LENGTH` which is exactly the byte width for these columns (binary(255) → 255, binary(0) → 0), and MySQLTableSchemaQuery (pkg/jdbc/jdbc.go:419) already selects from that , so maybe we can utilize this directly

utils/typeutils/datatype_test.go
```diff
@@ -118,6 +119,17 @@ func TestTypeFromValue(t *testing.T) {
...
+func TestTypeFromValueBytes(t *testing.T) {
+	defer func(version int) { constants.LoadedStateVersion = version }(constants.LoadedStateVersion)
+
+	assert.Equal(t, types.Binary, TypeFromValue([]byte{0xff, 0x00}))
+	assert.Equal(t, types.Binary, TypeFromValue([]byte{}))
+
+	constants.LoadedStateVersion = 7
+	assert.Equal(t, types.String, TypeFromValue([]byte{0xff, 0x00}), "older state detects bytes as text")
+	assert.Equal(t, types.String, TypeFromValue([]byte{}), "older state detects bytes as text")
+}
```
vaibhav-datazip, line 131: can we change this as well , in our test case and iterative assertion format , which has been followed across the codebase

utils/keys_hash_test.go
```diff
@@ -0,0 +1,31 @@
...
+	assert.Equal(t, GetKeysHash(map[string]any{"a": "ff", "b": "x"}, "a", "b"), GetKeysHash(composite, "a", "b"),
+		"composite keys hash the hex form of byte values")
+
+	old := constants.LoadedStateVersion
+	t.Cleanup(func() { constants.LoadedStateVersion = old })
+	constants.LoadedStateVersion = 7
+
+	key := []byte{0xff, 0x00, 0x80}
+	assert.Equal(t, fmt.Sprintf("%v", key), GetKeysHash(map[string]any{"id": key}, "id"),
+		"state written before version 8 hashed the printed form of a byte key")
```
vaibhav-datazip, line 28: here as well , the same test case like structure, also can we rename this file utils_test as this function is part of utils file

drivers/abstract/incremental.go
mihir-datazip: Done

types/data_types.go
mihir-datazip: Move to a common function that gives `isBytes`

utils/typeutils/fields.go
mihir-datazip: Done

mihir-datazip:
All six rows of the review's performance table meet or beat the "head + §5.1 patch" column when the latest changes are measured against `staging`. The baseline is `staging` at the fork point. The one newer `staging` commit, #1248, only renames the Java writer package and touches no Go hot path.

| Row | Review: base → head → head + §5.1 | Now: `staging` → latest changes | Met? |
|---|---|---|---|
| Iceberg validity per value, `fixed[16]` column | — → 252 ns, 9 allocs → 21 ns, 0 allocs | — → 19.5 ns, 0 allocs | ✅ |
| Iceberg batch, 10 `fixed[16]` + 10 `binary` | 11.6 ms as string → 12.2 ms, 1.58M allocs → 11.9 ms, 680k allocs | 9.31 ms as string → 7.52 ms, allocs −70% | ✅ faster than `staging` |
| Iceberg validity per value, long ← int32 | 9 ns → 114 ns → 14 ns (1.56×) | 12.0 → 16.3 ns (1.36×) | ✅ |
| Iceberg batch, Mongo/Kafka-JSON shape | 11.2 ms → 12.1 ms (+8%) → 11.3 ms (+1%) | 8.46 → 8.54 ms (+1%, not significant, p=0.35), same allocs | ✅ |
| Parquet batch, 10 `BINARY(16)` columns | 1.85M allocs against 1.45M as binary → allocs same as binary | 11.25 ms / 380.1k allocs against 11.31 ms / 380.1k as binary | ✅ identical |
| Legacy per `fixed[16]` / `binary` value | 45 / 43 ns → unchanged, fine | 39 / 34 ns, 2 allocs each | ✅ |

drivers/mysql/internal/datatype_conversion.go
mihir-datazip: Done

utils/typeutils/datatype_test.go
mihir-datazip: Done

utils/keys_hash_test.go
mihir-datazip: Done

PR #1184, opened 2026-09-03
fix: allow numbers in iceberg catalog names

### LINKED ISSUE #1032: issue: spec does not support having numbers in Catalog Name field in Iceberg as Destination
## Description

Currently, the UI spec does not allow creating a destination with a `catalog_name` having _numbers_ in it. Something as simple as `olake_iceberg2`.

Though the `CLI` can ingest into a destination with the `catalog_name` field having numbers.

_A more thorough approach would be:_ To check if even special characters are allowed, and accordingly update / remove the regex logic as the Catalog Service should throw an error in case the `catalog_name` is not valid, rather than the `spec` deciding whether or not the `catalog_name` is valid or not.

Kaileshwar16:
Fixes #1032

## Description

Iceberg destination specs currently restrict `catalog_name` using:

`^[a-z_]+$`

This causes valid catalog names containing numbers, such as
`olake_iceberg2`, to fail UI validation even though they are accepted
through the CLI/backend.

This PR updates the catalog name pattern to:

`^[a-z0-9_]+$`

across all Iceberg catalog configurations.

### Before fix
`olake_iceberg2` is rejected by the Catalog Name field.

https://drive.google.com/file/d/1rEWgiIL2IyHatBJooCYZ-WKm7yE31YRH/view?usp=drive_link

### After fix
Built a custom OLake driver from this branch and verified that
`olake_iceberg2` passes Catalog Name validation.

https://drive.google.com/file/d/1rP_YhcaOT3Kn6xss849v7G1LGuPwGnuD/view?usp=sharing

### ensure nothing breaks
sync run

https://drive.google.com/file/d/1QAul_PVfQUUqGQ0rx5OOFad2KLkwWBRJ/view?usp=drive_link

## Testing

- Validated `destination/iceberg/resources/spec.json`
- Tested the custom driver through the OLake UI
- Confirmed catalog names containing numbers are accepted

### DIFF
--- a/destination/iceberg/resources/spec.json
+++ b/destination/iceberg/resources/spec.json
@@ -53,7 +53,7 @@
                 "catalog_name": {
                   "type": "string",
                   "title": "Catalog Name",
-                  "pattern": "^[a-z_]+$",
+                  "pattern": "^[a-z0-9_]+$",
                   "description": "Name of the Iceberg catalog OLake Go registers tables under. Defaults to olake_iceberg if left empty."
                 },
                 "jdbc_url": {
@@ -126,7 +126,7 @@
                 "catalog_name": {
                   "type": "string",
                   "title": "Catalog Name",
-                  "pattern": "^[a-z_]+$",
+                  "pattern": "^[a-z0-9_]+$",
                   "description": "Name of the Iceberg catalog OLake Go registers tables under. Defaults to olake_iceberg if left empty."
                 },
                 "hive_uri": {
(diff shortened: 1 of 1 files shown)


nayanj98:
@Kaileshwar16 Can you please show the complete video that you were able to successfully connect the destination and also were able to sync the data. Once that is done will assign a reviewer.

Kaileshwar16:
@nayanj98 i used dummy values to establish connection so it does not get connected to destination, but scope of this issue is to allow allow numbers in catalog names and i did shown that, but if establishing connection is mandatory i am happy to do it

Kaileshwar16:
@nayanj98 i updated the link with new video showing the destination is created

nayanj98:
@Kaileshwar16 I had also requested to run the sync and show once that the sync is running perfectly. This is a necessity and we only assign a reviewer once you show that even with your changes the sync is running fine. This is one important requirement from our end as it helps the reviewer confirm that the sync is running fine based on your changes.

Kaileshwar16:
@nayanj98 i have completed the requirement, can u please check this.

PR #1186, opened 2026-09-03
chore(make): self-contained destination compose, s3 source shares its MinIO

mihir-datazip:
# Description

`destination/iceberg/local-test/docker-compose.yml` pulled its `minio` and `mc` services from `drivers/s3/docker-compose.yml` through `include:`, and both files declared the same compose project (`local-test`). Because of that, `make olake.s3.teardown` (`down --volumes --remove-orphans` on the s3 file) removed `spark-iceberg` and `iceberg-postgres` as orphans and wiped the ivy cache, after which Spark Connect no longer came up until the whole destination stack was refreshed.

This makes the destination compose self-contained and turns the s3 source into a consumer of the destination's MinIO:

- `minio` and `mc` are inlined into the destination compose, unchanged; `drivers/s3/docker-compose.yml` is deleted.
- Removed concept of SOURCE_DRIVERS and just kept the DRIVERS. We need to now take care that the driver.mk has the actual commands present else for new drivers the make targets might fail due to compose file not there
- `olake.minio.up` brings up only MinIO. `olake.s3.up` and `olake.destination.all.up` both depend on it, so `make -j` runs the step once instead of two `compose up` calls racing on the same container (the shape of CI's `make -j2 olake.s3.up olake.destination.all.up`).
- `olake.s3.wait` probes MinIO as before; `olake.s3.stop` and `olake.s3.teardown` are no-ops that point at the destination targets, which own the container's lifecycle. The s3 tests create their own `olake-s3-test` bucket, so they never depended on `mc`.

No workflow changes: CI calls `olake.<driver>.up` / `olake.<driver>.wait` generically.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] `docker compose -f destination/iceberg/local-test/docker-compose.yml config --services` resolves every service after removing the include.
- [x] `make help` still lists `s3` in `SOURCE_DRIVERS`; `make -n olake.s3.up|wait|stop|teardown|refresh` show the intended recipes.
- [x] `make -j2 -n olake.s3.up olake.destination.all.up` emits the `up -d minio` step exactly once.
- [x] Against a running stack: `make olake.s3.start` is a no-op `up` plus a successful probe, `make olake.s3.teardown` does nothing, `make olake.destination.all.start` recreates no container.
- [x] `make test.discover.s3` passes through `driver_test_setup` (`olake.s3.start` + `olake.destination.all.start` + jar).

# Screenshots or Recordings

N/A

## Documentation

## Related PR's (If Any):


Makefile
```diff
@@ -162,18 +162,17 @@ HELP_TARGETS :=
...
+CDC_DRIVERS := $(filter-out $(NON_CDC_DRIVERS),$(DRIVERS))
+DRIVER_PKGS := $(addsuffix /...,$(addprefix ./,$(DRIVERS)))
 CDC_PKGS := $(addsuffix /...,$(addprefix ./,$(CDC_DRIVERS)))
 
 # The drivers the integration suites cover, queried by CI (integration-tests.yml) so the list
 # lives in this file only: it is what the driver matrix fans out to on a push to master.
 .PHONY: print.source-drivers
 print.source-drivers:
-	@echo $(SOURCE_DRIVERS)
+	@echo $(DRIVERS)
```
sahitya-datazip: here print.source-drivers and print.drivers both echo `DRIVERS`

mihir-datazip: removed `source-drivers`

sahitya-datazip (approved):

PR #1190, opened 2026-09-04
fix: gate pr merging until workflows approved

vaibhav-datazip:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/.github/workflows/required-checks.yml
+++ b/.github/workflows/required-checks.yml
@@ -1,0 +1,188 @@
+name: Required Checks
+
+# The one status check a pull request must pass before it can be merged. It runs no tests itself:
+# it watches every other workflow run on the PR's head commit and turns green only once all of them
+# have finished successfully.
+#
+# Why one aggregate check instead of listing every job in the ruleset:
+#
+#   * Job names drift. The integration matrix names its jobs per driver ("Test postgres"), and a
+#     driver a PR does not touch never reports. A required check that never reports blocks the PR
+#     forever, with no re-run able to clear it.
+#   * A fork PR whose workflows are parked on "Approve and run" produces no check runs at all. This
+#     job is parked with them, so its check stays unreported, and the merge button stays disabled
+#     until a maintainer approves the runs AND the runs pass.
+
+on:
+  pull_request:
+
+# A new push supersedes the previous head commit, so the in-flight watcher for the old commit is
+# pointless. Cancelled, not left to burn a runner for an hour.
+concurrency:
+  group: required-checks-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  all-checks:
+    name: All Required Checks Passed
+    runs-on: ubuntu-latest
+    timeout-minutes: 75
+    permissions:
+      actions: read        # list the workflow runs for this commit
+      pull-requests: read  # list the PR's changed files
+      contents: read
+    steps:
+      # Deliberately no actions/checkout: nothing from the PR's tree is executed here, so a fork
+      # PR cannot change what this job does by editing a script in its own branch.
+      - name: Decide which workflows this PR must run
+        id: scope
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          # Two workflows are path-filtered, so they do not run on every PR and cannot be required
+          # unconditionally -- a docs-only PR would wait forever on a run that never starts. These
+          # globs are the *intersection* of the pull_request `paths:` in integration-tests.yml and
+          # security-ci.yaml: any file matching one of them triggers BOTH workflows, so when one is
+          # touched both are safe to demand. Deliberately narrower than either filter; being too
+          # narrow only means we require less, being too wide would deadlock the PR.
+          CODE_PATHS: |
+            *.go
+            *.java
+            *go.mod
+            *pom.xml
+        run: |
+          set -euo pipefail
+
+          files=$(gh api --paginate "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER/files" --jq '.[].filename')
+
+          code=false
+          while IFS= read -r pattern; do
+            [ -n "$pattern" ] || continue
+            for file in $files; do
+              case "$file" in $pattern) code=true; break 2 ;; esac
+            done
+          done <<< "$CODE_PATHS"
+
+          echo "code-changed=$code" >> "$GITHUB_OUTPUT"
+          echo "PR touches code paths: $code"
+
+      - name: Wait for every workflow on this commit
+        env:
+          GH_TOKEN: ${{ github.token }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+
+          # Workflows that trigger on every PR with no path filter.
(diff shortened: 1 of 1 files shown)


.github/workflows/required-checks.yml
```diff
@@ -0,0 +1,130 @@
+name: Required Checks
+
+# Require only "All Required Checks Passed" in the ruleset. Green when every PR
+# run for the commit finished success/skipped/neutral.
+
+on:
+  workflow_run:
+    workflows:
+      - "*"
```
mihir-datazip: GitHub only documents glob and `!` patterns for `branches`, `tags` and `paths`. Every `workflow_run.workflows` example in the docs is an exact name, and I could not find a thread where `"*"` works (https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run).

Please confirm once if '*' will work as expected here

.github/workflows/required-checks.yml
```diff
@@ -0,0 +1,130 @@
...
+      - "!Required Checks"
+    types: [requested, completed]
+
+concurrency:
+  group: required-checks-${{ github.event.workflow_run.head_sha }}
+  cancel-in-progress: true
```
mihir-datazip: This reopens the read-lag hole the fold below is written for. Two workflows completing seconds apart: the first gate run is cancelled, the survivor folds only its own payload, may still read the other run as in progress, posts pending, and no later event corrects it.

With `cancel-in-progress: false` the group serializes: the newest pending run executes after the running one, by which time the API has caught up. Gate runs take seconds, so the queueing is cheap.

.github/workflows/required-checks.yml
```diff
@@ -0,0 +1,130 @@
...
+
+# Require only "All Required Checks Passed" in the ruleset. Green when every PR
+# run for the commit finished success/skipped/neutral.
+
+on:
+  workflow_run:
+    workflows:
+      - "*"
+      - "!Required Checks"
+    types: [requested, completed]
```
mihir-datazip: Suggest dropping `requested`. Per the docs it does not fire on re-runs, and a missing check already blocks merge as "Expected", so all it buys is the badge reading "in progress" instead of "expected" during the first minute. It doubles gate runs and is the source of most of the concurrency churn.

.github/workflows/required-checks.yml
```diff
@@ -0,0 +1,130 @@
...
+          TRIGGER_CONCLUSION: ${{ github.event.workflow_run.conclusion }}
+        run: |
+          set -euo pipefail
+          : "${HEAD_SHA:?workflow_run payload carried no head SHA}"
+
+          runs=$(gh api "repos/${GITHUB_REPOSITORY}/actions/runs?head_sha=${HEAD_SHA}&per_page=100")
+
+          # Unpaginated so runs can be grouped in one pass; refuse rather than lose some.
+          total=$(jq -r '.total_count // 0' <<< "$runs")
+          if [ "$total" -gt 100 ]; then
```
mihir-datazip: The cap is partly self-inflicted: every gate run is itself an `event=workflow_run` run on this same SHA and counts toward `total_count` (two per source workflow per attempt while `requested` is on).

Either `gh api --paginate` and merge pages in jq, or pass `event=pull_request` server-side. Nothing in the repo uses `pull_request_target`, and adding one would be a deliberate security decision anyway, so the extra event handling in the jq could go with it.

.github/workflows/required-checks.yml
```diff
@@ -0,0 +1,130 @@
...
+            def approval: (.status | IN("action_required", "waiting")) or (.conclusion == "action_required");
+            def finished: (.status == "completed") and (approval | not);
+            def passing:  (.conclusion | IN("success", "skipped", "neutral"));
+
+            # Latest attempt per workflow, keyed by id so a rename is still the
+            # same workflow.
+            ( .workflow_runs
+              | map(select(.event == "pull_request" or .event == "pull_request_target"))
+              | group_by(.workflow_id)
+              | map(sort_by(.run_number, .run_attempt) | last)
```
mihir-datazip: Each run appears once in this list (the API reports its latest attempt), so the `run_attempt` tie-break does nothing. `max_by(.run_number)` reads clearer.

.github/workflows/required-checks.yml
```diff
@@ -0,0 +1,130 @@
...
+
+            | { path: env.TRIGGER_PATH, name: env.TRIGGER_NAME, status: env.TRIGGER_STATUS,
+                conclusion: (if (env.TRIGGER_CONCLUSION | length) > 0 then env.TRIGGER_CONCLUSION else "-" end)
+              } as $trigger
+
+            # Fold in the event payload, forward only: a run never un-completes,
+            # so an out-of-order `requested` cannot reopen one the API reports as
+            # finished. Covers read lag on the last completion, which no later
+            # event would correct.
+            | ( if ($trigger.path | length) == 0 then $api
```
mihir-datazip: The payload carries `workflow_run.workflow_id`, which is what the grouping above already keys on. Folding by id instead of `path` removes `TRIGGER_PATH`, the empty-path guard on this line, and the "empty" verdict below, which is unreachable once the trigger is always folded in.

.github/workflows/required-checks.yml
```diff
@@ -0,0 +1,130 @@
...
+          CHECK_NAME: All Required Checks Passed
+          EXTERNAL_ID: olake-required-checks
+          DETAILS_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          TRIGGER_PATH: ${{ github.event.workflow_run.path }}
+          TRIGGER_NAME: ${{ github.event.workflow_run.name }}
+          TRIGGER_STATUS: ${{ github.event.workflow_run.status }}
+          TRIGGER_CONCLUSION: ${{ github.event.workflow_run.conclusion }}
+        run: |
+          set -euo pipefail
+          : "${HEAD_SHA:?workflow_run payload carried no head SHA}"
```
mihir-datazip: Nit: `workflow_run` always carries `head_sha`; this guard can go.

.github/workflows/required-checks.yml
```diff
@@ -0,0 +1,130 @@
...
+
+          existing=$(gh api --paginate "repos/${GITHUB_REPOSITORY}/commits/${HEAD_SHA}/check-runs?per_page=100" \
+            --jq '.check_runs[] | select(.external_id == env.EXTERNAL_ID) | "\(.id) \(.status)"' \
+            | sort -n | tail -1)
+
+          method=POST
+          url="repos/${GITHUB_REPOSITORY}/check-runs"
+          # A completed check run cannot be reopened, so a flip back to pending
+          # needs a fresh one.
+          if [ -n "$existing" ] && { [ "${existing##* }" != completed ] || [ "$new_status" = completed ]; }; then
```
mihir-datazip: The pending flip on the POST branch already relies on a newer same-name check run superseding the older one for the required-check evaluation. If we lean on that, every publish can be a plain POST: drops the lookup call, the PATCH branch, `del(.head_sha)` and `EXTERNAL_ID`.

PR #1194, opened 2026-09-04
feat(mysql): support CDC without binlog_row_metadata=FULL

vikaxsh:
# Description
MySQL CDC currently requires binlog_row_metadata=FULL, which was added in MySQL 8.0.1, defaults to MINIMAL, and does not exist on 5.7 or any MariaDB version. That means we refuse CDC on 5.7, on MariaDB, and on every stock 8.0 server that hasn't been reconfigured — and on 5.7 we fail with a misleading sql: no rows in result set instead of a clear message.

This rebuilds the missing column metadata — names, ENUM/SET members, collations, signedness — from information_schema, cached per table. The binlog is still preferred when it carries the metadata; the fallback only kicks in when it doesn't. Column type bytes and ENUM/SET detection come from the mandatory part of TableMapEvent and are unchanged on both paths.

Decoding now goes through a single columnView resolved once per RowsEvent, so everything that differs between the two sources is confined to one constructor instead of six accessor calls spread through the row loop.

binlog_row_metadata drops from a hard requirement to a warning. log_bin=ON and binlog_format=ROW stay required.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -42,6 +42,7 @@ jobs:
       pull-requests: read
     outputs:
       drivers: ${{ steps.drivers.outputs.drivers }}
+      driver-matrix: ${{ steps.matrix.outputs.matrix }}
       jar-cached: ${{ steps.jar-cache.outputs.cache-hit }}
       go-cached: ${{ steps.go-cache.outputs.cache-hit }}
       go-checks-changed: ${{ steps.go-checks.outputs.changed }}
@@ -55,6 +56,25 @@ jobs:
       - name: Resolve driver matrix
         id: drivers
         uses: ./.github/actions/detect-drivers
+
+      # The driver matrix, plus the pseudo-driver "mysql-5.7". It is the mysql driver against a
+      # 5.7 server -- same module, same image, its own runner -- so it fans out beside the real
+      # drivers instead of extending the mysql job. 5.7 predates binlog_row_metadata, making it
+      # the only stack that covers the information_schema fallback in pkg/binlog. Added only when
+      # mysql is already in the matrix: a PR that touches nothing mysql should not boot a mysql
+      # stack to test it.
+      - name: Build driver matrix
+        id: matrix
+        env:
+          DRIVERS: ${{ steps.drivers.outputs.drivers }}
+        run: |
+          set -euo pipefail
+          matrix="$DRIVERS"
+          if jq -e 'index("mysql")' <<<"$DRIVERS" >/dev/null; then
+            matrix=$(jq -c '. + ["mysql-5.7"]' <<<"$DRIVERS")
+          fi
+          echo "matrix=$matrix" >> "$GITHUB_OUTPUT"
+          echo "Driver matrix: $matrix"
 
       - name: Detect go-check input changes
         id: go-checks
@@ -224,10 +244,36 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        driver: ${{ fromJSON(needs.preflight.outputs.drivers) }}
+        driver: ${{ fromJSON(needs.preflight.outputs.driver-matrix) }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v7
+
+      # Maps a matrix entry to the driver module it builds, the compose override it runs against
+      # and the suites it runs. Every step below reads these rather than matrix.driver, so the
+      # mysql-5.7 entry reuses the mysql image, the mysql buildx cache and the mysql test binary.
+      # EXTRA_COMPOSE_mysql lands in the job env, so every make call picks up the override without
+      # threading it through each one.
+      - name: Resolve driver variant
+        env:
+          MATRIX_DRIVER: ${{ matrix.driver }}
+        run: |
+          set -euo pipefail
+          case "$MATRIX_DRIVER" in
+            mysql-5.7)
+              echo "DRIVER=mysql" >> "$GITHUB_ENV"
+              echo "EXTRA_COMPOSE_mysql=drivers/mysql/docker-compose.5.7.yml" >> "$GITHUB_ENV"
+              # Sync only: it covers backfill, CDC and schema evolution over the fixture, which is
+              # where every fallback-sensitive type lives -- ENUM, SET, latin1/ucs2/utf16le columns
+              # and the unsigned maxima. Discover and 2PC are orthogonal to binlog metadata and
+              # already run on 8.0 in the sibling job.
+              echo "TEST_TARGET=test.sync.mysql" >> "$GITHUB_ENV"
+              ;;
+            *)
+              echo "DRIVER=$MATRIX_DRIVER" >> "$GITHUB_ENV"
+              echo "TEST_TARGET=test.integration.$MATRIX_DRIVER" >> "$GITHUB_ENV"
+              ;;
+          esac
 
       # The iceberg catalog's postgres runs as uid 999 and has to own its bind mount.
       - name: Set up Data Directories
@@ -244,7 +290,7 @@ jobs:
         id: containers
         background: true
         timeout-minutes: 15
-        run: make -j2 --output-sync=target olake.${{ matrix.driver }}.up olake.destination.all.up
+        run: make -j2 --output-sync=target olake.${{ env.DRIVER }}.up olake.destination.all.up
 
       - name: Set up Docker Buildx
         uses: ./.github/actions/buildx
@@ -258,7 +304,7 @@ jobs:
         id: image
         background: true
         timeout-minutes: 20
-        run: make docker.${{ matrix.driver }}.build DOCKER_BUILD="docker buildx build --progress=plain --cache-from type=gha,scope=olake-base --cache-from type=gha,scope=olake-${{ matrix.driver }} --load"
+        run: make docker.${{ env.DRIVER }}.build DOCKER_BUILD="docker buildx build --progress=plain --cache-from type=gha,scope=olake-base --cache-from type=gha,scope=olake-${{ env.DRIVER }} --load"
 
       # setup-go's cache keys on go.sum, which this repo does not track, so it would cache nothing.
       - name: Set up Go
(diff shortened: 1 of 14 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -42,6 +42,7 @@ jobs:
       pull-requests: read
     outputs:
       drivers: ${{ steps.drivers.outputs.drivers }}
+      driver-matrix: ${{ steps.matrix.outputs.matrix }}
```
mihir-datazip: rename this to `sources`

.github/workflows/integration-tests.yml
```diff
@@ -56,6 +57,25 @@ jobs:
         id: drivers
         uses: ./.github/actions/detect-drivers
 
+      # The driver matrix, plus the pseudo-driver "mysql-5.7". It is the mysql driver against a
+      # 5.7 server -- same module, same image, its own runner -- so it fans out beside the real
+      # drivers instead of extending the mysql job. 5.7 predates binlog_row_metadata, making it
+      # the only stack that covers the information_schema fallback in pkg/binlog. Added only when
+      # mysql is already in the matrix: a PR that touches nothing mysql should not boot a mysql
+      # stack to test it.
+      - name: Build driver matrix
```
mihir-datazip: Instead of doing this manually, just detect sources dynamically with multiple docker-compose and use the suffix in the docker-compose file to the sources matrix

Makefile
```diff
@@ -197,10 +197,17 @@ prepare.all: $(addprefix prepare.,$(DRIVERS))
 # at once, and `make -j olake.all.wait` collapses all the probes into one parallel
 # step, so slow boots (db2, spark) overlap with each other and with whatever
 # runs between the two -- what CI does.
+# The compose files one driver's stack is made of: its own, plus the override named by
+# EXTRA_COMPOSE_<d> when set. CI sets EXTRA_COMPOSE_mysql for the mysql-5.7 matrix entry, so
+# the same olake.mysql.* targets bring up a 5.7 server without a second set of targets; unset,
+# every stack stays a plain single-file one. Only files named exactly docker-compose.yml make a
+# directory a driver (see SOURCE_DRIVERS), so an override never becomes a driver of its own.
+compose_files = -f drivers/$(1)/docker-compose.yml $(if $(EXTRA_COMPOSE_$(1)),-f $(EXTRA_COMPOSE_$(1)))
```
mihir-datazip: This parameter should be called SOURCE_COMPOSE_FILE as we might also need similar thing for destination as well

.github/workflows/integration-tests.yml
```diff
@@ -276,30 +322,48 @@ jobs:
...
             -e TELEMETRY_DISABLED=true \
-            "olake/source-${{ matrix.driver }}:local" \
+            "olake/source-${{ env.DRIVER }}:local" \
             check --destination /testdata/iceberg_destination.json
 
-      # Integration + 2PC + (kafka) Rebalance in one go test, against the image built above.
+      # Fails loudly rather than silently re-running the 8.0 suite: every assertion in the suite
+      # passes against 8.0 too, so an override that stopped threading through must not look green.
+      # Runs only for the mysql-5.7 entry, the only one that sets an override.
+      - name: Verify server is 5.7 without binlog_row_metadata
```
mihir-datazip: This job can be abstracted in `driver.mk` instead of adding steps for driver version specific logic

pkg/binlog/filter.go
```diff
@@ -105,73 +108,173 @@ func (f ChangeFilter) FilterRowsEvent(ctx context.Context, e *replication.RowsEv
...
 	return nil
 }
 
-// convertRowToMap converts a binlog row to a map.
-func convertRowToMap(row []interface{}, tableMap *replication.TableMapEvent, columnTypes []string, converter func(value interface{}, columnType string) (interface{}, error)) (map[string]interface{}, error) {
-	columns := tableMap.ColumnNameString()
-	if len(columns) != len(row) {
-		return nil, fmt.Errorf("column count mismatch: expected %d, got %d", len(columns), len(row))
+// columnView is everything needed to decode one RowsEvent. Indexed by ordinal position.
+type columnView struct {
```
mihir-datazip, line 120: Can we somehow use single struct for []columnMeta and this?

vikaxsh: yeah we can make one struct but columnMeta models a row of the information_schema query, while the resolved view is that merged with the binlog's own metadata. some fields are only added after resolve column like Type, so for now we can keep as it is

Makefile
```diff
@@ -197,25 +202,40 @@ prepare.all: $(addprefix prepare.,$(DRIVERS))
 # at once, and `make -j olake.all.wait` collapses all the probes into one parallel
 # step, so slow boots (db2, spark) overlap with each other and with whatever
 # runs between the two -- what CI does.
+# The compose files one source stack is made of: the driver's own, plus the override named by
```
mihir-datazip: Cleanup old comments, also we can keep comments minimal as this is self-explanatory now

drivers/mysql/driver.mk
```diff
@@ -1,3 +1,27 @@
 # mysql fragment for the root Makefile (see the driver-fragment contract there).
 
 PROBE.mysql = docker exec olake_mysql-test mysql -h localhost -u root -proot1234 -e "SELECT 1"
+
+# Asserts the server that answered is the one the compose files asked for. MySQL's VERSION()
```
mihir-datazip: Same here, a lot of comments that is not really needed

.github/workflows/integration-tests.yml
```diff
@@ -56,6 +57,34 @@ jobs:
         id: drivers
         uses: ./.github/actions/detect-drivers
 
+      # Every source stack the suites run against: one entry per selected driver, plus one
```
mihir-datazip: No need to refer mysql here, it is generic step for all the drivers in future as well.
You can remove these comments as well as `Resolve source stacks` cleanly defines what we are doing

.github/workflows/integration-tests.yml
```diff
@@ -244,7 +307,7 @@ jobs:
         id: containers
         background: true
         timeout-minutes: 15
-        run: make -j2 --output-sync=target olake.${{ matrix.driver }}.up olake.destination.all.up
+        run: make -j2 --output-sync=target olake.${{ env.DRIVER }}.up olake.destination.all.up
```
mihir-datazip: I think you need to pass the compose file here right?

.github/workflows/integration-tests.yml
```diff
@@ -276,30 +339,38 @@ jobs:
...
+            "olake/source-${{ env.DRIVER }}:local" \
             check --destination /testdata/iceberg_destination.json
 
-      # Integration + 2PC + (kafka) Rebalance in one go test, against the image built above.
+      # Asserts the stack that came up is the one the variant asked for -- a compose override
+      # that stopped threading through would otherwise re-run the default server's suite and
+      # still go green. The rules are the driver's own (VERIFY_STACK.<d> in drivers/<d>/driver.mk);
+      # this step just calls the target, and it is a no-op for drivers that define none.
+      - name: Verify source stack
+        run: make olake.${{ env.DRIVER }}.verify
```
mihir-datazip: Ideally compose file should be passed here as well as it will pass even for wrong version

.github/workflows/integration-tests.yml
```diff
@@ -224,11 +253,45 @@ jobs:
...
+            echo "matrix entry '$MATRIX_DRIVER' resolves to neither a driver nor $override"
+            exit 1
+          fi
+
+          echo "DRIVER=$driver" >> "$GITHUB_ENV"
+          echo "EXTRA_COMPOSE_$driver=$override" >> "$GITHUB_ENV"
+          # Sync only. It covers backfill, CDC and schema evolution over the driver's fixture,
+          # which is where a server difference shows up; Discover and 2PC exercise no
+          # server-version-specific path and already run on the default stack in the sibling job.
+          echo "TEST_TARGET=test.sync.$driver" >> "$GITHUB_ENV"
```
mihir-datazip: We should get the compose file for this source in the matrix and use that in the downstream jobs instead

.github/workflows/integration-tests.yml
```diff
@@ -276,30 +339,38 @@ jobs:
...
+        run: make olake.${{ env.DRIVER }}.verify
+
+      # Integration + 2PC + (kafka) Rebalance in one go test, against the image built above --
+      # or, for the mysql-5.7 entry, the sync suite alone. See "Resolve driver variant".
       - name: Run tests
         env:
-          OLAKE_DRIVER_IMAGE: olake/source-${{ matrix.driver }}:local
-        run: make test.integration.${{ matrix.driver }}
+          OLAKE_DRIVER_IMAGE: olake/source-${{ env.DRIVER }}:local
+        run: make "$TEST_TARGET"
```
mihir-datazip: We should just call `test.integration.${{ matrix.driver }}` only, its better to have coverage for 2pc as well here.

.github/workflows/integration-tests.yml
```diff
@@ -224,11 +253,45 @@ jobs:
...
+          driver=${MATRIX_DRIVER%%-*}
+          suffix=${MATRIX_DRIVER#*-}
+          override="drivers/$driver/docker-compose.$suffix.yml"
+          if [ ! -f "$override" ]; then
+            echo "matrix entry '$MATRIX_DRIVER' resolves to neither a driver nor $override"
+            exit 1
+          fi
+
+          echo "DRIVER=$driver" >> "$GITHUB_ENV"
+          echo "EXTRA_COMPOSE_$driver=$override" >> "$GITHUB_ENV"
```
mihir-datazip: Where is EXTRA_COMPOSE used now?

.github/workflows/integration-tests.yml
vikaxsh: fixed

.github/workflows/integration-tests.yml
vikaxsh: used in make olake.mysql.up for passing extra compose

mihir-datazip: We should use the same name as in makefile to avoid confusion because  `EXTRA_COMPOSE_` gives a sense we are running more than one compose for source

Makefile
```diff
@@ -197,25 +199,34 @@ prepare.all: $(addprefix prepare.,$(DRIVERS))
 # at once, and `make -j olake.all.wait` collapses all the probes into one parallel
 # step, so slow boots (db2, spark) overlap with each other and with whatever
 # runs between the two -- what CI does.
+# The driver's compose file, plus the override named by EXTRA_COMPOSE_<d> when set -- CI passes
+# it per make call so the same olake.<d>.* targets bring up an alternate server. The destination
+# half is $(DEST_COMPOSE), which would grow a DEST_COMPOSE_FILE counterpart if it needs the same.
+SOURCE_COMPOSE_FILE = -f drivers/$(1)/docker-compose.yml $(if $(EXTRA_COMPOSE_$(1)),-f $(EXTRA_COMPOSE_$(1)))
```
mihir-datazip: I meant renaming this `EXTRA_COMPOSE_$(1)` to `SOURCE_COMPOSE_$(1)`

vikaxsh: done

mihir-datazip: image

Also can we remove passing of empty flags in CI steps so it remains small commands for default test runs?

vikaxsh: done

mihir-datazip (approved):
LGTM

PR #1202, opened 2026-09-07
feat: source type in telemetry sync completed event and more config conditions

vaibhav-datazip:
# Description

- `Sync Completed` telemetry now carries `source_type`, matching what `Sync Started` already sends.

- Postgres CDC wrapped `constants.ErrNonRetryable` twice on the LSN validation path
   -  once inside `validateGlobalState`
   -  again at the caller 
```   
producing `failed with non retryable error: invalid global state: failed with non retryable error: lsn mismatch...`.
```

The sentinel now lives only at the call site (and uses `%w` instead of `%s`).

- Added `errs.Precondition` classification to every check in every driver's `Config.Validate` (postgres, mysql, mssql, mongodb, oracle, kafka, db2, s3) plus the shared SSL/PEM validators, and to the postgres CDC state checks. These previously reported `unclassified` with `error_type: *errors.errorString` in the failure event, since the chain walk can only surface evidence that already exists in the error.
- 
Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/utils/ssl.go
+++ b/utils/ssl.go
@@ -37,8 +37,6 @@ type SSLConfig struct {
 	ClientKey  string `mapstructure:"client_key,omitempty" json:"client_key,omitempty" yaml:"client_key,omitempty"`
 }
 
-// codeSSLConfigInvalid marks an ssl block the user filled in wrong; no server was reached.
-// One code per condition: two checks a user would fix differently never share one.
 const (
 	// The ssl block itself is missing a field.
 	codeSSLMissing         = "config.ssl_missing"
--- a/utils/ssl.go
+++ b/utils/ssl.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/datazip-inc/olake/utils/errs"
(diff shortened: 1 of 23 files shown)


mihir-datazip (dismissed):

utils/ssl.go
```diff
@@ -36,20 +37,46 @@ type SSLConfig struct {
 	ClientKey  string `mapstructure:"client_key,omitempty" json:"client_key,omitempty" yaml:"client_key,omitempty"`
 }
 
+// codeSSLConfigInvalid marks an ssl block the user filled in wrong; no server was reached.
```
mihir-datazip: codeSSLConfigInvalid does not exist, update the comment

vaibhav-datazip: removed other unnecessary comments as well, and improved the mssql code handling for figuring out the error easily

mihir-datazip (approved):

PR #1220, opened 2026-09-09
Fix/s3 same second incremental

### LINKED ISSUE #1131: issue: the driver should handle same-second arrivals in s3 source
## Summary

The S3 driver's incremental cursor is a **whole-second timestamp compared with a strict `>`**. Any object whose `LastModified` falls in the same second as the cursor is skipped — permanently, silently, and with exit code 0. The sync reports success and reads 0 records.

`tests/s3/s3_util_test.go` currently works around this by sleeping 1s before every upload that a later incremental sync is supposed to pick up. That guard makes the IT deterministic but hides a real product bug; **the driver should close the window itself.**

## Mechanism

Three pieces combine:

**1. Discovery truncates the timestamp to whole seconds.** `objstorage.ObjectInfo.LastModified` is a `time.Time` and both S3 and MinIO expose millisecond precision, but the driver formats it away:

https://github.com/datazip-inc/olake/blob/0f78e5212940990895542f9ba7a5caaf70b3259e/drivers/s3/internal/s3.go#L144

So the collision window is not a nanosecond tie — it is up to a **full second wide**.

**2. The cursor is seeded from the max of those truncated values.** `FetchMaxCursorValues` returns `max(LastModified)` over every discovered file in the stream, and the abstract layer stores it as the stream cursor *before* backfill runs:

https://github.com/datazip-inc/olake/blob/0f78e5212940990895542f9ba7a5caaf70b3259e/drivers/s3/internal/incremental.go#L28-L39
https://github.com/datazip-inc/olake/blob/0f78e5212940990895542f9ba7a5caaf70b3259e/drivers/abstract/incremental.go#L37-L43

**3. The incremental filter is a strict string `>`.** Equality means "already synced":

https://github.com/datazip-inc/olake/blob/0f78e5212940990895542f9ba7a5caaf70b3259e/drivers/s3/internal/incremental.go#L129-L141

A file stamped exactly at the cursor second was never actually read — it just happens to compare equal.

## Why the loss is permanent, not a one-sync delay

After the first sync, `HasCompletedBackfill` is true and the cursor is non-nil, so subsequent runs skip backfill entirely and go straight to `StreamIncrementalChanges`. The cursor only ever advances from the `_last_modified_time` value of records that were actually **read** (`getMaxIncrementCursorFromData`); a skipped file contributes nothing. So:

- if no newer file arrives, the cursor stays put and the file is skipped again on every future sync;
- if a newer file *does* arrive, the cursor moves **past** the missed file, and it is skipped again on every future sync.

Either way the file is never read. There is no error, no warning, no non-zero exit — only a `Debugf` line saying the file was skipped.

## Reproduction

```
t0.100  PUT s3://bucket/stream/seed_1.csv        -> LastModified 10:00:00.100 -> "10:00:00Z"
t0.200  sync #1: discover lists seed_1
                 cursor := "2026-08-18T10:00:00Z"
                 backfill reads seed_1                                    ✅ 3 rows
t0.700  PUT s3://bucket/stream/insert_1.csv      -> LastModified 10:00:00.700 -> "10:00:00Z"
t5.000  sync #2: filter "10:00:00Z" > "10:00:00Z" -> false
                 "no new files to process (all files <= cursor)"          ❌ 0 rows, exit 0
t9.000  sync #3..N: identical                                             ❌ never read
```

`insert_1.csv` is gone from the pipeline forever.

## Blast radius

- **Affected:** every S3 source in `incremental` mode, in every release. `_last_modified_time` is the only cursor field the driver registers, so there is no alternative cursor a user can pick to avoid this.
- **Not affected:** `full_refresh` (re-reads everything), and any state where backfill did not complete (the abstract layer re-seeds the cursor and re-runs backfill in that case).
- **Real-world likelihood is not negligible.** Any producer that writes a batch of objects and a sync that lands mid-batch hits it; upstream jobs that flush several files at once are the common shape. Destination behaviour makes it worse to notice: with Iceberg the previously synced rows are still there so the table just looks stale, and with the Parquet destination the sync writes no files at all.
- **Observed in CI:** `TestS3Compat/*/compare/pq_inc` flaked with "only one run produced parquet files" against v0.9.2 on warm runs (~1s per case). Cold runs (~7s+) never reproduced it, which is exactly the second-boundary signature.

## Current workaround (to be removed by the fix)

https://github.com/datazip-inc/olake/blob/0f78e5212940990895542f9ba7a5caaf70b3259e/tests/s3/s3_util_test.go#L810-L830

`putFile` takes a `wait bool`, and the three uploads that a later sync must observe — `insert_1`, `update_1`, `evolve_1` — pass `true` and sleep a second first. This costs ~3s per S3 variant and, more importantly, means the IT cannot catch a regression in this area.

## Proposed fix
We should re think the better way to track increments which should ideally not depend on timestamps. But if there is none we might need to do an equality check instead of strict `>` timestamp and then figure out what data needs to be synced from here onwards

## Acceptance criteria
- [ ] A file uploaded in the same second as the previous sync's cursor is read by the next incremental sync.
- [ ] That file is read **exactly once** — a third sync with no new files reads 0 records (no duplicates from the `>=`).
- [ ] Unit coverage in `drivers/s3/internal` for the filter: strictly-newer, same-second-unseen, same-second-seen, and the legacy-state fallback.
- [ ] The `wait bool` parameter, the three `true` call sites, and the TODO are removed from `tests/s3/s3_util_test.go`, and `TestS3Integration` passes warm.

---

https://github.com/datazip-inc/olake/blob/0f78e5212940990895542f9ba7a5caaf70b3259e/tests/s3/s3_util_test.go#L822-L827

---

iSparshP:
# Description

Fixes #1131

OLake's S3 incremental source can silently skip files that arrive in the same second as the current incremental cursor.

S3 `LastModified` timestamps are exposed at second-level precision, while the existing incremental filter used a strict `>` comparison. As a result, a newly arrived file with the same timestamp as the cursor was considered already processed and was skipped permanently.

This change fixes the issue by:

- Changing the incremental timestamp comparison to `>=`.
- Tracking the file keys already synced at the current cursor timestamp.
- Reading same-second files only when their keys have not already been processed.
- Seeding and advancing the synced-key state during backfill and incremental sync.
- Removing the 1-second sleep workaround from the S3 integration tests.

This ensures that each file is processed exactly once, including files arriving within the same second.

No interface changes are required and other drivers are unaffected.

## Type of change

# How Has This Been Tested?

The following tests were run successfully:

- [x] Unit tests covering strictly-newer, same-second-unseen, same-second-seen, and legacy-state scenarios.
- [x] S3 end-to-end integration tests covering full-load and incremental insert/update flows.

### Unit tests

```bash
cd drivers/s3
go test ./internal/ -run 'TestIncrementalSyncFiltering|TestCursorTimestampComparison' -v

### DIFF
--- a/drivers/s3/internal/incremental.go
+++ b/drivers/s3/internal/incremental.go
@@ -14,6 +14,48 @@ import (
 // ============================================
 // This file contains all incremental sync logic using cursor-based filtering
 
+// syncedKeysField is a companion state entry to the LastModified cursor. It
+// holds the file keys already synced whose LastModified equals the cursor's
+// second. S3 (and the S3 List API in general) only exposes LastModified at
+// whole-second granularity, so a same-second collision window is up to a full
+// second wide and cannot be disambiguated by timestamp alone. Remembering the
+// keys already read at that second lets us use a `>=` filter and still read
+// each file exactly once. See issue #1131.
+const syncedKeysField = "_last_modified_time_synced_keys"
+
+// syncedKeysFromState reconstructs the synced-key set from state. The value is
+// stored as []string in-process but round-trips through JSON as []any across
+// runs, so both are handled. A missing value (legacy state written before this
+// fix) yields an empty set, which degrades to reading same-second boundary
+// files once on the first post-upgrade sync.
+func syncedKeysFromState(v any) map[string]bool {
+	set := map[string]bool{}
+	switch keys := v.(type) {
+	case []string:
+		for _, k := range keys {
+			set[k] = true
+		}
+	case []any:
+		for _, k := range keys {
+			if s, ok := k.(string); ok {
+				set[s] = true
+			}
+		}
+	}
+	return set
+}
+
+// keysAtSecond returns the keys of files whose LastModified equals second.
+func keysAtSecond(files []FileObject, second string) []string {
+	var keys []string
+	for _, file := range files {
+		if file.LastModified == second {
+			keys = append(keys, file.FileKey)
+		}
+	}
+	return keys
+}
+
 // FetchMaxCursorValues returns the maximum LastModified timestamp for all files in the stream
 // This is used by the abstract layer to track incremental sync progress
 func (s *S3) FetchMaxCursorValues(_ context.Context, stream types.StreamInterface) (any, any, error) {
@@ -31,6 +73,13 @@ func (s *S3) FetchMaxCursorValues(_ context.Context, stream types.StreamInterfac
 		if file.LastModified > maxLastModified {
 			maxLastModified = file.LastModified
 		}
+	}
+
+	// Seed the synced-key set with every file at the max second. Backfill reads
+	// all discovered files, so those keys are covered once backfill completes;
+	// this stops the first incremental sync from re-reading them.
+	if configuredStream, ok := stream.(*types.ConfiguredStream); ok && s.state != nil {
+		s.state.SetCursor(configuredStream, syncedKeysField, keysAtSecond(files, maxLastModified))
 	}
 
 	logger.Infof("Max cursor value for stream %s: %s (from %d files)", streamName, maxLastModified, len(files))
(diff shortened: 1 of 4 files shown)


nayanj98:
@iSparshP Can you please upload the video else we will not be considering this PR

nayanj98:
Closing this PR as the video requirement was not fulfilled.

PR #1225, opened 2026-09-09
fix: give NonRetryable sync failures a distinct exit code and log line

### LINKED ISSUE #1196: NonRetryable CDC failures give process supervisors no way to distinguish them from transient errors
## Environment

- Olake version: v0.9.5 (commit `1005110bab4522a132a120ebd8064d147dda63f1`, also present on current `master` as of this report)
- Source: Postgres, CDC (`pgoutput`)
- Destination: Iceberg, REST catalog (AWS S3 Tables), legacy gRPC/Java writer
- Deployment pattern: `olake sync ...` invoked repeatedly as a fresh process from inside a long-lived container (`while true; do ./olake sync ...; sleep 10; done`), state on a PersistentVolumeClaim so progress survives across invocations.

## Related

This came up while hitting the `validateGlobalState` fatal described in #853, #854, #998, and #1084 (I left field data on #321, the open discussion about fail-vs-full-load policy, since it's directly relevant there). This issue is scoped to something separate from all of those: none of them address whether an operator running `olake sync` under a process supervisor can tell this class of failure apart from a normal one.

## Description

`constants.ErrNonRetryable` is checked internally, `utils.RetryOnBackoff` stops its own retry loop when it sees this error, which shows it's meant to require a human to step in rather than be retried. But nothing about the process's exit code or log format signals that externally. It looks like any other sync error.

We run `olake sync` from inside a shell loop that retries on any non-zero exit, which is a completely ordinary way to run a long-lived sync job (the same shape as a Kubernetes restart policy or systemd `Restart=always`). When we hit a `NonRetryable` LSN mismatch, that loop retried it every 10 seconds for about 11 hours, because there was nothing to catch and treat differently. The failure never restarted the container, so `kubectl get pods` showed `Running`, `0 restarts`, the entire time. We only found out because an unrelated external monitor was watching replication-slot WAL retention on the source and caught the growth.

## Ask

Give `NonRetryable` failures a signal an external supervisor can actually act on: a distinct process exit code, or at minimum a clearly-greppable log line (something like `FATAL: manual intervention required`, consistently formatted) that a supervisor or alerting rule can match on to stop retrying and page a human, instead of silently looping.

Happy to share more detail on our exact deployment if useful.

pujitha24:
# Description

`olake sync` exits with the same generic code (1) and no distinguishing log line whether a
failure is transient or a `constants.ErrNonRetryable` failure that needs a human (e.g. a
Postgres LSN mismatch requiring `clear destination`). A process supervisor running `olake sync`
in a retry loop (shell loop, systemd `Restart=always`, a Kubernetes restart policy) can't tell
the two apart, so it keeps retrying a failure that will never succeed.

This adds a distinct exit code (`2`) and a consistently-formatted, greppable log line
(`FATAL: manual intervention required: ...`) for `ErrNonRetryable` failures, so a supervisor can
match on either and stop retrying instead of looping forever.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] `go build ./constants/... ./utils/... ./protocol/... .` — the root package (`.`) is
      `connector.go`'s package and pulls in the full `destination/iceberg` +
      `destination/parquet` dependency tree; builds clean.
- [x] `go vet` on the same packages — clean.
- [x] `go test ./constants/... ./utils/... ./protocol/... .` — all pass, including a new
      `TestIsNonRetryable` covering: nil error, an unrelated error, an error wrapped with `%w`,
      and an error wrapped with `%s` (as `drivers/postgres/internal/cdc.go` already does at the
      `invalid global state` site), confirming the exit-code path is reachable even where the
      wrap chain doesn't survive `errors.Is`.
- [x] `gofmt -l` on every changed file — clean.
- [x] `golangci-lint` v1.64.8 (the version this repo's own `make olake.lint` / CI installs, per
      `.github/workflows/golang-ci.yml`, confirmed by reading the CI job's own install log) run
      against the changed packages — clean, no findings.
- [ ] Not run: the full `go build ./...` across every driver module (mongodb, oracle, mysql,
      db2, mssql, kafka, s3) — the sandbox this was built in ran out of local disk space partway
      through and could not fit the full dependency set. This change only touches shared code
      (`constants`, `utils`, `protocol`, `connector.go`); no changed function's signature is used
      differently by any driver, and every driver's `main()` calls the same unmodified
      `olake.RegisterDriver` entry point, so no driver-specific breakage is expected, but this
      specific check could not be executed here.

## Notes for reviewers

- Traced `constants.ErrNonRetryable` end to end: e.g.
  `drivers/postgres/internal/cdc.go` → `protocol/sync.go`'s `RunE` → `RootCmd.Execute()` →
  `connector.go`'s `RegisterDriver`. The error message substring survives every layer, so the new
  exit-code check fires for real non-retryable failures from any driver (Postgres, Kafka, etc. —
  every call site that wraps `constants.ErrNonRetryable` funnels through this one entry point).
- Two pre-existing things this PR does not touch, flagged for visibility since they affect how
  much weight the new signal should carry:
  - `drivers/abstract/cdc.go` wraps *any* error from backfill (including plausibly-transient
    ones) in `constants.ErrNonRetryable` unconditionally. That was cosmetic before (both paths
    hit the same generic exit code 1); after this change such a failure will now also produce
    the new "manual intervention required" exit code 2, which may be premature for a transient
    cause. Out of scope here since it's a separate, pre-existing tagging issue.
  - A panic recovered by `safego.Recovery` still always exits with code 1, regardless of whether
    the underlying cause would have been non-retryable — this PR is scoped to the
    `ErrNonRetryable` return-value path only, as the issue asks for.

# Screenshots or Recordings

N/A — this is a process exit-code / log-line change, no UI surface.

## Documentation

## Related PR's (If Any):

None.

Fixes #1196

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -7,6 +7,7 @@ import (
 	_ "github.com/datazip-inc/olake/destination/parquet" // registering parquet destination
 	"github.com/datazip-inc/olake/drivers/abstract"
 	protocol "github.com/datazip-inc/olake/protocol"
+	"github.com/datazip-inc/olake/utils"
 	"github.com/datazip-inc/olake/utils/logger"
 	"github.com/datazip-inc/olake/utils/safego"
 )
@@ -18,6 +19,12 @@ func RegisterDriver(driver abstract.DriverInterface) {
 	err := protocol.CreateRootCommand(true, driver).Execute()
 	if err != nil {
 		protocol.ReportFailure(err)
+		// constants.ErrNonRetryable means retrying will not fix this: give a process supervisor
+		// (shell retry loop, systemd, Kubernetes restart policy) a distinct exit code and a
+		// greppable log line so it can stop retrying and page a human instead of looping forever.
+		if utils.IsNonRetryable(err) {
+			logger.FatalNonRetryable(err)
+		}
 		logger.Fatal(err)
 	}
 
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -89,6 +89,12 @@ var ParallelCDCDrivers = []DriverType{MongoDB, MSSQL}
 var ErrNonRetryable = fmt.Errorf("failed with non retryable error")
 var ErrGlobalContextGroup = fmt.Errorf("global context group error")
 
+// ExitCodeNonRetryable is the process exit code used when a sync fails with ErrNonRetryable,
+// so a process supervisor (shell retry loop, systemd Restart=always, Kubernetes restart policy)
+// can tell a failure that needs a human apart from one worth retrying, instead of looping on it
+// forever the way a generic exit code 1 would.
+const ExitCodeNonRetryable = 2
+
 // DriversRequiringIncrementalFormatter are drivers that require special formatting for incremental value
 var DriversRequiringIncrementalFormatter = []DriverType{Oracle, DB2, MSSQL}
 
--- a/utils/logger/logger.go
+++ b/utils/logger/logger.go
@@ -67,6 +67,18 @@ func Fatalf(format string, v ...interface{}) {
 	os.Exit(1)
 }
 
+// FatalNonRetryable logs a consistently-formatted, greppable FATAL line for a failure that
+// requires manual intervention (constants.ErrNonRetryable) and exits with
+// constants.ExitCodeNonRetryable instead of the generic exit code 1 used for ordinary failures.
+// This lets a process supervisor (shell retry loop, systemd, Kubernetes restart policy) match on
+// the exit code or the log line and stop retrying instead of looping on an error retrying can
+// never fix. zerolog's Fatal level always exits with code 1 itself, so this logs at Error level
+// and exits explicitly.
+func FatalNonRetryable(err error) {
+	logger.Error().Msgf("FATAL: manual intervention required: %s", err)
+	os.Exit(constants.ExitCodeNonRetryable)
+}
+
 // Error writes record into os.stdout with log level ERROR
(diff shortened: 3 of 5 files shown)


nayanj98:
@pujitha24 Can you please request for assigning of the issue on the issue itself so that I can assign you the issue. Also we have rebased your branch to staging, please make sure to please merge into the same base branch instead of master.

pujitha24:
Thanks for the heads up. The branch is already based on staging (see the merge commit in the history), so that should be all set on this end.

I'll head over to issue #1196 and leave a comment requesting the assignment there.

nayanj98:
@pujitha24 Yes please drop a comment on the issue so that I can quickly assign and others also know that you are working on this issue! Once this is done I will assign a reviewer for the PR.

pujitha24:
Sorry, I said I'd drop a comment on #1196 and never followed through — my mistake. Turns out I don't have a posting flow set up for anything outside this PR thread, just for replies here, so I can't get it onto the issue itself from where I'm sitting. This PR is the implementation for #1196, so hopefully that's enough to go on for assignment; if not, could someone add the comment on your end, or let me know another way to get it there?

nayanj98:
@pujitha24 Assigning @vaibhav-datazip to review your PR

vaibhav-datazip:
@pujitha24 the pr and even the replies look like bot generated, we dont accept contributions of completely bot generated pr. please close this pr

pujitha24:
Understood, thanks for taking a look — closing this out.

PR #1229, opened 2026-09-10
feat: derive stream delete formats from target query engines

vikaxsh:
# Description

Lets a client declare which engines will read the destination Iceberg tables, and
derives each stream's delete format from that instead of always defaulting to `eq`.

Equality deletes are cheapest (no index bootstrap) but only a few engines read them;
positional deletes are the universal denominator. Picking by hand meant picking wrong.

- `spec --available-query-engines` serves the read-capability matrix: 13 engines, the
  delete formats each resolves, and what OLake can write. The engine names it returns are
  exactly the values `--target-query-engines` accepts, locked by a round-trip test.
- `discover --target-query-engines spark,duckdb,hive` intersects those capabilities with
  what OLake can write and records the result per stream as `available_update_types`
  (cheapest first), seeding `update_type` from element 0.
- The engines are a pure discover-time input, never persisted. Only the derived list is
  written, so a caller passes the flag on every discover.
- Re-discover re-picks any stream whose stored `update_type` the current engines can't
  read, and logs the flip. Catalogs without engines behave exactly as before.
- Sync validates `update_type` against the stream's recorded list, catching a hand-edited
  streams file.

`dv` stays out of the writable set until the deletion-vector writer lands.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):

### DIFF
--- a/types/catalog.go
+++ b/types/catalog.go
@@ -3,10 +3,12 @@ package types
 import (
 	"fmt"
 	"reflect"
+	"slices"
 	"strings"
 
 	"github.com/datazip-inc/olake/constants"
 	"github.com/datazip-inc/olake/utils"
+	"github.com/datazip-inc/olake/utils/logger"
 )
 
 // Message is a dto for olake output row representation
@@ -80,14 +82,22 @@ type StreamMix struct {
 	StreamWithPosUpdateType int `json:"stream_with_pos_update_type_count"`
 }
 
-func GetWrappedCatalog(streams []*Stream, driver string) *Catalog {
+func GetWrappedCatalog(streams []*Stream, driver string, engines []QueryEngine) *Catalog {
 	catalog := &Catalog{
 		Streams:         []*ConfiguredStream{},
 		SelectedStreams: make(map[string][]StreamMetadata),
 	}
+	// The default delete format is the cheapest one every target engine can read.
+	available := AvailableUpdateTypes(engines)
+	updateType := PreferredUpdateType(available)
 
 	// Loop through each stream and populate Streams and SelectedStreams
 	for _, stream := range streams {
+		stream.AvailableUpdateTypes = available
+		if stream.DefaultStreamProperties != nil {
+			stream.DefaultStreamProperties.UpdateType = updateType
+		}
+
 		// Create ConfiguredStream and append to Streams
 		catalog.Streams = append(catalog.Streams, &ConfiguredStream{
 			Stream: stream,
@@ -103,7 +113,7 @@ func GetWrappedCatalog(streams []*Stream, driver string) *Catalog {
 			StreamName:      stream.Name,
 			AppendMode:      utils.Ternary(driver == string(constants.Kafka), true, false).(bool),
 			Normalization:   IsDriverRelational(driver),
-			UpdateType:      string(UpdateTypeEquality),
+			UpdateType:      string(updateType),
 			SelectedColumns: selectedCols,
 		})
 	}
@@ -116,7 +126,7 @@ func GetWrappedCatalog(streams []*Stream, driver string) *Catalog {
 // 2. SelectedColumns: Retain columns present in both old and new schemas, add NEW columns if sync_new_columns is true
 // 3. SyncMode: Use from oldCatalog if the stream exists in old catalog
 // 4. Everything else: Keep as new catalog
-func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
+func mergeCatalogs(oldCatalog, newCatalog *Catalog, engines []QueryEngine) *Catalog {
 	if oldCatalog == nil {
 		return newCatalog
 	}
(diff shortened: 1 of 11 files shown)


types/query_engine.go
```diff
@@ -0,0 +1,139 @@
...
+	{Engine: QueryEngineTrino, Label: "Trino", Supports: []UpdateType{UpdateTypeEquality, UpdateTypePosition, UpdateTypeDeletionVector}},
+	{Engine: QueryEnginePresto, Label: "Presto", Supports: []UpdateType{UpdateTypeEquality, UpdateTypePosition}},
+	{Engine: QueryEngineStarRocks, Label: "StarRocks", Supports: []UpdateType{UpdateTypeEquality, UpdateTypePosition}},
+	{Engine: QueryEngineHive, Label: "Apache Hive", Supports: []UpdateType{UpdateTypePosition}},
+	{Engine: QueryEngineDuckDB, Label: "DuckDB", Supports: []UpdateType{UpdateTypePosition}},
+	{Engine: QueryEngineAthena, Label: "AWS Athena", Supports: []UpdateType{UpdateTypePosition}},
+	{Engine: QueryEngineSnowflake, Label: "Snowflake", Supports: []UpdateType{UpdateTypePosition}},
+	{Engine: QueryEngineBigQuery, Label: "Google BigQuery", Supports: []UpdateType{UpdateTypePosition}},
+	{Engine: QueryEngineDatabricks, Label: "Databricks", Supports: []UpdateType{UpdateTypePosition, UpdateTypeDeletionVector}},
+	{Engine: QueryEngineDremio, Label: "Dremio", Supports: []UpdateType{UpdateTypePosition}},
```
hash-data, line 62: cant we have map of updateType -> query engine support, which will be easy

types/catalog.go
```diff
@@ -80,14 +82,22 @@ type StreamMix struct {
...
+	available := AvailableUpdateTypes(engines)
+	updateType := PreferredUpdateType(available)
 
 	// Loop through each stream and populate Streams and SelectedStreams
 	for _, stream := range streams {
+		stream.AvailableUpdateTypes = available
+		if stream.DefaultStreamProperties != nil {
+			stream.DefaultStreamProperties.UpdateType = updateType
+		}
+
```
hash-data, line 100: why we cant do it in abstract? where discover is being populating streams

types/query_engine.go
```diff
@@ -0,0 +1,139 @@
...
+	Label  string      `json:"label"`
+	// Supports are the delete formats the engine resolves at read time, independent of
+	// what OLake can write (writableUpdateTypes).
+	Supports []UpdateType `json:"supports"`
+}
+
+// writableUpdateTypes are the delete formats OLake can produce, cheapest first: equality
+// needs no index, positional needs a full identifier -> RowLocation index. Deletion
+// vectors are absent until the dv writer lands.
+var writableUpdateTypes = []UpdateType{UpdateTypeEquality, UpdateTypePosition}
```
hash-data: add dv as well

types/catalog.go
```diff
@@ -187,6 +198,36 @@ func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
...
+	available := AvailableUpdateTypes(engines)
+	if metadata.UpdateType != "" && slices.Contains(available, UpdateType(metadata.UpdateType)) {
+		return
+	}
+
+	preferred := PreferredUpdateType(available)
+	if preferred == "" {
+		// No format satisfies every engine; discover fails before reaching here, so leave
+		// the value untouched rather than blanking a working configuration.
+		return
```
hash-data: dont we need to make it empty so that ui fails? we need to make sure ui fails in this case

types/query_engine.go
```diff
@@ -0,0 +1,139 @@
...
+	return available[0]
+}
+
+func readableByAll(engines []QueryEngine, updateType UpdateType) bool {
+	for _, engine := range engines {
+		spec, found := lookupEngine(engine)
+		if !found || !slices.Contains(spec.Supports, updateType) {
+			return false
+		}
+	}
```
hash-data, line 124: check can be found && slices contains

types/catalog.go
```diff
@@ -187,6 +198,36 @@ func mergeCatalogs(oldCatalog, newCatalog *Catalog) *Catalog {
...
+	if len(engines) == 0 {
+		return
+	}
+
+	available := AvailableUpdateTypes(engines)
+	if metadata.UpdateType != "" && slices.Contains(available, UpdateType(metadata.UpdateType)) {
+		return
+	}
+
+	preferred := PreferredUpdateType(available)
```
hash-data: so we are setting eq if eq and pos both supported

types/query_engine.go
vikaxsh: on ui we need to list query enginnes

types/catalog.go
vikaxsh: abstract doesn't know about query engines and it is dest concern

types/catalog.go
vikaxsh: done

types/catalog.go
vikaxsh: default is eq everywhere

types/query_engine.go
vikaxsh: we can keep current version

PR #1233, opened 2026-09-16
fix: keep microsecond precision for timestamp in legacy writer

### LINKED ISSUE #1188: Iceberg legacy writer truncates timestamptz precision from microseconds to milliseconds
## Environment

- Olake version: v0.9.5 (commit `1005110bab4522a132a120ebd8064d147dda63f1`, also present on current `master` as of this report)
- Destination: Iceberg, REST catalog (AWS S3 Tables), normalized (typed-column) write mode, legacy gRPC/Java writer
- Source: Postgres, CDC (pgoutput)

## Description

Every `timestamptz` column written through the Iceberg legacy writer's normalized path loses sub-millisecond precision. Postgres stores `timestamptz` at microsecond resolution, and Iceberg's own `timestamptz` logical type is defined at microsecond resolution too, but the value that actually reaches Iceberg has been rounded down to the millisecond.

Confirmed on real data: the same row read from Postgres directly vs. read back from the synced Iceberg table shows the timestamp truncated, e.g. Postgres `2023-05-18 03:52:45.573605+00` arrives in Iceberg as `2023-05-18 03:52:45.573000+00`. The trailing 605 microseconds are gone, every time, on every table with a `timestamptz` column we've checked.

## Root cause (read from source, not just observed)

`destination/iceberg/legacy-writer/writer.go`, `toProtoFieldValue()`:

```go
case "timestamptz":
	t, err := typeutils.ReformatDate(val, true)
	if err != nil {
		return nil, fmt.Errorf("failed to reformat rawValue[%v] of type[%T] as timestamp value: %s", val, val, err)
	}
	return &proto.IcebergPayload_IceRecord_FieldValue{
		Value: &proto.IcebergPayload_IceRecord_FieldValue_LongValue{LongValue: t.UnixMilli()},
	}, nil
```

`Write()` (same file) calls this once per `timestamptz` field for every record in the normalized-mode path (the mode used whenever the destination schema carries typed columns rather than a single stringified-JSON blob):

```go
fv, err := toProtoFieldValue(field.IceType, val)
```

`val` here is the CDC-decoded value for that field. For a `timestamptz` column, that's a `time.Time` with full microsecond precision, coming straight off pgoutput's timestamp decode (Postgres itself doesn't store sub-microsecond precision, so this is already the ceiling). `ReformatDate(val, true)` passes it straight through unchanged for the `time.Time` case (`utils/typeutils/reformat.go:168-169`, `case time.Time: return v, nil`), so precision is intact right up to the `t.UnixMilli()` call. That call is where it's lost. `time.Time.UnixMilli()` truncates to millisecond resolution by construction; there is no `UnixMicro()` or nanosecond variant used anywhere in this switch.

For comparison, every other numeric case in the same switch (`int`, `long`, `float`, `double`) round-trips full precision via `typeutils.Reformat*`. `timestamptz` is the only case that discards precision on the way out.

## Steps to reproduce

1. Configure a Postgres CDC stream on a table with a `timestamptz` column whose values carry non-zero microseconds (e.g. `2026-01-01 00:00:00.123456+00`).
2. Point it at an Iceberg REST destination using the legacy (Java gRPC) writer in normalized mode.
3. Run `sync`, then read the same row back from Iceberg.

## Expected behavior

The `timestamptz` value round-trips at full (microsecond) precision, matching Iceberg's own `timestamptz` type definition and what the Parquet legacy path already achieves elsewhere in this codebase (`types/type_schema.go`'s `ToParquet()` declares these fields as `parquet.Timestamp(parquet.Microsecond)`).

## Actual behavior

Every `timestamptz` value is silently rounded down to the millisecond on write. No error, no warning. The value is simply wrong at the microsecond level from the very first sync.

## Suggested fix

Replace `t.UnixMilli()` with a microsecond-resolution equivalent (`t.UnixMicro()`) in the `case "timestamptz":` branch of `toProtoFieldValue()`. I don't have visibility into what the Java-side gRPC receiver assumes about the unit of this `LongValue` for a `timestamptz` field, so I can't confirm `UnixMicro()` is a complete fix on its own without also checking or adjusting the receiving side. The Go-side call site is the definite, source-confirmed origin of the precision loss; the receiver-side unit contract needs a look from whoever owns that code.

Happy to test a fix against our setup if that's useful.

faridkotia:
# Description

The Iceberg legacy (Java gRPC) writer floored every `timestamptz` to millisecond precision. Go encoded the value as epoch millis (`t.UnixMilli()`) and the Java decoder read it back with `Instant.ofEpochMilli`. Both sides used the same wrong unit, so nothing errored — the value was silently truncated: `2023-05-18 03:52:45.573605+00` → `2023-05-18 03:52:45.573000+00`.

Postgres `timestamptz` and Iceberg's `timestamptz` are both microsecond-resolution, and the Arrow writer already used `Timestamp_us`, so the legacy writer was the outlier. That divergence is what #1073 tracked.

Changes:
- `legacy-writer/writer.go` — encode `t.UnixMicro()`
- `SchemaConvertor.java` — decode via `DateTimeUtil.timestamptzFromMicros(...)`
- `types/data_types.go` — `IcebergTypeToDatatype("timestamptz")` returns `TimestampMicro`
- Removed `TS_MS_FIELDS` and its unreachable `long` branch (millisecond-era relic)
- Removed the arrow-vs-legacy precision split in `tests/s3/s3_util_test.go` and the TODO that
  generated #1073

The Go binary and the Java JAR must be released together — they agree on the wire unit, and a
one-sided upgrade misreads every timestamp by 1000x.

Fixes #1188 #1073

## Type of change

# How Has This Been Tested?

Postgres (CDC, pgoutput) → Iceberg, **JDBC catalog on MinIO**, legacy writer (`arrow_writes` unset),
normalized mode. Read back via Spark.

- [x]  Precision round trip: source `2023-05-18 03:52:45.573605+00` reads back as
      `2023-05-18 03:52:45.573605`. Issue's own repro value `2026-01-01 00:00:00.123456` also exact.
      `_olake_timestamp` lands at the correct current time with micros, confirming both sides agree
      on the unit.
- [x]  Backward compatibility: incremental sync with state against a table already
      holding rows written by the old build. Old millisecond rows (`.654`) read back correctly
      alongside new microsecond rows (`.654753`) — no 1000x shift, no corruption.
- [x] Schema creation after `TS_MS_FIELDS` removal: new table with a `timestamptz` and
      a real `bigint`. `_olake_timestamp` / `_cdc_timestamp` still created as `timestamp`, `bigint`
      still `bigint` — schema identical to before.
- [x]  Unit tests pass. One expectation updated (`TestResolveColumnType`), which asserted
      the old `TimestampMilli` label; filter behaviour unchanged since `ParseFilterValue` treats all
      timestamp variants identically.

Not yet covered:
- [ ]  Arrow writer regression check, to confirm the shared-table path is unaffected
- [ ] Non-normalized (stringified-JSON) write mode
- [ ]  Pre-1970 / far-future timestamps end-to-end (`DateTimeUtil` verified in isolation down to `1966-02-02T02:26:50.987655Z`, not through a sync)
- [ ]  CDC delete / upsert path, and schema evolution adding a `timestamptz` column
- [ ] Non-Postgres sources (MySQL, MongoDB, Oracle, MSSQL, DB2)
- [ ] partitioned tables and filters

# Screenshots or Recordings

https://github.com/user-attachments/assets/e3b85c12-c4f9-4559-9a72-6fb98a3931a4

final table
Screenshot 2026-09-17 at 2 54 30 AM

## Documentation


vaibhav-datazip:
Please resolve the conflicts, we will start the review once that's done

faridkotia:
@vaibhav-datazip the merge conflicts have been resolved. Kindly check.

vaibhav-datazip (approved):
LGTM

mihir-datazip (approved):

mihir-datazip (changes requested):
The millisecond split is gone everywhere, but s3 is the only integration test that seeds sub-millisecond timestamps, so it's the only one that exercises this fix. Every other suite seeds whole seconds (e.g. `'2023-01-01 12:00:00+00'`), and those round-trip the same through the old millis encoding and the new micros one, so they pass with or without this PR.

Could we give each suite one sub-millisecond value (e.g. `.573605`, the repro value from the description) on its micro-capable timestamp columns, in both the seed and update statements, and move the expected `time.Date(...)` nanos to match?

- **postgres**: `col_timestamp`, `col_timestamptz` (just the literals). `col_timestamptz` is the exact path this PR fixes.
- **mysql**: `created_date DATETIME` / `created_timestamp TIMESTAMP` are precision 0, so they need `DATETIME(6)` / `TIMESTAMP(6)` first.
- **mssql**: `col_datetime2`, `col_datetimeoffset`, `col_datetime2_nullable`, `created_at` (all `(6)`). `DATETIME` and `SMALLDATETIME` can't hold micros.
- **oracle**: `col_timestamp`, `col_timestamptz`, `col_timestampltz` (`DATE` can't).
- **db2**: `col_timestamp`.
- **kafka**: `timestamp_value` JSON string.
- **mongodb**: skip it, BSON Date only stores milliseconds.

faridkotia:
Agreed, s3 is the only suite actually exercising this. I'll add a sub-millisecond value (.573605) to the micro-capable timestamp columns in each suite's seed and update statements, and shift the expected time.Date(...) nanos to match — including the MySQL DATETIME(6)/TIMESTAMP(6) and MSSQL/Oracle precision bumps. Skipping mongodb, since BSON Date is millisecond-only.
Thankyou

PR #1248, opened 2026-09-21
chore(iceberg): rename Java writer package from io.debezium to io.olake

vikaxsh:
# Description

The Iceberg Java writer still used `io.debezium.server.iceberg` as its package, a leftover from its debezium-server-iceberg origin. Every Java log line printed that name, which made users think OLake runs Debezium. It doesn't.

- Rename the Java package and folders to `io.olake.iceberg` and the gRPC proto package to `io.olake.iceberg.rpc`. The Go and Java gRPC code was regenerated with the original protoc versions, so the diff contains only the rename.
- Update the Maven groupId, name and mainClass, the log4j2 logger, and the client-factory class names in `java_client.go`.
- Reword one error message and the writer README, which wrongly said data is sent "in Debezium format".
- Add `.gitignore` exceptions so `io/olake/` isn't ignored by the `olake` binary rule.
- Keep the Apache 2.0 attribution to debezium-server-iceberg in `Olake-changes-notice.txt`, as the license requires.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Scenario A
- [ ] Scenario B

# Screenshots or Recordings

## Documentation

## Related PR's (If Any):


.gitignore
```diff
@@ -1,4 +1,7 @@
 olake
+# `olake` matches any depth (driver binaries); keep the Iceberg writer's io.olake Java packages.
```
mihir-datazip: Just update `olake` to `drivers/*/olake` instead of adding manual entries for negative ignore

mihir-datazip (approved):

